PR #1, opened 2020-11-18
Sync with upstream

ddn0:
I don't think I can do a fast-forward merge in GitHub, so once this is approved, I'll just manually fast-forward master.


gurbinder533 (approved):
LGTM.

PR #4, opened 2020-12-15
Support new conan package names 

arthurp:


arthurp:
This builds correctly in AWS. Before this chance it did not.

PR #5, opened 2020-12-31
Wrap Jaccard and some "algorithms" from PropertyFileGraph.h

arthurp:
Ported from KatanaGraph/katana-enterprise#735


libgalois/src/analytics/jaccard/jaccard.cpp
```diff
@@ -0,0 +1,250 @@
...
+    break;
+  case JaccardPlan::kSorted:
+    r = JaccardImpl<IntersectWithSortedEdgeList>(
+        pg_result.value(), compare_node, plan);
+    break;
+  }
+
+  if (!r) {
+    // Undo property creation.
+    if (auto r1 = pfg->RemoveNodeProperty(output_property_name); !r1) {
```
gurbinder533: Do we need this? If we do not mark properties as persistent that should suffice, isn't it?

libgalois/src/analytics/jaccard/jaccard.cpp
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  return r;
+}
+
+constexpr static const double EPSILON = 1e-6;
+
+galois::Result<void>
+galois::analytics::JaccardAssertValid(
+    galois::graphs::PropertyFileGraph* pfg, size_t compare_node,
```
gurbinder533: `uint32_t` for compare_node? or `GNode`.

libgalois/src/analytics/jaccard/jaccard.cpp
```diff
@@ -0,0 +1,250 @@
...
+      graph.end()) {
+    return galois::ErrorCode::AssertionFailed;
+  }
+
+  return galois::ResultSuccess();
+}
+
+galois::Result<JaccardStatistics>
+galois::analytics::JaccardStatistics::Compute(
+    galois::graphs::PropertyFileGraph* pfg, size_t compare_node,
```
gurbinder533: ditto.

lonestar/analytics/cpu/jaccard/jaccard_cli.cpp
```diff
@@ -99,75 +64,58 @@ main(int argc, char** argv) {
...
+  if (!stats_result) {
+    GALOIS_LOG_FATAL(
+        "could not make compute statistics: {}", stats_result.error());
+  }
+
+  stats_result.value().Print();
 
   if (!skipVerify) {
-    if (graph.GetData<NodeValue>(base) == 1.0) {
+    if (galois::analytics::JaccardAssertValid(
```
gurbinder533, line 117: Isn't this returning `Galois::Result`, so should check `.value`?

gurbinder533 (approved):
LGTM, except for a few minor things.

libgalois/src/analytics/jaccard/jaccard.cpp
arthurp: Ah. Yes. I meant to ask about this.

I think we do need it. The reason is that leaving the property on the `PropertyFileGraph` will: a) force a rerun of the algorithm to use a different property name to avoid conflicts, b) waste memory since the old property is referenced and will hang around in memory. These are both issue in a Jupyter notebook where having a failure in a given cell is usually handled by only rerunning that one cell.

We could fix this by changing the property creation semantics of algorithms to be "add or overwrite", but I think that could make it easy to accidentally destroy your data due to the silent overwrite.

libgalois/src/analytics/jaccard/jaccard.cpp
arthurp: GNode doesn't really exist any more. size_t is wrong though. I'll fix it.

lonestar/analytics/cpu/jaccard/jaccard_cli.cpp
arthurp: The assert valid routines don't return a bool any more (at the request of dnn0). They just error out if the checks see invalidities. So this is correct now.

libgalois/src/analytics/jaccard/jaccard.cpp
arthurp: I'll remove it for now and create an issue to track this and fix it eventually. One way or the other.

lonestar/analytics/cpu/jaccard/jaccard_cli.cpp
gurbinder533: Got it. I didn't look at the parallelSTL loop. Makes sense.

libgalois/src/analytics/jaccard/jaccard.cpp
gurbinder533: oh ok. Makes sense to remove it if it produces invalid results. I was thinking of just doing a single routine on a graph but one can do multiple routines and these properties will continue to grow.

PR #7, opened 2020-12-31
Add Github Actions CI

arthurp:


.github/workflows/cpp.yaml
```diff
@@ -0,0 +1,255 @@
...
+      run: CLANG_FORMAT=clang-format-10 external/katana/scripts/check_format.sh .
+
+    - name: Check general text formatting
+      run: |
+        external/katana/scripts/check_general_text_format.py lib* cmake/ external/katana/cmake/ katana/ deploy/ \
+          demosite/ config external/katana/lib* external/katana/python/ external/katana/config
+
+    - name: Check guards
+      run: |
+        external/katana/scripts/check_ifndefs.py --root=$PWD lib*
```
ddn0: Remove `external` (throughout)?

.github/workflows/python.yaml
```diff
@@ -0,0 +1,221 @@
...
+
+        INPUT_URL="https://repo.anaconda.com/miniconda/Miniconda3-py38_4.8.3-${PLAT}.sh"
+        mkdir -p "$(dirname $MINICONDA_FILE)"
+        curl -fL --output "$MINICONDA_FILE" "$INPUT_URL"
+      if: steps.cache-miniconda.outputs.cache-hit != 'true'
+    - name: Download graphs
+      run: |
+        GRAPH_FILE="$HOME/.cache/graph/inputs.tar.gz"
+        GRAPH_LOCATION="$HOME/.cache/graph/inputs"
+        INPUT_URL="https://katana-ci-public.s3.us-east-1.amazonaws.com/inputs/katana-inputs-v14.tar.gz"
```
ddn0: IIRC, there is some code in python codebase that can download inputs on demand. If that's true, I might consider just always relying on that rather using all this cache logic because downloading the inputs takes a relatively small amount of time anyway.

ddn0 (approved):
Happy for this to land now as well if that is what it will take to get actions to run on PRs at all.

PR #9, opened 2020-12-31
Fix CI

arthurp:


arthurp:
@ddn0 Can you set all the CI jobs, except the MacOS ones, as being required to merge? Just like enterprise. I'd do it, but I don't think I can because I'm not an admin.

arthurp:
Oh and set review required too. It would let me merge without review right now.

tylershunt:
I did it!

tylershunt (approved):
LGTM - workflow cancelation is only important when we're trying to prevent long queues on our private runners anyway.

arthurp:
> LGTM - workflow cancelation is only important when we're trying to prevent long queues on our private runners anyway.

I would like workflow cancellation. There are queues for the Github runners too. But there isn't a secure way to do it that I can see.

.github/workflows/python.yaml
```diff
@@ -68,7 +66,11 @@ jobs:
       run: |
         brew install ccache curl
       # TODO: Does brew install need bzip2, libz-dev, and make?
-    
+
```
ddn0, line 69: The TODO above seems orphaned.

ddn0 (approved):

.github/workflows/python.yaml
arthurp: I'm gonna do some CI clean up once I have it working. I'll get this then.

PR #11, opened 2020-12-31
tsuba: move http client code to enterprise

tylershunt:
There's nothing particularly secret about this code; this move is mainly
to simplify my workflow since changes to the HTTP client will only be
tested by enterprise CI.

There are some pragmatic reasons to do this:
 * The http server code lives in enterprise
 * Gets rid of scary warning message about consistency (that open-only
   users can't help anyway) that doesn't matter unless there are
   concurrent users.


tylershunt:
Stay tuned for a sibling PR in `katana-enterprise` where I bring the code I deleted back from the dead.

arthurp:
Just to confirm: This doesn't remove any HTTP client that would, say, allow the use of KF graphs from HTTP URLs. This is only related to the name-server and distributed consistency system supported by enterprise. Correct?

tylershunt:
> Just to confirm: This doesn't remove any HTTP client that would, say, allow the use of KF graphs from HTTP URLs. This is only related to the name-server and distributed consistency system supported by enterprise. Correct?

That's correct. This is only related to the name-server and distributed consistency. Although the functionality that reads graphs from HTTP URLs is also enterprise only atm.

And there is a general HTTP client code in `libsupport` (this code is the only user of that right now), my plan was to leave that where it is.

arthurp (approved):
LGTM.

Let's definitely leave the HTTP code in libsupport for the moment. I would love to see the HTTP graph loading code move to open source at some point. Seems useful and not in any way secret.

tylershunt:
After #9 since I rebased on there to make sure CI works.

PR #14, opened 2021-01-02
Deduplicate CI and setup scripts.

arthurp:
This moves changes from enterprise `setup_ubuntu.sh` to the open repo.


.github/workflows/check_docs_ci.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
ddn0: When these snippets are run by github actions, they are run with `set -e`; we should probably at least do the same here.

If we want to be more paranoid, `set -euo pipefail` is a good default for most scripts. If you add `-u` you may have to do something like `${1-}` to ensure that `$1` is always defined. Or simply allow bash to handle that as an error since that's what happens below. I don't mind being terse here since the only CI workflows call these scripts.

.github/workflows/configure_cpp_ci.sh
```diff
@@ -0,0 +1,84 @@
+#! /bin/bash
```
ddn0: ditto about `set -e` and co, and `#!/bin/bash` for consistency

.github/workflows/cpp.yaml
```diff
@@ -47,19 +47,11 @@ jobs:
...
-      run: |
-        [[ -f ~/.profile ]] && source ~/.profile
-        .github/workflows/setup_conan.sh
-        conan remove --locks
-        # try to install, but if it doesn't work try to rebuild from scratch
-        # because changes to conan flags can cause it to get stuck
-        conan install -if $HOME/build --build=missing config || \
-          (rm -rf ~/.conan && .github/workflows/setup_conan.sh && \
-             conan install -if $HOME/build --build=missing config)
+      run: .github/workflows/setup_conan_ci_install.sh $HOME/build
```
ddn0: if we are going to add `ci` to names everywhere, can this be `setup_conan_install_ci.sh` so it matches the other scripts?

Or simply drop `ci` altogether since it can be understood from the placement of these files in `.github/workflows`.

If you are worried about the name `.github/workflows/check_docs.sh` shadowing `scripts/check_docs.sh` (and somewhat similarly for `setup_conan_ci_install.sh` and `setup_conan.sh`, you can put all the CI related stuff into one wrapper script, e.g., `.github/workflows/ci_wrap.sh check_docs ...; ` (etc.)

.github/workflows/download_miniconda.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
ddn0, line 3: set -e

.github/workflows/check_docs_ci.sh
```diff
@@ -0,0 +1,15 @@
...
+BUILD_DIR=$1
+shift 1
+
+if [ -z "$BUILD_DIR" ]; then
+  echo "ERROR: BUILD_DIR should be argument 1."
+  exit 1
+fi
+
+CHECK_DOCS="$(cd "$(dirname "$0")"/../../scripts || exit 1; pwd)"/check_docs.sh
```
ddn0: We don't add `|| exit 1` elsewhere and the script will fail anyway if `CHECK_DOCS` doesn't point to `check_docs.sh`. Perhaps drop?

I also have a slight preference for just having `REPO_ROOT="$(cd "$(dirname "$0")/../.."); pwd)"` at the top of all these scripts so everything can be referred to consistently.

(One can get pretty sophisticated/cryptic with getting the path [1, 2]. Anything simple and clear is fine with me. Anything that needs to be robust probably shouldn't be done in bash anyway.)


[1] https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself
[2] https://stackoverflow.com/questions/4774054/reliable-way-for-a-bash-script-to-get-the-full-path-to-itself

.github/workflows/setup_conan_ci_install.sh
```diff
@@ -0,0 +1,20 @@
+#! /bin/bash
```
ddn0: #!/bin/bash

.github/workflows/setup_miniconda.sh
```diff
@@ -0,0 +1,7 @@
+# This is a bash INCLUDE file. Source it don't run it.
```
ddn0, line 1: maybe `activate_miniconda.sh` is a better name because this doesn't install or configure anything unlike the other `setup_` scripts.

.github/workflows/check_docs_ci.sh
arthurp: I'll take your simple consistent approach. I really wish the BSD runtime had `realpath`. Lacking it really makes scripting correctly harder.

.github/workflows/cpp.yaml
arthurp: Since these directories are never added to `PATH` I'm not really worried about duplcate names. I'll go with your initial thought.

arthurp:
@ddn0 I think I have addressed all your concerns. Thanks for the review. Lots of good advise.

BTW, I considered adding `set -x` to the bash scripts used in CI so that what is executed is obvious in the action log. Do you think that would be a good idea?

ddn0:
> BTW, I considered adding set -x to the bash scripts used in CI so that what is executed is obvious in the action log. Do you think that would be a good idea?

Yeah `set -x` is a good idea.

ddn0 (approved):

tylershunt (approved):

PR #18, opened 2021-01-04
Wrap algorithm utilities

arthurp:


libgalois/include/galois/analytics/sssp/sssp.h
```diff
@@ -134,412 +134,32 @@ using SsspEdgeWeight = galois::PODProperty<Weight>;
...
+GALOIS_EXPORT Result<bool> SsspValidate(
+    graphs::PropertyFileGraph* pfg, size_t start_node,
+    const std::string& edge_weight_property_name,
+    const std::string& output_property_name);
 
-    if (plan.algorithm() == SsspPlan::kAutomatic) {
-      plan = SsspPlan::Automatic(&graph.GetPropertyFileGraph());
-    }
+struct GALOIS_EXPORT SsspStatistics {
+  /// The maximum distance across all nodes.
```
gurbinder533, line 147: Let's add "source node" similar to bfs here as well.

libgalois/include/galois/analytics/sssp/sssp.h
```diff
@@ -134,412 +134,32 @@ using SsspEdgeWeight = galois::PODProperty<Weight>;
...
-      TopoTileAlgo(&graph, source);
-      break;
-    case SsspPlan::kDeltaStepBarrier:
-      DeltaStepAlgo<UpdateRequest, OBIMBarrier>(
-          &graph, source, ReqPushWrap(), OutEdgeRangeFn{&graph}, plan.delta());
-      break;
-    default:
-      return galois::ErrorCode::InvalidArgument;
-    }
+  float average_distance() { return total_distance / n_reached_nodes; }
```
gurbinder533: float(total_distance)?

libgalois/include/galois/analytics/sssp/sssp.h
arthurp: I intentionally left it out because it cannot be computed from the graph since multiple nodes may have 0 distance. I think it would have to be passed to Compute which makes it questionably useful as a "statistic". In BFS, it can be computed as the sole 0 in the graph.

libgalois/include/galois/analytics/sssp/sssp.h
arthurp: Actually I think it should return double.

libgalois/include/galois/analytics/sssp/sssp.h
arthurp: We could consider attaching the source node as metadata on the property itself. Similar to the [field metadata in Arrow tables](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv45fieldNSt6stringENSt10shared_ptrI8DataTypeEEbNSt10shared_ptrIK16KeyValueMetadataEE). If we actually used Arrow Tables under the hood we could just _use_ their metadata system directly. But with our current system we will have to roll our own.
With metadata we could store information about the algorithm configuration which generated the data. The more I think about this, the more I like it. It's kinda of a provenance. We could also store the weight property used for SSSP so that could be looked up later.

gurbinder533: Yeah, makes sense. Good point.

gurbinder533:
This looks good to me.

arthurp:
> This looks good to me.

Good. I think the API is good at this point.

This PR however is not. ;-) It's not even ready for review. It doesn't pass the tests. There is something up with the SSSP validation code.

lonestar/analytics/cpu/bfs/bfs_cli.cpp
```diff
@@ -137,46 +119,37 @@ main(int argc, char** argv) {
...
-            &graph, source)) {
+    if (stats.n_reached_nodes < graph.num_nodes()) {
+      std::cerr << (graph.num_nodes() - stats.n_reached_nodes)
+                << " unvisited nodes; this is an error if the graph is "
+                   "strongly connected\n";
+    }
+    if (auto r = BfsValidate(pfg.get(), "level"); r && r.value()) {
       std::cout << "Verification successful.\n";
     } else {
       GALOIS_DIE("verification failed");
```
gurbinder533, line 139: GALOIS_LOG_FATAL instead of GALOIS_DIE.

lonestar/analytics/cpu/bfs/bfs_cli.cpp
```diff
@@ -137,46 +119,37 @@ main(int argc, char** argv) {
...
-      },
-      galois::loopname("Sanity check"), galois::no_stats());
-
-  // report sanity stats
-  galois::gInfo("# visited nodes is ", num_visited.reduce());
-  galois::gInfo("Max distance is ", max_dist.reduce());
-  galois::gInfo("Sum of visited distances is ", sum_dist.reduce());
+  auto stats_result = BfsStatistics::Compute(pfg.get(), "level");
+  if (!stats_result) {
+    std::cerr << stats_result.error().message() << "\n";
```
gurbinder533, line 124: Why not GALOIS_LOG_FATAL?

lonestar/analytics/cpu/bfs/bfs_cli.cpp
```diff
@@ -137,46 +119,37 @@ main(int argc, char** argv) {
...
       GALOIS_DIE("verification failed");
     }
   }
 
   if (output) {
-    std::vector<uint32_t> results = MakeResults(graph);
-    assert(results.size() == graph.size());
+    auto r = pfg->NodePropertyTyped<uint32_t>("level");
+    if (!r) {
+      std::cerr << r.error().message() << "\n";
```
gurbinder533, line 146: GALOIS_LOG_FATAL?

libgalois/include/galois/graphs/PropertyFileGraph.h
```diff
@@ -202,14 +203,68 @@ class GALOIS_EXPORT PropertyFileGraph {
...
+    }
+
+    auto array =
+        std::dynamic_pointer_cast<typename arrow::CTypeTraits<T>::ArrayType>(
+            chunked_array->chunk(0));
+    if (!array) {
+      return ErrorCode::TypeError;
+    }
+    return array;
+  }
```
arthurp, line 266: @danielmawhirter Could I get you to look at these new functions? There was some debate and I wanted to make sure this was OK with people.

danielmawhirter: Looks good to me! Only note is I tend to LOG something at an error's origin, but not sure if that's standardized through the codebase.

arthurp: The logs are not actually useful because they will not be visible in all cases (for instance they don't reliably show up in Jupyter notebooks when called from the Python API). Also the logged message will be spurious if the error is handled by the caller and not propagated further. The idea going forward is to attach message and location information to the error return value itself. I think we should probably push forward with that soon and start standardizing.

gurbinder533 (approved):
LGTM. Let's merge this as soon as possible.

PR #20, opened 2021-01-05
adding k shortest paths implementation

uditagar:
This is an implementation for finding k shortest paths (paths can contain cycles) from start to end node.


lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+using OutEdgeRangeFn = SSSP::OutEdgeRangeFn;
+using TileRangeFn = SSSP::TileRangeFn;
+
+using PSchunk = katana::PerSocketChunkFIFO<kChunkSize>;
+using OBIM = katana::OrderedByIntegerMetric<UpdateRequestIndexer, PSchunk>;
+using OBIM_Barrier = katana::OrderedByIntegerMetric<
+    UpdateRequestIndexer, PSchunk>::with_barrier<true>::type;
+
+bool
+CheckIfReachable(Graph* graph, const GNode& source) {
```
gurbinder533: `CheckReachability`?

uditagar: will do

lonestar/analytics/cpu/k-shortest-paths/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+2. Run `cd <BUILD>/lonestar/analytics/cpu/k-shortest-paths; make -j`
+
+RUN
+--------------------------------------------------------------------------------
+
+The following are a few example command lines.
+
+-`$ ./k-shortest-paths-cpu <path-to-graph> --algo=deltaStep --delta=13 --edgePropertyName=value --numPaths=10 --startNode=1 --reportNode=100 -t 40 
+-`$ ./k-shortest-paths-cpu <path-to-graph> --algo=deltaTile --delta=13 --edgePropertyName=value --numPaths=10 --startNode=1 --reportNode=100 -t 40`
```
roshandathathri: Why two identical lines?

lonestar/analytics/cpu/k-shortest-paths/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+PERFORMANCE  
+--------------------------------------------------------------------------------
+
+* deltaStep/deltaTile algorithms typically performs the best on high diameter
+  graphs, such as road networks. Its performance is sensitive to the *delta* parameter, which is
+  provided as a power-of-2 at the commandline. *delta* parameter should be tuned
+  for every input graph
+* topo/topoTile algorithms typically perform the best on low diameter graphs, such
```
roshandathathri: Drop this point

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+struct NodeCount {
+  using ArrowType = arrow::CTypeTraits<uint32_t>::ArrowType;
+  using ViewType = katana::PODPropertyView<std::atomic<uint32_t>>;
+};
+
+struct NodeMax {
+  using ArrowType = arrow::CTypeTraits<uint32_t>::ArrowType;
+  using ViewType = katana::PODPropertyView<std::atomic<uint32_t>>;
+};
```
roshandathathri, line 89: Replace with:
```
using NodeCount = UInt32Property;
using NodeMax = UInt32Property;
```

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+struct EdgeWeight : public katana::PODProperty<uint32_t> {};
```
roshandathathri: Why not?
`using EdgeWeight = UInt32Property;`

lonestar/analytics/cpu/k-shortest-paths/README.md
```diff
@@ -0,0 +1,39 @@
...
+  as social networks and RMAT graphs
+* All algorithms rely on CHUNK_SIZE for load balancing,
```
roshandathathri: Change `CHUNK_SIZE` to `kChunkSize`

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+bool
+CheckReachability(Graph* graph, const GNode& source) {
+  katana::InsertBag<GNode> current_bag;
+  katana::InsertBag<GNode> next_bag;
+
+  current_bag.push(source);
+  graph->GetData<NodeCount>(source) = 1;
+
+  while (true) {
+    if (current_bag.begin() == current_bag.end())
```
roshandathathri: `while (current_bag.begin() != current_bag.end())`

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+      break;
+
+    katana::do_all(
+        katana::iterate(current_bag),
+        [&](GNode n) {
+          for (auto edge : graph->edges(n)) {
+            auto dest = *(graph->GetEdgeDest(edge));
+            if (graph->GetData<NodeCount>(dest) == 0) {
+              graph->GetData<NodeCount>(dest) = 1;
```
roshandathathri, line 203: This is just BFS right? Maybe add an option for asynchronous `for_each`?

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+    current_bag.clear();
+    std::swap(current_bag, next_bag);
+  }
+
+  if (graph->GetData<NodeCount>(reportNode) == 0) {
+    return false;
+  }
+
+  katana::do_all(katana::iterate(*graph), [&graph](GNode n) {
+    graph->GetData<NodeCount>(n) = 0;
```
roshandathathri, line 219: Are you sure you don't want to do this if you're returning false?

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+  katana::GAccumulator<size_t> bad_work;
+  //! [reducible for self-defined stats]
+  katana::GAccumulator<size_t> wl_empty_work;
+
+  graph->GetData<NodeCount>(source) = 1;
+
+  katana::InsertBag<Item> init_bag;
+
+  Path* path = new Path();
+  path->last = NULL;
```
roshandathathri: `nullptr`

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+          path->last = item.path;
+          path_pointers->push(path);
+
+          if (ddata_count < numPaths) {
+            katana::atomicAdd<uint32_t>(ddata_count, (uint32_t)1);
+            katana::atomicMax<uint32_t>(ddata_max, new_dist);
+          }
+
+          if (dst == reportNode)
+            bag->push(std::make_pair(new_dist, path));
```
roshandathathri: `if (...) {}`

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+  return true;
+}
+
+//delta stepping implementation for finding a shortest path from source to report node
+template <typename Item, typename OBIMTy, typename PushWrap, typename EdgeRange>
+void
+DeltaStepAlgo(
+    Graph* graph, const GNode& source, const PushWrap& pushWrap,
+    const EdgeRange& edgeRange,
+    katana::InsertBag<std::pair<uint32_t, Path*>>* bag,
```
roshandathathri: A more descriptive name please. Like report_node_bag?

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+            katana::atomicAdd<uint32_t>(ddata_count, (uint32_t)1);
+            katana::atomicMax<uint32_t>(ddata_max, new_dist);
+          }
+
+          if (dst == reportNode)
+            bag->push(std::make_pair(new_dist, path));
+
+          if ((graph->GetData<NodeCount>(reportNode) < numPaths) ||
+              ((graph->GetData<NodeCount>(reportNode) >= numPaths) &&
+               (graph->GetData<NodeMax>(reportNode) > new_dist))) {
```
roshandathathri: A more clear way to write this:
```
bool terminate = ...;
if (!terminate) {
  ...
  pushWrap(..);
}
```
More comments would help too.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+            const Path* const_path = path;
```
roshandathathri, line 285: Why is this required?

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
roshandathathri: There's another way to terminate. `for_each` has support for a parallel break/exit. 
So a more clear way to write this is:
```
for_each(..., [&](const Item& item, auto& ctx) {
for (auto ii : edgeRange(item)) {
  GNode dst = *(graph->GetEdgeDest(ii));
  ...
  if (dst == reportNode) {
    ... 
    bool terminate = ...;
    if (terminate) {
      ctx.breakLoop();
      return;
    }
  }
  ...
  pushWrap(...);
}
}, katana::parallel_break(), ...);
```

However, I'm not sure if this would be more performant than the non-break option above. So please benchmark both before you choose one.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+    katana::gInfo("Using delta-step of ", (1 << stepShift), "\n");
+    KATANA_LOG_WARN(
+        "Performance varies considerably due to delta parameter.\n");
+    KATANA_LOG_WARN("Do not expect the default to be good for your graph.\n");
+  }
+
+  katana::do_all(katana::iterate(graph), [&graph](GNode n) {
+    graph.GetData<NodeMax>(n) = 0;
+    graph.GetData<NodeCount>(n) = 0;
+  });
```
roshandathathri: Please push this to a function called `Initialize*()`

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+    default:
+      std::abort();
+    }
+  }
+
+  execTime.stop();
+
+  if (reachable) {
+    std::multimap<uint32_t, Path*> paths_map;
```
roshandathathri, line 423: Is this an ordered map?

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+    for (uint32_t iter = 0; iter < num; iter++) {
+      katana::gPrint(" ", report);
+      GNode parent = report;
+
+      const Path* path = it_report->second;
+      it_report++;
+
+      while (path->last != NULL) {
+        parent = path->parent;
+        katana::gPrint(" ", parent);
```
roshandathathri: nit: I think it's better to print a path from source to report node. Please write a recursive function to print the path in such a way.

Also please print the distance of the path.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+        path = path->last;
+      }
+      katana::gPrint("\n");
+    }
+
+    katana::do_all(
+        katana::iterate(path_pointers), [&](Path* p) { delete (p); });
```
roshandathathri: I had originally typed this: "Drop `path_pointers` and change all `Path*` to `std:;shared_ptr<Path>`. We should avoid using plain/C-style pointers". However, with the use of custom allocation, `std::shared_ptr` cannot be used. So, this is fine for now.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,345 @@
...
+          auto& ddata_max = graph->GetData<NodeMax>(dst);
+
+          Distance ew = graph->GetEdgeData<EdgeWeight>(ii);
+          const Distance new_dist = item.distance + ew;
+
+          if ((ddata_count >= numPaths) && (ddata_max <= new_dist))
+            continue;
+
+          Path* path;
+          path = new Path();
```
roshandathathri: Concurrent allocation using `new` would be bad in performance. Use our custom allocators instead. Try something like:
```
class PathAlloc {
public:
  Path* NewPath() {
     Path* path = allocator_.allocate(1);
     allocator_.construct(path, Path());
     return path;
  }

  void DeletePath(Path* path) {
     allocator_.destroy(path);
     allocator_.deallocate(path, 1);
  }

private:
  katana::FixedSizeAllocator<Path> alloc_;
};
```

Then instead of calling `new Path()` and `delete (path)`, call `path_alloc.NewPath()` and `path_alloc.DeletePath(path)`. Declare `PathAlloc path_alloc;` before `for_each`.

lonestar/analytics/cpu/k-shortest-paths/README.md
uditagar: not identical. It just shows two different algo types

roshandathathri: Sorry for missing that!

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: sorry, I don't understand this one. If you can further explain, that would be good. Thank you

uditagar: this will probably not allow atomic adds, I think

roshandathathri: Yes, I had missed that. This is fine as is for now.

But I think we should add `std::atomic` variants for property views in Properties.h in the long run. @ddn0

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: yes, because there is nothing more to do if there is no path from source to report node and thus the code terminates, when false is returned,

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: I think my intention was not to terminate but to check whether I want to put this new extended path to the worklist. And I guess your second suggestion would probably exit the for loop, right? I might be wrong here too. thanks

roshandathathri: Yes, both yield the same results.

uditagar: but I am not trying to check for termination. I do admit that I should have written it in a better way, like your first suggestion.

roshandathathri: Why not? It's safe to terminate as soon as you have the required number of paths for the report node.

uditagar: the issue is that there is still a chance that I can a get a shorter path later on. So I need to run it until all items in the worklist are processed.

roshandathathri: Makes sense. Yes, it's not a termination condition. I get it now.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: yes, it's ordered

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: I am getting runtime errors with this approach. It will be good if you can give me some suggestion. thanks

roshandathathri: What error are you getting?

It definitely supports concurrent allocations. Check delaunay triangulation for an example usage.

uditagar: Thanks man. I will go through it then.

uditagar: I get this segmentation fault: `Thread 1 "k-shortest-path" received signal SIGSEGV, Segmentation fault.
katana::PerBackend::getRemote (this=0x55555592a4c0 <katana::getPTSBackend()::b>, thread=0, 
    offset=3968) at /home/udit/udit_katana_public_Jan5/libgalois/src/PerThreadStorage.cpp:155
155	  char* rbase = heads[thread].load(std::memory_order_relaxed);`

roshandathathri: I can help debug this. Can you please push your code?

uditagar: I just pushed it. Thank you.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,457 @@
...
+const char* const ALGO_NAMES_BFS[] = {"asyncBFS", "syncBFS"};
+
+static cll::opt<AlgoBFS> algoBFS(
+    "algoBFS", cll::desc("Choose an algorithm:"),
+    cll::values(clEnumVal(asyncBFS, "asyncBFS"), clEnumVal(syncBFS, "syncBFS")),
+    cll::init(syncBFS));
+
+struct Path {
+  uint32_t parent;
+  const Path* last;
```
roshandathathri:
```cpp
const Path* last{nullptr};
```

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,457 @@
...
+  //! [reducible for self-defined stats]
+  katana::GAccumulator<size_t> wl_empty_work;
+
+  graph->GetData<NodeCount>(source) = 1;
+
+  katana::InsertBag<Item> init_bag;
+
+  //Path* path = new Path();
+  Path* path;
+  path_alloc.NewPath();
```
roshandathathri:
```cpp
Path* path = path_alloc.NewPath();
```

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,457 @@
...
+  path->last = nullptr;
```
roshandathathri, line 243: Drop line. It is not required with the default initialization I mentioned above.

However, what is the value of parent for this path?

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+          Distance ew = graph->GetEdgeData<EdgeWeight>(ii);
+          const Distance new_dist = item.distance + ew;
+
+          if ((ddata_count >= numPaths) && (ddata_max <= new_dist))
+            continue;
+
+          Path* path;
+          //path = new Path();
+          path_alloc.NewPath();
```
roshandathathri: `Path* path = path_alloc.NewPath();`

roshandathathri (commented):
Try if these changes fixes the issue.

uditagar:
> Try if these changes fixes the issue.

Thanks, I will check them now.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: this path pointer corresponds to the source node, hence no parent to keep track of.

uditagar:
Thank you for the comments. It seems I made some stupid mistake there. It seems to be working now.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
roshandathathri: It's not good to leave the parent uninitialized. Maybe just assign source as the parent?

roshandathathri:
No worries. `NewPath()` returns a pointer just like `new Path()`

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
uditagar: ok, I will do that

uditagar:
I have incorporated all the comments now. Thank you

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,452 @@
...
+    cll::desc("Number of paths to compute from source to report node (default "
+              "value 1)"),
+    cll::init(1));
+enum AlgoSSSP { deltaTile = 0, deltaStep, deltaStepBarrier };
+
+const char* const ALGO_NAMES_SSSP[] = {
+    "deltaTile", "deltaStep", "deltaStepBarrier"};
+
+static cll::opt<AlgoSSSP> algoSSSP(
+    "algoSSSP", cll::desc("Choose an algorithm:"),
```
roshandathathri: -> "Choose an algorithm for SSSP"

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,452 @@
...
+    cll::values(
+        clEnumVal(deltaTile, "deltaTile"), clEnumVal(deltaStep, "deltaStep"),
+        clEnumVal(deltaStepBarrier, "deltaStepBarrier")),
+    cll::init(deltaTile));
+
+enum AlgoBFS { asyncBFS = 0, syncBFS };
+
+static cll::opt<AlgoBFS> algoBFS(
+    "algoBFS", cll::desc("Choose an algorithm:"),
```
roshandathathri: -> "Choose an algorithm for reachability"

BFS and SSSP is confusing. Use "reachability" instead of bfs.

lonestar/analytics/cpu/k-shortest-paths/SSSP.cpp
```diff
@@ -0,0 +1,452 @@
...
+    cll::values(clEnumVal(asyncBFS, "asyncBFS"), clEnumVal(syncBFS, "syncBFS")),
```
roshandathathri: Drop BFS in the string names

roshandathathri (approved):
Thanks, Udit. LGTM

PR #21, opened 2021-01-06
Add doc on merging and dealing with CI

ddn0:


CONTRIBUTING.md
```diff
@@ -345,6 +372,15 @@ if (auto r = ReturnsAResult(); !r) {
...
 # Continuous Integration
 
+## Dealing with CI Errors
+
+If the error is due to a transient external failure, you can re-run jobs in the
+GitHub UI.
+
+The CI itself just runs commands that can easily be duplicated in any developer
+environment. You see the exact commands by looking at the CI workflow
+definition files under the `.github` directory.
```
arthurp: Maybe mention that using a single-use VM (AWS or local) can be useful to duplicate CI runs. For myself, I wouldn't be willing to run the apt install stuff on my main OS. It adds a bunch of package repos and changes package versions. We definitely wouldn't want people running those scripts on the shared machines.

arthurp (approved):
Other than issue with duplicating CI runs, LGTM. Fix as you think is needed and merge as you like.

CONTRIBUTING.md
ddn0: How about the following instead?

When debugging a CI failure, it is good to confirm that tests pass locally in your developer environment first. You can also run many of the source checks locally as well. Take a look at the GitHub workflow definitions under `.github` directory to see what script and build parameters are used.

arthurp: That works. I kind of misunderstood what you wrote the first time I think. And this is clearer.

PR #22, opened 2021-01-07
Wrap SSSP and BFS for Python and test

arthurp:


libgalois/include/galois/Properties.h
```diff
@@ -182,6 +182,26 @@ ConstructPropertyViews(const std::vector<arrow::Array*>& arrays) {
       arrays, std::make_index_sequence<std::tuple_size_v<PropTuple>>());
 }
 
+namespace {
+/// This function works around a bug in NumPyBuffer (arrows wrapper around numpy
+/// arrays) by const casting the values pointer of mutable buffers.
```
ddn0: nit: can you replace "by const casting the values pointer of mutable buffers." (what) with something like "which claims a buffer is mutable but doesn't support mutable_data()." (why) or something along those lines?

libgalois/include/galois/Properties.h
```diff
@@ -182,6 +182,27 @@ ConstructPropertyViews(const std::vector<arrow::Array*>& arrays) {
       arrays, std::make_index_sequence<std::tuple_size_v<PropTuple>>());
 }
 
+namespace {
+/// Get the mutable values pointer of a mutable ArrayData.
+/// This function works around a bug in NumPyBuffer (arrowss wrapper around numpy
```
ddn0: nit: "arrow's"

libgalois/include/galois/Properties.h
```diff
@@ -202,19 +223,40 @@ class PODPropertyView {
...
+          sizeof(T));
+      return ErrorCode::ArrowError;
+    }
+    if (array.offset() < 0) {
+      GALOIS_LOG_DEBUG("arrow error: immutable buffers not supported");
+      return ErrorCode::ArrowError;
+    }
+    if (array.data()->buffers.size() <= 1 ||
+        !array.data()->buffers[1]->is_mutable()) {
+      GALOIS_LOG_DEBUG("arrow error: Offset not supported");
```
ddn0: seems like this message should be exchanged with the one above

libgalois/include/galois/analytics/bfs/bfs.h
```diff
@@ -70,12 +68,17 @@ using BfsNodeDistance = galois::PODProperty<uint32_t>;
 /// not exist before the call.
 GALOIS_EXPORT Result<void> Bfs(
     graphs::PropertyFileGraph* pfg, size_t start_node,
-    const std::string& output_property_name,
-    BfsPlan algo = BfsPlan::Automatic());
+    const std::string& output_property_name, BfsPlan algo = {});
 
+/// Do a quick validation of the results of a BFS computation where the results
+/// are stored in property_name. This function does not do an exhaustive check.
+/// The results are approximate and make have false-negatives.
```
ddn0: "may have"

libgalois/include/galois/analytics/bfs/bfs.h
```diff
@@ -70,12 +68,17 @@ using BfsNodeDistance = galois::PODProperty<uint32_t>;
...
+/// @return True iff the results pass the validation checks, false if they do
+///   not, or an error resulting from the check algorithm itself.
```
ddn0: FWIW, I don't know how callers would behave differently between `false` from "had an error" here. Perhaps, `Result<void>`

libgalois/include/galois/analytics/sssp/sssp.h
```diff
@@ -52,6 +52,27 @@ class SsspPlan : Plan {
 public:
   SsspPlan() : SsspPlan{kCPU, kAutomatic, 0, 0} {}
 
+  SsspPlan(const galois::graphs::PropertyFileGraph* pfg) : Plan(kCPU) {
+    // TODO: What to do about const cast? We know we don't modify pfg, but there
```
ddn0: TODO initials

libgalois/include/galois/analytics/sssp/sssp.h
```diff
@@ -52,6 +52,27 @@ class SsspPlan : Plan {
...
+    //  is no way to construct a const PropertyGraph.
+    auto graph =
+        galois::graphs::PropertyGraph<std::tuple<>, std::tuple<>>::Make(
+            const_cast<galois::graphs::PropertyFileGraph*>(pfg), {}, {});
+    if (!graph)
```
ddn0: nit: if braces

libgalois/include/galois/analytics/bfs/bfs.h
arthurp: I see your point. We could go for a "throw on failure" function (It will literally raise an exception in Python). If so, I want to rename this to `BfsAssertValid` to make it clear that it's a check that will throw on failure. Thoughts?

ddn0: Raising an exception in the python API seems reasonable but the C++ code we have been writing hasn't been throwing exceptions so far. Is it the case that we are intending a strict or automated mirroring between the Python API and the C++ one?

If we do want to have an exception throwing version in the C++ library, I have a moderate preference to also have the `Result<void>` returning function too.

arthurp: Sorry. I was confusing. By raise exception, I just meant return an error object/code in C++. I see error codes as just a manual implementation of exceptions. The Python wrapper converts returns error objects into exceptions. I know we don't use C++ exceptions.

I still think it should be called `BfsAssertValid`. It converts a negative check into an error which is the fundamental action of an assertion.

ddn0: Cool. Got it.

`BfsAssertValid` seems reasonable to me.

ddn0:
LGTM; @gurbinder533 ?

lonestar/analytics/cpu/sssp/sssp_cli.cpp
```diff
@@ -140,7 +140,7 @@ main(int argc, char** argv) {
 
   std::cout << "Running " << AlgorithmName(algo) << " algorithm\n";
```
gurbinder533, line 148: Line47: Can you please change `clEnumVal` to `clEnumValN` for `algo` options? Also, in sssp/CMakeLists.txt add` --algo` explicitly to the tests. 
I think the same goes for bfs_cli.cpp.

arthurp: Done!

gurbinder533 (approved):
Awesome.

PR #26, opened 2021-01-08
Refactor topology API to make it consistent

arthurp:
* Lift topology access implementations to only exist on `GraphTopology`
* Allow direct access to topology on `PropertyFileGraph` just like `PropertyGraph` (and the Python API)

I'm aware this isn't important. But I had it mostly done before I realized I was rat-holing. So I finished it.

This is not a review priority, but I would love to get it in.

### DIFF
--- a/libgalois/include/katana/PropertyFileGraph.h
+++ b/libgalois/include/katana/PropertyFileGraph.h
@@ -9,6 +9,7 @@
 #include <arrow/chunked_array.h>
 #include <arrow/type_traits.h>
 
+#include "katana/Details.h"
 #include "katana/ErrorCode.h"
 #include "katana/LargeArray.h"
 #include "katana/config.h"
@@ -18,7 +19,14 @@ namespace katana {
 
 /// A graph topology represents the adjacency information for a graph in CSR
 /// format.
-struct GraphTopology {
+struct KATANA_EXPORT GraphTopology {
+  using Node = uint64_t;
+  using Edge = uint32_t;
+  using node_iterator = boost::counting_iterator<Node>;
+  using edge_iterator = boost::counting_iterator<Edge>;
+  using edges_range = StandardRange<edge_iterator>;
+  using iterator = node_iterator;
+
   std::shared_ptr<arrow::UInt64Array> out_indices;
   std::shared_ptr<arrow::UInt32Array> out_dests;
 
@@ -31,11 +39,44 @@ struct GraphTopology {
            out_dests->Equals(*other.out_dests);
   }
 
-  std::pair<uint64_t, uint64_t> edge_range(uint32_t node_id) const {
+  // Edge accessors
+
+  // TODO(amp): [[deprecated("use edges(node)")]]
+  std::pair<Edge, Edge> edge_range(Node node_id) const {
     auto edge_start = node_id > 0 ? out_indices->Value(node_id - 1) : 0;
     auto edge_end = out_indices->Value(node_id);
     return std::make_pair(edge_start, edge_end);
   }
+
+  /**
+   * Gets the edge range of some node.
+   *
+   * @param node an iterator pointing to the node to get the edge range of
+   * @returns iterable edge range for node.
+   */
+  edges_range edges(node_iterator node) const { return edges(*node); }
+  // TODO(amp): [[deprecated("use edges(Node node)")]]
+
+  /**
+   * Gets the edge range of some node.
+   *
+   * @param node node to get the edge range of
+   * @returns iterable edge range for node.
+   */
+  edges_range edges(Node node) const {
+    auto [begin_edge, end_edge] = edge_range(node);
+    return MakeStandardRange<edge_iterator>(begin_edge, end_edge);
+  }
+
+  // Standard container concepts
+
+  node_iterator begin() const { return node_iterator(0); }
+
+  node_iterator end() const { return node_iterator(num_nodes()); }
+
+  size_t size() const { return num_nodes(); }
+
+  bool empty() const { return num_nodes() == 0; }
 };
 
 /// A property graph is a graph that has properties associated with its nodes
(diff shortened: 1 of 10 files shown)


arthurp:
@ddn0 feel free to hand this review off or delay. This isn't a priority and I don't mind rebasing it.

gurbinder533 (approved):
LGTM.

libgalois/include/galois/analytics/Utils.h
```diff
@@ -13,57 +13,25 @@ namespace galois::analytics {
...
-    samples[trial] =
-        std::distance(graph.edges(node).begin(), graph.edges(node).end());
-    sample_total += samples[trial];
-  }
-  std::sort(samples.begin(), samples.end());
-  double sample_average = static_cast<double>(sample_total) / num_samples;
-  double sample_median = samples[num_samples / 2];
-  return sample_average / 1.3 > sample_median;
-}
+GALOIS_EXPORT bool isApproximateDegreeDistributionPowerLaw(
```
ddn0: `IsApproximate....`?

libgalois/include/galois/graphs/PropertyFileGraph.h
```diff
@@ -31,11 +38,51 @@ struct GraphTopology {
...
+    return MakeStandardRange<edge_iterator>(begin_edge, end_edge);
+  }
+
+  /**
+   * Gets the first edge of some node.
+   *
+   * @param node node to get the edge of
+   * @returns iterator to first edge of node
+   */
+  edge_iterator edge_begin(uint32_t node) const { return *edges(node).begin(); }
```
ddn0: shouldn't this also take a `const node_iterator&` instead of `uint32_t`?

libgalois/include/galois/graphs/PropertyFileGraph.h
```diff
@@ -352,6 +399,56 @@ class GALOIS_EXPORT PropertyFileGraph {
...
+    return topology().edges(node);
+  }
+
+  /**
+   * Gets the first edge of some node.
+   *
+   * @param node node to get the edge of
+   * @returns iterator to first edge of node
+   */
+  edge_iterator edge_begin(Node node) const {
```
ddn0: FWIW, my original intent was to keep the new graph interface relatively orthogonal with the existing uses of `edge_begin`/`edge_end` maintained for legacy/deprecation reasons.

I don't know if it that helpful to have so many different ways to do the same thing. `edge_begin` and `edge_end` were created before C++ had `for (T item : container)`. I think `edges` should suit most new use cases better.

arthurp: I killed `edge_{begin,end}` except on `PropertyGraph`.

arthurp:
I applied changes requested and added some TODO to deprecate some methods down the road.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -13,57 +13,25 @@ namespace katana::analytics {
 //! Used to pick random non-zero degree starting points for search algorithms
 //! This code has been copied from GAP benchmark suite
 //! (https://github.com/sbeamer/gapbs/blob/master/src/benchmark.h)
-template <typename Graph>
-class SourcePicker {
-  static const uint32_t kRandSeed;
+class KATANA_EXPORT SourcePicker {
```
ddn0, line 19: Not related to this PR and no need to fix now, but this is a good a time as any to give my yearly "`utils` is a horrible name" speech. Utility files/classes don't have a clear focus so they tend to just accumulate unrelated functionality and become a site of future circular dependencies.

Now, I'm off my soapbox.

Some of these functions might make more sense being declared closer to PropertyFileGraph. (And SourcePicker seems undergeneralized since the seed is just fixed.)

libgalois/src/analytics/Utils.cpp
```diff
@@ -0,0 +1,56 @@
...
+katana::analytics::IsApproximateDegreeDistributionPowerLaw(
+    const PropertyFileGraph& graph) {
+  if (graph.num_nodes() < 10)
+    return false;
+  uint32_t averageDegree = graph.num_edges() / graph.num_nodes();
+  if (averageDegree < 10)
+    return false;
+  SourcePicker sp(graph);
+  uint32_t num_samples = 1000;
+  if (num_samples > graph.num_nodes())
```
ddn0: nit: if-braces always

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -9,16 +9,26 @@
...
 
+#include "katana/Details.h"
 #include "katana/ErrorCode.h"
 #include "katana/LargeArray.h"
 #include "katana/config.h"
 #include "tsuba/RDG.h"
 
 namespace katana {
 
+class PropertyFileGraph;
```
ddn0: This forward declaration doesn't seem necessary...

ddn0 (approved):

libgalois/src/analytics/Utils.cpp
arthurp: Yep. I'll fix it. Looking forward though, can we get the formatter to fix this? It seems silly to do it manually.

ddn0: It's a `clang-tidy` option (but not a `clang-format` option as far as I know). There is a way to run `clang-tidy` in an automatically fix mode, but it is slow and not all the rewrite rules were actually correct or idempotent.

libgalois/include/katana/analytics/Utils.h
arthurp: I fixed the SourcePicker issue and added a TODO for the "utils" issue.

libgalois/src/analytics/Utils.cpp
arthurp: And I discovered clang-tidy can do this and isn't running automatically in my setup. NVM

PR #33, opened 2021-01-12
Deg count

ayaabdelgawad:
Added a module that will `calculate_degree()`, calculating the in and out degrees of all nodes in a given graph and add them as two new node properties

Also, fixed typo in documentation of python/katana/loops.py

### DIFF
--- a/python/katana/lonestar/analytics/calculate_degree.py
+++ b/python/katana/lonestar/analytics/calculate_degree.py
@@ -1,0 +1,62 @@
+import numpy as np
+import pyarrow
+
+from katana.property_graph import PropertyGraph
+from katana.datastructures import LargeArray, AllocationPolicy
+from katana.loops import do_all, do_all_operator
+from katana.atomic import atomic_add
+
+
+@do_all_operator()
+def initialize_in_degree(nin, nid):
+    nin[nid] = 0
+
+
+@do_all_operator()
+def count_in_and_out_degree(graph: PropertyGraph, nout, nin, nid):
+    out_degree = 0
+    for edge in graph.edges(nid):
+        out_degree += 1
+        dst = graph.get_edge_dst(edge)
+        atomic_add(nin, dst, 1)
+    nout[nid] = out_degree
+
+
+@do_all_operator()
+def count_weighted_in_and_out_degree(graph: PropertyGraph, nout, nin, weight_array, nid):
+    out_degree = 0
+    for edge in graph.edges(nid):
+        weight = weight_array[edge]
+        out_degree += weight
+        dst = graph.get_edge_dst(edge)
+        atomic_add(nin, dst, weight)
+    nout[nid] = out_degree
+
+
+def calculate_degree(graph: PropertyGraph, in_degree_property, out_degree_property, weight_property=None):
+    """
+    Calculate the (potentially weighted) in and out degrees of a graph.
+    The function will modify the given graph by adding two new node properties,
+    one for the in degree and one for the out degree. Nothing is returned.
+    Parameters:
+        graph: a PropertyGraph
+        in_degree_property: the property name for the in degree
+        out_degree_property: the property name for the out degree
+        weight_property: an edge property to use in calculating the weighted degree
+    """
+    num_nodes = graph.num_nodes()
+    nout = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+    nin = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+
+    do_all(range(num_nodes), initialize_in_degree(nin.as_numpy()), steal=False)
+
+    # are we calculating weighted degree?
+    if not weight_property:
+        count_operator = count_in_and_out_degree(graph, nout.as_numpy(), nin.as_numpy())
+    else:
+        count_operator = count_weighted_in_and_out_degree(
+            graph, nout.as_numpy(), nin.as_numpy(), graph.get_edge_property(weight_property)
+        )
+    do_all(range(num_nodes), count_operator, steal=True)
+
+    graph.add_node_property(pyarrow.table({in_degree_property: nin, out_degree_property: nout}))
--- a/python/katana/loops.py
+++ b/python/katana/loops.py
@@ -78,7 +78,7 @@ def for_each_operator(typ=None, nopython=True, **kws):
     >>> @for_each_operator()
     ... def f(arg0, ..., argn, element, ctx): ...
 
-    Decorator to declare an operator for use with a `do_all`.
+    Decorator to declare an operator for use with a `for_each`.
     The operators have some restructions; see below.
     If the operator has any arguments other than the element and context arguments expected from the loop, those
     arguments must be bound by calling the function to create a closure:


python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+        atomic_add(nin, dst, weight)
+    nout[nid] = outdeg
+
+""" 
+   Calculate the (potentially weighted) in & out degrees of a graph
+   Will add the in & out degrees as 2 new node properties of the graph
+   weight_property: an edge property to use in calculating the weighted degree
+   inProp: the property name for the in degree
+   outProp: the property name for the out degree
+"""
```
arthurp: Doc strings go inside the function. https://www.pythonforbeginners.com/basics/python-docstrings

Also please write in full sentences. The documentation will eventually show up to users.

python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+def calculate_degree(graph: PropertyGraph, weight_propery=None, inProp="in_degree", outProp="out_degree"):
+    num_nodes = graph.num_nodes()
+    nout = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+    nin = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+
+    do_all(
+        range(num_nodes),
+        initialize_in_deg(nin.as_numpy()),
+        steal=False #is this ok?
```
arthurp: Yes. That's OK. It might even make it slightly faster.

python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+    nout[nid] = outdeg
+
+""" 
+   Calculate the (potentially weighted) in & out degrees of a graph
+   Will add the in & out degrees as 2 new node properties of the graph
+   weight_property: an edge property to use in calculating the weighted degree
+   inProp: the property name for the in degree
+   outProp: the property name for the out degree
+"""
+def calculate_degree(graph: PropertyGraph, weight_propery=None, inProp="in_degree", outProp="out_degree"):
```
arthurp: `in_degree_property` and `out_degree_property`. Also move those two before `weight_property` so it's easier to leave unweighted while setting the output property names. I'm not sure I like having default output names. I could be talked into it, but since the user of the function will need to reference the property later, it's best for them to explicitly provide a name.

`propery` is misspelled.

python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+    else:
+        do_all(
+            range(num_nodes),
+            count_weighted_in_and_out_deg(graph, nout.as_numpy(), nin.as_numpy(), graph.get_edge_property(weight_propery)),
+            steal=True
+        )
+        
+
+    graph.add_node_property(pyarrow.table({inProp: nin}))
+    graph.add_node_property(pyarrow.table({outProp: nout}))
```
arthurp: Nit: Add them both at the same time. (By putting them both in the dict.)

python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+            range(num_nodes),
+            count_in_and_out_deg(graph, nout.as_numpy(), nin.as_numpy()),
+            steal=True
+        )
+    else:
+        do_all(
+            range(num_nodes),
+            count_weighted_in_and_out_deg(graph, nout.as_numpy(), nin.as_numpy(), graph.get_edge_property(weight_propery)),
+            steal=True
+        )
```
arthurp: Nit: Instead of writing the loop twice, store the operator in a variable.

```python
if weight_propery == None:
    calculate_operator = count_in_and_out_deg(graph, nout.as_numpy(), nin.as_numpy())
else:
    calculate_operator = count_weighted_in_and_out_deg(graph, nout.as_numpy(), nin.as_numpy(), graph.get_edge_property(weight_propery))
do_all(
        range(num_nodes),
        calculate_operator,
        steal=True
    )
```

python/katana/loops.py
```diff
@@ -78,7 +78,7 @@ def for_each_operator(typ=None, nopython=True, **kws):
     >>> @for_each_operator()
     ... def f(arg0, ..., argn, element, ctx): ...
 
-    Decorator to declare an operator for use with a `do_all`.
+    Decorator to declare an operator for use with a `for_each`.
```
arthurp, line 81: Good catch. Thanks.

python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+    nin = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+
+    do_all(
+        range(num_nodes),
+        initialize_in_deg(nin.as_numpy()),
+        steal=False #is this ok?
+    )
+
+    # are we calculating weighted degree?
+    if weight_propery == None:
```
arthurp: The python convention is to use the fact that `None` is treated as False, while any non-empty string is treated as True. The empty string is a terrible property name, so you can just write: `if weight_property:`

python/katana/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,68 @@
...
+def count_in_and_out_deg(graph: PropertyGraph, nout, nin, nid):
+    outdeg = 0
+    for edge in graph.edges(nid):
+        outdeg += 1
+        dst = graph.get_edge_dst(edge)
+        atomic_add(nin, dst, 1)
+    nout[nid] = outdeg
+
+@do_all_operator()
+def count_weighted_in_and_out_deg(graph: PropertyGraph, nout, nin, weightArr, nid):
```
arthurp: Use `weight_array` or just `weight`. It might also be good to use `in_degree` and `out_degree` for clarity.

arthurp (changes requested):
Overall this is great. Mostly I'm gonna ask for style changes. Functionally the code looks pretty good. The changes I'm asking for here should be quick since they are just small refactors (find and replace!!).

I'm pushing for clean and consistent. If you think I'm totally wrong about something feel free to push back. :-)

* All variables (incl parameters) should be "snail_case". Nothing in our Python style is "lowerCaseCamel". (Classes are "UpperCaseCamel".)
* Please expand `deg` to `degree` everywhere. Editors can autocomplete and its clearer to a reader of the code.
* The function is called `calculate_degree` so name the module the same thing.
* Please run the code formatter: `./scripts/check_python_format.sh -fix python/katana/lonestar/analytics/deg_count.py`

Well done!

python/katana/lonestar/analytics/calculate_degree.py
```diff
@@ -0,0 +1,62 @@
...
+        weight_property: an edge property to use in calculating the weighted degree
+    """
+    num_nodes = graph.num_nodes()
+    nout = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+    nin = LargeArray[np.uint64](num_nodes, AllocationPolicy.INTERLEAVED)
+
+    do_all(range(num_nodes), initialize_in_degree(nin.as_numpy()), steal=False)
+
+    # are we calculating weighted degree?
+    if not weight_property:
```
arthurp, line 54: nit: You could remove the not by swapping the code in the if and else branches.

arthurp (approved):
LGTM.

Well done! You even figured out our dumb scripts. :-)

arthurp:
The CI got stuck. I'm not sure why. I re-ran it from the actions tab and things seem to work now.

@gurbinder533, do you want to review as well?

@ayaabdelgawad, don't forget to squash all your commits when you merge (you can have github do it for you if you want). A single commit is appropriate (or two if you want to split out the doc change).

gurbinder533 (approved):
LGTM. Nice work.

PR #34, opened 2021-01-14
Small improvements to DynamicBitset

scober:
`bitwise_not` seems like a logical inclusion given that there is already `bitwise_{or,xor,and}`. And I added an assert to `DynamicBitset::test()` because I have been bitten by that before.

### DIFF
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -182,13 +182,14 @@ public:
 
   /**
    * Check a bit to see if it is currently set.
-   * Using this is recommeneded only if set() and reset()
+   * Using this is recommended only if set() and reset()
    * are not being used in that parallel section/phase
    *
    * @param index Bit to check to see if set
    * @returns true if index is set
    */
   bool test(size_t index) const {
+    assert(index < num_bits);
     size_t bit_index = index / bits_uint64;
     uint64_t bit_offset = 1;
     bit_offset <<= (index % bits_uint64);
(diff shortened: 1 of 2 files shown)


libgalois/include/katana/DynamicBitset.h
```diff
@@ -182,13 +182,14 @@ class KATANA_EXPORT DynamicBitset {
...
    * Check a bit to see if it is currently set.
-   * Using this is recommeneded only if set() and reset()
+   * Using this is recommended only if set() and reset()
    * are not being used in that parallel section/phase
    *
    * @param index Bit to check to see if set
    * @returns true if index is set
    */
   bool test(size_t index) const {
+    KATANA_ASSERT(index < num_bits);
```
roshandathathri: Why not just `assert()`? That would not be checked on release builds, which I think is desirable in this case.

roshandathathri (approved):
LGTM

libgalois/include/katana/DynamicBitset.h
scober: Does `KATANA_ASSERT` not get compiled out of release builds? I agree that this would be best as a debug-only check.

roshandathathri:
```
//! Like assert but unconditionally executed
#define KATANA_ASSERT(cond, ...)      
```
So yes, it is for release asserts.

scober: Whoops, thanks for catching that.

PR #39, opened 2021-01-18
Feature/out of core gr2kg

tylershunt:
graph-convert: support out-of-core GR->KG conversions

I do this by reading the GR header the performing a partial copy where all of the GR file except for the node data are copied into an RDG directory. The result is a valid topology file which is then connected to the RDG Metadata.

This is exposed in graph convert as a special case optimization for the case where the edge type is void.

```
$ graph-convert --edgeType=void -gr2kg \
      s3://example-graphs/road-europe.gr s3://new_graph/rdg
```

Caveats:
  * doesn't work across different back-ends (e.g. both source and dest must be in the local file system or in s3, gs etc).
  * When in s3 both source and dest have to be in the same region
  * Azure implementation is still TODO


libtsuba/include/tsuba/RDG.h
```diff
@@ -74,6 +74,11 @@ class KATANA_EXPORT RDG {
 
   katana::Result<void> UnbindTopologyFileStorage();
 
+  /// Inform this RDG that it's topology is in storage at this location
```
ddn0, line 78: "its", "graph"

ddn0: Is this register or really just updating the topology altogether?

E.g., is this any different than how we add and remove properties?

libtsuba/include/tsuba/file.h
```diff
@@ -39,6 +39,18 @@ KATANA_EXPORT katana::Result<void> FileStat(
 KATANA_EXPORT katana::Result<void> FileStore(
     const std::string& uri, const uint8_t* data, uint64_t size);
 
+/// Copy a slice of a file from one place to another, out-of-core if possible
```
ddn0: qq: the out-of-core-ness would suggest that there should be a max/target "working memory" parameter. No reason to add it now, but is that the correct model of how this operation should work?

libtsuba/include/tsuba/tsuba.h
```diff
@@ -51,6 +52,11 @@ KATANA_EXPORT katana::Result<RDGHandle> Open(
 KATANA_EXPORT katana::Result<RDGHandle> Open(
     const std::string& rdg_name, uint64_t version, uint32_t flags);
 
+/// Generate a new canonically named topology file name in the
+/// directory associated with handle. Exported to support
+/// out-of-core conversion
+KATANA_EXPORT katana::Uri NewTopologyFileName(RDGHandle handle);
```
ddn0: We have been calling these factory methods `MakeFoo` AFAIK

libtsuba/src/LocalStorage.cpp
```diff
@@ -50,6 +52,32 @@ tsuba::LocalStorage::WriteFile(
...
+katana::Result<void>
+tsuba::LocalStorage::CopyFile(
+    std::string source_uri, std::string dest_uri, uint64_t begin,
+    uint64_t size) {
+  CleanUri(&source_uri);
+  CleanUri(&dest_uri);
+
+  std::ifstream ifile(source_uri, std::ios_base::binary);
+  if (!ifile) {
+    KATANA_LOG_ERROR("failed to open soure file");
```
ddn0: "source"

tools/graph-convert/graph-convert.cpp
```diff
@@ -2868,18 +2875,90 @@ AppendEdgeData<void>(
...
     using Graph = katana::FileGraph;
     using GNode = Graph::GraphNode;
     using EdgeData = katana::LargeArray<EdgeTy>;
     using edge_value_type = typename EdgeData::value_type;
 
+    if constexpr (std::is_same<EdgeTy, void>::value) {
+      // the property graph topology file format is very close to gr, so we can
+      // use this shortcut. This shortcut also avoids reading the graph on one
+      // host so it's important to do this to support large graphs
+      fmt::print(stderr, "attempting out-of-core conversion\n");
```
ddn0, line 2949: for later: I'm wondering if we should add some callback/status API so that callers can implement some progress tracking / monitoring on expensive operations like partitioning and loading, e.g., `ExpensiveOperation(....., [](size_t n, size_t total) { print("did {n}th of {total}"); })`

libtsuba/include/tsuba/GR.h
```diff
@@ -0,0 +1,38 @@
+#ifndef KATANA_LIBTSUBA_TSUBA_GR_H_
```
ddn0: I think it might be better to name this "topology" something rather than "gr" as the latter is a little cryptic.

ddn0 (approved):

libtsuba/include/tsuba/file.h
tylershunt: In general, yes. Depending on copy location (and certainly for local filesystem copy) the data takes a round trip through memory. S3 is an interesting case since the API allows you to copy from one s3 object directly into another (one of the main reasons I didn't consider a working memory here)

libtsuba/include/tsuba/GR.h
```diff
@@ -0,0 +1,38 @@
...
+
+/// The metadata block at the head of every CSR file
+struct GRHeader {
+  uint64_t version{0};
+  uint64_t edge_type_size{0};
+  uint64_t num_nodes{0};
+  uint64_t num_edges{0};
+};
+
+/// The header and out index array of every CSR file the length of out_indexes
```
gurbinder533: `file. The length`

libtsuba/include/tsuba/GR.h
```diff
@@ -0,0 +1,38 @@
...
+  GRHeader header;
+  uint64_t out_indexes[];  // NOLINT needed for layout
+};
+
+constexpr uint64_t
+GRFileSize(const GRHeader header) {
+  uint64_t edge_size =
+      header.version == 1 ? sizeof(uint32_t) : sizeof(uint64_t);
+  return sizeof(header) + ((header.num_nodes - 1) * sizeof(uint64_t)) +
+         (header.num_edges * edge_size) +
```
gurbinder533: iirc, the gr files have padding as well: [](https://github.com/IntelligentSoftwareSystems/Galois/blob/master/libgalois/include/galois/graphs/LC_CSR_Graph.h#L925)

libtsuba/include/tsuba/file.h
ddn0: Maybe drop the reference to out of core then? At most we can allude to the fact that certain backends have remote-copy capabilities or even scope this API to just remote copies (i.e., same backend).

gurbinder533 (approved):
LGTM.

libtsuba/include/tsuba/GR.h
tylershunt: oh! nice catch; luckily since I never read the edge values the padding never matters; I'll update this to be more correct though.

libtsuba/include/tsuba/file.h
tylershunt: I think I'll re-name it `FileRemoteCopy` and update the comment with those restrictions.

libtsuba/include/tsuba/RDG.h
tylershunt: It is a little awkward. All of the existing operations for properties/topology say "here's data to be your new property/topology, write it out when we write out the next version". In this case I needed something that says "I've already written out the topology, here's where I stored it, update your metadata to reflect that"

ddn0: Maybe if we get more of these use-cases, one way to organize them is as "updating the metadata" versus "updating the data."

I think this will be more specific than "register". I don't know how important it is now though.

tylershunt:
Decided to add some common bit related functions to libsupport as well. (Mostly to support the fix for @gurbinder533 's gripe about padding.

libtsuba/include/tsuba/RDG.h
tylershunt: I renamed the function `SetTopologyFilename` which I think at least limits the scope of what users should expect from this function.

tools/graph-convert/graph-convert.cpp
tylershunt: Seems reasonable to me; an alternative would be to enrich the logs and tools for parsing them in real time with something like `prometheus`?

tylershunt: Or I guess call backs like this are general enough to be the mechanism that we would use to add prometheus/metrics support.

ddn0: Yeah; users can do whatever they want and the core library doesn't need to take on any specific logging library as a dependency.

If/when we go down this route, I have some preference for a pretty generic callback function: `ExpensiveOperation(..., [](FooContext ctx) { })` as that will give us some flexibility about the actual parameters passed and give the callee a handle to query the system if necessary.

PR #42, opened 2021-01-18
Average degree connectivity

pamvc:
This branch has the code for average_neighbor degree and average_degree connectivity.

Pushed the changes in the latest commit.
Average_neighbor_degree does not have any syntax errors, but the answers are not exactly the same as networkx's. Average_degree_connectivity has not been fully tested yet, because average_neighbor_degree took most of my time to debug but but I'm confident in my algorithm and with another set of eyes I think it could be finalized rather quickly.


python/galois/lonestar/analytics/deg_count.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
```
arthurp: This is already in master. Please rebase against master so it doesn't show up here.

python/galois/lonestar/analytics/average_degree_connectivity.py
```diff
@@ -0,0 +1,142 @@
...
+#average degree connectivity for degree k by calling 
+# avg_n_sum / value of key (degree) which the number of nodes with that degree
+#We store this in a result dictionary where
+#key = degree and value = degree's average degree connectivity
+
+
+#QUESTIONS: ask about correct way to pass parameters :) example: line 76
+#TODO: fix indentation
+
+#method to get the sum of the average neighbor degree for nodes with degree k
```
arthurp: Documentation for how a function is used (so what it's arguments should be and what it returns) should be in a doc string. Please move your function documentation comments into doc strings throughout.

Nit: Also, this is not actually a method. It's a function.

python/galois/lonestar/analytics/average_degree_connectivity.py
```diff
@@ -0,0 +1,142 @@
...
+            in_dict[deg] = newvalue
+
+        else: 
+            #add key and value 
+            #create an array with no finite size 
+            nid_array = []
+            #add nid to value array
+            nid_array.append(node_id)
+            #set nid array as this degrees value
+            in_dict[deg] = nid_array
```
arthurp: The branches of this if are nearly identical. Also there is a function that is very helpful for things like this [`dict.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault). You should be able to write:
```
nid_array = in_dict.setdefault(deg, []) 
newvalue.append(node_id)
```
No need to assign it back into the dict because it is added by `setdefault`.

python/galois/lonestar/analytics/average_degree_connectivity.py
```diff
@@ -0,0 +1,142 @@
...
+        else: 
+            #add key and value 
+            #create an array with no finite size
+            nid_array = []
+            #add nid to value array
+            nid_array.append(node_id)
+            #set nid array as this degrees value
+            out_dict[deg] = nid_array
+
+    return out_dict
```
arthurp: I think this is identical to `fill_graph_in_deg` other than the property name. Just have one version that takes the property name as an argument (or takes the array as an argument if you prefer).

python/galois/lonestar/analytics/average_degree_connectivity.py
```diff
@@ -0,0 +1,142 @@
...
+
+            #get value (array of NID's)
+            value = outDict.get(degree)
+            avg_degree_connectivity = avg_n_sum(value) / len(value)
+
+            #add key value pair
+            result_dict[degree] = avg_degree_connectivity
+
+    for key, value in result_dict.items():
+        print(key, ':', value)
```
arthurp: Return the result instead of printing it. If you need to print the value write a driver program that calls the function.

python/galois/lonestar/analytics/average_neighbor_degree.py
```diff
@@ -0,0 +1,86 @@
...
+
+    if source == "in" and target == "in":
+        deg_array = graph.get_node_property("in_degree_property")
+    else: 
+        deg_array = graph.get_node_property("out_degree_property")
+
+    result_array = helper(graph, deg_array, weight, weight_property)
+
+    for node_id, avg_neighbor in enumerate(result_array):
+        print(node_id, ' : ', avg_neighbor)
```
arthurp: Return the result instead of printing it. If you need to print the value write a driver program that calls the function.

python/galois/lonestar/analytics/average_neighbor_degree.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return result_array
+
+
+
+def average_neighbor_degree(graph: PropertyGraph, source, target, weight, weight_property=None):
+   
+    calculate_degree(graph, in_degree_property="in_degree_property", out_degree_property="out_degree_property", weight_property = weight)
+
+    if source == "in" and target == "in":
```
arthurp: If these two have to be the same, then you should only have one argument. Also instead of using string constants use an enum. If you have trouble ask @ayaabdelgawad. ;-)

python/galois/lonestar/analytics/average_degree_connectivity.py
```diff
@@ -0,0 +1,142 @@
...
+
+    #ask about correct way to pass in parameters :)
+    in_dict = fill_graph_in_deg(graph)
+    out_dict = fill_graph_out_deg(graph)
+
+    #create map that will hold the results 
+    #where key = deg and value = avg_degconn
+    result_dict = dict()
+
+    if source == "in" and target == "in":
```
arthurp: If these two have to be the same, then you should only have one argument. Also instead of using string constants use an enum. If you have trouble ask @ayaabdelgawad. ;-)

arthurp (changes requested):
Looking fairly good. But it needs some clean up.

* Please run `scripts/check_python_format.sh -fix [files]` on your code to fix up the formatting. You will need to install black: `conda install -c conda-forge black=19.10b0`
* Remove all the comments that are just development notes. There appear to be a lot that are out of date or meaningless to anyone else reading your code. Documenting what's going on it good. But no need for your questions to live on in the code.
* Use full words for things. So `get_avg_degconn` should be `get_average_degree_connectivity`. Also `get` usually implies that it's already computed and just need to be retrieved. So `average_degree_connectivity` would be better.

pamvc:
Pushed the changes in the latest commit. 
Average_neighbor_degree does not have any syntax errors, but the answers are not exactly the same as networkx's. Average_degree_connectivity has not been fully tested yet, because average_neighbor_degree took most of my time to debug but but I'm confident in my algorithm and with another set of eyes I think it could be finalized rather quickly.

PR #46, opened 2021-01-20
Average Degree Connectivity (from @pamvc)

arthurp:
This version does not produce correct results. I am taking ownership of this PR and will finish up work on it as time permits.

Replaces and Closes #42. This initial commit in this PR is a flattened version that PR.


arthurp:
This code doesn't really add anything to the project at this point. So I'm closing this PR. We can grab this code later if we ever have a specific use for it.

PR #48, opened 2021-01-21
Avoid weak vtables

ddn0:
If a virtual class has no out-of-line definitions, its vtable is emitted
in each translation unit it is referenced. RTTI tables follow the same
pattern so if we want consistent type info across users of a library
like libgalois, we should avoid having multiple vtables for the same
class.

Only clang has a diagnostic for this issue.

I also considered applying this throughout the codebase but it turns out
some external dependencies do not compile with -Wweak-vtable at all.

### DIFF
--- a/cmake/Modules/KatanaBuildCommon.cmake
+++ b/cmake/Modules/KatanaBuildCommon.cmake
@@ -259,6 +259,40 @@ if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING)
   find_package(benchmark REQUIRED)
 endif ()
 
+###### Common Functions ######
+
+function(set_common_katana_library_options)
+  set(no_value_options ALWAYS_SHARED)
+  set(one_value_options)
+  set(multi_value_options)
+
+  cmake_parse_arguments(X "${no_value_options}" "${one_value_options}" "${multi_value_options}" ${ARGN})
+
+  set(target ${X_UNPARSED_ARGUMENTS})
+  # Some careful defines and build directives to ensure things work on Windows
+  # (see config.h):
+  #
+  # 1. When we build our shared library, KATANA_EXPORT => dllexport
+  # 2. When someone uses our shared library, KATANA_EXPORT => dllimport
+  # 3. When we build a static library, KATANA_EXPORT => ""
+  #
+  # In the world of ELF, 1 and 2 can both be handled with visibility("default")
+  if (BUILD_SHARED_LIBS OR X_ALWAYS_SHARED)
+    target_compile_definitions(${target} PRIVATE KATANA_SHARED_LIB_BUILDING)
+  else ()
+    target_compile_definitions(${target} PRIVATE KATANA_STATIC_LIB)
+  endif()
+
+  # Having a single definition of a vtable has two benefits:
+  #
+  # 1. Reduces the size of intermediate object files
+  # 2. More importantly, gives canonical typeinfo pointers so that derived objects in
+  #    different libraries can resolve to the same base class typeinfo.
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    target_compile_options(${target} PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:-Wweak-vtables>")
+  endif ()
+endfunction ()
+
 ###### Test Inputs ######
 
 if (KATANA_GRAPH_LOCATION)
--- a/libgalois/include/katana/Executor_Deterministic.h
+++ b/libgalois/include/katana/Executor_Deterministic.h
@@ -91,7 +91,7 @@ template <typename OptionsTy>
 using DItem =
     DItemBase<typename OptionsTy::value_type, OptionsTy::useLocalState>;
 
-class FirstPassBase : public SimpleRuntimeContext {
+class KATANA_EXPORT FirstPassBase : public SimpleRuntimeContext {
 protected:
   bool firstPassFlag;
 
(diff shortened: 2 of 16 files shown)


cmake/Modules/KatanaBuildCommon.cmake
```diff
@@ -259,6 +259,40 @@ if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING)
...
 
+###### Common Functions ######
+
+function(set_common_katana_library_options)
+  set(options ALWAYS_SHARED)
+  set(one_value_args)
+  set(multi_value_args)
+
+  cmake_parse_arguments(X "${options}" "${one_value_args}" "${multi_value_args}" ${ARGN})
+
```
arthurp: `"${no_value_options}" "${single_value_options}" "${multi_value_options}"`?

libgalois/include/katana/Executor_Deterministic.h
```diff
@@ -91,23 +91,25 @@ template <typename OptionsTy>
...
 
-  void resetFirstPass(void) { firstPassFlag = false; }
+  void setFirstPass() { firstPassFlag = true; }
+
+  void resetFirstPass() { firstPassFlag = false; }
 
   virtual void alwaysAcquire(Lockable*, katana::MethodFlag) = 0;
 
-  virtual void subAcquire(Lockable* lockable, katana::MethodFlag f) {
+  void subAcquire(Lockable* lockable, katana::MethodFlag f) override {
```
arthurp, line 112: Moving from potentially dynamic polymorphism to only static polymorphism seems dangerous. Could you verify that all uses of this class are static? Maybe you could make the superclass private to prevent casts to the super type that could cause problems. (I don't know if this is practical, but it seems worth investigating.)

libgalois/include/katana/Executor_Deterministic.h
```diff
@@ -272,7 +278,7 @@ class DeterministicContextBase<OptionsTy, false, true>
       this->notReady = true;
   }
 
-  virtual void alwaysAcquire(Lockable* lockable, katana::MethodFlag m) {
+  void alwaysAcquire(Lockable* lockable, katana::MethodFlag m) override {
```
arthurp, line 281: Same dynamic to only static polymorphism issue.

libgalois/include/katana/Executor_Deterministic.h
```diff
@@ -192,7 +198,7 @@ class ReaderContext : public katana::UnionFindNode<ReaderContext>,
 
   bool propagate() { return this->find()->isReady(); }
 
-  virtual void alwaysAcquire(Lockable*, katana::MethodFlag) {
+  void alwaysAcquire(Lockable*, katana::MethodFlag) override {
     KATANA_DIE("unreachable");
```
arthurp: If this is unreachable then it probably shouldn't even be a method on this class. Can it be moved out of the base class into the "mutable contexts" branch of the heirarchy?

Also, `KATANA_LOG_FATAL`?

arthurp (approved):
I think I may have the same issues in Plan structures used by the new algorithms APIs (e.g., [libgalois/include/katana/analytics/connected_components/connected_components.h](https://github.com/KatanaGraph/katana/blob/ca8deaeb736c845125591ce764e096881720d3c1/libgalois/include/katana/analytics/connected_components/connected_components.h#L16)). Do those need fixing? It would be very reasonable to move the static constructor functions into the implementation file.

Other than the nits below, LGTM.

roshandathathri (approved):
LGTM

libgalois/include/katana/Executor_Deterministic.h
arthurp: I didn't know what `override` meant. NVM.

ddn0: I chatted with @arthurp offline and we have clarified the issue and it is not relevant here.

witchel (commented):
Can you explain this during a meeting?  I'd like to know when to explicitly `= default;` a destructor.

ddn0:
> I think I may have the same issues in Plan structures used by the new algorithms APIs (e.g., libgalois/include/katana/analytics/connected_components/connected_components.h). Do those need fixing? It would be very reasonable to move the static constructor functions into the implementation file.

The actual situations where this actually matters are rare though very cryptic to debug. (A current issue with enabling `visibility=hidden` on our macOS builds could stem from something like this.)

I think it is a good practice to have, relatively cheap (in code and runtime cost) to do correctly and there is a simple compiler diagnostic to notify us if we are doing it wrong, which is why I went ahead and anchored all our virtual classes.

Regarding `Plan`, if we aren't using virtual functions, I think it is less important to have canonical vtables/rtti because there isn't the same expectation about `dynamic_cast`s, etc.

arthurp:
> there is a simple compiler diagnostic to notify us if we are doing it wrong

You are referring to `-Wweak-vtable`?

BTW, I kinda think we should have a `KATANA_ADDITIONAL_CHECKS` build option that enables additional warning like this and maybe runs clang-tidy (cmake supports doing that). Ideally we could even pick and choose which checks it would run.

ddn0:
> You are referring to -Wweak-vtable?

Yup.

> BTW, I kinda think we should have a KATANA_ADDITIONAL_CHECKS build option that enables additional warning like this and maybe runs clang-tidy (cmake supports doing that). Ideally we could even pick and choose which checks it would run.

I briefly looked into running clang-tidy in CI. I stopped either because it was annoyingly slow or the effort to clean up legacy code (I forget which). I don't mind having it enabled as a CI pass if we can figure out a way to do it sanely.

I have found that the output of linting tools can be version dependent, so having some containerized/reproducible version for local development is a must-have.

Some another check that might be useful is include-what-you-use:
- https://cliutils.gitlab.io/modern-cmake/chapters/features/utilities.html

tylershunt (approved):
👍  for knowledge

arthurp:
> I briefly looked into running clang-tidy in CI. I stopped either because it was annoyingly slow or the effort to clean up legacy code (I forget which). I don't mind having it enabled as a CI pass if we can figure out a way to do it sanely.

I just meant a run-at-home option that would make it easier for people to run secondary checks.

ddn0:
> I just meant a run-at-home option that would make it easier for people to run secondary checks.

Depending on your IDE, you can get clang-tidy suggestions all the time: https://github.com/ddn0/dotfiles

arthurp:
> > I just meant a run-at-home option that would make it easier for people to run secondary checks.
> 
> Depending on your IDE, you can get clang-tidy suggestions all the time: https://github.com/ddn0/dotfiles

I was talking about collecting any and all tools or modes, in one place that's easy to run, that might make you ask for changes on the PR. ;-) I mention this because I would never have known about `-Wweak-vtable` unless I ended up on this PR, so I'm sure there are checkers I'm not aware of.

(I have setup tidy in my IDE.)

PR #50, opened 2021-01-21
macos-setup: only install if missing package

danielmawhirter:


ddn0 (approved):
I really have a love-hate relationship with homebrew

PR #53, opened 2021-01-26
Rename NodeProperty/NodeTable to node_property

ddn0:
Properties are tables but they are really a specific kind of table: one
where rows correspond to nodes or edges of a property graph. Match API
with this mental model.

I debated PropertyFileGraph::GetNodeProperty(int i) between
PropertyFileGraph::node_property(int i). On balance, I favored
GetNodeProperty(int i) because there is a corresponding
GetNodePropertyTyped method that is certainly not a field access and
thus isn't a good candidate to be named node_property_typed, and it
seemed reasonable to me to keep these two related accessors with the
same naming style.

### DIFF
--- a/libgalois/include/katana/PropertyFileGraph.h
+++ b/libgalois/include/katana/PropertyFileGraph.h
@@ -54,7 +54,7 @@ struct KATANA_EXPORT GraphTopology {
    * @param node an iterator pointing to the node to get the edge range of
    * @returns iterable edge range for node.
    */
-  edges_range edges(node_iterator node) const { return edges(*node); }
+  edges_range edges(const node_iterator& node) const { return edges(*node); }
   // TODO(amp): [[deprecated("use edges(Node node)")]]
 
   /**
@@ -118,10 +118,10 @@ public:
     std::shared_ptr<arrow::Schema> (PropertyFileGraph::*schema_fn)() const;
     std::shared_ptr<arrow::ChunkedArray> (PropertyFileGraph::*property_fn)(
         int i) const;
-    std::vector<std::shared_ptr<arrow::ChunkedArray>> (
+    const std::shared_ptr<arrow::Table>& (
         PropertyFileGraph::*properties_fn)() const;
     Result<void> (PropertyFileGraph::*add_properties_fn)(
-        const std::shared_ptr<arrow::Table>& table);
+        const std::shared_ptr<arrow::Table>& props);
     Result<void> (PropertyFileGraph::*remove_property_fn)(int i);
 
     std::shared_ptr<arrow::Schema> schema() const { return (g->*schema_fn)(); }
@@ -130,13 +130,13 @@ public:
       return (g->*property_fn)(i);
     }
 
-    std::vector<std::shared_ptr<arrow::ChunkedArray>> Properties() const {
+    const std::shared_ptr<arrow::Table>& properties() const {
       return (g->*properties_fn)();
     }
 
     Result<void> AddProperties(
-        const std::shared_ptr<arrow::Table>& table) const {
-      return (g->*add_properties_fn)(table);
+        const std::shared_ptr<arrow::Table>& props) const {
+      return (g->*add_properties_fn)(props);
     }
 
     Result<void> RemoveProperty(int i) const {
@@ -242,24 +242,24 @@ public:
   /// Determine if two PropertyFileGraphss are Equal
   bool Equals(const PropertyFileGraph* other) const {
     return topology().Equals(other->topology()) &&
-           rdg_.node_table()->Equals(*other->node_table()) &&
-           rdg_.edge_table()->Equals(*other->edge_table());
+           rdg_.node_properties()->Equals(*other->node_properties()) &&
+           rdg_.edge_properties()->Equals(*other->edge_properties());
   }
 
   std::shared_ptr<arrow::Schema> node_schema() const {
-    return rdg_.node_table()->schema();
+    return rdg_.node_properties()->schema();
   }
 
   std::shared_ptr<arrow::Schema> edge_schema() const {
-    return rdg_.edge_table()->schema();
+    return rdg_.edge_properties()->schema();
   }
 
-  std::shared_ptr<arrow::ChunkedArray> NodeProperty(int i) const {
-    return rdg_.node_table()->column(i);
+  std::shared_ptr<arrow::ChunkedArray> GetNodeProperty(int i) const {
+    return rdg_.node_properties()->column(i);
   }
 
-  std::shared_ptr<arrow::ChunkedArray> EdgeProperty(int i) const {
-    return rdg_.edge_table()->column(i);
+  std::shared_ptr<arrow::ChunkedArray> GetEdgeProperty(int i) const {
+    return rdg_.edge_properties()->column(i);
   }
 
   /**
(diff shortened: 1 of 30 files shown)


arthurp (approved):
Yay!! Thanks. LGTM.

The irony of changing the name of something away from "table" and at the same time making it's type `Table` is pretty intense. ;-)

ddn0:
> The irony of changing the name of something away from "table" and at the same time making it's type Table is pretty intense. ;-)

yeah. lol.

surprisingly common though.

Lifetime of a variable name:
1. Obviously name it after its type: `Thing thing;` 
2. Oh, there are going to be more of these, name it after its specific occurence: `Thing pocket_lint;`

arthurp:
> > The irony of changing the name of something away from "table" and at the same time making it's type Table is pretty intense. ;-)
> 
> yeah. lol.
> 
> surprisingly common though.
> 
> Lifetime of a variable name:
> 
>     1. Obviously name it after its type: `Thing thing;`
> 
>     2. Oh, there are going to be more of these, name it after its specific occurence: `Thing pocket_lint;`

3. Oh we need to specialize the object and there may be more, `PocketLint foo`

witchel (approved):
This is surely a victory, thank you.

Nice judicious use of props!

This whole init the upper limit idiom in for loops is freaking me out.

witchel:
This is probably not the right forum for these questions, but I have some pointed questions for PropertyFileGraph and since this PR is laying it bare, I thought I'd ask.

- Is it wise to export partition_metadata?  I mean, it makes sense for Distribution to see it, but allowing access to a raw structure (tsuba::PartitionMetadata) seems undisciplined.
- Same question for master_nodes and mirror_nodes.  It makes sense that Distribution and RDG both need to see this data, but why is there a public interface for it in PropertyFileGraph?
- Why is local_to_global_vector of type arrow::ChunkedArray?  Seems like a suboptimal choice for a vector intended for random access.
- The type system here is a bit different from the (re)partitioner.  Perhpas the timing is coming for everyone to enjoy LocalNodeIDs, and the other contents of PartitionTypes.h?

ddn0:
> 3. Oh we need to specialize the object and there may be more, `PocketLint foo`

4. We need to make it generic, `using PocketLint = Thing<Sizer, Placer>;` and the cycle begins again but now meta.

ddn0:
> Is it wise to export partition_metadata? I mean, it makes sense for Distribution to see it, but allowing access to a raw structure (tsuba::PartitionMetadata) seems undisciplined.
>

Not enough context for me to comment.

> Same question for master_nodes and mirror_nodes. It makes sense that Distribution and RDG both need to see this data, but why is there a public interface for it in PropertyFileGraph?
>

The presence of master and mirror nodes in PropertyFileGraph is a little surprising to me as well.

My mental model is:
- PropertyFileGraph: single machine graph with properties; may be annoying to use for applications directly but does capture all the data dependencies; main purpose: definition of what a "graph" physically is
- PropertyGraph: a convenience layer on top of PropertyFileGraph to make it easier to use for some applications
- currently unnamed entity: PropertyFileGraph + Distribution;
- DistGraph: convenience layer on top of PropertyFileGraph + Distribution to make it easier .....

I haven't been following the partitionable stuff closely, so some of the above thinking could be stale.

> Why is local_to_global_vector of type arrow::ChunkedArray? Seems like a suboptimal choice for a vector intended for random access.
>

Not enough context for me to comment.

> The type system here is a bit different from the (re)partitioner. Perhpas the timing is coming for everyone to enjoy LocalNodeIDs, and the other contents of PartitionTypes.h?

Sounds reasonable to me though we do have to make sure the dependencies flow in the same direction as PropertyFileGraph ideally shouldn't have a dependence on partitioning.

arthurp:
> The presence of master and mirror nodes in PropertyFileGraph is a little surprising to me as well.

When `PropertyFileGraph` was introduced in the summer (when I was an intern), it was actually supposed to become the distributed version of the graph with the local version just being the degenerate 1-node case. I guess that didn't happen.

> My mental model is:
> 
> * PropertyFileGraph: single machine graph with properties; may be annoying to use for applications directly but does capture all the data dependencies; main purpose: definition of what a "graph" physically is
> * PropertyGraph: a convenience layer on top of PropertyFileGraph to make it easier to use for some applications
> * : PropertyFileGraph + Distribution;
> * DistGraph: convenience layer on top of PropertyFileGraph + Distribution to make it easier .....

I approve of this mental model. I think we should make it true. Not in this PR though.

witchel:
> Sounds reasonable to me though we do have to make sure the dependencies flow in the same direction as PropertyFileGraph ideally shouldn't have a dependence on partitioning.

For sure.

I'm messing a bit with PropertyFileGraph, so let me ask you this question.  Take partitioned_metadata.  Distribution passes this to RDG via PropertyFileGraph.  What is a better way to do this?  I'm not crazy about friend classes or methods.

roshandathathri:
> My mental model is:
> 
> * PropertyFileGraph: single machine graph with properties; may be annoying to use for applications directly but does capture all the data dependencies; main purpose: definition of what a "graph" physically is
> * PropertyGraph: a convenience layer on top of PropertyFileGraph to make it easier to use for some applications
> * currently unnamed entity: PropertyFileGraph + Distribution;
> * DistGraph: convenience layer on top of PropertyFileGraph + Distribution to make it easier .....

I agree with this mental model but I don't think the "unnamed entity" is required.

ddn0:
> When PropertyFileGraph was introduced in the summer (when I was an intern), it was actually supposed to become the distributed version of the graph with the local version just being the degenerate 1-node case. I guess that didn't happen.

In my proposed mental model that would be PropertyFileGraph + degenerate Distribution.

> I'm messing a bit with PropertyFileGraph, so let me ask you this question. Take partitioned_metadata. Distribution passes this to RDG via PropertyFileGraph. What is a better way to do this? I'm not crazy about friend classes or methods.

If you are talking about a small tactical thing, I don't know. `friend` might be appropriate here to at least limit the blast radius of this somewhat awkward dependency.

Would it be possible to pull out `PropertyFileGraph` out of `Distribution`?

FWIW, `Distribution` itself was extracted from `DistGraph`. Before the partitioner and applications all just used `DistGraph` which I thought was pretty confusing because there were certain fields that were only relevant/used during partitioning.

witchel:
> My mental model is:
> 
> * PropertyFileGraph: single machine graph with properties; may be annoying to use for applications directly but does capture all the data dependencies; main purpose: definition of what a "graph" physically is
> * PropertyGraph: a convenience layer on top of PropertyFileGraph to make it easier to use for some applications
> * currently unnamed entity: PropertyFileGraph + Distribution;
> * DistGraph: convenience layer on top of PropertyFileGraph + Distribution to make it easier .....

Thanks for the discussion.

I guess it says something about me that I find DistGraph harder to use than PropertyFileGraph.  I don't find the template-based property model natural, but I'm not very good with templates.

Should I not use PropertyFileGraphs in the distributed case?  Because I'm currently doing that.

arthurp:
> Should I not use PropertyFileGraphs in the distributed case? Because I'm currently doing that.

As the master of the Python wrapper. PropertyFileGraph is good. It's easy to wrap. And in general don't use templates in anything that could be considered and API (for instance the signature of an algorithms entry point). Templates are a pain to wrap. That's why PropertyFileGraph is underlying implementation with templatized version available for C++ algorithms to use internally.

ddn0:
> I guess it says something about me that I find DistGraph harder to use than PropertyFileGraph. I don't find the template-based property model natural, but I'm not very good with templates.

> As the master of the Python wrapper. PropertyFileGraph is good.

Agreed. I don't think `DistGraph` is for everyone, and for library code especially, PFG (+ Distribution) is more appropriate IMHO.

roshandathathri:
> I don't think `DistGraph` is for everyone, and for library code especially, PFG (+ Distribution) is more appropriate IMHO.

Now I see the need for the unnamed entity. `DistFileGraph` may be a good name for it.

roshandathathri:
> Should I not use PropertyFileGraphs in the distributed case? Because I'm currently doing that.

Yes, I think it makes sense to restrict PFG and PG for single machine use only.

arthurp:
> Yes, I think it makes sense to restrict PFG and PG for single machine use only.

If this is the case then we should remove any features from then related to distributed execution, like the master node information. Those should be moved to the Distribution, maybe. The RDG still needs to handle them, but the Distribution can directly use the RDG.

roshandathathri:
> If this is the case then we should remove any features from then related to distributed execution, like the master node information. Those should be moved to the Distribution, maybe. The RDG still needs to handle them, but the Distribution can directly use the RDG.

Agree but before we do that, we need to add an untemplated `DistFileGraph` class

arthurp:
Speaking of the naming of classes, why is "File" the marker of a non-templated version? It seems like a weird artifact of the past. I would prefer something like GenericPropertyGraph and GenericDistributedGraph. Though "Generic" doesn't feel right either. Thoughts?

roshandathathri:
> Speaking of the naming of classes, why is "File" the marker of a non-templated version? It seems like a weird artifact of the past. I would prefer something like GenericPropertyGraph and GenericDistributedGraph. Though "Generic" doesn't feel right either. Thoughts?

Good point. I'm fine with "Generic". Another alternative is to drop "Generic" or "File" from untemplated classes and instead add "Typed" or something similar to the templated classes.

ddn0:
New issue for the discussion about renaming PropertyGraph: https://github.com/KatanaGraph/katana/issues/54

roshandathathri (approved):
LGTM

PR #55, opened 2021-01-29
PropertyFileGraph improved encapsulation

witchel:
This PR is mostly a grab bag of features that I need for log processing in the enterprise repo.

o In PropertyFileGraph I've done the following.
  - Moved partition_metadata, master_nodes, and mirror_nodes out of the public interface.
  - ...and made Distribution a friend
  - Added node iterators and created public interfaces to get masters, outgoing mirrors, and incoming mirrors
  - Some const correctness (e.g., local_to_global_vector)

o Handle nulls property moving between Arrow and ImportData
  - Ongoing discussions with Daniel about moving the log code away from ImportData.

o _(Moved to enterprise repo because it requires DistMemSys)_ Provide a graph-info tool.  I couldn't get any of the tools in the tools directory to work for me, so I wrote a simple and not very performant program to print out some graph metadata.  You can see sample output below.  This might be a case of, "Hey Emmett, I don't think anyone will enjoy these, so maybe don't leave your snacks in the break room."

```
URI              : ../../graphs/ldbc01-2
Nodes            :  0 :  226389
Edges            :  0 :  738984
Masters          :  0 :   99668
Outgoing mirrors :  0 :       0
Incoming mirrors :  0 :  126721
Nodes            :  1 :  239787
Edges            :  1 :  738981
Masters          :  1 :  227920
Outgoing mirrors :  1 :       0
Incoming mirrors :  1 :   11867
Node Properties  : 29
birthday       :int64    browserUsed    :utf8     City           :uint8    Comment        :uint8   
Company        :uint8    content        :utf8     Continent      :uint8    Country        :uint8   
creationDate   :int64    firstName      :utf8     Forum          :uint8    gender         :utf8    
id             :int64    imageFile      :utf8     language       :utf8     lastName       :utf8    
length         :int32    locationIP     :utf8     Message        :uint8    name           :utf8    
Organisation   :uint8    Person         :uint8    Place          :uint8    Post           :uint8   
Tag            :uint8    TagClass       :uint8    title          :utf8     University     :uint8   
url            :utf8    
Edge Properties  : 19
classYear      :int32    CONTAINER_OF   :uint8    creationDate   :int64    HAS_CREATOR    :uint8   
HAS_INTEREST   :uint8    HAS_MEMBER     :uint8    HAS_MODERATOR  :uint8    HAS_TAG        :uint8   
HAS_TYPE       :uint8    IS_LOCATED_IN  :uint8    IS_PART_OF     :uint8    IS_SUBCLASS_OF :uint8   
joinDate       :int64    KNOWS          :uint8    LIKES          :uint8    REPLY_OF       :uint8   
STUDY_AT       :uint8    workFrom       :int32    WORK_AT        :uint8   
```


libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -186,34 +227,28 @@ class KATANA_EXPORT PropertyFileGraph {
...
+    auto pm = rdg_.part_metadata();
+    return topology_.nodes(pm.num_owned_, pm.num_nodes_with_edges_);
+  }
+  GraphTopology::nodes_range incoming_mirrors() const {
+    auto pm = rdg_.part_metadata();
+    return topology_.nodes(pm.num_nodes_with_edges_, pm.num_nodes_);
   }
 
-  const std::shared_ptr<arrow::ChunkedArray>& local_to_global_vector() {
+  // ChunkedArray seems like a bad choice for a vector intended for random access
```
ddn0: perhaps remove "seems" or add something that is more indicative of "yes, this is bad; it should probably change, but I haven't gotten to it", e.g.,

TODO(witchel): ChunkedArray is inherited from arrow::Table interface but this is really a ChunkedArray of one chunk, change to arrow::Array.

(assuming that this is the reason it is a ChunkedArray to begin with)

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -254,11 +291,17 @@ class KATANA_EXPORT PropertyFileGraph {
...
+  // num_rows() == num_nodes() (all local nodes)
   std::shared_ptr<arrow::ChunkedArray> GetNodeProperty(int i) const {
+    if (i >= rdg_.node_properties()->num_columns())
+      return nullptr;
     return rdg_.node_properties()->column(i);
   }
 
+  // num_rows() == num_edges() (all local edges)
   std::shared_ptr<arrow::ChunkedArray> GetEdgeProperty(int i) const {
+    if (i >= rdg_.edge_properties()->num_columns())
```
ddn0: nit: braces around if-clauses

tools/graph-info/graph-info.cpp
```diff
@@ -0,0 +1,147 @@
...
+      fmt::print("{:16} : {}\n", "Version", prefix.version());
+    }
+  } else {
+    PrintDist(src_uri, this_host, num_hosts);
+  }
+  // dist_mem_sys->Fini()  prints Stat output, so do our own shutdown
+  if (auto fini_good = tsuba::Fini(); !fini_good) {
+    KATANA_LOG_ERROR("tsuba::Fini: {}", fini_good.error());
+  }
+  katana::internal::destroySystemNetworkInterface();
```
ddn0: - destroying `dist_mem_sys` should handle this case for you too.
- I'm missing why you wouldn't just call `dist_mem_sys->Fini()` directly as it is supposed to handle the cleanup of everything it initializes (tsuba, katana::getSystemNetworkInterface).

libgalois/include/katana/PropertyFileGraph.h
witchel: Right on.  @roshandathathri , is there any reason to keep this as a chunked array?

tools/graph-info/graph-info.cpp
witchel: I wrote a comment.
`  // dist_mem_sys->Fini()  prints Stat output, so do our own shutdown`
Maybe I'm too sensitive to aesthetics, but the banner output from the Stats printing was confusing and untidy.

ddn0: I see. I saw the comment but I didn't realize it was related to omitting Stats printing altogether.

My suggestion is to file an issue to remove/make optional Stats printing throughout. And either use `dist_mem_sys->Fini()` and deal with the ugly output in the interim or keep the code referencing internal APIs and be prepared to fix things if there is a change to initialization behavior.

witchel: Ok, that makes sense.

witchel: I added an optional flag to Fini for stats printing.  I know Boolean parameters are often not ideal.  I made it optional, which I hope does not violate our coding standards.

witchel:
Hmm, I put graph-info in external/katana/tools because it seemed like it would rely on open APIs.  But it uses DistMemSys, which seems to be enterprise only.  So I guess I should move it to tools in enterprise?

libgalois/include/katana/PropertyFileGraph.h
roshandathathri: No, let's change it to `arrow::Array`. Thanks!

witchel: Will do, but in another PR.  Issue #57 .

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -109,6 +114,42 @@ class KATANA_EXPORT PropertyFileGraph {
...
+  }
+  void set_partition_metadata(const tsuba::PartitionMetadata& meta) {
+    rdg_.set_part_metadata(meta);
+  }
+
+  /// Per-host vector of master nodes
+  ///
+  /// master_nodes()[this_host].empty() is true
+  /// master_nodes()[host_i][x] contains LocalNodeID for my master and
+  ///   host_i has a mirror
```
roshandathathri: "contains LocalNodeID for my masters"

nit: "masters that have a mirror on host_i" or "masters for which host_i has a mirror"

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -109,6 +114,42 @@ class KATANA_EXPORT PropertyFileGraph {
...
+  const std::vector<std::shared_ptr<arrow::ChunkedArray>>& master_nodes()
```
roshandathathri, line 132: Don't see a reason for this to be `ChunkedArray`

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -109,6 +114,42 @@ class KATANA_EXPORT PropertyFileGraph {
...
+  }
+  void set_master_nodes(std::vector<std::shared_ptr<arrow::ChunkedArray>>&& a) {
+    rdg_.set_master_nodes(std::move(a));
+  }
+
+  /// Per-host vector of mirror nodes
+  ///
+  /// mirror_nodes()[this_host].empty() is true
+  /// mirror_nodes()[host_i][x] contains LocalNodeID for my mirrors and
+  ///   host_i has the master
```
roshandathathri: same nit: "mirrors that have the master on host_i" or "mirrors for which host_i has the master"

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -109,6 +114,42 @@ class KATANA_EXPORT PropertyFileGraph {
...
+  const std::vector<std::shared_ptr<arrow::ChunkedArray>>& mirror_nodes()
```
roshandathathri, line 145: Don't see a reason for this to be `ChunkedArray`

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -240,6 +276,8 @@ class KATANA_EXPORT PropertyFileGraph {
   }
 
   /// Determine if two PropertyFileGraphss are Equal
+  // NB: It might be useful to have a version where property tables
+  // with permuted columns compare as equal
```
roshandathathri: I did not catch the drift of this. What kind of permutations?

Also, what's NB?

roshandathathri (approved):
LGTM

libgalois/include/katana/PropertyFileGraph.h
ddn0: nota bene: Latin for "please note"

ddn0: Though, most of the time, I find you can basically drop "NB" and keep the same intent without adding a dependency on Latin :p

roshandathathri: Haha, yes :)

witchel: But the NB enabled this delightful discussion.

My point about permutations is this.  Let's say I have two node properties, A and B.  In the `node_properties` table, column A comes first.  If I update an A property, writing the table requires removing the column, updating it, and adding it back to the table.  Now, column B comes first.  Even if I change the values in A to be what they were originally, Arrow won't compare the two node property tables as equal.

I propose listing all the columns by name and checking them individually for equality.  I think that is a better notion of `Equal` than what we have now (arrow::Array::Equal).  I know I would use it.  Maybe @arthurp would.

libgalois/include/katana/PropertyFileGraph.h
witchel: Nice one.  This stuff is delicate.

libgalois/include/katana/PropertyFileGraph.h
witchel: And again.

libgalois/include/katana/PropertyFileGraph.h
roshandathathri: I understand now. Yes, I agree with your proposal. That makes more sense.

libgalois/include/katana/PropertyFileGraph.h
```diff
@@ -419,6 +471,8 @@ class KATANA_EXPORT PropertyFileGraph {
 
   bool empty() const { return topology().empty(); }
 
+  // NB: num_nodes in repartitioner is of type LocalNodeID
+  /// Return the number of local nodes and edges
```
arthurp: I think this note also applies to size. So it should be there too. Also please make this into a doc comment. It's pretty important for API users.

libgalois/include/katana/PropertyFileGraph.h
witchel: I have added this functionality to this PR.

libgalois/include/katana/PropertyFileGraph.h
witchel: I flipped the lines and made the comment about num_nodes a doc comment.  At no additional charge, I removed the NB: prefix that I so dearly love (but others find confusing).  I'm not sure what "So it should be there too" means.

arthurp: The `size` method returns the same value as `num_nodes`. So it too will be only local nodes in some cases. that's all I meant. The way you rewrote the comments makes me think I may have misunderstood in the first though though.

In the distributed case, is `num_nodes` (and `size`) the total number of global nodes in the graph, or is it the number of local nodes present in the current process?

witchel: So you want me to add a comment to PropertyFileGraph::size that it only returns the number of local nodes.  I didn't notice that size returns num_nodes.  I can add this.

The comment is trying to say that num_nodes returns the number of local nodes and edges.  That is not the number of global nodes in the distributed case.  Do you have an improved way to phrase the comment?

arthurp: "Returns the number of local nodes" is just fine. Has it said that from the beginning? If so, sorry. I just failed to read it right and none of this was an issue.

Just a nit, but doc comments only associate with the function below them, so if you want the documentation to come up in IDEs (e.g., when you mouse over a method name) you need to have a separate comment on each function (i.e., `num_nodes` and `num_edges`).

witchel: Fair enough.  I have improved the comments again.

PR #56, opened 2021-01-29
Make setup_ubuntu.sh work on 16.04.

arthurp:
Limit package versions based on Python version.
Only use apt-add-repository --no-update when available.
Also, avoid argument splitting in run_as_original_user.

This does **not** fix actually building Katana on 16.04. It is still missing a GCC 7 (it has GCC 6 max). @bosconi you are welcome to add those fixes on this PR (yes, you can push to my repo, the PR existing allows it), or as a separate PR.

### DIFF
--- a/scripts/setup_ubuntu.sh
+++ b/scripts/setup_ubuntu.sh
@@ -21,10 +21,14 @@ for arg in "$@"; do
 done
 
 run_as_original_user() {
+  # The second "$@" is expanded in this script and becomes the arguments to a
+  # bash script `"$@"` that simply executes its command-line arguments. This
+  # protects characters like < from being interpreted as redirections if $@
+  # were expanded directly.
   if [[ -n "${ORIGINAL_USER}" ]]; then
-    su - "${ORIGINAL_USER}" bash -c "$*"
+    su - "${ORIGINAL_USER}" bash -c '"$@"' -- "$@"
   else
-    bash -c "$*"
+    bash -c '"$@"' -- "$@"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/setup_ubuntu.sh
```diff
@@ -22,9 +22,9 @@ done
 
 run_as_original_user() {
   if [[ -n "${ORIGINAL_USER}" ]]; then
-    su - "${ORIGINAL_USER}" bash -c "$*"
+    su - "${ORIGINAL_USER}" bash -c '"$@"' -- "$@"
   else
-    bash -c "$*"
+    bash -c '"$@"' -- "$@"
```
ddn0, line 31: I need a shell primer: what does repeating `$@` do?

arthurp: The script running in the subshell is `"$@"` which expands to the arguments the script receives without splitting or interpreting the arguments any further. Bash will try to run the whole set of arguments as a command line with the first argument used as the command. The arguments to this script are `"$@"` but in the function scope. These expand to all the arguments without splitting (as above) and pass those as arguments to the script. This means that no splitting will happen after the arguments are split for the call to the function.

I made this fix because I needed to pass `"pip<=20.3"` (or whatever) as an argument to a function called via `run_as_original_user` and the nested shell was splitting the the argument and interpreting the `<` as an input redirect.

arthurp: Basically I was implementing what in C++ they might call "perfect forwarding". Passing through the arguments to a subshell without any change.

This isn't an idiom I found somewhere, by the way. There may be a better way to do it. I just thought this up in the moment.

ddn0: Ah. I had to play around with an example to understand.

The second "$@" is expanded in the script and becomes arguments to a simple "trampoline" script (`bash -c '"$@"'`) that simply executes its command line arguments.

Do you think a comment like:
```bash
# The second "$@" is evaluated in this script and becomes the argument to a bash script `bash -c '"$@"'`
# that simply executes its command-line arguments. This protects characters like < from being interpreted
# as pipe redirections if $@ were stringified directly.
```

would be useful?

arthurp: Sure. I added a version of this that is modified to use the actual shell terminology (to help people search for the ideas). Take a look.

bosconi (commented):
Thanks for this. Looks good. I'll make sure we loop back once 16.04 support has left the building, so we can drop the Python 3.6 support.

bosconi (approved):
Once more with the right box checked.

PR #59, opened 2021-02-01
Fix SSSP infinity values.

arthurp:
Hopefully these will also clean up the overflow error in #55.

### DIFF
--- a/libgalois/include/katana/analytics/BfsSsspImplementationBase.h
+++ b/libgalois/include/katana/analytics/BfsSsspImplementationBase.h
@@ -37,6 +37,7 @@ struct BfsSsspImplementationBase {
 
   constexpr static const bool kTrackWork = false;
 
+  // TODO(amp): Remove / 4
   constexpr static const Dist kDistanceInfinity =
       std::numeric_limits<Dist>::max() / 4;
 
--- a/python/katana/lonestar/analytics/sssp.py
+++ b/python/katana/lonestar/analytics/sssp.py
@@ -18,13 +18,19 @@ from katana.timer import StatTimer
 from katana.galois import setActiveThreads
 
 
-UpdateRequest = np.dtype([("src", np.uint64), ("dist", np.uint32),])
+def dtype_info(t):
+    t = np.dtype(t)
+    if t.kind == "f":
+        return np.finfo(t)
+    else:
+        return np.iinfo(t)
 
 
 def create_distance_array(g: PropertyGraph, source, length_property):
-    inf_distance = numba.types.uint64.maxval
     a = np.empty(len(g), dtype=dtype_of_pyarrow_array(g.get_edge_property(length_property)))
-    a[:] = inf_distance
+    # TODO(amp): Remove / 4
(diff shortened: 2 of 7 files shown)


python/katana/lonestar/analytics/sssp.py
```diff
@@ -18,13 +18,25 @@
...
+    else:
+        return np.iinfo(t)
 
 
 def create_distance_array(g: PropertyGraph, source, length_property):
-    inf_distance = numba.types.uint64.maxval
     a = np.empty(len(g), dtype=dtype_of_pyarrow_array(g.get_edge_property(length_property)))
-    a[:] = inf_distance
+    a[:] = dtype_info(a.dtype).max
     a[source] = 0
```
witchel, line 34: Just noting that `inf_distance `makes some intuitive sense where `dtype_info(a.dtype).max` does not.

witchel (approved):
I'm out of my depth in this code, but if this unblocks #55, I'm all for it.

python/katana/lonestar/analytics/sssp.py
```diff
@@ -18,13 +18,25 @@
 from katana.galois import setActiveThreads
 
 
-UpdateRequest = np.dtype([("src", np.uint64), ("dist", np.uint32),])
+UpdateRequest = np.dtype(
+    [
+        ("src", np.uint64),
```
gurbinder533: "src" is nodeId, right? Shouldn't it be uint32?

libgalois/include/katana/analytics/BfsSsspImplementationBase.h
```diff
@@ -38,7 +38,7 @@ struct BfsSsspImplementationBase {
   constexpr static const bool kTrackWork = false;
 
   constexpr static const Dist kDistanceInfinity =
-      std::numeric_limits<Dist>::max() / 4;
+      std::numeric_limits<Dist>::max();
```
gurbinder533: The same value of infinity should be used for bfs as well, right?

gurbinder533 (approved):
LGTM. Minor comments.

libgalois/include/katana/analytics/BfsSsspImplementationBase.h
arthurp: It shouldn't matter. But I'll change BFS too. The same weird `max/4` pattern appears in other applications in both the CPU and distributed lonestar suites. I have not tried to fix them in general because I don't know why this number was chosen.

gurbinder533: Okay. Just use the same value for shared-memory bfs and sssp and we can fix others when we encounter them.

arthurp: So I had to roll back all the `/4` changes because it caused the tests to fail because the check data contains this constant and I didn't want to to make such a tiny change to the inputs on short notice. @witchel is waiting on this PR. I'll create a new PR to fix things like this in a bit. Is that OK?

arthurp:
These commits were merged as part of #55. Closing.

PR #63, opened 2021-02-03
support: add ViewCast arrow utility

tylershunt:
Wraps up the common operation of checking an array's type then applying
a cast to a compatible type in order to access its contents.


witchel (approved):
Sweet!

PR #65, opened 2021-02-04
PropertyGraph and Arrow tweaks.  Introduce Time.h

witchel:
Once again, open repo changes in service of an enterprise checkin.  It includes some PropertyGraph improvements that should be non-controversial.  One reason I moved some code into the .cpp file was to cut down on build time.

I did add some Arrow utility code into ArrowInterchange.h.  It might make sense to have a dedicated file, but I say let's wait until there is critical mass.

Do you think now is a good time to revamp the benchmark timing code and promote it to libsupport?  Neither did I, but I had to measure some time and I just had a cup of coffee and then Time.h happened.  At least it is more C++-y than my code from last June.  Who was that guy?


witchel:
Anyone know why 18.04, Debug, g++-9 doesn't like `#include <arrow/stl.h>`?
https://github.com/KatanaGraph/katana/pull/65/checks?check_run_id=1828146984#step:10:16

ddn0:
> Anyone know why 18.04, Debug, g++-9 doesn't like #include <arrow/stl.h>?

I think it is a dependency thing:
- libtsuba depends on libarrow
- libtsuba depends on libsupport

Until now, libsupport has not depended on libarrow directly.

witchel:
> > Anyone know why 18.04, Debug, g++-9 doesn't like #include <arrow/stl.h>?
> 
> I think it is a dependency thing:
> 
> * libtsuba depends on libarrow
> * libtsuba depends on libsupport
> 
> Until now, libsupport has not depended on libarrow directly.

Yep, thanks. I realized this after I wrote the comment and added the cmake dependence to libsupport.

witchel:
Hate to nudge, but with #812 passing CI, I'd like to commit both of these some time kinda soon.

libsupport/CMakeLists.txt
```diff
@@ -32,6 +33,18 @@ target_include_directories(katana_support PUBLIC
 
 set_common_katana_library_options(katana_support)
 
+find_package(Arrow REQUIRED)
```
ddn0, line 36: Can we remove the corresponding arrow configuration in `libtsuba`?

libsupport/include/katana/ArrowInterchange.h
```diff
@@ -276,6 +276,110 @@ TableBuilder::AddColumn(const ColumnOptions& options) {
...
+        }
+      }
+      return;
+    }
+  }
+}
+
+template <typename ArrowArrayType>
+void
+PrintCmpElts(
```
ddn0: "Elements"

"Cmp"?

Given the rather precise formatting used in this function, it is hard for me to see how this function or the above would be generally reusable. Would it be possible to just keep this in the utility and wait to see if someone else really wants to share/reuse this?

ddn0 (commented):
cc @danielmawhirter on arrow stuff

libsupport/include/katana/ArrowInterchange.h
witchel: Sure thing.  I realize it was a stretch, but these functions answer the "what is in this array" and "when these two arrays aren't equal, why not?" questions.  That came up multiple times in multiple files for me, but I'll take my snacks out of the break room.

libsupport/CMakeLists.txt
witchel: I'll take the `arrow::arrow` out of libtsuba.  It still needs `arrow::parquet`.

witchel: I think this is the minimal change, but I'm not sure about the `else `case.

ddn0: On reflection, redundant import of both arrow and parquet in both libraries might not be the end of the world either (since parquet and arrow go together).

The `else` case looks okay to me.

ddn0 (approved):

libsupport/CMakeLists.txt
witchel: Cool.  I'm happy to leave it like it is, or add back arrow and parquet in both places.

libsupport/src/ArrowInterchange.cpp
```diff
@@ -0,0 +1,59 @@
...
+  auto res = builder.Reserve(length);
+  KATANA_LOG_ASSERT(res.ok());
+  for (int64_t i = 0; i < length; ++i) {
+    builder.UnsafeAppend(indices_vec[i]);
+  }
+  std::shared_ptr<arrow::Array> indices;
+  res = builder.Finish(&indices);
+  KATANA_LOG_ASSERT(res.ok());
+  return IndexedTake(original, indices);
+}
```
danielmawhirter, line 59: Glad to see these are useful! Philosophical question. These should only ever fail with OOM, should such instances be wrapped in Result, or using Assert? I imagine it's a case where tearing everything down is a reasonable behavior?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -427,7 +423,8 @@ class KATANA_EXPORT PropertyGraph {
         .property_fn = &PropertyGraph::GetEdgeProperty,
         .properties_fn = &PropertyGraph::edge_properties,
         .add_properties_fn = &PropertyGraph::AddEdgeProperties,
-        .remove_property_fn = &PropertyGraph::RemoveEdgeProperty,
+        .remove_property_int = &PropertyGraph::RemoveEdgeProperty,
+        .remove_property_str = &PropertyGraph::RemoveEdgeProperty,
     };
   }
```
danielmawhirter, line 430: Naive question, is the Edge vs Node property view the primary reason this interface uses function pointers?

libsupport/src/ArrowInterchange.cpp
```diff
@@ -0,0 +1,59 @@
...
+  KATANA_LOG_ASSERT(res.ok());
+  for (int64_t i = 0; i < length; ++i) {
+    builder.UnsafeAppend(i);
+  }
+  std::shared_ptr<arrow::Array> indices;
+  res = builder.Finish(&indices);
+  KATANA_LOG_ASSERT(res.ok());
+  auto chunked = IndexedTake(original, indices);
+  return chunked->chunk(0);
+}
```
danielmawhirter, line 39: I'm imagining cases where it would be desirable to Unchunk multiple arrays, in which case caching the indices would be valuable. No change needed now

danielmawhirter (approved):

libsupport/src/ArrowInterchange.cpp
witchel: Thank you for the code!  I think there are many parts of our code base that could fail if memory is exhausted.  I think it is acceptable to die from an exception in that case then litter the code with checks.  If the OS kills us because of OOM, I don't know if we even get a signal.

libgalois/include/katana/PropertyGraph.h
witchel: These aren't so necessary when you are just manipulating an arrow table.  But when you want to write a routine that modifies `Properties`, these allow you to write a single routine that you can use for both node and edge properties.

libsupport/src/ArrowInterchange.cpp
witchel: Sure that sounds cool.  When you find an application for that, please structure this code to make it fit.

tylershunt (approved):
LGTM

libsupport/CMakeLists.txt
witchel: Sigh.  It took @tylershunt to realize that I can't have libsupport depend on arrow and libtsuba depend on parquet.  Now, libsupport depends on both libtsuba and parquet.  I think that'll do it.

PR #67, opened 2021-02-06
Refactor python tests and build

arthurp:
This PR removes some of the cmake weirdness used in the previous build system and allows the python build to happen as part of the normal cmake build. In fact, the python tests will now run as part of the ctest run as well.


arthurp:
Important questions:

* Should "Conda Package and Python Formatting / build_and_package" be kept or disabled? It takes a really long time (40-60 minutes). 
* Is disabling that test ok? (22980eb)
* Is the "fake out of tree" build of the python packages OK? It builds a symlink tree (real directories, but full of symlinks instead of copies) and builds there, to avoid polluting the source tree.

.github/workflows/python.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
 #        access_token: ${{ github.token }}
 
   build_and_package:
+#    name: Build and Test Conda Packages
```
ddn0: keep or delete?

.github/workflows/python.yaml
```diff
@@ -28,6 +29,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         submodules: 'recursive'
+        # We need history so that we have tags to provide version information.
```
ddn0, line 31: qq: I'm not a git guru. I would have thought that `fetch-depth: 0` would be contrary to fetching "history" though. What am I missing?

.github/workflows/python.yaml
```diff
@@ -36,7 +38,7 @@ jobs:
     - name: Cache conda packages
       uses: actions/cache@v2
       with:
-        key: ${{matrix.os}}-conda-pkgs-${{hashFiles('**/conda_recipe/meta.yaml')}}
+        key: ${{matrix.os}}-conda-pkgs-${{hashFiles('**/conda_recipe/meta.yaml', '**/*environment.yml')}}
```
ddn0: - "yml", "yaml"?
- and is `*environment.yml` a standard conda naming scheme?

.github/workflows/python.yaml
```diff
@@ -178,3 +175,127 @@ jobs:
...
+          ccache --zero-stats
+          export CMAKE_BUILD_PARALLEL_LEVEL=2
+          cmake --build $HOME/build --parallel 2
+          ccache --show-stats
+          ccache --cleanup
+
+      - name: Test python_env.sh
+        if: startsWith(matrix.os, 'ubuntu-')
+        run: |
+          . .github/workflows/activate_miniconda.sh
```
ddn0, line 270: are we also running `pytest` somewhere too?

cmake/Modules/KatanaBuildCommon.cmake
```diff
@@ -41,6 +42,8 @@ set(BUILD_TESTING ON CACHE BOOL "Build tests")
 # "/usr/local/lib" is not a default search location for ld.so
 set(CMAKE_INSTALL_PREFIX "/usr" CACHE STRING "install prefix")
 
+set(BUILD_BINDING "" CACHE STRING "Semi-colon separated list of language bindings to build (e.g., 'python'). Default: none")
```
ddn0: is `BUILD_BINDING` a well-known cmake variable? If not, some preference for `KATANA_BUILD_BINDING` or `KATANA_BUILD_BINDINGS` or `KATANA_LANG_BINDINGS`

Python.md
```diff
@@ -16,8 +16,8 @@ Installing Katana-Python
 Install [libnuma](https://github.com/numactl/numactl) as required in your distribution
```
ddn0: - There are some stray references to "Galois" in this file still.

Is the correct way to build from scratch to just do: `cmake -DBUILD_BINDING=python .. && make`?

cmake/Modules/KatanaPythonSetupSubdirectory.cmake
```diff
@@ -0,0 +1,265 @@
+include_guard(DIRECTORY)
+
```
ddn0, line 2: Maybe a comment here outlining here that we are creating a symlinked build directory because of X, Y and Z.

python/katana_setup.py
```diff
@@ -0,0 +1,220 @@
...
+
+    is_clang = any("clang" in s for s in extension_options["compiler"])
+
+    extension_options["extra_compile_args"].extend(
+        [
+            # TODO(amp): This warning appears in generated code and might or might not be an issue,
+            #  leave it as a warning for now.extension_options["compiler"]
+            "-Wno-error=uninitialized" if is_clang else "-Wno-error=maybe-uninitialized",
+            # Unused stuff is very common in generated code and is safe.
+            "-Wno-unused-variable",
```
ddn0, line 147: I ran into some compiler warning (promoted to an error) about redundant moves. If we are only using cc to glue APIs together, it might make sense to use `-Wno-error` here as there is not much we can do if our dependencies are not always perfect C++ (lol) and compiler warnings change relatively frequently.

python/katana_setup.py
```diff
@@ -0,0 +1,220 @@
...
+
+    if not extension_options["extra_compile_args"]:
+        extension_options["extra_compile_args"] = ["-std=c++17", "-Werror"]
+
+    if extension_options["compiler"]:
+        compiler = " ".join(extension_options["compiler"])
+        os.environ["CXX"] = compiler
+        os.environ["CC"] = compiler
+
+        # Because of odd handling of the linker in setuptools with C++ the
```
ddn0, line 133: It is probably for the best we use the same program. We can get into issues mixing programs from different toolchains: https://gist.github.com/jlblancoc/99521194aba975286c80f93e47966dc5

ddn0 (commented):
Sorry for the delay.

> Is disabling that test ok? (22980eb)

Seems ok. Though, if the sanitizer failures start leaking out of that unit test, we will have to find a better workaround

> Should "Conda Package and Python Formatting / build_and_package" be kept or disabled? It takes a really long time (40-60 minutes).

Depends on what process we want to release new packages. If it is done out of this repo, there should be something that tests the packaging process here.

> Is the "fake out of tree" build of the python packages OK? It builds a symlink tree (real directories, but full of symlinks instead of copies) and builds there, to avoid polluting the source tree.

Seems fine to me.

.github/workflows/python.yaml
arthurp: I'll uncomment it. It forces a change in the "required checks" of the repo. So we will need to disable the requirement of the check for a little while and then reenable it once the PR lands. That's why I commented it out. But the names here are terrible, so I want to change them.

.github/workflows/python.yaml
arthurp: The standard name is actually just `environment.yaml`. Ideally it would be in the project root, so you can just run `conda env update` but given that we don't really use conda much (other than me) I don't think moving it to the root makes sense. I'll rename it to just `environment.yaml` though.

.github/workflows/python.yaml
arthurp: pytest runs as part of ctest if the python wrappers are enabled.

Python.md
arthurp: Yes. I'll update the documentation a bit.

python/katana_setup.py
arthurp: I'm not sure what you are referring to here. Am I already doing what you are asking for? Or do I need to make sure the programs are all the same in some additional way?

ddn0: The PR is fine as-is. I was just noting a potential issue that could/did happen when unbundling build tools.

.github/workflows/python.yaml
arthurp: Your instinct is correct. However, the checkout action interprets this parameter strangely. See https://github.com/actions/checkout#checkout-v2. I will add a comment.

.github/workflows/python.yaml
arthurp: Irritatingly, it appears that most conda stuff defaults to `.yaml`, but this one case uses `.yml`. So they are just different for now. Sigh. `environment.yml` is the standard. As is `meta.yaml`. Yuck.

arthurp:
> > Should "Conda Package and Python Formatting / build_and_package" be kept or disabled? It takes a really long time (40-60 minutes).
> 
> Depends on what process we want to release new packages. If it is done out of this repo, there should be something that tests the packaging process here.

I enabled it, but I think we shouldn't mark it as required so that people can merge before it finishes if required.

arthurp:
> > Is disabling that test ok? ([22980eb](https://github.com/KatanaGraph/katana/commit/22980eb48573f9c8b6e9683d79dc757651dba17f))
> 
> Seems ok. Though, if the sanitizer failures start leaking out of that unit test, we will have to find a better workaround

It's leaking already. But only in the [enterprise repo tests](https://github.com/KatanaGraph/katana-enterprise/pull/850/checks?check_run_id=1937854581#step:14:33388). @ddn0 Can you help? I'm really not sure how to create a correct suppression file. I did try.

ddn0:
:tada:

ddn0 (approved):

arthurp:
I have rebased and cleaned up and I'm ready to merge. @gurbinder533 can you review this final version?

gurbinder533 (approved):
Thanks, @arthurp. I will test this soon.

PR #78, opened 2021-02-21
Add louvain clustering using property graphs

gurbinder533:


libgalois/include/katana/PropertyGraph.h
```diff
@@ -423,6 +423,9 @@ class KATANA_EXPORT PropertyGraph {
   Result<void> RemoveEdgeProperty(int i);
   Result<void> RemoveEdgeProperty(const std::string& prop_name);
 
+  Result<void> RemoveAllNodeProperties();
+  Result<void> RemoveAllEdgeProperties();
```
arthurp: These functions make me nervous because you should almost never do this because the user probably expects any unrelated properties to be untouched by most algorithms. Can you add documentation that warns user about that?

libgalois/include/katana/analytics/louvain_clustering/louvain_clustering.h
```diff
@@ -0,0 +1,104 @@
...
+  uint32_t max_iterations() const { return max_iterations_; }
+  uint32_t min_graph_size() const { return min_graph_size_; }
+
+  static LouvainClusteringPlan DoAll(
+      bool enable_vf, double c_threshold, double threshold,
+      uint32_t max_iterations, uint32_t min_graph_size) {
+    return {kCPU,      kDoAll,         enable_vf,     c_threshold,
+            threshold, max_iterations, min_graph_size};
+  }
+};
```
arthurp: This doesn't follow our conventions. They are documented here: https://github.com/KatanaGraph/katana/blob/master/docs/GraphOperationAPIs.md#plan

The most important thing for python wrapping is the `kDefault*` constants.

libgalois/include/katana/analytics/louvain_clustering/louvain_clustering.h
```diff
@@ -0,0 +1,104 @@
...
+
+struct KATANA_EXPORT LouvainClusteringStatistics {
+  /// Total number of unique clusters in the graph.
+  uint64_t total_clusters;
+  /// Total number of clusters with more than 1 node.
+  uint64_t total_non_trivial_clusters;
+  /// The number of nodes present in the largest cluster.
+  uint64_t largest_cluster_size;
+  /// The ratio of nodes present in the largest cluster.
+  double largest_cluster_ratio;
```
arthurp: I think "proportion" is better than "ratio".

libgalois/include/katana/analytics/louvain_clustering/louvain_clustering.h
```diff
@@ -0,0 +1,104 @@
...
+    PropertyGraph* pg, const std::string& edge_weight_property_name,
+    const std::string& output_property_name, LouvainClusteringPlan plan = {});
+
+KATANA_EXPORT Result<void> LouvainClusteringAssertValid(
+    PropertyGraph* pg, const std::string& edge_weight_property_name,
+    const std::string& output_property_name);
+
+struct KATANA_EXPORT LouvainClusteringStatistics {
+  /// Total number of unique clusters in the graph.
+  uint64_t total_clusters;
```
arthurp: `n_clusters`?

libgalois/include/katana/analytics/louvain_clustering/louvain_clustering.h
```diff
@@ -0,0 +1,104 @@
...
+  /// Total number of clusters with more than 1 node.
+  uint64_t total_non_trivial_clusters;
```
arthurp: `n_non_trivial_clusters`?

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+
+// Maintain community information
+template <typename EdgeWeightType>
+struct CommunityType {
+  std::atomic<uint64_t> size;
+  std::atomic<EdgeWeightType> degree_wt;
+  EdgeWeightType internal_edge_wt;
+};
+
+typedef katana::LargeArray<uint64_t> LargeArray;
```
arthurp: This `typedef` is a bad idea. It will pretty much always make code that uses it confusing.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+
+using LCPreviousCommunityId = katana::PODProperty<uint64_t>;
+using LCCurrentCommunityId = katana::PODProperty<uint64_t>;
+using LCColorId = katana::PODProperty<int64_t>;
+template <typename EdgeWeightType>
+using LCDegreeWeight = katana::PODProperty<EdgeWeightType>;
+
+template <typename EdgeWeightType>
+using LCEdgeWeight = katana::PODProperty<EdgeWeightType>;
```
arthurp: I'm guessing `LC` = "Louvain Clustering". If so, these should be in the implementation of that algorithm instead of in the general clustering header file. If they are actually generic and could apply to another clustering algorithm then replace `LC` with `Clustering` or something.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+    katana::gPrint("Number of Edges: ", graph.num_edges(), "\n");
+  }
+
+  static void PrintGraphCharateristics(const katana::PropertyGraph& pfg) {
+    katana::gPrint("/******************************************/\n");
+    katana::gPrint("/************ Graph Properties ************/\n");
+    katana::gPrint("/******************************************/\n");
+    katana::gPrint("Number of Nodes: ", pfg.num_nodes(), "\n");
+    katana::gPrint("Number of Edges: ", pfg.num_edges(), "\n");
+  }
```
arthurp: These are not used and should not be used, so let's remove them.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+
+    [[maybe_unused]] uint64_t num_unique_clusters =
+        RenumberClustersContiguously(graph);
+    auto mod = CalModularityFinal<Graph, EdgeWeightType>(graph);
+  }
+
+  /*
+ * Copy Graph
+ */
+  katana::Result<void> CopyGraph(
```
arthurp: The name implies that this function copies a `PropertyGraph`. I don't think it does that. Can we rename this and provide useful documentation?

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+using LCCurrentCommunityId = katana::PODProperty<uint64_t>;
+using LCColorId = katana::PODProperty<int64_t>;
+template <typename EdgeWeightType>
+using LCDegreeWeight = katana::PODProperty<EdgeWeightType>;
+
+template <typename EdgeWeightType>
+using LCEdgeWeight = katana::PODProperty<EdgeWeightType>;
+
+template <typename _Graph, typename _EdgeType, typename _CommunityType>
+struct ClusteringImplementationBase {
```
arthurp, line 52: I do not have a solid understanding of the function in this class, but can you confirm that each is potentially useful for other clustering algorithms? And if not move them to the appropriate algorithm implementations. The functions look kind of special purpose.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+  }
+}
+
+/// \cond DO_NOT_DOCUMENT
+katana::Result<void>
+katana::analytics::LouvainClusteringAssertValid(
+    [[maybe_unused]] katana::PropertyGraph* pg,
+    [[maybe_unused]] const std::string& edge_weight_property_name,
+    [[maybe_unused]] const std::string& property_name) {
+  // TODO(gill): This should have real checks. amp has no idea what to check.
```
arthurp: It's the than amp has no idea what to check, but in this case gill doesn't either. ;-)

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+    Graph graph_curr = graph_result.value();
+
+    /*
+    * Vertex following optimization
+    */
+    if (plan.is_enable_vf()) {
+      [[maybe_unused]] uint64_t num_nodes_to_fix = Base::VertexFollowing(
+          &graph_curr);  // Find nodes that follow other nodes
+
+      [[maybe_unused]] uint64_t num_unique_clusters =
```
arthurp: This is definitely used so drop `[[maybe_unused]]`.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+    if (!graph_result) {
+      return graph_result.error();
+    }
+    Graph graph_curr = graph_result.value();
+
+    /*
+    * Vertex following optimization
+    */
+    if (plan.is_enable_vf()) {
+      [[maybe_unused]] uint64_t num_nodes_to_fix = Base::VertexFollowing(
```
arthurp: It looks like this might be called for the side effect. If that is the case, I would prefer you remove the variable it's assigned to entirely so no one thinks the variable is the important part and deletes the call. A call without an assignment is much more clearly for side-effects.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+          }
+          curr_mod = curr_mod_result.value();
+          break;
+        }
+        default:
+          return katana::ErrorCode::InvalidArgument;
+        }
+      }
+
+      [[maybe_unused]] uint64_t num_unique_clusters =
```
arthurp: drop `[[maybe_unused]]`?

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+  if (auto r = impl.LouvainClustering(
+          pfg, edge_weight_property_name, clusters_orig, plan);
+      !r) {
+    return r.error();
+  }
+
+  // Remove all the existing node/edge properties
+  if (auto r = pfg->RemoveAllNodeProperties(); !r) {
+    return r.error();
+  }
```
arthurp: I don't think dropping all properties of a users graph is a good idea. That is not something anyone would expect a graph algorithm to do.

If you need to clean up from the algorithms mess maybe a vector of [TemporaryPropertyGuard](https://github.com/KatanaGraph/katana/blob/bc9a6c665bc8aeb1920c3ebb8567a6563c5162d3/libgalois/include/katana/analytics/Utils.h#L74) would be a simple way to handle it. When the vector destructs all the properties will delete. It could be scoped to this function.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+  using ClusterBase = katana::analytics::ClusteringImplementationBase<
+      Graph, EdgeWeightType, CommTy>;
+  ClusterBase cluster_base{};
+  auto graph_result =
+      Graph::Make(pg, {property_name}, {edge_weight_property_name});
+  if (!graph_result) {
+    return graph_result.error();
+  }
+  auto graph = graph_result.value();
+  return cluster_base.template CalModularityFinal<Graph, EdgeWeightType>(graph);
```
arthurp: The constructor of `ClusterBase` needs no arguments and is not kept around then `CalModularityFinal` can probably be a static method.

lonestar/analytics/cpu/louvain_clustering/louvain_clustering_cli.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  LouvainClusteringPlan plan = LouvainClusteringPlan();
+  switch (algo) {
+  case LouvainClusteringPlan::kDoAll:
+    plan = LouvainClusteringPlan::DoAll(
+        enable_vf, c_threshold, threshold, max_iterations, min_graph_size);
+    break;
+  default:
+    std::cerr << "Invalid algorithm\n";
+    abort();
```
arthurp: `KATANA_LOG_FATAL`

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+  using Graph = _Graph;
+  using GNode = typename Graph::Node;
+  using EdgeTy = _EdgeType;
+  using CommunityType = _CommunityType;
+
+  /*
+ * Typedefs
+ */
+  constexpr static const uint64_t INF_VAL =
+      std::numeric_limits<uint64_t>::max() / 2 - 1;
```
arthurp: Can this be `std::numeric_limits<uint64_t>::max()` instead? If not, please document why this constant is used.

Also `kInfVal`.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+  constexpr static const uint64_t UNASSIGNED =
+      std::numeric_limits<uint64_t>::max();
+  constexpr static const double DOUBLE_MAX =
+      std::numeric_limits<double>::max() / 4;
```
arthurp: Can this be `std::numeric_limits<douple>::max()` instead? If not, please document why this constant is used and change it's name. It's definitely not `DOUBLE_MAX`.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+
+  typedef katana::LargeArray<uint64_t> LargeArray;
+  typedef katana::LargeArray<EdgeTy> largeArrayEdgeTy;
```
arthurp: These should have better names or just be removed. The original type isn't that long. You can `using katana::LargeArray` to shorten the types a bit without hiding the element types.

Also `largeArrayEdgeTy` should be capitalized.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+    if (auto r = pfg_to->AddEdgeProperties(edge_data_table); !r) {
+      return r;
+    }
+    return katana::ResultSuccess();
+  }
+
+  /*
+ * Graph construction routines to make
+ * coarser graphs.
+ */
```
arthurp: Either improve or delete these utterly useless comments that just restate the function names.

arthurp (changes requested):
Lots of comments. I don't know the algorithm at all. But there are still a good number of things that stick out to me.

Nit: I would prefer to have the addition of `RemoveAll*Properties` as a separate commit (in the same PR) from adding the algorithm. It's not that important though.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
uditagar: probably can drop this. I don't think this is used anywhere at the moment.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
uditagar: can drop this as well, since this is not being used anywhere

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+    /**
+   * Add the node's current cluster to be considered
+   * for movement as well
+   */
+    cluster_local_map[graph.template GetData<LCCurrentCommunityId>(n)] =
+        0;                 // Add n's current cluster
+    counter.push_back(0);  // Initialize the counter to zero (no edges incident
+                           // yet)
+    num_unique_clusters++;
+
```
uditagar: we can skip this. The for loop can handle this. For example, if there is an edge from n to a node m (or itself) in its own community, then the node's current cluster/community will get added to the cluster map.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -0,0 +1,680 @@
...
+  uint64_t VertexFollowing(Graph* graph) {
+    using GNode = typename Graph::Node;
+    // Initialize each node to its own cluster
+    katana::do_all(katana::iterate(*graph), [&](GNode n) {
+      graph->template GetData<LCCurrentCommunityId>(n) = n;
+    });
+
+    // Remove isolated and degree-one nodes
+    katana::GAccumulator<uint64_t> isolatedNodes;
```
uditagar: should be isolated_nodes

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,615 @@
...
+  using CommTy = CommunityType<EdgeWeightType>;
+  using CommArray = katana::LargeArray<CommTy>;
+
+  using Graph = katana::TypedPropertyGraph<NodeData, EdgeData>;
+  using GNode = typename Graph::Node;
+
+  using Base = katana::analytics::ClusteringImplementationBase<
+      Graph, EdgeWeightType, CommTy>;
+
+  katana::Result<double> algoLouvainWithoutLockingDoAll(
```
uditagar: I thought we were using the for_each implementation, since the do_all implementation does not give good modularity scores.

We can also add it later.

libgalois/include/katana/PropertyGraph.h
gurbinder533: Yeah, these make me nervous as well. Let me see if I can use TemporaryGuard for these properties.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
gurbinder533: Yes, these will be used for Leiden clustering as well.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
gurbinder533: :P lol

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
gurbinder533: Good point.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
gurbinder533: for_each implementation needs read/write locks on the nodes which we don't have yet.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
gurbinder533: I don't understand what you mean.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
uditagar: ok

libgalois/include/katana/analytics/ClusteringImplementationBase.h
uditagar: I was just saying that since the entry is zero, we don't need to add n's communtiy to the map at the start. We can do that inside the for loop (when we find an edge from n to its community)

uditagar: we can leave it like this as well, it's fine.

uditagar (commented):
looks fine to me. Probably need to address Arthur's comments. Thank you.

libgalois/include/katana/PropertyGraph.h
gurbinder533: Removed the dangerous "RemoveAll" from the property graph.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
gurbinder533: Added more meaningful comments :).

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -57,52 +55,29 @@ struct ClusteringImplementationBase {
...
-  }
 
-  static void PrintGraphCharateristics(const katana::PropertyGraph& pfg) {
-    katana::gPrint("/******************************************/\n");
-    katana::gPrint("/************ Graph Properties ************/\n");
-    katana::gPrint("/******************************************/\n");
-    katana::gPrint("Number of Nodes: ", pfg.num_nodes(), "\n");
-    katana::gPrint("Number of Edges: ", pfg.num_edges(), "\n");
-  }
+  using CommArray = katana::LargeArray<CommunityType>;
```
arthurp: "Comm" usually means "communication". `CommunityArray`?

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -182,36 +166,46 @@ struct ClusteringImplementationBase {
     return;
   }
 
+  /**
+   * Computes the constant teem 1/(2 * total internal edge weight)
```
arthurp: "teem"? "term"?

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -182,36 +166,46 @@ struct ClusteringImplementationBase {
...
+      local_weight += graph.template GetData<DegreeWeight<EdgeWeightType>>(n);
     });
-    /* This is twice since graph is symmetric */
+    //This is twice since graph is symmetric
     double total_edge_weight_twice = local_weight.reduce();
     return 1 / total_edge_weight_twice;
   }
 
+  /**
+   * Computes the constant teem 1/(2 * total internal edge weight)
```
arthurp: "teem"

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -352,8 +360,11 @@ LouvainClusteringWithWrap(
       std::is_integral_v<EdgeWeightType> ||
       std::is_floating_point_v<EdgeWeightType>);
 
+  std::vector<std::string> temp_node_property_names = {
+      "CurrentId", "PreviousId", "DegreeWt"};
```
arthurp: These names need to have a random number added to them to avoid name conflicts with existing user properties. Or you could prefix them with something like `_katana_` and document that we reserve all property names which begin with that.

arthurp (approved):
LGTM overall. My only concern is the property naming conflict issue. Address it as you see fit.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
uditagar: second this

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -0,0 +1,631 @@
...
+    katana::StatTimer TimerClusteringTotal("Timer_Clustering_Total");
+    TimerClusteringTotal.start();
+
+    auto graph_result = Graph::Make(pfg);
+    if (!graph_result) {
+      return graph_result.error();
+    }
+    Graph graph = graph_result.value();
+
+    CommArray c_info;    // Community info
```
uditagar: probably will need to update this `CommArray` too, if you change it in the header file

uditagar (approved):

PR #79, opened 2021-02-21
ci: Add note about CI branches

ddn0:


arthurp (approved):
Please add a note that all ("most"?) of the `scripts/check_*` scripts have a `-fix` option that will correct the same errors they check for (for the most part). Otherwise, LGTM.

(Please also take a glance at https://katanagraph.atlassian.net/browse/ENG-204. I think it will help with this as well. In the longer term.)

PR #88, opened 2021-03-01
Simplify backtrace and other nits

witchel:
This is a collection of small fixes.

BuildGraph: allow creation of empty Arrow arrays
 o It is useful to build empty Arrow arrays from empty vectors
cut and paste fix in RDG::DoMake
 o Simple bug fix to return the error from the checked variable
Simplify backtrace logic
 o Only have one thread print backtrace and eliminate useless code.  We could work harder to only print the backtrace on 1 host if all fail, but it dosn't quite seem worthwhile.
edge_dest is a valid operation on GraphTopology
 o Seemed like a useful operation worthy of a name.


tylershunt (approved):
LGTM

PR #90, opened 2021-03-02
Add version management script. Consistently use it.

arthurp:
This implements RFC-ENG-013 and indirectly RFC-ENG-005.


scripts/setup_macos.sh
```diff
@@ -27,3 +27,5 @@ brew_install_if_missing libcypher-parser
 # https://github.com/KatanaGraph/homebrew-dependencies
 #brew tap KatanaGraph/dependencies
 brew uninstall apache-arrow || true
+
+pip3 install PyGithub packaging
```
ddn0, line 32: What are your thoughts on venv, poetry or pipenv? The last one we have used elsewhere to manage python dependencies.

scripts/katana_version/version.py
```diff
@@ -0,0 +1,388 @@
...
+    tag_version = None
+    version_tags = [m for m in (VERSION_TAG_RE.match(t) for t in git.get_tags_of(commit, repo_path)) if m]
+    if len(version_tags) > 1:
+        logger.warning(f"There is more than one version tag at the given commit. Picking one arbitrarily.")
+    if version_tags:
+        tag_version = version.Version(version_tags[0].group("version"))
+    return tag_version
+
+
+def compute_commit_count(config, commit, last_version_commit, repo_path, pretend_master, core_branch):
```
ddn0, line 223: I suppose I should be happy that this isn't a bash script.

scripts/katana_version/version.py
```diff
@@ -0,0 +1,388 @@
...
+            version_str = version_file.read()
+    ver = version.Version(version_str.strip())
+
+    if no_dev:
+        return ver
+
+    return add_dev_to_version(ver)
+
+
+def get_tag_version(config, commit, repo_path):
```
ddn0, line 211: Since we don't have the automated actions that calls this script (unless I'm mistaken), can you add a README with some common commands and scenarios? That might be a good place to summarize the format, etc.

scripts/setup_macos.sh
arthurp: I would prefer to use Conda if we want a python environment tool. We are probably forced to use it anyway because it's a very common system in data science. If you just want a dependency file instead of having the package names in the script we can use a pip requirements file. The same requirements file could also be used in pipenv/conda/venv if people want that in their own dev environment.

scripts/katana_version/version.py
arthurp: Probably. I really overengineered this, but I think it's a lot better than a bash script that tries to do all this.

scripts/katana_version/version.py
arthurp: Yea. That's a good point. I meant to copy in some of the RFC text into the documentation. I'll add something to CONTRIBUTING.md.

scripts/setup_macos.sh
ddn0: How about a requirements file (for each python app) and some `find -name requirements.txt | xargs pip install -r ` thing at the least? This would at least allow dependencies to be written closer to the script that needs them.

And probably a pointer in that file that says that you might have to adjust the conda dependency file as well? Or would https://pypi.org/project/conda-merge/ be helpful here?

arthurp: That could be useful. I would produce a single file to use to install the packages from multiple inputs. This file would probably need to be committed to the repo (like a lock file). If we are going that direction, I think just having lock files might be better: https://github.com/conda-incubator/conda-lock. It's somewhat experimental, but I played with it and it works and doesn't need to be installed on the user machine since it generates a lock file that can be used by conda directly. Installing with such a lock file also reduces the install time for one a katana-open development environment from ~60sec to ~10sec.

CONTRIBUTING.md
```diff
@@ -113,7 +113,159 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
 features or plugins to streamline this process (e.g.,
 [vim-fugitive](https://github.com/tpope/vim-fugitive),
-[magit](https://magit.vc/)).
+[magit](https://magit.vc/)). Git also provides `git commit --fixup` and
+`git rebase --autosquash` which can streamline this process.
+
+# Build and Release Versions
+
+Release versions of Katana are a semantic version with an release candidate tag
+(`rc#`) append for release candidates.
```
ddn0: "appended"

CONTRIBUTING.md
```diff
@@ -113,7 +113,159 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+
+Build/development versions of Katana are in the form:
+`0.1.10rc1.dev+40.0.04ac6f`. This is a release version followed with `.dev` and
```
ddn0: This is a lot to parse at once.

How about:

"... are in the form: <version> .dev + <number> . <number> . <hash>. For example, 0.1.10rc1.dev+40.0.04ac6f is .... "

?

CONTRIBUTING.md
```diff
@@ -113,7 +113,159 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+a local version (after the `+`) made up of the number of commits on the nearest
+major branch (e.g., `master` or `release/v*`), a `0`, and a abbreviated commit
```
ddn0: "an abbreviated commit"

CONTRIBUTING.md
```diff
@@ -113,7 +113,159 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+* Otherwise, the content of `config/version.txt` specifies the upcoming release
+  and the version of the commit is computed based on the commit history and
+  branches.
+
+Day to day development is on master which is considered to be a development
+build working up to the next release. The version in `config/version.txt` will
+be the upcoming semantic version (e.g., `0.1.11`). Master should at least build
+and pass tests, but may be unstable or buggy at times. This is especially true
+between commits in the same PR, because the intermediate commits may never have
+been tested after they are rebased onto master by Github during the PR “merge”
```
ddn0: just noting: it looks like these are non-ASCII quotation marks (if that matters)

CONTRIBUTING.md
```diff
@@ -113,7 +113,159 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+* If the commit has a tag in the form `v{version number}` (e.g., `v0.1.10rc1`)
+  then that is the version regardless of other information.
+* Otherwise, the content of `config/version.txt` specifies the upcoming release
+  and the version of the commit is computed based on the commit history and
+  branches.
+
+Day to day development is on master which is considered to be a development
+build working up to the next release. The version in `config/version.txt` will
+be the upcoming semantic version (e.g., `0.1.11`). Master should at least build
+and pass tests, but may be unstable or buggy at times. This is especially true
```
ddn0: stylistic suggestion: move the general principle up and shorten the examples to keep it punchy

E.g.,

The master branch should at least build and pass tests but may be unstable from commit to commit. It may be the case that an regression is introduced post-merge or that a PR is more clearly expressed with intermediate commits that do not pass all tests. We prefer a clear history over "atomic" commits. During our release process, we specifically address stability and regression issues.

ddn0 (approved):
I played around with some `script/version` commands and they seem to behave as expected.

scripts/setup_macos.sh
arthurp: I have created two issues based on this discussion: #111 and #112.

CONTRIBUTING.md
```diff
@@ -113,7 +113,161 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
 push --force`  to update your branch. Depending on your IDE, there are various
 features or plugins to streamline this process (e.g.,
 [vim-fugitive](https://github.com/tpope/vim-fugitive),
-[magit](https://magit.vc/)).
+[magit](https://magit.vc/)). Git also provides `git commit --fixup` and
+`git rebase --autosquash` which can streamline this process.
+
+# Build and Release Versions
+
+Release versions of Katana are a semantic version with an release candidate tag
```
bosconi: as -> a

CONTRIBUTING.md
```diff
@@ -113,7 +113,161 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+(`rc#`) appended for release candidates.
+
+Build/development versions of Katana are are in the form:
```
bosconi: in -> of

CONTRIBUTING.md
```diff
@@ -113,7 +113,161 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+`[release].dev+N.0.######`.
+
+* a release version followed
```
bosconi: followed by?

CONTRIBUTING.md
```diff
@@ -113,7 +113,161 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+* `.dev`
+* `+`
+* the number of commits on the nearest major branch (e.g., `master` or `release/v*`)
+* `0`
+* an abbreviated commit id
+
+The `0` is to leave space for the enterprise commit count in enterprise builds
```
bosconi: "to leave space" -> "a placeholder"? Should we be even more explicit and say that for enterprise releases only it is replaced with the enterprise repo commit count?

CONTRIBUTING.md
```diff
@@ -113,7 +113,161 @@ You can use `git rebase --interactive` to reorganize your commits and  `git
...
+subcommand).
+
+**To show the current version**, run `scripts/version show`. This will print out
+the version of the working tree. To see the version of a specific commit run
+`scripts/version show {commit}`. This will print out the version of a clean
+checkout of that commit.
+
+**To create a release branch**, run `scripts/version release_branch {next version}`
+and merge the PRs it creates as soon as possible (to limit the number
+of commits which have odd version numbers). This will create a new release
```
bosconi: odd as in not even?

bosconi (approved):
Small document things but LGTM!

PR #95, opened 2021-03-03
Support gcc-7

ddn0:


python/tests/test_loops.py
```diff
@@ -144,7 +144,7 @@ def f(out, i):
 def test_for_each_no_push(modes):
     @for_each_operator()
     def f(out, i, ctx):
-        out[i] += i + 1
+        out[i] = i + 1
```
arthurp: This test is definitely thread safe as it was before. Let's not change it without a reason.

arthurp (approved):
I think you have a stray change to a test that does not need fixing. Other than that LGTM.

python/tests/test_loops.py
ddn0: It failed in CI (twice).

https://github.com/KatanaGraph/katana/runs/2018400129?check_suite_focus=true

Unless you mean `test_for_each_no_push` is significantly different than `test_for_each`.

arthurp: It actually is different because there is no push so there can be no race. However in the end it doesn't really matter. I just don't like changing tests without a good reason.

ddn0: I don't mind reverting `+=` to `=` but it seems weird to me to write `+=` when you don't actually mean to accumulate anything and the sibling test is definitely was not safe with `+=`. If the idea is that each location is only accessed by a single iteration, `=` seems clearer to me.

arthurp: I think I probably wrote it the original way to create more complex code to "test more", but it's not at all clear that's better. I may also have been thinking about defeating the optimizer, but again, it's not clear that's needed here.

PR #103, opened 2021-03-08
lint and fix all bash scripts

aneeshdurg:


aneeshdurg:
I think most of the failures are coming from bad lints where variables that held multiple arguments got turned into a single argument. Going to split these changes up a bit.

.github/workflows/check_docs.sh
```diff
@@ -5,6 +5,6 @@ KATANA_REPO_ROOT="$(cd "$(dirname "$0")/../.."; pwd)"
 BUILD_DIR=$1
 shift 1
 
-cmake -S . -B $BUILD_DIR \
-  -DCMAKE_TOOLCHAIN_FILE=$BUILD_DIR/conan_paths.cmake && \
-  cd $BUILD_DIR && $KATANA_REPO_ROOT/scripts/check_docs.sh
+cmake -S . -B "$BUILD_DIR" \
+  -DCMAKE_TOOLCHAIN_FILE="$BUILD_DIR"/conan_paths.cmake && \
```
arthurp, line 9: I prefer to quote whole arguments instead of just the variables being expanded. So unless someone else objects I would prefer `"-DCMAKE_TOOLCHAIN_FILE=$BUILD_DIR/conan_paths.cmake"`. I think it makes the argument boundaries clearer.

This comment applies throughout the PR. I can mark all the spot if you need, but I suspect you can find them easily enough.

.github/workflows/setup_conan_install.sh
```diff
@@ -5,13 +5,14 @@ KATANA_REPO_ROOT="$(cd "$(dirname "$0")/../.."; pwd)"
 BUILD_DIR="$1"
 shift 1
 
+# shellcheck disable=SC1090
 [[ -f ~/.profile ]] && source ~/.profile
 
-$KATANA_REPO_ROOT/scripts/setup_conan.sh
+"$KATANA_REPO_ROOT"/scripts/setup_conan.sh
```
arthurp, line 11: I would also prefer to quote entire command paths.

.github/workflows/setup_miniconda.sh
```diff
@@ -1,7 +1,12 @@
-# This is a bash INCLUDE file. Source it don't run it.
+#!/bin/bash
```
arthurp: I'm not 100% sure we are assuming bash is available everywhere. Let's try to be Bourne Shell compatible if possible. It's useful on OSX because the default shell is `zsh`.

scripts/check_go_format.sh
```diff
@@ -23,20 +23,21 @@ emit_prunes() {
...
 
 if [ -z "$FILES" ]; then
     echo "no files to fix!" >&2
     exit 0
 fi
 
-if [ -n "${FIX}" ]; then
-  ${GOFMT} -s -w ${FILES}
+if [ -n "$FIX" ]; then
+  $GOFMT -s -w "$FILES"
```
arthurp: This is wrong. It will combine all the file names into a single argument. To do this correctly, I see two options: a) pipe the files into these `GOFMT` commands and use xargs. b) make `FILES` an array and construct it with [readarray](https://helpmanual.io/builtin/readarray/) on the results of find.

scripts/check_go_lint.sh
```diff
@@ -3,38 +3,37 @@
...
 
 # only run linter in the root of go modules
-for root in ${ROOTS}; do
-  NEWROOTS=$(find ${root} -name .git -prune -o -name 'go.mod' -printf '%h\n' | sort -u | xargs)
-  if [ -n "${NEWROOTS}" ]; then
-    GOROOTS="${NEWROOTS} ${GOROOTS}"
+for root in "${ROOTS[@]}"; do
+  NEWROOTS=$(find "$root" -name .git -prune -o -name 'go.mod' -printf '%h\n' | sort -u | xargs)
+  if [ -n "$NEWROOTS" ]; then
+    GOROOTS="$NEWROOTS $GOROOTS"
```
arthurp: You used arrays elsewhere. It would also allow this to be space safe. Which it currently is not.

scripts/check_python_format.sh
```diff
@@ -14,17 +14,18 @@ if [ "$1" == "-fix" ]; then
...
 }
 
-FILES=$(find ${ROOTS} $(emit_prunes) -name '*.py' -print | xargs)
+# shellcheck disable=SC2046
+FILES=$(find "${ROOTS[@]}" $(emit_prunes) -name '*.py')
 
-if [ -n "${FIX}" ]; then
-  ${PYFMT} ${FILES}
+if [ -n "$FIX" ]; then
+  $PYFMT "$FILES"
```
arthurp: Same issue as the `GOFMT` lines above.

scripts/check_python_lint.sh
```diff
@@ -38,19 +39,22 @@ is_test() {
...
-while read -d '' filename; do
+while read -r -d '' filename; do
   if is_test "${filename}"
   then
     args="${TEST_ARGS}"
   else
     args="${ARGS}"
   fi
-  if ! ${LINT} ${args} "${filename}"
+  if ! ${LINT} "${args}" "${filename}"
```
arthurp, line 49: This is treating `args` as a single argument. You would need to use an array to split it space safely. Or just remove the quotes. I don't see any clear way a space could get into an argument.

scripts/rename_galois_to_katana.sh
```diff
@@ -190,74 +188,76 @@ fi
...
   ### Include directories
   for f in lib*/include/galois; do
-    if [[ ! -e ${f} ]]; then
+    if [[ ! -e "$f" ]]; then
       # Glob failed
       continue
     fi
     # use rsync so we get consistent behavior even if target directory exists
-    rsync -a --remove-source-files ${f}/. $(dirname ${f})/katana
+    targetdir=$(dirname "$f")
```
arthurp: If you want to be space safe you need to quote the `$(...)` too. This issue appears several times.

scripts/setup_ubuntu.sh
```diff
@@ -96,9 +96,9 @@ GIT=git
 LIBRARIES="libxml2-dev llvm-10-dev"
 
 apt install -yq --allow-downgrades \
-  $DEVELOPER_TOOLS \
+  "$DEVELOPER_TOOLS" \
```
arthurp, line 99: This is supposed to be word split. Either use an array or remove the quotes. Same with `LIBRARIES` below.

arthurp (changes requested):
There are a number of issues here. Mainly it appears there are a number of places where you quoted variables that are supposed to be word split.

Did you test all the scripts and test them with multiple files? It seems like many of them would fail to work properly with the changes. Yes, we would ideally have tests for all the shell scripts. No we don't. ;-)

BTW, I think it would be safe and wise to skip fixing `scripts/rename_galois_to_katana.sh` entirely. It was a one use script that really only exists now as an artifact of how we did the rename. Maybe mark the whole file as ignored by shellcheck and add a comment.

arthurp:
I failed to read your comment before reviewing. Sorry. You marked it as ready to review, so I assumed you had fixed all the major issues. :-) You can just resolve a number of my comments as soon as you fix this issue you commented on.

scripts/rename_galois_to_katana.sh
```diff
@@ -5,7 +5,7 @@ set -xeuo pipefail
 MODE=${1:-}
 
 if [[ ${MODE} == "contents" ]]; then
-  while read -d '' filename; do
+  while read -r -d '' filename; do
```
ddn0: you can just delete this file; it was only needed for historical reasons.

scripts/check_go_format.sh
ddn0: My usual point when I think it is the time to stop using a shell script is when you need arrays or maps.

At that point, either (a) you prioritize transparency over generality ("yup, it doesn't handle paths with spaces in it, but if you want to know what the CI is doing it is running that command.") or (b) you notice that the users (and the execution environment) of script have expanded beyond the initial authoring group ("I ran this on RHEL4, it printed a bunch of stuff and nothing works").

If we think a particular script falls into (b), I'd suggest rewriting it in vanilla python, which doesn't have perfect dependency or version management, but will be more comprehensible to a larger set of future developers who will have to maintain it.

arthurp: This is a good point. What do you mean by "vanilla python" though? Do you mean Python using only the standard library? Which version of Python? I think defining the "shell + 1" language and environment would actually be quite useful.

ddn0: "vanilla python" being python using just the standard library and supporting all versions from last supported distro (nominally Ubuntu 18.04*) forward.

So python 3.6 I believe.

[*] I still consider supporting releases before 18.04 as best-effort, but if folks feel strongly about making earlier releases a hard requirement, that's fine too.

scripts/rename_galois_to_katana.sh
roshandathathri: Are you sure? We may need to port Galois code to Katana code in the future.

ddn0: I didn't intend for this script to support ongoing porting. I was mainly concerned with making my changes automated and verifiable, so I think of it as a one-off sort of thing.

If we are expecting to accept more Galois code, I think we should ask for more changes like tests, documentation and consistent APIs, so maintaining a namespace renamer doesn't loom high on my list.

I don't mind keeping it if it is worth the maintenance burden, but if it were just me, I'd delete it :D

scripts/check_go_format.sh
aneeshdurg: I am 100% in favor of migrating this to python. Sounds like a separate PR though. Do linter scripts have to support pre 18.04 distros though? It seems reasonable to say we don't support developing on 16.04 directly.

scripts/rename_galois_to_katana.sh
aneeshdurg: I'll delete it, it's always in the git history if we realize we really need it later.

PR #104, opened 2021-03-08
Add random-walks app to katana library

gurbinder533:


libgalois/include/katana/analytics/random_walks/random_walks.h
```diff
@@ -0,0 +1,106 @@
...
+        number_of_edge_types};
+  }
+};
+
+/// Compute the random-walks for pg. The pg is expected to be
+/// symmetric.
+/// The parameters can be specified, but have reasonable defaults. Not all parameters
+/// are used by the algorithms.
+/// The generated random-walks generated are return in Katana::InsertBag.
+KATANA_EXPORT Result<katana::InsertBag<katana::gstl::Vector<uint32_t>>>
```
arthurp: I'll review when you actually ask me too. But I thought I would warn you: There is Python support for `InsertBag` but only for primitive types and simple structs of primitive types. This type would need to be fully converted before it can be returned. If we can simplify it to be a bit more C++ and less Katana (maybe use `std::vector<std::vector<...>>`) that would make Python wrapping easier.

libgalois/src/analytics/random_walks/random_walks.cpp
```diff
@@ -158,22 +99,27 @@ struct Node2VecAlgo {
...
-          for (uint32_t walk_iter = 2; walk_iter <= walkLength; walk_iter++) {
+          for (uint32_t walk_iter = 2; walk_iter <= plan_.walk_length();
+               walk_iter++) {
             uint32_t curr = walk[walk_iter - 1];
             uint32_t prev = walk[walk_iter - 2];
 
+            //check if n has no neighbor
+            if (degree[curr] == 0) {
+              continue;
+            }
```
uditagar: this is not required for the symmetric case. And for un-symmetric case, if curr has degree 0, then the walk is stuck there. So probably can use `break` instead of `continue`

libgalois/src/analytics/random_walks/random_walks.cpp
```diff
@@ -307,28 +258,34 @@ struct Edge2VecAlgo {
...
+          types_vec.push_back(nbr_pair.second);
 
-          for (uint32_t walk_iter = 2; walk_iter <= walkLength; walk_iter++) {
+          for (uint32_t walk_iter = 2; walk_iter <= plan_.walk_length();
+               walk_iter++) {
             uint32_t curr = walk[walk.size() - 1];
+            //check if n has no neighbor
+            if (degree[curr] == 0) {
+              return;
+            }
```
uditagar, line 270: same as here https://github.com/KatanaGraph/katana/pull/104/files#r589808929

libgalois/include/katana/analytics/random_walks/random_walks.h
uditagar: In that  case, we would need to create a new `std::vector` and move the elements of the `InsertBag` to it. We still need `InsertBag` for our random-walk generation.

uditagar (commented):
Looks good, except one minor comment. Thanks

libgalois/include/katana/analytics/random_walks/random_walks.h
gurbinder533: @arthurp Thank you for pointing that out. Let us do a quick performance evaluation using std::vector.

arthurp: I could see an argument that using `vector` is also better for C++ because it is a well known type. Internally you would still use InsertBag and just convert before you return. Is that what you are going to benchmark?

uditagar: I thought what he meant was using `std::vector` instead of `gstl::Vector` during the random-walk computation. I will do that.

gurbinder533: Yes, I meant just replacing gstl::Vector altogether but I see @arthurp point of just copying over results before returning. We didn't notice any significant performance degradation with using std::vector though.

uditagar (approved):
This looks fine. Thank you.

PR #105, opened 2021-03-09
Add context to Result

ddn0:
```
Allow arbitrary string data to be attached to a katana::Result. Because
this change affects most of the codebase, I paid additional attention to
the implementation details. Because there is at most one error live at a
time (usually), the route I took is a thread-local buffer to store the
additional context information.

The implementation keeps extra state in order to identify potential
misuse of error information like storing one on the heap. This is only
best-effort though. Depending on experience, we can later strengthen the
checks or get rid of them altogether without changing the API.

Attached are microbenchmark results on my machine. The benchmark program
runs a recursive chain of function calls fn1 o fn2 o ... where o is
function composition. fn1 (GoWithHandler) and fn2 (Go) simulate handling
or just propagating errors upwards respectively.

The parameters to the benchmark are as follows (and in order):

- Number of threads
- Number of chains to run
- Handle ratio: what proportion of calls are GoWithHandler versus Go. A
  value of 1 indicates all calls are GoWithHandler. A value of 3
  indicates, on average, every third call is GoWithHandler.
- Max recursive depth: the probability of terminating the recursive
  calls increases until the max depth. The expected termination depth is
  depth - 1.
- Expected failure depth: the expected depth that we would return an
  error as opposed to terminating successfully.

So "Variant/4/1024/3/32/24" indicates that the benchmark ran with 4
threads with approximately 1 of 3 calls being GoWithHandler. The
expected call depth of successful chains is 32, while failures would be
expected to appear around depth 24. This means we expect all runs to end
with an error.

For ".../32/31", we expect about a quarter of runs to end in an error.

For ".../32/40", we expect all runs to terminate successfully.

Variants:

- ReturnException: throw exceptions to signal errors
- ...ErrorCodeResult: return Result with std::error_code to signal
  an error. This is the state of katana::Result before this commit.
- ...StringResult: A Result is an std::error_code with a std::string
  for additional error context. When returning errors, just return an
  error code.
- ...KatanaResult: A Result is an std::error_code with a thread-local
  buffer for additional error context. When returning errors, just
  return an error code. This is the implementation introduced by this
  commit.
- ...WithContext: When returning errors, add a relatively long string as
  an error context.

Representative results:

  $ libsupport/test/result-bench --benchmark_filter=/4/1024/3/32
  2021-03-08 16:41:27
  Running libsupport/test/result-bench
  Run on (12 X 4500 MHz CPU s)
  CPU Caches:
    L1 Data 32K (x6)
    L1 Instruction 32K (x6)
    L2 Unified 256K (x6)
    L3 Unified 12288K (x1)
  Load Average: 0.46, 0.33, 0.43
  -------------------------------------------------------------------------------------------------
  Benchmark                                                       Time             CPU   Iterations
  -------------------------------------------------------------------------------------------------
  ReturnException/4/1024/3/32/24/real_time                 15325505 ns       290385 ns           40
  ReturnException/4/1024/3/32/31/real_time                 16655975 ns       296925 ns           40
A ReturnException/4/1024/3/32/40/real_time                   867247 ns       174371 ns          803
B ReturnErrorCodeResult/4/1024/3/32/24/real_time             803555 ns       171352 ns          874
  ReturnErrorCodeResult/4/1024/3/32/31/real_time            1090889 ns       192249 ns          659
  ReturnErrorCodeResult/4/1024/3/32/40/real_time            1179030 ns       196463 ns          609
C ReturnStringResult/4/1024/3/32/24/real_time                937029 ns       181062 ns          707
  ReturnStringResult/4/1024/3/32/31/real_time               1462163 ns       228739 ns          453
  ReturnStringResult/4/1024/3/32/40/real_time               1477620 ns       216901 ns          536
D ReturnKatanaResult/4/1024/3/32/24/real_time                801050 ns       169369 ns          890
  ReturnKatanaResult/4/1024/3/32/31/real_time                967240 ns       175805 ns          700
  ReturnKatanaResult/4/1024/3/32/40/real_time               1024511 ns       182287 ns          658
E ReturnStringResultWithContext/4/1024/3/32/24/real_time    1280537 ns       207760 ns          581
  ReturnStringResultWithContext/4/1024/3/32/31/real_time    1736147 ns       235750 ns          425
  ReturnStringResultWithContext/4/1024/3/32/40/real_time    1546701 ns       228684 ns          468
F ReturnKatanaResultWithContext/4/1024/3/32/24/real_time     850376 ns       175213 ns          834
  ReturnKatanaResultWithContext/4/1024/3/32/31/real_time    1208895 ns       202613 ns          472
  ReturnKatanaResultWithContext/4/1024/3/32/40/real_time    1097040 ns       191791 ns          656

Unsurprisingly, using exceptions is best when there are no errors (A).
The new katana::Result is no worse than its previous implementation (B
vs D), and when returning error contexts, it is significantly better
(>10%) than using a string (C vs D and E vs F), which justifies the
additional work in its implementation.
```


ddn0:
@arthurp I'm about to embark on changing references from `katana.cpp.libstd.boost cimport std_result` to a newer katana_result type. Are there any gotchas I should know about?

```
python/katana/analytics/_triangle_count.cpp:5355:116: error: could not convert 'katana::analytics::TriangleCount(__pyx_v_pg->__pyx_obj_6katana_14property_graph_PropertyGraph::underlying.std::shared_ptr<katana::PropertyGraph>::<anonymous>.std::__shared_ptr<katana::PropertyGraph, __gnu_cxx::_S_atomic>::get(), __pyx_v_plan->__pyx_obj_6katana_9analytics_15_triangle_count_TriangleCountPlan::underlying_)' from 'basic_result<[...],katana::ErrorInfo,katana::internal::abort_policy>' to 'basic_result<[...],std::error_code,boost::outcome_v2::policy::error_code_throw_as_system_error<long unsigned int, std::error_code, void>>'
 5355 |         __pyx_t_1 = __pyx_f_6katana_9analytics_15_triangle_count_handle_result_int(katana::analytics::TriangleCount(__pyx_v_pg->underlying.get(), __pyx_v_plan->underlying_)); if (unlikely(__pyx_t_1 == ((uint64_t)-1) && __Pyx_ErrOccurredWithGIL())) __PYX_ERR(0, 123, __pyx_L4_error)
      |                                                                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                                                    |
      |                                                                                                                    basic_result<[...],katana::ErrorInfo,katana::internal::abort_policy>
error: command 'ccache' failed with exit status 1
```

arthurp:
> Are there any gotchas I should know about?

Cython does not handle non-default constructable types well. This is because it generates declarations for all variables at the top level of every function (old C style). To avoid this issue, I had to create functions which take the value that is not default constructable as an argument and do what I need. This avoids generating a local variable of the offending type.

There isn't any real way to write templated functions in Cython. So you will see that I had to write the same function several times to get it at each type I need. Referencing C++ template types does work. I don't think C++ templated functions with explicit template arguments are supported (if the template arguments are inferred, it's fine of course).

These is a bit of colaboration between `external/katana/python/katana/__init__.py` and `external/katana/python/katana/cpp/libstd/boost.pxd` to provide and use the Exception types into which the error categories are mapped. It's not too confusing, but if you don't know it's there it will look pretty odd.

Hopefully that puts you on the right path, but I'm sure there will be other surprises.

arthurp (approved):
LGTM. This is great!!!  Thanks.

If you have a regex or other way to search for the log-then-error pattern, please put it in slack or something when you make this change. I assume there are a bunch of more cases where the logging needs to be moved into the error handling. And I'm sure people have code in branches that will need updating for this.

That is not to say that all code has to be updated immediately, just that we should encourage it.

libtsuba/src/GlobalState.cpp
```diff
@@ -123,16 +123,22 @@ tsuba::OneHostOnly(const std::function<katana::Result<void>()>& cb) {
...
+    res = cb();
     if (!res) {
-      KATANA_LOG_ERROR("OneHostOnly operation failed: {}", res.error());
       failed = true;
     }
   }
 
   if (Comm()->Broadcast(0, failed)) {
+    if (!res) {
+      return KATANA_ERROR(
```
arthurp: Should this be: `res.WithContext(ErrorCode::MpiError, "failure in single host execution")`

We have reclassifying `WithContext` and this seems like the exact case to use it.

libsupport/include/katana/ErrorCode.h
```diff
@@ -6,6 +6,45 @@
...
+/// std::error_category::default_error_condition and
+/// std::error_category::equivalent.
+///
+/// The way to create a new std::error_code is to:
+/// - Define an error enum
+/// - Define a new error category that is a subclass of std::error_category
+/// - Tell the STL about your enum by specializing std::is_error_code_enum
+///   to your enum if an enum class.
+/// - Tell the STL how to create a std::error_code from your enum by defining
+///   make_error_code for your enum. This function will be found by ADL so
```
roshandathathri: What's ADL?

roshandathathri (approved):
LGTM

Didn't check python changes

libsupport/include/katana/ErrorCode.h
ddn0: Argument dependent lookup. I've updated the text to spell out the acronym.

It just means that you have a situation like this:
```cpp
namespace N1 {
struct A {};
void Fn(A);
}

namespace N2 {
struct B {};
void Fn(B);
}

N1::A a;
N2::B b;
Fn(a); // Calls N1::Fn
Fn(b); // Calls N2::Fn
```

roshandathathri: Thanks

libsupport/include/katana/Result.h
```diff
@@ -1,39 +1,247 @@
...
+      std::is_error_code_enum_v<ErrorEnum> ||
+          std::is_error_condition_enum_v<ErrorEnum>,
+      ErrorInfo>
+  WithContext(ErrorEnum err, F && fmt_string, Args && ... args) {
+    Reclassify(make_error_code(err));
+    return WithContext(
+        std::forward<F>(fmt_string), std::forward<Args>(args)...);
+  }
+
+  friend std::ostream& operator<<(std::ostream& out, const ErrorInfo& ei) {
```
tylershunt, line 165: What about making this type `fmt`-able as well? Something like https://github.com/KatanaGraph/katana/blob/abc81273802823c66947a4e555631cc502705a32/libsupport/include/katana/Uri.h#L64

tylershunt (approved):
Looks great, very excited for this.

libsupport/include/katana/Result.h
ddn0: `fmt` already knows how to call `operator<<` if necessary.

One reason to configure `fmt` specifically is if you don't want the same output as `operator<<`:

https://github.com/KatanaGraph/katana/blob/master/libsupport/include/katana/Logging.h#L20

But if you are happy with `operator<<`, `fmt` will also be happy with calling it.

I've added something to the unit test to indicate the result should be the same if you print with `fmt` and `operator<<`.

tylershunt: oh nice! I had no idea; glad I asked so that I could learn about it. It turns out that `fmt` is just very good (which we already knew)

PR #114, opened 2021-03-11
Add a min non-zero capacity for PODResizeableArray

roshandathathri:


libgalois/include/katana/PODResizeableArray.h
```diff
@@ -138,7 +140,7 @@ class PODResizeableArray {
         data_ = NULL;
         capacity_ = 0;
       }
-    } else if (size_ < capacity_) {
+    } else if ((size_ < capacity_) && (size_ < kMinNonZeroCapacity)) {
```
ddn0: We don't want to realloc if `size >= kMinNonZeroCapacity`?
 
and below, regarding the statement `capacity_ = size_`. If `size < kMinNonZeroCapacity`, capacity_ could be less than kMinNonZeroCapacity (and non zero), which seems non-intuitive.

roshandathathri: Yes, sorry. That was a mistake. It should be `size_ >= kMinNonZeroCapacity`

libgalois/include/katana/PODResizeableArray.h
```diff
@@ -139,11 +141,19 @@ class PODResizeableArray {
         capacity_ = 0;
       }
     } else if (size_ < capacity_) {
-      _Tp* new_data_ = static_cast<_Tp*>(
-          realloc(reinterpret_cast<void*>(data_), size_ * sizeof(_Tp)));
-      KATANA_LOG_DEBUG_ASSERT(new_data_);
-      data_ = new_data_;
-      capacity_ = size_;
+      if (size_ >= kMinNonZeroCapacity) {
```
ddn0: I feel like it would be simpler to write something like:
```cpp
else if (size_ < capacity_) {
  new_cap = std::max(size_, kMinNonZeroCapacity);
  new_data_ = ....

  capacity_ = new_cap;
}

```

ddn0 (approved):

PR #115, opened 2021-03-12
Impove error output

ddn0:
Fix two bugs:

One, the error context is not reset upon creation of new errors, leading
to an accumulation of unrelated error messages. Instead, reset the state
of error info thread-local buffer when a new error info is created.

Two, when an error is created without a context and then a context is
added, e.g.,

  ErrorInfo ei = ErrorCode::InvalidArgument;
  ei = ei.WithContext("more");

the resulting message is simply "more". Instead prepend some
representation of the original error like "more: invalid argument".

### DIFF
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -117,6 +117,13 @@ public:
   ErrorInfo(ErrorEnum && err)
       : ErrorInfo(make_error_code(std::forward<ErrorEnum>(err))) {}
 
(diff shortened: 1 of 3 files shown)


ddn0:
I spent too much time performance tuning and not enough time looking at error messages :D

roshandathathri (approved):
LGTM

arthurp (approved):
LGTM

tylershunt (approved):
LGTM

PR #119, opened 2021-03-13
RDG creates a valid local_to_global_id_

witchel:
Small PR to support enterprise work.
 o RDG constructor should create a valid local_to_global_id_.  It does now.
 o Convenience function for creating an empty arrow chunked array.
 o New error code for failing a graph update.

### DIFF
--- a/libsupport/src/ArrowInterchange.cpp
+++ b/libsupport/src/ArrowInterchange.cpp
@@ -57,3 +57,16 @@ katana::Shuffle(const std::shared_ptr<arrow::ChunkedArray>& original) {
   KATANA_LOG_ASSERT(res.ok());
   return IndexedTake(original, indices);
 }
+
+std::shared_ptr<arrow::ChunkedArray>
+katana::EmptyChunkedArray(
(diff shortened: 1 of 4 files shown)


libsupport/src/ArrowInterchange.cpp
```diff
@@ -57,3 +57,17 @@ katana::Shuffle(const std::shared_ptr<arrow::ChunkedArray>& original) {
   KATANA_LOG_ASSERT(res.ok());
   return IndexedTake(original, indices);
 }
+
+std::shared_ptr<arrow::ChunkedArray>
+katana::EmptyChunkedArray(
+    const std::shared_ptr<arrow::DataType>& type, int64_t length) {
+  // Creating an int64 array, but this array shouldn't be used directly
```
tylershunt: Stale comment? I think you've parameterized the type if I understand this code correctly.

tylershunt (approved):
LGTM

libsupport/src/ArrowInterchange.cpp
witchel: Busted!

PR #125, opened 2021-03-22
tsuba: provide interfaces to read and write parquet

tylershunt:
I plan to use these interfaces to support out-of-core properties for import.

### DIFF
--- a/libtsuba/src/AddProperties.cpp
+++ b/libtsuba/src/AddProperties.cpp
@@ -5,117 +5,26 @@
 #include "katana/Result.h"
 #include "tsuba/Errors.h"
 #include "tsuba/FileView.h"
+#include "tsuba/ParquetReader.h"
 
 namespace {
 
-katana::Result<std::shared_ptr<arrow::ChunkedArray>>
-ChunkedStringToLargeString(const std::shared_ptr<arrow::ChunkedArray>& arr) {
-  arrow::LargeStringBuilder builder;
-
-  for (const auto& chunk : arr->chunks()) {
-    std::shared_ptr<arrow::StringArray> string_array =
-        std::static_pointer_cast<arrow::StringArray>(chunk);
-    for (uint64_t i = 0, size = string_array->length(); i < size; ++i) {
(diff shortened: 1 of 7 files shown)


ddn0 (approved):

libtsuba/src/ParquetReader.cpp
```diff
@@ -0,0 +1,209 @@
...
+Result<std::unique_ptr<tsuba::ParquetReader>>
+tsuba::ParquetReader::Make(std::optional<Slice> slice) {
+  return std::unique_ptr<ParquetReader>(new ParquetReader(slice));
+}
+
+Result<std::shared_ptr<arrow::Table>>
+tsuba::ParquetReader::ReadFromUriSliced(const katana::Uri& uri) {
+  auto slice = slice_.value();
+  if (slice.offset < 0 || slice.length < 0) {
+    return tsuba::ErrorCode::InvalidArgument;
```
scober, line 82: Maybe add some context? I'm worried this will be confusing because `slice` isn't actually an argument to this function.

scober (approved):
LGTM

libtsuba/src/ParquetReader.cpp
tylershunt: My assumption was that it was private and the relevant context was "don't call this directly", but fair enough, I'll add a quick comment.

libtsuba/src/AddProperties.cpp
```diff
@@ -5,117 +5,26 @@
 #include "katana/Result.h"
 #include "tsuba/Errors.h"
 #include "tsuba/FileView.h"
+#include "tsuba/ParquetReader.h"
 
 namespace {
 
-katana::Result<std::shared_ptr<arrow::ChunkedArray>>
-ChunkedStringToLargeString(const std::shared_ptr<arrow::ChunkedArray>& arr) {
```
witchel, line 12: Really nice to see all this arrow code exiting AddProperties.

witchel (commented):
I'm late to the party, but I like to tag along.

PR #126, opened 2021-03-22
Add dependency checks to katana_setup.py

arthurp:
This makes direct calls to setup.py produce reasonable errors when
dependencies are not satified. This should also make it easier to
build python packages for unusual environments (like dpkg native
packages with pip python packages).

### DIFF
--- a/python/katana_setup.py
+++ b/python/katana_setup.py
@@ -1,17 +1,13 @@
+import logging
+import subprocess
 import sys
 import os
-import warnings
+import tempfile
+from distutils.errors import CompileError
+from functools import lru_cache
 from pathlib import Path
-import numpy
 import setuptools
-import Cython.Build
 import configparser
-from packaging.version import Version
-
-try:
-    from sphinx.setup_command import BuildDoc
-except ImportError:
-    BuildDoc = None
 
 import generate_from_jinja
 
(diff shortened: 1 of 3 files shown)


arthurp:
@DataBoyTX I have you as a reviewer because you ran into this problem last week. If you could try `python setup.py build` again and see if the output tells you what you need to do to correctly build Katana, that would be lovely. I just want to get eyes on the output to make sure it makes sense to people not steeped in this system.

python/katana_setup.py
```diff
@@ -202,7 +322,9 @@ def setup(*, source_dir, package_name, doc_package_name, **kwargs):
         package_data={"": [str(f) for f in pxd_files]},
         package_dir={"": str(source_dir)},
         tests_require=["pytest"],
-        setup_requires=setup_requires,
+        # NOTE: Do not use setup_requires. It doesn't work properly for our needs because it does install the packages
```
ddn0: "it does install"?
would either "it doesn't install" or "it installs" be clearer?

python/katana_setup.py
```diff
@@ -33,6 +29,111 @@ def unique_list(l):
...
+@lru_cache(maxsize=None)
+def require_python_module(module_name, ge_version=None, lt_version=None):
+    v_str = ""
+    if ge_version:
+        v_str += f">={ge_version}"
+    if lt_version:
+        v_str += f"<{lt_version}"
+    if ge_version or lt_version:
+        v_str = f" ({v_str})"
+    print(f"Checking for Python package '{module_name}'{v_str}...", end="")
```
ddn0, line 52: qq: why the single quotes around `module_name` here but not in the exception messages below?

ddn0 (approved):

DataBoyTX:
changes looked good to me.   thanks

PR #129, opened 2021-03-25
Add max_global_node_id field to PartitionMetadata

scober:
The simplest way to support sparse global node ids is to maintain a `max_global_node_id` in addition to a `num_global_nodes`. Add this field to `PartitionMetadata` and support writing it to disk and reading it from disk when present.

### DIFF
--- a/libtsuba/src/RDGPartHeader.cpp
+++ b/libtsuba/src/RDGPartHeader.cpp
@@ -416,6 +416,7 @@ tsuba::to_json(json& j, const tsuba::PartitionMetadata& pmd) {
       {"is_outgoing_edge_cut", pmd.is_outgoing_edge_cut_},
       {"is_incoming_edge_cut", pmd.is_incoming_edge_cut_},
       {"num_global_nodes", pmd.num_global_nodes_},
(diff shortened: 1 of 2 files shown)


libtsuba/src/RDGPartHeader.cpp
```diff
@@ -425,18 +426,43 @@ tsuba::to_json(json& j, const tsuba::PartitionMetadata& pmd) {
...
+  }
+  if (auto it = j.find("num_edges"); it != j.end()) {
+    it->get_to(pmd.num_edges_);
+  }
+  if (auto it = j.find("num_owned"); it != j.end()) {
+    it->get_to(pmd.num_owned_);
+  }
+  if (auto it = j.find("cartesian_grid"); it != j.end()) {
+    it->get_to(pmd.cartesian_grid_);
+  }
```
tylershunt: Would prefer to leave these required. We can do `if (..find)...` for new fields with the aim of removing them all eventually when we have better migration tools.

scober: How does it look now?

tylershunt (approved):

PR #132, opened 2021-03-26
Fix handling of clones without an origin remote

arthurp:


scripts/katana_version/__init__.py
```diff
@@ -112,9 +112,11 @@ def _find_katana_remotes(dir):
         if not origin_remote:
             non_upstream_remotes = remotes - {upstream_remote}
             if len(non_upstream_remotes) == 1:
-                origin_remote = list
+                origin_remote = list(non_upstream_remotes)[0]
             else:
-                raise ConfigurationError(f"{name}: Could not find an origin remote.")
+                raise ConfigurationError(
+                    f"{name}: Could not find an origin remote. Make sure you have a remote named 'origin'."
```
roshandathathri: Why is this required?

roshandathathri: Are you changing this? If not, let me know and I can approve

arthurp: Let me check. It's definitely required to do the version management actions, but it may not be needed to get a version. In which case crashing here is a bug.

arthurp: Working on it.

roshandathathri (approved):
LGTM

PR #133, opened 2021-03-26
Switch CI to use Mambaforge instead of miniconda.

arthurp:
This reduces CI environment setup time noticably.

https://github.com/conda-forge/miniforge#mambaforge


arthurp:
Unscientific measurements (just comparing a few samples) shows that this saves 3 minutes on python builds jobs and more than 15 minutes on conda packaging jobs (packaging creates lots of environments).

ddn0 (approved):

PR #144, opened 2021-04-06
support: drop trailing slash for real

tylershunt:
`+` and `?` are both greedy, but the `+` operator is apparently more
greedy. To make this regex actually drop this slash when it's around
I need another `?` which is how you tell `+` not to be greedy.

### DIFF
--- a/libsupport/src/Uri.cpp
+++ b/libsupport/src/Uri.cpp
@@ -17,7 +17,7 @@
 namespace {
 
 // get scheme and path, always drop trailing slash
-const std::regex kUriRegex("(?:([a-zA-Z0-9]+)://)?(.+)/?");
+const std::regex kUriRegex("^(?:([a-zA-Z0-9]+)://)?(.+?)/?$");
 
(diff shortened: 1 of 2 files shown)


libsupport/src/Uri.cpp
```diff
@@ -17,7 +17,7 @@
 namespace {
 
 // get scheme and path, always drop trailing slash
-const std::regex kUriRegex("(?:([a-zA-Z0-9]+)://)?(.+)/?");
+const std::regex kUriRegex("^(?:([a-zA-Z0-9]+)://)?(.+?)/?$");
```
ddn0, line 20: Can you add a unit test?

ddn0 (approved):

libsupport/src/Uri.cpp
tylershunt: Done.

libsupport/src/Uri.cpp
```diff
@@ -17,7 +17,7 @@
...
```
witchel, line 21: Why did you add the start (`^`) and end (`$`) markers?

PR #152, opened 2021-04-08
meta: increase version number

tylershunt:
As pointed out 0.1.11 is already out there.


arthurp (changes requested):
Can we use `scripts/version` for this? It would be a good test and good to be consistent. It will create a new PR, but that's fine.

tylershunt:
> Can we use `scripts/version` for this? It would be a good test and good to be consistent. It will create a new PR, but that's fine.

Oh! sure! let me do that.

arthurp:
There is some documentation here: https://github.com/KatanaGraph/katana/blob/master/CONTRIBUTING.md#version-management-tool. but I just realized it is actually missing the command you need for this. The command you need is `scripts/version bump <new version>`.

PR #158, opened 2021-04-11
Add timer guard to allow easier timing of functions with multiple ret…

arthurp:
Simultaneous: https://github.com/KatanaGraph/katana-enterprise/pull/989

### DIFF
--- a/libgalois/include/katana/Timer.h
+++ b/libgalois/include/katana/Timer.h
@@ -104,6 +104,25 @@ public:
   uint64_t get_usec() const { return 0; }
 };
 
+//! Start and stop a TimeAccumulator in the scope of the guard.
+template <typename T>
+class [[nodiscard]] TimerGuard {
+  // The template parameter is required because TimeAccumulator::start and stop
+  // are not virtual. The template parameter is inferrable, so it's not a big
+  // deal, and there is no need to change the original class.
+  T& timer_;
+
+public:
+  TimerGuard(T & timer) : timer_(timer) { timer_.start(); }
+
+  TimerGuard(const TimerGuard&) = delete;
+  TimerGuard(TimerGuard &&) = delete;
+  TimerGuard& operator=(const TimerGuard&) = delete;
+  TimerGuard& operator=(TimerGuard&&) = delete;
+
+  ~TimerGuard() { timer_.stop(); }
+};
+
 template <typename F>
 void
 timeThis(const F& f, const char* const name) {
--- a/libgalois/include/katana/Timer.h
+++ b/libgalois/include/katana/Timer.h
@@ -123,17 +123,5 @@ public:
   ~TimerGuard() { timer_.stop(); }
 };
 
-template <typename F>
-void
-timeThis(const F& f, const char* const name) {
-  StatTimer t("Time", name);
-
-  t.start();
-
-  f();
-
-  t.stop();
-}
-
 }  // end namespace katana
 #endif
--- a/libgalois/include/katana/Profile.h
+++ b/libgalois/include/katana/Profile.h
@@ -53,7 +53,11 @@ profileVtune(const F& func, const char* region) {
 
   __itt_resume();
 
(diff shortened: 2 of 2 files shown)


libgalois/include/katana/Timer.h
```diff
@@ -104,6 +104,24 @@ class CondStatTimer<false> {
   uint64_t get_usec() const { return 0; }
 };
 
+template <typename T>
+class TimerGuard {
+public:
+  T timer;
+
+  template <typename... Args>
+  TimerGuard(Args&&... args) : timer(std::forward<Args>(args)...) {
```
ddn0: Shouldn't this just take a reference to a timer rather than construct a new timer?

libgalois/include/katana/Timer.h
```diff
@@ -104,6 +104,24 @@ class CondStatTimer<false> {
...
+  }
+
+  TimerGuard(const TimerGuard&) = delete;
+  TimerGuard(TimerGuard&&) = delete;
+  TimerGuard& operator=(const TimerGuard&) = delete;
+  TimerGuard& operator=(TimerGuard&&) = delete;
+
+  ~TimerGuard() { timer.stop(); }
+};
+
```
ddn0, line 125: comment on `timeThis` below: if we add `TimerGuard`, I think it would be good to remove `timeThis` so we have just one way to do this.

libgalois/include/katana/Timer.h
arthurp: I didn't want to keep a reference in the field of this class because that can create stale reference issues, and would force the client code to use two lines, one for the timer and one for the guard. The timer classes are not copyable or movable, so there isn't a clean way I can see to do this other than what I did here.

ddn0: I was thinking that TimerGuard was similar in spirit to std::lock_guard: https://en.cppreference.com/w/cpp/thread/lock_guard/lock_guard so it would enable things like:

```cpp
void Func() {
  TimeAccumulator t;
  {
    TimerGuard guard(t);
    Fn1();
  }

  {
    TimerGuard guard(t);
    Fn2();
  }
}
```

but it sounds like you want

```cpp
void Func() {
  TimerGuard t(TimeAccumulator());
  Fn1();
}
```

?

Also would it make sense to add `[[nodiscard]]` to the constructor?

arthurp: Yes. I was implementing the latter. Because the majority of the time there is only one region, but honestly the extra line of code isn't too bad and your idea is more generic. My only concern is making sure that `TimerGuard t(TimeAccumulator());` is an error if we expect the former case, since it could cause stack corruption, maybe C++ handles this these days, I'll have to check.

I will reimplement with your approach and with `[[nodiscard]]`.

libgalois/include/katana/Timer.h
```diff
@@ -104,17 +104,24 @@ class CondStatTimer<false> {
...
 };
 
-template <typename F>
-void
-timeThis(const F& f, const char* const name) {
-  StatTimer t("Time", name);
+//! Start and stop a TimeAccumulator in the scope of the guard.
+template <typename T>
+class [[nodiscard]] TimerGuard {
+  // The template parameter is required because TimeAccumulator::start and stop
```
ddn0, line 110: Maybe it is just me, but I don't think it is too unexpected that TimerGuard is a templated class, unless you mean that TimerGuard really should only be used with TimeAccumulator?

Even in that case, I think `TimerGuard` being ambivalent to being used with `TimeAccumulator` versus any class that has `start` and `stop` seems reasonable.

ddn0 (approved):

libgalois/include/katana/Timer.h
arthurp: That's fair. I just wanted to document the design choice since the class very much is designed to be used with `TimeAccumulator`. Though as you say it's not actually specific to that.

PR #160, opened 2021-04-13
Skip gram porting to katana

uditagar:
This is a porting of the skip gram based embedding model to the katana code base. The base code written in C is available here: https://github.com/tmikolov/word2vec/blob/master/word2vec.c


lonestar/analytics/cpu/skip-gram/Embedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+    while (ss >> val) {
+      walk.push_back(val);
+    }
+
+    random_walks->push_back(std::move(walk));
+  }
+}
+
+void
+BuildVocab(
```
gurbinder533, line 82: Parallelize this function. Use katana::gstl::Map<uint32_t, uint32_t>. See CC for example: https://github.com/KatanaGraph/katana/blob/master/libgalois/src/analytics/connected_components/connected_components.cpp

lonestar/analytics/cpu/skip-gram/Embedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+static cll::opt<uint32_t> numNegSamples(
+    "numNegSamples", cll::desc("Number of negative samples (default value 5)"),
+    cll::init(5));
+
+static cll::opt<uint32_t> numIterations(
+    "numIterations",
+    cll::desc("Number of Training Iterations (default value 5)"), cll::init(5));
+
+static cll::opt<uint32_t> minCount(
```
gurbinder533: mininumFrequency.

lonestar/analytics/cpu/skip-gram/Embedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+      }
+    }*/
+    //  of << "\n";
+  }
+
+  of.close();
+}
+
+void
+RefineRandomWalks(
```
gurbinder533, line 172: Document the functionality in the comment above.

lonestar/analytics/cpu/skip-gram/Embedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  uint32_t num_trained_tokens;
+
+  BuildVocab(random_walks, &vocab, &vocab_multiset, &num_trained_tokens);
+
+  std::vector<std::vector<uint32_t>> refined_random_walks;
+
+  RefineRandomWalks(random_walks, &refined_random_walks, vocab);
+
+  HuffmanCoding huffman_coding(&vocab, &vocab_multiset);
```
gurbinder533, line 209: Shouldn't this be optional?

gurbinder533 (commented):
Please add a few lines to document the intent of all the functions.

lonestar/analytics/cpu/skip-gram/Embedding.cpp
uditagar: yeah, probably. I don't know how to do it easily at the moment. thank you

tylershunt:
I think this is no longer active; closing. Please re-open if I'm wrong!

PR #166, opened 2021-04-16
tsuba: add an interface to read number of rows

tylershunt:
Call should be cheap because it just looks at the metadata.


scober (approved):
LGTM

PR #174, opened 2021-04-22
Fix version script to handle more corner cases.

arthurp:
This is based on an issue encountered by @narendrachaudhary51.

### DIFF
--- a/scripts/katana_version/__main__.py
+++ b/scripts/katana_version/__main__.py
@@ -1,8 +1,10 @@
 import argparse
+import datetime
 import logging
 import os
 import platform
 import re
+import sys
 import threading
 from enum import Enum
 from os import environ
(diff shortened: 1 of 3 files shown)


scripts/katana_version/__main__.py
```diff
@@ -189,35 +224,118 @@ def provenance_subcommand(args):
...
+        "--prefix",
+        "-p",
+        help="Prefix to print before the first key-value pair. Use the source luck.",
+        dest="prefix",
+        type=str,
+    )
+    parser.add_argument(
+        "--suffix",
+        "-s",
+        help="Suffix to print after the last key-value pair. Use the source luck.",
```
ddn0: "source luck"? I don't understand.

scripts/katana_version/__main__.py
```diff
@@ -189,35 +224,118 @@ def provenance_subcommand(args):
         values.update(
             katana_enterprise_hash=git.get_hash(git.HEAD, katana_enterprise_repo_path, exclude_dirty=(SUBMODULE_PATH,))
         )
+        values.update(katana_enterprise_head=git.get_hash(git.HEAD, katana_enterprise_repo_path, pretend_clean=True))
+
+    # Sort by the key for easier reading later
+    values = dict(sorted(values.items(), key=lambda kv: kv[0]))
```
ddn0, line 235: qq: I suppose the reason for this manual json encoding is because yaml is not part of the python standard library and we wanted a uniform way to print out json, yaml, python, and `#define`-based output?

arthurp: Yes. As the commit message implies, this was not an efficient use of time. So I didn't try to optimize. I just wanted to push it to stop fiddling.

scripts/katana_version/__main__.py
arthurp: I meant "Use the source, Luke." Just a joke meaning "This is not well documented, read the source if you want to know more."

scripts/katana_version/__main__.py
```diff
@@ -165,20 +167,53 @@ def provenance_subcommand(args):
...
+        re.compile(r".*PATH"),
+    }
 
     for var in environment_to_capture:
-        if var in environ:
-            values[var.lower()] = environ[var]
+        if isinstance(var, re.Pattern):
+            vars = [k for k in environ.keys() if var.fullmatch(k)]
+        else:
+            vars = [var]
```
ddn0, line 194: This might be too paranoid so feel free to ignore, but perhaps we can have a defensive check here to make sure we definitely do not include any environment variable that has `(?![a-z0-9])key` (or something) in its name.

ddn0 (approved):

PR #184, opened 2021-04-26
Don't swap src and dst when returning edge iterators

aneeshdurg:


roshandathathri:
`FindEdgeWithLabel` needs to be fixed too

roshandathathri (approved):
Thanks!

PR #190, opened 2021-04-27
Basic support for global edge ids

scober:
The basic in-memory and on-storage graph representations need to be updated to support differentiated global and local edge ids.

### DIFF
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -240,11 +240,24 @@ public:
 
   // TODO(witchel): ChunkedArray is inherited from arrow::Table interface but this is
   // really a ChunkedArray of one chunk, change to arrow::Array.
-  const std::shared_ptr<arrow::ChunkedArray>& host_to_owned_global_ids() const {
-    return rdg_.host_to_owned_global_ids();
+  const std::shared_ptr<arrow::ChunkedArray>& host_to_owned_global_node_ids()
+      const {
+    return rdg_.host_to_owned_global_node_ids();
   }
-  void set_host_to_owned_global_ids(std::shared_ptr<arrow::ChunkedArray>&& a) {
-    rdg_.set_host_to_owned_global_ids(std::move(a));
+  void set_host_to_owned_global_node_ids(
+      std::shared_ptr<arrow::ChunkedArray>&& a) {
+    rdg_.set_host_to_owned_global_node_ids(std::move(a));
+  }
+
+  // TODO(witchel): ChunkedArray is inherited from arrow::Table interface but this is
+  // really a ChunkedArray of one chunk, change to arrow::Array.
+  const std::shared_ptr<arrow::ChunkedArray>& host_to_owned_global_edge_ids()
+      const {
+    return rdg_.host_to_owned_global_edge_ids();
+  }
+  void set_host_to_owned_global_edge_ids(
+      std::shared_ptr<arrow::ChunkedArray>&& a) {
+    rdg_.set_host_to_owned_global_edge_ids(std::move(a));
   }
 
   // TODO(witchel): ChunkedArray is inherited from arrow::Table interface but this is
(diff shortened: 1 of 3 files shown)


libgalois/include/katana/PropertyGraph.h
```diff
@@ -240,11 +240,24 @@ class KATANA_EXPORT PropertyGraph {
...
+  }
+  void set_host_to_owned_global_node_ids(
+      std::shared_ptr<arrow::ChunkedArray>&& a) {
+    rdg_.set_host_to_owned_global_node_ids(std::move(a));
+  }
+
+  // TODO(witchel): ChunkedArray is inherited from arrow::Table interface but this is
+  // really a ChunkedArray of one chunk, change to arrow::Array.
+  const std::shared_ptr<arrow::ChunkedArray>& host_to_owned_global_edge_ids()
+      const {
```
witchel, line 255: Well, I hate to see my TODOs multiply.  If I gave you a version of MarshalVector that returned an Array instead of a ChunkedArray, would you use it?

libtsuba/src/RDG.cpp
```diff
@@ -451,12 +472,16 @@ tsuba::RDG::DoMake(const katana::Uri& metadata_dir) {
   }
 
   KATANA_LOG_DEBUG(
-      "ReadPartMetadata master sz: {} mirrors sz: {} h2owned sz: {} l2u sz: "
+      "ReadPartMetadata master sz: {} mirrors sz: {} h2on sz: {} h20e sz: {} "
```
witchel: Why do we print this all the time in debug mode?

libtsuba/src/RDG.cpp
```diff
@@ -268,15 +274,30 @@ tsuba::RDG::WritePartArrays(const katana::Uri& dir, tsuba::WriteGroup* desc) {
...
+    auto h2on_res = StoreArrowArrayAtName(
+        host_to_owned_global_node_ids_, dir, kHostToOwnedGlobalNodeIDsPropName,
+        desc);
+    if (!h2on_res) {
+      return h2on_res.error();
+    }
+    next_properties.emplace_back(tsuba::PropStorageInfo{
+        .name = kHostToOwnedGlobalNodeIDsPropName,
+        .path = std::move(h2on_res.value()),
+        .persist = true,
```
witchel, line 294: real nit, but `h2on `and `h2oe `look an awful lot alike.

witchel (approved):
Persistence is the key.

libgalois/include/katana/PropertyGraph.h
scober: Sure. I just copied this comment from the one above `host_to_owned_global_node_ids()` so I wasn't going to way in on design here.

libtsuba/src/RDG.cpp
scober: Good question. I think @tylershunt may have written the first version of this print?

libtsuba/src/RDG.cpp
scober: Yeah, I'll try to come up with a better name.

libtsuba/src/RDG.cpp
```diff
@@ -43,7 +43,8 @@ namespace {
 // special partition property names
 const char* kMirrorNodesPropName = "mirror_nodes";
 const char* kMasterNodesPropName = "master_nodes";
-const char* kHostToOwnedGlobalIDsPropName = "host_to_owned_global_ids";
+const char* kHostToOwnedGlobalNodeIDsPropName = "host_to_owned_global_node_ids";
```
roshandathathri, line 46: This will break backward compatibility. Please treat it the same way kDeprecatedLocalToGlobalIDPropName is treated.

roshandathathri (approved):
LGTM

libtsuba/src/RDG.cpp
scober: Good catch

PR #193, opened 2021-04-28
Adding deterministic louvain

uditagar:


libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -173,6 +173,200 @@ struct LouvainClusteringImplementation
     return prev_mod;
   }
 
+  katana::Result<double> LouvainDeterministic(
+      katana::PropertyGraph* pfg, double lower,
+      double modularity_threshold_per_round, uint32_t& iter) {
+    katana::StatTimer TimerClusteringTotal("Timer_Clustering_Total");
+    TimerClusteringTotal.start();
```
arthurp: Nit: This timer is not stopped if the algorithm exits early (due to an error). I introduced a way to fix this a bit ago. You can write:

```
      katana::StatTimer timer(...);
      katana::TimerGuard timer_guard(timer);
```

And when `timer_guard` goes out of scope, the timer will stop.

libgalois/include/katana/analytics/louvain_clustering/louvain_clustering.h
```diff
@@ -81,6 +82,22 @@ class LouvainClusteringPlan : public Plan {
         max_iterations,
         min_graph_size};
   }
+
+  static LouvainClusteringPlan Deterministic(
```
arthurp, line 97: Document the algorithms a bit.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -173,6 +173,200 @@ struct LouvainClusteringImplementation
     return prev_mod;
   }
 
+  katana::Result<double> LouvainDeterministic(
```
arthurp, line 179: This function looks very similar to the one above. Can we abstract things a bit and reduce the duplication? The simplest approach would be to lift some common elements into a shared function or three. But it might also be possible to combine the functions entirely and just have a flag parameter to switch modes. (or template parameter if we must for some reason.)

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -173,6 +173,200 @@ struct LouvainClusteringImplementation
...
+    c_update_add.deallocate();
+
+    c_update_subtract.destroy();
+    c_update_subtract.deallocate();
+
+    local_target.destroy();
+    local_target.deallocate();
+
+    in_bag.destroy();
+    in_bag.deallocate();
```
arthurp: `LargeArray` does all this in it's destructor. There is no need for this, here or in the original version. I wonder why this was written in the first place.

arthurp (changes requested):
Can you write a test for the new algorithm? I know non-deterministic algorithms are super hard to test, but a deterministic version should be easier. Testing is important.

Looking pretty good though. Just a few things I would like to address.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
uditagar: I am not sure of how to do that at the moment. I will think more on this.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
uditagar: Can you point me to an example? I will use then. Thank you.

uditagar:
> Can you write a test for the new algorithm? I know non-deterministic algorithms are super hard to test, but a deterministic version should be easier. Testing is important.
> 
> Looking pretty good though. Just a few things I would like to address.

I am not sure how to do that too. If you can give me some pointers, I will go ahead and do it then. Thanks

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
arthurp: https://github.com/KatanaGraph/katana-enterprise/blob/c0f853e2267b34129dd78d848e07a8035c9ea81f/libquery/src/MatchQuery.cpp#L2549-L2550

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
arthurp: For the moment, you can just document it with a `TODO` and move on. I'm not sure how to pull out the duplication either. Or maybe there isn't as much duplication as it appears. I may not be understanding.

arthurp:
> I am not sure how to do that too. If you can give me some pointers, I will go ahead and do it then. Thanks

I was sure I responded to this. I guess I didn't. Sorry. 

When I asked for testing I meant a test like the other algorithm tests. For example, BFS has this:

https://github.com/KatanaGraph/katana/blob/b13d62a04a0d91f88e39db85bbbcec11db2e7ff4/lonestar/analytics/cpu/bfs/CMakeLists.txt#L1-L5

The idea is to provide a CLI to call the function, pick an input file (there are various already), and then create a verified output (that is known to be correct) that can be used to check the output of the algorithm during testing.

I know we probably don't have idea. Don't worry about it for now. But it's something we would like to have eventually.

arthurp (approved):
LGTM.

Please, create an issue or TODO to remind us that this algorithm needs tests eventually.

gurbinder533 (approved):
Looks good. 
One way to create ground truth files for this is to run this on a single machine and single thread and make sure high-level statistics match (or are better) some other frameworks such as neo4j and then write out the community ids to a file and then check against those in the tests.

uditagar:
> Looks good.
> One way to create ground truth files for this is to run this on a single machine and single thread and make sure high-level statistics match (or are better) some other frameworks such as neo4j and then write out the community ids to a file and then check against those in the tests.

This is fine. But the issue is that any change to the implementaton will then change the outcome as well. Let me think over it. I have already created a issue for this for myself. https://github.com/KatanaGraph/katana/issues/204

gurbinder533:
> > Looks good.
> > One way to create ground truth files for this is to run this on a single machine and single thread and make sure high-level statistics match (or are better) some other frameworks such as neo4j and then write out the community ids to a file and then check against those in the tests.
> 
> This is fine. But the issue is that any change to the implementaton will then change the outcome as well. Let me think over it. I have already created a issue for this for myself. #204

Yes, I thought more abt this. So the way to get around it is to have another phase that picks the minimum node_id as the cluster id for all the nodes similar to what we do in connected components.

PR #196, opened 2021-04-29
Make libnuma into a soft dependency

### LINKED ISSUE #185: Invalid version error during installation
Following discussion from https://github.com/KatanaGraph/katana/issues/70.
I do the following steps to install katana. I get invalid version errors at the corresponding steps.

1. git clone https://github.com/KatanaGraph/katana.git
2. SRC_DIR=/home/katana
3. conda env create --name katana-dev --file $SRC_DIR/conda_recipe/environment.yml
4. conda activate katana-dev 
5. conda install -c conda-forge numactl-devel-cos6-x86_64
6. BUILD_DIR=/home/katana/build
7. mkdir -p $BUILD_DIR
8. cmake -S $SRC_DIR -B $BUILD_DIR -DCMAKE_BUILD_TYPE=Release -DKATANA_LANG_BINDINGS=python -DBUILD_SHARED_LIBS=ON
9. conda build -c conda-forge $SRC_DIR/conda_recipe/
         **ERROR - anaconda3/lib/python3.8/site-packages/conda/models/version.py", line 164, in __init__
    raise InvalidVersionSpec(vstr, "empty version string")
conda.exceptions.InvalidVersionSpec: Invalid version '': empty version string**

10. cd $BUILD_DIR
11. make -j 8
         **ERROR - /home/katana/build/katana_python_build/scripts/katana_version/version.py", line 177, in get_explicit_version
    explicit_version = tag_version or get_config_version(
  File "/home/katana/build/katana_python_build/scripts/katana_version/version.py", line 197, in get_config_version
    with open(Path(__file__).parent.parent.parent / CONFIG_VERSION_PATH, "rt") as version_file:**

12. make clean
13.  rm CMakeCache.txt
14. cd $SRC_DIR
15. cmake -S $SRC_DIR -B $BUILD_DIR -DCMAKE_BUILD_TYPE=Release
16. cd $BUILD_DIR
17. make -j 8
       **Done**

arthurp:
This uses `dlopen` to get the 3 symbols we need from `libnuma`. If the load fails we just assume NUMA information is not available and use the existing "assume sockets are nodes" fallback. All this allows katana built with libnuma to run on machines that don't have it.

Fixes: #185 (because I ran into it while doing this PR).

Fixes: ENG-324


arthurp:
Should I add a CI job which runs `conda_recipe/test_install_conda_packages.py`?

libgalois/src/HWTopoLinux.cpp
```diff
@@ -57,41 +58,69 @@ struct cpuinfo {
...
-    if (!numaAvail)
-      katana::gWarn(
+  static bool numa_avail = false;
+
+  if (!warn_once) {
+    warn_once = true;
+    LoadLibNuma();
+    numa_avail = dynamic_numa_available && dynamic_numa_available() >= 0;
+    numa_avail = numa_avail && dynamic_numa_num_configured_nodes() > 0;
+    if (!numa_avail) {
```
ddn0, line 104: We actually have `KATANA_LOG_WARN_ONCE`: https://github.com/KatanaGraph/katana/blob/master/libsupport/include/katana/Logging.h#L139

conda_recipe/test_install_conda_packages.py
```diff
@@ -0,0 +1,37 @@
+#! /bin/python
+
+import argparse
+from pathlib import Path
+from subprocess import Popen, PIPE
+from sys import stderr
+import tarfile
+
+
+def main():
```
ddn0: This test is great! It would be good to put this in the CI; otherwise, it will have a tendency to rot.

I'm wondering if these files might make sense under `tests/` or at least outside of `conda_recipe/`. I think if there is a common pattern running tests, even if we might skip some depending on some configuration, it will be one less thing to remember.

ddn0 (approved):

conda_recipe/test_install_conda_packages.py
arthurp: Should I recreate the top-level `tests` directory? We killed it because it contained just Python tests before. But a top-level `tests` would make sense for packaging tests like that and for formatting checkers, since we run them as tests in CI.

I'll look into running this in CI as a dependent job after the conda packaging job. I suspect we don't want to make it required though, since conda packaging still takes an insanely long time in CI (45 minutes).

libgalois/src/HWTopoLinux.cpp
arthurp: I looked at that, but this doesn't actually just warn once. It actually checks once for numa as well. I'll use `std::call_once` in place of this hack though, and rename the flag to make that clear.

ddn0: I think the warning at Line 123 could just be replaced by `KATANA_LOG_WARN_ONCE` if you wanted. But yeah, this one is commingled with the numa check too.

conda_recipe/test_install_conda_packages.py
ddn0: Having the check be run but optional for merging is reasonable.

Having `tests/install_conda_packages_test.py` wouldn't look too weird to me.

Alternatively, `tests/python/....`,  `tests/check_foo.sh`, `tests/check_all.sh`

I'm also happy with leaving the checking scripts in `scripts` . I'm mainly looking for a common way to run various tests if possible and an obvious path for people to add more (so people don't complain about the burden of adding more things like this).

arthurp: I'm moving them to tests. In part because we have had several people say "How are you all locally running tests?" and a "tests" directory might help with that.

.github/workflows/python.yaml
```diff
@@ -110,6 +110,28 @@ jobs:
...
+  test_conda_package_install:
+    needs: build_and_test_conda_package
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        image:
+          - ubuntu:16.04
+          - ubuntu:latest
+          - centos:7
+          - centos:latest
```
arthurp, line 122: @ddn0 Did I over do the docker image matrix? Should I include anything else? They take about 5min each and all run in parallel after `build_and_test_conda_package` finishes.

ddn0: I might add ubuntu:18.04 and ubuntu:20.04 explicitly (in addition to `ubuntu:latest`). But I don't have a strong preference.

PR #197, opened 2021-04-30
galois/stats: don't write anything if stream is empty

tylershunt:
All hosts follow this code path, and we were causing problems when
different hosts stepped on each other; sometimes overwriting stats with
a zero length file, sometimes triggering and error from google storage
because of a race.

### DIFF
--- a/libgalois/src/Statistics.cpp
+++ b/libgalois/src/Statistics.cpp
@@ -256,7 +256,11 @@ katana::StatManager::Print() {
   std::ostringstream out;
   PrintStats(out);
 
-  if (auto res = tsuba::FileStore(impl_->outfile_, out.str()); !res) {
+  std::string stats = out.str();
+  if (stats.empty()) {
+    return;
+  }
+  if (auto res = tsuba::FileStore(impl_->outfile_, stats); !res) {
     KATANA_LOG_ERROR("printing stats: {}", res.error());
   }
 }


libgalois/src/Statistics.cpp
```diff
@@ -256,7 +256,11 @@ katana::StatManager::Print() {
   std::ostringstream out;
   PrintStats(out);
 
-  if (auto res = tsuba::FileStore(impl_->outfile_, out.str()); !res) {
+  std::string stats = out.str();
+  if (stats.empty()) {
+    return;
+  }
+  if (auto res = tsuba::FileStore(impl_->outfile_, stats); !res) {
```
ddn0, line 263: This seems fine, but if there was a problem with concurrent writes before, don't we still have this problem just rarer?

Given the trouble related to stats reporting, perhaps we should just switch over to just structured logging (of stats) rather than the existing Statistics code?

ddn0 (approved):

libgalois/src/Statistics.cpp
tylershunt: I'm in favor of switching wholesale to structured logging. The semantics of this call are confusing; as far as I can tell in the old version (ofstream instead of tsuba), all hosts were opening the stream but we were depending on only one of them ever writing it. I think as long as that holds this will work without a race.

ddn0: I remember the graph-query had a race in this case where multiple hosts opened the file and that would cause truncation even if only one person wrote to it. I think `ostream` with `write` always truncates? 

cc @roshandathathri

roshandathathri: Yes, I'm fine with this solution. But I think it might be better to enforce the invariant that only host 0 prints the stats. That can be done in `DistStatManager`. What do you think?

PR #198, opened 2021-04-30
Fix type of accessing inEdgeData

roshandathathri:
I have no idea why C++ compilers do not catch this type error! 
@arthurp

### DIFF
--- a/libgalois/include/katana/LC_CSR_CSC_Graph.h
+++ b/libgalois/include/katana/LC_CSR_CSC_Graph.h
@@ -395,7 +395,7 @@ public:
    */
   template <
       bool A = EdgeDataByValue, typename std::enable_if<!A>::type* = nullptr>
-  edge_iterator getOutEdge(edge_iterator ni) const {
+  EdgeIndData::value_type getOutEdge(edge_iterator ni) const {
     return inEdgeData[*ni];
   }
 


roshandathathri:
FYI: It was implicitly converting from `T` to `boost::counting_iterator<T>`

aneeshdurg (approved):
Good catch!

tylershunt:
Man C++ is just the worst.

ddn0:
Isn't this just because boost::counting_iterator<T> is implicitly convertible to T and vice versa?

Was the issue a truncation of a 64-bit values? If so, hopefully explicit ID types should help or using our own iterators classes rather than counting_iterator, the latter is a little about types.

roshandathathri:
> Isn't this just because boost::counting_iterator is implicitly convertible to T and vice versa?

Yes but that's a problem right? Is implicit conversion from `T` to `boost::counting_iterator<T>` then using the `*` operator an identity transformation?

> Was the issue a truncation of a 64-bit values?

Don't think so. 

> hopefully explicit ID types should help

Yes, definitely they will.

ddn0:
> Yes but that's a problem right? Is implicit conversion from T to boost::counting_iterator<T> then using the * operator an identity transformation?

It should be:

  https://godbolt.org/z/8oqaaq4K8

Though I'm not sure what issue you are trying to resolve.

roshandathathri:
Look at the enterprise repo PR to get the full picture:
https://github.com/KatanaGraph/katana-enterprise/pull/1083

arthurp:
I have been hoping that we would be able to get rid of the edge iterators entirely (except for iterable ranges like `graph->edges(node_id)`, of course, but no one will ever use those directly because `for-each` loops are a thing in C++). I really don't see any advantage to using iterators at all when IDs are fixed to be ints anyway. Do I misunderstand something?

PR #206, opened 2021-05-05
Improve Python API style.

arthurp:
* Change `setActiveThreads` to `set_active_threads`
* `__version__` is now a `str` instread of `bytes`


gurbinder533 (approved):
LGTM.

PR #207, opened 2021-05-05
Fix stats for sssp and bfs

gurbinder533:


libgalois/include/katana/analytics/sssp/sssp.h
```diff
@@ -127,14 +127,12 @@ KATANA_EXPORT Result<void> SsspAssertValid(
...
   /// The maximum distance across all nodes.
   double max_distance;
-  /// The sum of all node distances.
-  double total_distance;
-  /// The number of nodes reachable from the source node.
-  uint32_t n_reached_nodes;
-
-  double average_distance() const { return total_distance / n_reached_nodes; }
+  /// The average distances on visited nodes from the source node.
+  double average_visited_distance;
```
arthurp, line 135: This is now the same as `BfsStatistics` except for the type of `max_distance`. And the type here is really just a type that will (for the most part) hold the result regardless of the type of the input graph weights. Could we rename this to `TraversalStatistics` and use it for both SSSP and BFS?

arthurp (approved):
A comment of something I would like to see. But if it would take to much time, go ahead and merge. This is fine.

libgalois/include/katana/analytics/sssp/sssp.h
gurbinder533: Noted. Will do it later.

PR #210, opened 2021-05-06
Update inputs to v19 and improve python inputs download.

arthurp:
This change is needed by https://github.com/bowu/katana-enterprise/tree/no_template_dfs_fsm (@bowu).

### DIFF
--- a/python/katana/example_utils.py
+++ b/python/katana/example_utils.py
@@ -1,26 +1,34 @@
 import os
+from pathlib import Path
 import tarfile
 import urllib.request
+import shutil
 
 __all__ = ["get_input"]
 
 
-def get_cache_directory():
-    # Use the CI paths if they exist, just to make CI integration easier.
-    ci_inputs_path = os.environ["HOME"] + "/.cache/graph"
-    if os.path.isdir(ci_inputs_path):
-        return ci_inputs_path
-    # Otherwise use a more specific path name
-    return os.environ["HOME"] + "/.cache/galois"
-
-
-def get_inputs_directory():
-    cache_dir = get_cache_directory()
-    inputs_dir = cache_dir + "/inputs"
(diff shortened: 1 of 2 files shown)


arthurp:
Looks like I messed up something in my changes to the downloader. Probably directory deletion and creation. Sigh. @danielmawhirter can you fix it? If not, you could just roll back my larger changes and just update the version number without changing anything else in the Python code.

python/katana/example_utils.py
```diff
@@ -1,26 +1,34 @@
...
+    # Otherwise use a cache directory
+    if not inputs_dir:
+        inputs_dir = Path(os.environ["HOME"]) / ".cache" / "katana" / "inputs"
+    if not invalidate and inputs_dir.is_dir() and (inputs_dir / "propertygraphs" / "ldbc_003").is_dir():
         return inputs_dir
+    try:
+        shutil.rmtree(inputs_dir)
+    except OSError:
+        inputs_dir.unlink()
+    inputs_dir.mkdir(parents=True, exist_ok=True)
```
danielmawhirter:
```python
if inputs_dir.is_dir() and (inputs_dir / "propertygraphs" / "ldbc_003").is_dir():
  if not invalidate:
    return inputs_dir
  try:
    shutil.rmtree(inputs_dir)
  except OSError:
    inputs_dir.unlink()
inputs_dir.mkdir(parents=True, exist_ok=True)
```
My best guess here is that this is the order of operations you want, since the "does not exist" case seemed to be raising `FileNotFoundError` on both `rmtree` and `unlink`. If that's wrong, then I think the reversion you describe is the best option

arthurp: I think you may be right. You can make that change and push to my branch. You have access. Or just do the roll back. Up to you.

danielmawhirter (approved):
Ok, I think that worked! Merge at your leisure

arthurp:
@danielmawhirter Thanks so much for handling this!!! This looks great.

PR #213, opened 2021-05-07
tsuba: add option for parquet writer to write blocked files

tylershunt:
Large files are sometimes hard to swallow for downstream processors,
and writing many files means we can do the transformation in parallel.

### DIFF
--- a/libsupport/include/katana/ArrowInterchange.h
+++ b/libsupport/include/katana/ArrowInterchange.h
@@ -302,6 +302,12 @@ KATANA_EXPORT void DiffFormatTo(
     const std::shared_ptr<arrow::ChunkedArray>& a1,
     size_t approx_total_characters = 150);
 
+/// Estimate the amount of memory this array is using
+/// n.b. Estimate is best effort when array is a slice or a variable type like
+///   large_string; it will be an upper bound in those cases
+KATANA_EXPORT uint64_t
+ApproxArrayMemUse(const std::shared_ptr<arrow::Array>& array);
+
 }  // namespace katana
 
 #endif
--- a/libtsuba/src/ParquetWriter.cpp
+++ b/libtsuba/src/ParquetWriter.cpp
@@ -1,9 +1,6 @@
 #include "tsuba/ParquetWriter.h"
 
-#include <arrow/array/array_binary.h>
-#include <arrow/chunked_array.h>
-#include <parquet/properties.h>
-
+#include "katana/ArrowInterchange.h"
 #include "katana/Result.h"
 #include "tsuba/Errors.h"
 #include "tsuba/FaultTest.h"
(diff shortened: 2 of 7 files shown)


libsupport/include/katana/ArrowInterchange.h
```diff
@@ -302,6 +302,12 @@ KATANA_EXPORT void DiffFormatTo(
     const std::shared_ptr<arrow::ChunkedArray>& a1,
     size_t approx_total_characters = 150);
 
+/// Estimate the amount of memory this array is using
+/// n.b. Estimate is best effort when array is a slice or a variable type like
+///   large_string; it will be an upper bound in those cases
+KATANA_EXPORT uint64_t
+ApproxArrayMemUse(const std::shared_ptr<arrow::Array>& array);
```
witchel, line 309: Upper bound?  I'd think LargeStrings could be as large as memory.

libtsuba/src/ParquetWriter.cpp
```diff
@@ -205,3 +256,40 @@ tsuba::ParquetWriter::StoreParquet(
...
+    if (!desc_res) {
+      return desc_res.error();
+    }
+    our_desc = std::move(desc_res.value());
+    desc = our_desc.get();
+  }
+
+  katana::Result<void> ret = katana::ResultSuccess();
+  for (uint64_t i = 0, num_tables = tables_.size(); i < num_tables; ++i) {
+    ret = StoreParquet(tables_[i], uri + fmt::format(".{}", i), desc);
```
ddn0: do these names/urls need to be sortable in a way that matches the original table order?

libsupport/include/katana/ArrowInterchange.h
witchel: Oh, read the code, now I get it.

libtsuba/src/ParquetWriter.cpp
```diff
@@ -1,9 +1,14 @@
 #include "tsuba/ParquetWriter.h"
 
+#include <memory>
+
 #include <arrow/array/array_binary.h>
+#include <arrow/array/data.h>
 #include <arrow/chunked_array.h>
+#include <arrow/type.h>
 #include <parquet/properties.h>
```
witchel: Did you intend to add all these arrow headers?

libtsuba/src/ParquetWriter.cpp
tylershunt: They don't for the usecase driving this but that's a good idea, I'll add some leading zeros.

libtsuba/src/ParquetWriter.cpp
```diff
@@ -205,3 +256,40 @@ tsuba::ParquetWriter::StoreParquet(
...
+      return desc_res.error();
+    }
+    our_desc = std::move(desc_res.value());
+    desc = our_desc.get();
+  }
+
+  katana::Result<void> ret = katana::ResultSuccess();
+  for (uint64_t i = 0, num_tables = tables_.size(); i < num_tables; ++i) {
+    ret = StoreParquet(tables_[i], uri + fmt::format(".{:06}", i), desc);
+    if (!ret) {
```
witchel, line 276: Fixed-length format!  My dreams have come true.

witchel (commented):
Looks good.

tylershunt:
rebased temporarily to avoid some problematic commits since I'm using this branch

libtsuba/src/ParquetWriter.cpp
tylershunt: Thanks for the call out. I did not. One of these days I'm going to learn how to turn off this feature. the pluggin is too good to drop though.

tylershunt:
Ready to merge I think unless reviewers want to see something else, is blocking https://github.com/KatanaGraph/katana-enterprise/pull/1121

ddn0 (approved):

PR #220, opened 2021-05-09
python: Wrap property persistence, upsert, equals, path.

arthurp:
also improve documentation.


python/tests/test_property_graph.py
```diff
@@ -18,6 +18,38 @@ def test_load(property_graph):
...
+        property_graph = PropertyGraph(tmpdir)
+    assert property_graph.num_nodes() == 29092
+    assert property_graph.num_edges() == 39283
+    assert len(property_graph.node_schema()) == 31
+    assert len(property_graph.edge_schema()) == 19
+
+    assert property_graph == old_property_graph
+
+
+# TODO(amp): Reinstant this test once it matches the actual RDG semantics.
```
bowu: Maybe we can invite a reviewer who worked on `PropertyGraph`?

arthurp: This is below the level of `PropertyGraph`, it's an RDG issue. The issue is that I don't understand the current semantics of `InformPath`, and since we are moving to the log, I don't think there is value in figuring out at the moment. We should just update this stuff once we stabilize on the log and we know what that implies about the direct URL access API.

@tylershunt Is that a reasonable approach?

python/katana/property_graph.pyx.jinja
```diff
@@ -47,9 +49,12 @@ cdef class PropertyGraph:
 
         :param path: the path or URL from which to load the graph. This support local paths or s3 URLs.
         :type path: str
-        :param partition_id_to_load: The partition number in the graph to load. If this is None (default), then the partition corresponding to this host's network id will be loaded.
-        :param node_properties: A list of node property names to load into memory. If this is None (default), then all properties are loaded.
-        :param edge_properties: A list of edge property names to load into memory. If this is None (default), then all properties are loaded.
+        :param partition_id_to_load: The partition number in the graph to load. If this is None (default), then the
```
gurbinder533, line 52: Why are we talking about partitions ids, etc. for property graphs? Shouldn't this be in DistPropertyGraph?

python/katana/property_graph.pyx.jinja
```diff
@@ -78,19 +83,38 @@ cdef class PropertyGraph:
...
             pg = handle_result_PropertyGraph(
                 handle_result_GraphComponents(CGraph.ConvertGraphML(path_str, chunk_size, False))
                     .ToPropertyGraph())
         return PropertyGraph.make(pg)
 
-    def write(self, path, command_line) :
+    def write(self, path = None, str command_line = "katana.property_graph.PropertyGraph"):
         """
-        Write the property graph out the specified path or URL (or the original path it was loaded from if path is nor provided). Provide lineage information in the form of a command line.
+        Write the property graph out the specified path or URL (or the original path it was loaded from if path is
```
gurbinder533: graph out the specified --> graph out to the specified ...

python/tests/test_property_graph.py
witchel: I can inform about `InformPath`.  Under some circumstances, the PropertyGraph does not know from where it was read.  If we call `Commit()` on such a graph, it does not know where to write itself back.  `InformPath `informs the PropertyGraph from where it was read.  At this point, I can't remember how this situation arises and if it is still a concern.

python/katana/property_graph.pyx.jinja
```diff
@@ -195,8 +212,6 @@ cdef class PropertyGraph:
 
     cpdef uint64_t get_edge_dst(PropertyGraph self, uint64_t e):
```
gurbinder533: "get_edge_dest". We are using "dest" instead of "dst". It will be good to be consistent.

witchel (commented):
Thanks for wrapping these functions!  Let me know if there is any more info I can provide.

gurbinder533 (approved):
LGTM. A few minor comments.

arthurp:
>  Let me know if there is any more info I can provide.

@witchel Can you check and make sure the documentation on the new Cython functions are correct? As a way of checking to make sure I didn't misunderstand the semantics of those operations.

python/katana/property_graph.pyx.jinja
arthurp: No idea. It's in the C++ API, so I passed it through. I agree with you, BTW. When we remove the partition stuff from PropertyGraph we can remove it here too.

python/tests/test_property_graph.py
arthurp: The issue I'm seeing here is that if you load a file and then set a new path with `InformPath` it doesn't seem to actually change where the output will be written. But it's not important for the moment. I suspect all this will change a bit with advances in the storage format.

witchel: That is weird, especially if you are interfacing with the `PropertyGraph `directly and not a variant like `AttributedGraph`.  Actually, now that I'm thinking about it, this problem happened when I was loading a `DistGraph`.

python/katana/property_graph.pyx.jinja
witchel: I'm pretty sure this is Simon's code to allow him to load partitions independently from host IDs.  I mean, I don't think anyone is particularly proud of the RDG/PropertyGraph/OtherGraph interface, but we are moving in the right direction.

python/tests/test_property_graph.py
arthurp: > Actually, now that I'm thinking about it, this problem happened when I was loading a `DistGraph`.

Interesting. I will leave this test disabled and let you create an issue for this problem if needed. It doesn't seem like a fatal issue at the moment, though it is something we should address eventually.

PR #230, opened 2021-05-13
Make some noise on problematic shutdown

witchel:


tylershunt (approved):
This is fine; an alternative would be to attach the scheme using `WithContext(...)` if that's the info you're trying to get at here.

witchel:
That is so weird.  I called WARN because I thought the function did not return a Result.  My bad, and I changed the code.  It must have been too late last night.

PR #241, opened 2021-05-24
Update Python.md

gliga:
Adding a missing ) to balance parenthesis in text.

### DIFF
--- a/Python.md
+++ b/Python.md
@@ -15,7 +15,7 @@ Building Katana-Python
 ----------------------
 
 Install [libnuma](https://github.com/numactl/numactl) development files as required in your distribution
-(`sudo apt install libnuma-dev` for Ubuntu or Debian, `sudo yum install numactl-devel` for CentOS or RHEL).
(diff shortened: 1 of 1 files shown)


Python.md
```diff
@@ -15,7 +15,7 @@ Building Katana-Python
 ----------------------
 
 Install [libnuma](https://github.com/numactl/numactl) development files as required in your distribution
-(`sudo apt install libnuma-dev` for Ubuntu or Debian, `sudo yum install numactl-devel` for CentOS or RHEL).
+(`sudo apt install libnuma-dev`) for Ubuntu or Debian, `sudo yum install numactl-devel` for CentOS or RHEL).
```
ddn0, line 18: - The new `)` unbalances the `)` at the end of this line.
- The PR description mentions replacing "conda" with "Conda" but there are no changes that do so (yet).

gliga: Updated my commit message.

ddn0: > The new ) unbalances the ) at the end of this line.

I'm a little confused. Before this PR, we had two parentheses (left, right) and they were balanced. With this change, we have three parentheses (left, right, right) and they are not balanced.

PR #243, opened 2021-05-25
OpaqueID improvements

scober:
`OpaqueIDLinear` is now an iterator, which allows it be passed to `katana::iterate` directly.

`OpaqueIDLinear` now has a static field called `sentinel`.

### DIFF
--- a/libsupport/include/katana/OpaqueID.h
+++ b/libsupport/include/katana/OpaqueID.h
@@ -181,6 +181,16 @@ public:
 
   using DifferenceType = std::make_signed_t<_Value>;
 
+  // iterator traits
+  using difference_type = DifferenceType;
+  using value_type = _IDType;
+  using pointer = _IDType*;
+  using reference = _IDType&;
+  using iterator_category = std::random_access_iterator_tag;
+
+  // must provide a dereference to be an iterator
+  reference operator*() { return *static_cast<_IDType*>(this); }
+
   _IDType& operator++() {
     ++this->value_;
     return *static_cast<_IDType*>(this);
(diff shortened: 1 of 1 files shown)


libsupport/include/katana/OpaqueID.h
```diff
@@ -216,6 +237,16 @@ struct OpaqueIDLinear : public OpaqueIDOrderedWithValue<_IDType, _Value> {
...
+
+  /// Be careful! std::numeric_limits is only specialized for a handful of
+  /// numeric types and the default behavior of std::numeric_limits<T> is to
+  /// return T().
+  ///
+  /// So if your _Value is not a language-provided numeric type
+  /// you should specialize std::numeric_limits<_Value>, specialize this
+  /// sentinel function, or be prepared to use _Value as your sentinel value
+  static constexpr _IDType sentinel =
+      _IDType(std::numeric_limits<_Value>::max());
```
arthurp: Use `boost::math::tools::max_value`. That will fail instead of producing 0 in cases where it is not specialized. It will work for anything that specializes either that function or `numeric_limits`.

If subclasses override this, will it prevent static errors? If not, maybe this needs an `enable_if` template of some kind to allow using `OpaqueIDLinear` on types without `numeric_limits`. Though I'm not sure there is any reason to support that.

arthurp (approved):
One change. But otherwise LGTM.

libsupport/include/katana/OpaqueID.h
scober: I think we could safely wait and find out

PR #244, opened 2021-05-25
build: Remove elfutils

ddn0:
It is not a direct dependency of katana.


ddn0:
Test:
https://github.com/KatanaGraph/katana/pull/242 runs into a conda issue but this PR doesn't

arthurp (approved):
LGTM. However, I am working on a better fix because the issue appear to be an overly strict dependency on elfutils. And without elfutils conda builds will not be able to display nice back traces. You can merge if you like, but I will hopefully have a better fix tonight.

Issue #25, opened 2021-01-07
Rework "NA" (i.e., null, invalid) handling in PropertyViews


arthurp:
Currently handling of Arrow invalid values (also called nulls and NAs) in [PropertyViews](https://github.com/KatanaGraph/katana/blob/master/libgalois/include/galois/Properties.h), is ad hoc and probably not what people would expect. Currently reading the valid bit works (`IsValid`), but the valid bits are never updated even when the user writes to the property data.

We don't really expect to use NAs much at all and there are lots of tools to eliminate NAs (e.g., in pandas). So my approach would be to remove all support for NAs and add checks to fail as early as possible if an NA is encountered. Arrow has specific support for arrays that never contain NA; they are represented with no null_bitmap. We should be able to use that to eliminate many dangerous cases.

The main reason we may need to do this is that in the python ecosystem the arrow arrays we use may originate from other libraries that has NAs (like pandas) and those NAs will be maintained by pyarrow. So we need to be aware of them and provide useful errors so users know to eliminate NAs.

danielmawhirter:
I'm definitely open to improvements here, as the current state isn't the greatest. A couple interesting notes:

1. Many of the analytics applications expect every index to be valid, so PropertyViews reflect that by treating null values as default-initialized. This is a semantics error, but we currently ignore it to avoid a re-work of those applications. For testing purposes, changing the null check to an `assert()` in Properties.h will show failing tests for every application that makes this assumption.
2. Querying is a different story, and while it's not meaningful in the context of this repo necessarily, we do already fail fast for improper null accesses in that domain.
3. Writing into arrow arrays / parquet files in a reasonable way is a challenge. The simple case of writing and updating the null bitmask isn't too complicated in serial (there's some work to be done for concurrency). The more complicated case shows up in String, or any non-fixed-size datatype, where dynamic allocations would be required for writes.

Because of point 1, we've already been able to load and access data with nulls while preserving the assumption of validity in analytics. But point 3 still leaves significant challenges.

Do you have an overall vision of how to address this?

arthurp:
To be clear, nothing I'm talking about here will affect `PropertyFileGraph`. So dynamically typed access or access from Python will be unchanged. This issue is only with the `PropertyGraph` interface.

[...]

PR #246, opened 2021-05-26
enable ArrowVisitor on arrow::NullType

danielmawhirter:
Correctly handling NullType allows for typeless-nulls, like the Cypher `null` keyword.
A Scalar (`std::shared_ptr<arrow::Scalar>`) can be verified to be null in several ways:
`scalar->type->Equals(arrow::null())` (typeless null)
`scalar->type->id() == arrow::Type::NA` (typeless null)
`dynamic_cast<arrow::NullScalar*>(scalar.get()) != nullptr` (typeless null)
`scalar->is_valid == false` (possibly typed null)


danielmawhirter:
Resolves #10 while I'm here, json_checker moved to enterprise a while ago

danielmawhirter:
Relevant to #25

witchel (approved):
The code looks fine, but are we generating the typeless nulls?  Because in GraphUpdate I always generate null's using the property schema type.

danielmawhirter:
> The code looks fine, but are we generating the typeless nulls? Because in GraphUpdate I always generate null's using the property schema type.

Up to this point we haven't, partially due to this limitation. Query has a use case that wants them now though (Cypher `null` keyword), so IMO it's time to enable them

PR #248, opened 2021-05-26
Workarounds to compile the code with nvcc (CUDA)

srogatch:
Small workarounds to fix the compilation of katana::Result with nvcc in CUDA

### DIFF
--- a/libgalois/include/katana/Properties.h
+++ b/libgalois/include/katana/Properties.h
@@ -149,7 +149,7 @@ template <typename PropTuple>
 Result<std::tuple<>>
 ConstructPropertyViews(
     const std::vector<arrow::Array*>&, std::index_sequence<>) {
-  return std::tuple<>();
+  return Result<std::tuple<>>(std::tuple<>());
 }
 
 template <typename PropTuple, size_t head, size_t... tail>
@@ -298,9 +298,7 @@ public:
   using value_type = uint8_t;
 
   static Result<BooleanPropertyReadOnlyView> Make(
-      const arrow::BooleanArray& array) {
-    return BooleanPropertyReadOnlyView(array);
-  }
+      const arrow::BooleanArray& array);
 
   bool IsValid(size_t i) const {
     KATANA_LOG_DEBUG_ASSERT(i < (size_t)array_.length());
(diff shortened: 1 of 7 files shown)


amberhassaan (approved):
No comments.

libgalois/include/katana/Properties.h
```diff
@@ -149,7 +149,7 @@ template <typename PropTuple>
 Result<std::tuple<>>
 ConstructPropertyViews(
     const std::vector<arrow::Array*>&, std::index_sequence<>) {
-  return std::tuple<>();
+  return Result<std::tuple<>>(std::tuple<>());
```
roshandathathri, line 152: `Result` constructor should handle this conversion. I'm surprised an explicit conversion is required. Did any build fail due to this? if so, maybe we should try to fix this in the `Result` constructor.

libgalois/include/katana/Properties.h
```diff
@@ -298,9 +298,7 @@ class BooleanPropertyReadOnlyView {
   using value_type = uint8_t;
 
   static Result<BooleanPropertyReadOnlyView> Make(
-      const arrow::BooleanArray& array) {
-    return BooleanPropertyReadOnlyView(array);
-  }
+      const arrow::BooleanArray& array);
```
roshandathathri, line 301: Single line functions are better defined in the header so that they _can_ be inlined by the compiler. Does this not compile for host code using nvcc?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -290,18 +290,7 @@ class KATANA_EXPORT PropertyGraph {
...
-    }
-    auto uri_res = katana::Uri::Make(input_path);
-    if (!uri_res) {
-      return uri_res.error();
-    }
-
-    rdg_.set_rdg_dir(uri_res.value());
-    return ResultSuccess();
-  }
+  Result<void> InformPath(const std::string& input_path);
```
roshandathathri, line 293: Ditto about inlining.

roshandathathri (commented):
I have no objections to merging but please answer the questions below.

libgalois/include/katana/Properties.h
srogatch: The CUDA build fails if this header file is included into a `.cu` file compiled with `nvcc`. For some reason `nvcc` doesn't see the implicit constructor. We can't do much because `Result` definition simply redirects to Boost:
```
/// A Result is a T or an ErrorInfo.
template <class T>
using Result = BOOST_OUTCOME_V2_NAMESPACE::std_result<
    T, ErrorInfo, internal::abort_policy>;
```

libgalois/include/katana/Properties.h
srogatch: No, it doesn't compile with `nvcc`. My earlier measurements show that non-inline functions often outperform inline functions likely because the branch prediction kicks in for non-inline functions. Doesn't `g++`/`clang` have whole program optimization / link-time code generation option that would still make inline the functions defined not in headers?

libgalois/include/katana/PropertyGraph.h
srogatch: The same problem as is in the above comment.

PR #254, opened 2021-05-27
Fix: Clearing (clear) and initializing (resize) bits

gliga:
Ensuring that we clear bits (when clear is invoked) and initializing bits when resizing is done (as they would be random otherwise)

### DIFF
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -109,6 +109,7 @@ public:
    */
   void clear() {
     num_bits_ = 0;
+    // TODO(roshan) parallelize this.
     std::fill(bitvec_.begin(), bitvec_.end(), 0);
     bitvec_.clear();
   }
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -86,8 +86,11 @@ public:
     KATANA_LOG_DEBUG_ASSERT(
         kNumBitsInUint64 == 64);  // compatibility with other devices
     num_bits_ = n;
+    size_t old_size = bitvec_.size();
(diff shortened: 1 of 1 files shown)


libgalois/include/katana/DynamicBitset.h
```diff
@@ -86,8 +86,11 @@ class KATANA_EXPORT DynamicBitset {
     KATANA_LOG_DEBUG_ASSERT(
         kNumBitsInUint64 == 64);  // compatibility with other devices
     num_bits_ = n;
+    size_t old_size = bitvec_.size();
     bitvec_.resize((n + kNumBitsInUint64 - 1) / kNumBitsInUint64);
-    reset();
+    if (bitvec_.size() > old_size) {
+      std::fill(bitvec_.begin() + old_size, bitvec_.end(), 0);
+    }
```
tylershunt, line 93: Just a note that this changes the semantics of `resize` since the old version invoked `reset` to clear everything. I think these semantics for `resize` make (more) sense, but I'm not sure if there were parts of the code that depended on `resize` also clearing all bits so thought I would call it out.

roshandathathri: Yes, the callers may need to be fixed if the CI tests fail.

libgalois/include/katana/DynamicBitset.h
```diff
@@ -106,6 +109,7 @@ class KATANA_EXPORT DynamicBitset {
    */
   void clear() {
     num_bits_ = 0;
+    std::fill(bitvec_.begin(), bitvec_.end(), 0);
```
roshandathathri, line 113: nit: It's better to move this into a private function and we may want to parallelize this in the future. You can add a TODO on me for that. Something like:
`\\ TODO(roshan) parallelize this`

roshandathathri (approved):
LGTM

PR #259, opened 2021-05-28
[KAT-516] Introduce Types to PropertyGraph

roshandathathri:
Differentiate between types and non-type/arbitrary properties.
    Currently, assumes all boolean or uint8 properties are types.

    Constructs a new array to store types of all nodes and another
    for all edges. The types of a node are encoded using a TypeID.

    Encode TypeID to identify/contain a unique combinations of
    types. Currently, the encoding uses 8 bits, so it supports up
    to 254 unique combinations of types (excluding unknown and
    invalid types).

### DIFF
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -99,19 +99,24 @@ struct KATANA_EXPORT GraphTopology {
 /// comprise the physical representation of the logical property graph.
 class KATANA_EXPORT PropertyGraph {
 public:
-  /// TypeID uniquely identifies/contains a combination/set of types
-  /// TypeID is represented using 8 bits
-  /// TypeID for nodes is distinct from TypeID for edges
-  using TypeID = uint8_t;
-  using ArrowTypeID = arrow::UInt8Type;
-  static constexpr TypeID kUnknownType = TypeID{0};
-  static constexpr TypeID kInvalidType = std::numeric_limits<TypeID>::max();
-  /// A set of TypeIDs
-  using TypeIDsSet = std::bitset<std::numeric_limits<TypeID>::max() + 1>;
-  /// A map from TypeID to the set of the type names it contains
-  using TypeIDToTypeNamesMap = std::vector<std::unordered_set<std::string>>;
-  /// A map from the type name to the set of the TypeIDs that contain it
-  using TypeNameToTypeIDsMap = std::unordered_map<std::string, TypeIDsSet>;
+  /// TypeSetID uniquely identifies/contains a combination/set of types
+  /// TypeSetID is represented using 8 bits
+  /// TypeSetID for nodes is distinct from TypeSetID for edges
+  using TypeSetID = uint8_t;
+  using ArrowTypeSetID = arrow::UInt8Type;
+  static constexpr TypeSetID kUnknownType = TypeSetID{0};
+  static constexpr TypeSetID kInvalidType =
+      std::numeric_limits<TypeSetID>::max();
+  /// A set of TypeSetIDs
+  using SetOfTypeSetIDs =
+      std::bitset<std::numeric_limits<TypeSetID>::max() + 1>;
+  /// A set of type names
+  using SetOfTypeNames = std::unordered_set<std::string>;
+  /// A map from TypeSetID to the set of the type names it contains
+  using TypeSetIDToSetOfTypeNamesMap = std::vector<SetOfTypeNames>;
+  /// A map from the type name to the set of the TypeSetIDs that contain it
+  using TypeNameToSetOfTypeSetIDsMap =
+      std::unordered_map<std::string, SetOfTypeSetIDs>;
 
   // Pass through topology API
   using node_iterator = GraphTopology::node_iterator;
@@ -139,24 +144,24 @@ private:
   // caller of SetTopology.
   GraphTopology topology_;
 
-  /// A map from the node TypeID to
+  /// A map from the node TypeSetID to
   /// the set of the node type names it contains
-  TypeIDToTypeNamesMap node_type_id_to_type_names_;
-  /// A map from the edge TypeID to
+  TypeSetIDToSetOfTypeNamesMap node_type_set_id_to_type_names_;
+  /// A map from the edge TypeSetID to
   /// the set of the edge type names it contains
-  TypeIDToTypeNamesMap edge_type_id_to_type_names_;
+  TypeSetIDToSetOfTypeNamesMap edge_type_set_id_to_type_names_;
 
   /// A map from the node type name
-  /// to the set of the node TypeIDs that contain it
-  TypeNameToTypeIDsMap node_type_name_to_type_ids_;
+  /// to the set of the node TypeSetIDs that contain it
+  TypeNameToSetOfTypeSetIDsMap node_type_name_to_type_set_ids_;
   /// A map from the edge type name
-  /// to the set of the edge TypeIDs that contain it
-  TypeNameToTypeIDsMap edge_type_name_to_type_ids_;
+  /// to the set of the edge TypeSetIDs that contain it
+  TypeNameToSetOfTypeSetIDsMap edge_type_name_to_type_set_ids_;
 
-  /// The node TypeID for each node in the graph
-  std::shared_ptr<arrow::NumericArray<ArrowTypeID>> node_type_id_;
-  /// The edge TypeID for each edge in the graph
-  std::shared_ptr<arrow::NumericArray<ArrowTypeID>> edge_type_id_;
+  /// The node TypeSetID for each node in the graph
+  std::shared_ptr<arrow::NumericArray<ArrowTypeSetID>> node_type_set_id_;
+  /// The edge TypeSetID for each edge in the graph
+  std::shared_ptr<arrow::NumericArray<ArrowTypeSetID>> edge_type_set_id_;
 
   // Keep partition_metadata, master_nodes, mirror_nodes out of the public interface,
   // while allowing Distribution to read/write it for RDG
@@ -268,11 +273,11 @@ public:
       const std::vector<std::string>& node_properties,
       const std::vector<std::string>& edge_properties) const;
 
-  /// Construct node & edge TypeIDs from node & edge properties
-  /// Also constructs metadata to convert between types and TypeIDs
+  /// Construct node & edge TypeSetIDs from node & edge properties
+  /// Also constructs metadata to convert between types and TypeSetIDs
   /// Assumes all boolean or uint8 properties are types
   /// TODO(roshan) move this to be a part of Make()
-  Result<void> ConstructTypeIDs();
+  Result<void> ConstructTypeSetIDs();
 
   const std::string& rdg_dir() const { return rdg_.rdg_dir().string(); }
 
@@ -346,70 +351,80 @@ public:
   }
 
   /// \returns the number of node types
-  size_t GetNodeTypesNum() const { return node_type_name_to_type_ids_.size(); }
+  size_t GetNodeTypesNum() const {
+    return node_type_name_to_type_set_ids_.size();
+  }
 
   /// \returns the number of edge types
-  size_t GetEdgeTypesNum() const { return edge_type_name_to_type_ids_.size(); }
+  size_t GetEdgeTypesNum() const {
+    return edge_type_name_to_type_set_ids_.size();
+  }
 
   /// \returns true if a node type with @param name exists
   /// NB: no node may have this type
   /// TODO(roshan) build an index for the number of nodes with the type
   bool HasNodeType(const std::string& name) const {
-    return node_type_name_to_type_ids_.count(name) == 1;
+    return node_type_name_to_type_set_ids_.count(name) == 1;
   }
 
   /// \returns true if an edge type with @param name exists
   /// NB: no edge may have this type
   /// TODO(roshan) build an index for the number of edges with the type
   bool HasEdgeType(const std::string& name) const {
-    return edge_type_name_to_type_ids_.count(name) == 1;
+    return edge_type_name_to_type_set_ids_.count(name) == 1;
   }
 
-  /// \returns the set of node TypeIDs that contain
+  /// \returns the set of node TypeSetIDs that contain
   /// the node type with @param name
   /// (assumes that the node type exists)
-  const TypeIDsSet& NodeTypeNameToTypeIDs(const std::string& name) const {
-    return node_type_name_to_type_ids_.at(name);
+  const SetOfTypeSetIDs& NodeTypeNameToTypeSetIDs(
+      const std::string& name) const {
+    return node_type_name_to_type_set_ids_.at(name);
   }
 
-  /// \returns the set of edge TypeIDs that contain
+  /// \returns the set of edge TypeSetIDs that contain
   /// the edge type with @param name
   /// (assumes that the edge type exists)
-  const TypeIDsSet& EdgeTypeNameToTypeIDs(const std::string& name) const {
-    return edge_type_name_to_type_ids_.at(name);
+  const SetOfTypeSetIDs& EdgeTypeNameToTypeSetIDs(
+      const std::string& name) const {
+    return edge_type_name_to_type_set_ids_.at(name);
   }
 
-  /// \returns the number of node TypeIDs (including kUnknownType)
-  size_t GetNodeTypeIDsNum() const {
-    return node_type_id_to_type_names_.size();
+  /// \returns the number of node TypeSetIDs (including kUnknownType)
+  size_t GetNodeTypeSetIDsNum() const {
+    return node_type_set_id_to_type_names_.size();
   }
 
-  /// \returns the number of edge TypeIDs (including kUnknownType)
-  size_t GetEdgeTypeIDsNum() const {
-    return edge_type_id_to_type_names_.size();
+  /// \returns the number of edge TypeSetIDs (including kUnknownType)
+  size_t GetEdgeTypeSetIDsNum() const {
+    return edge_type_set_id_to_type_names_.size();
   }
 
   /// \returns the set of node type names that contain
-  /// the node TypeID @param node_type_id
-  /// (assumes that the node TypeID exists)
-  const std::unordered_set<std::string>& NodeTypeIDToTypeNames(
-      TypeID node_type_id) const {
-    return node_type_id_to_type_names_.at(node_type_id);
+  /// the node TypeSetID @param node_type_set_id
+  /// (assumes that the node TypeSetID exists)
+  const SetOfTypeNames& NodeTypeSetIDToTypeNames(
+      TypeSetID node_type_set_id) const {
+    return node_type_set_id_to_type_names_.at(node_type_set_id);
   }
 
   /// \returns the set of edge type names that contain
-  /// the edge TypeID @param edge_type_id
-  /// (assumes that the edge TypeID exists)
-  const std::unordered_set<std::string>& EdgeTypeIDToTypeNames(
-      TypeID edge_type_id) const {
-    return edge_type_id_to_type_names_.at(edge_type_id);
+  /// the edge TypeSetID @param edge_type_set_id
+  /// (assumes that the edge TypeSetID exists)
+  const SetOfTypeNames& EdgeTypeSetIDToTypeNames(
+      TypeSetID edge_type_set_id) const {
+    return edge_type_set_id_to_type_names_.at(edge_type_set_id);
   }
 
-  /// \return returns the node TypeID for @param node
-  TypeID GetNodeTypeID(Node node) const { return node_type_id_->Value(node); }
+  /// \return returns the node TypeSetID for @param node
+  TypeSetID GetNodeTypeSetID(Node node) const {
+    return node_type_set_id_->Value(node);
+  }
 
-  /// \return returns the edge TypeID for @param edge
-  TypeID GetEdgeTypeID(Edge edge) const { return edge_type_id_->Value(edge); }
+  /// \return returns the edge TypeSetID for @param edge
+  TypeSetID GetEdgeTypeSetID(Edge edge) const {
+    return edge_type_set_id_->Value(edge);
+  }
 
   // Return type dictated by arrow
   int32_t GetNodePropertyNum() const {
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -97,6 +97,29 @@ struct KATANA_EXPORT GraphTopology {
 /// manages the serialization of the various partitions and properties that
 /// comprise the physical representation of the logical property graph.
 class KATANA_EXPORT PropertyGraph {
+public:
+  /// TypeID uniquely identifies/contains a combination/set of types
+  /// TypeID is represented using 8 bits
+  /// TypeID for nodes is distinct from TypeID for edges
+  using TypeID = uint8_t;
+  using ArrowTypeID = arrow::UInt8Type;
+  static constexpr TypeID kUnknownType = TypeID{0};
+  static constexpr TypeID kInvalidType = std::numeric_limits<TypeID>::max();
+  /// A map from TypeID to the set of the type names it contains
+  using TypeIDToTypeNamesMap = std::vector<std::unordered_set<std::string>>;
+  /// A map from the type name to the set of the TypeIDs that contain it
+  using TypeNameToTypeIDMap =
+      std::unordered_map<std::string, std::unordered_set<TypeID>>;
+
+  // Pass through topology API
+  using node_iterator = GraphTopology::node_iterator;
+  using edge_iterator = GraphTopology::edge_iterator;
+  using edges_range = GraphTopology::edges_range;
+  using iterator = GraphTopology::iterator;
+  using Node = GraphTopology::Node;
+  using Edge = GraphTopology::Edge;
+
+private:
   PropertyGraph(std::unique_ptr<tsuba::RDGFile> rdg_file, tsuba::RDG&& rdg);
 
   /// Validate performs a sanity check on the the graph after loading
@@ -113,6 +136,25 @@ class KATANA_EXPORT PropertyGraph {
   // The topology is either backed by rdg_ or shared with the
   // caller of SetTopology.
   GraphTopology topology_;
+
+  /// A map from the node TypeID to
+  /// the set of the node type names it contains
+  TypeIDToTypeNamesMap node_type_id_to_type_names_;
+  /// A map from the edge TypeID to
+  /// the set of the edge type names it contains
+  TypeIDToTypeNamesMap edge_type_id_to_type_names_;
+
+  /// A map from the node type name
+  /// to the set of the node TypeIDs that contain it
+  TypeNameToTypeIDMap node_type_name_to_type_ids_;
+  /// A map from the edge type name
+  /// to the set of the edge TypeIDs that contain it
+  TypeNameToTypeIDMap edge_type_name_to_type_ids_;
+
+  /// The node TypeID for each node in the graph
+  std::shared_ptr<arrow::NumericArray<ArrowTypeID>> node_type_id_;
+  /// The edge TypeID for each edge in the graph
+  std::shared_ptr<arrow::NumericArray<ArrowTypeID>> edge_type_id_;
 
   // Keep partition_metadata, master_nodes, mirror_nodes out of the public interface,
   // while allowing Distribution to read/write it for RDG
@@ -224,6 +266,12 @@ public:
       const std::vector<std::string>& node_properties,
       const std::vector<std::string>& edge_properties) const;
 
+  /// Construct node & edge TypeIDs from node & edge properties
+  /// Also constructs metadata to convert between types and TypeIDs
+  /// Assumes all boolean or uint8 properties are types
+  /// TODO(roshan) move this to be a part of Make()
+  Result<void> ConstructTypeIDs();
+
   const std::string& rdg_dir() const { return rdg_.rdg_dir().string(); }
 
   uint32_t partition_id() const { return rdg_.partition_id(); }
@@ -295,6 +343,74 @@ public:
     return edge_properties()->schema();
   }
 
+  /// \returns the number of node types
+  size_t GetNodeTypesNum() const { return node_type_name_to_type_ids_.size(); }
+
+  /// \returns the number of edge types
+  size_t GetEdgeTypesNum() const { return edge_type_name_to_type_ids_.size(); }
+
+  /// \returns true if a node type with @param name exists
+  /// NB: no node may have this type
+  /// TODO(roshan) build an index for the number of nodes with the type
+  bool HasNodeType(const std::string& name) const {
+    return node_type_name_to_type_ids_.count(name) == 1;
+  }
+
+  /// \returns true if an edge type with @param name exists
+  /// NB: no edge may have this type
+  /// TODO(roshan) build an index for the number of edges with the type
+  bool HasEdgeType(const std::string& name) const {
+    return edge_type_name_to_type_ids_.count(name) == 1;
+  }
+
+  /// \returns the set of node TypeIDs that contain
+  /// the node type with @param name
+  /// (assumes that the node type exists)
+  const std::unordered_set<TypeID>& NodeTypeNameToTypeIDs(
+      const std::string& name) const {
+    return node_type_name_to_type_ids_.at(name);
+  }
+
+  /// \returns the set of edge TypeIDs that contain
+  /// the edge type with @param name
+  /// (assumes that the edge type exists)
+  const std::unordered_set<TypeID>& EdgeTypeNameToTypeIDs(
+      const std::string& name) const {
+    return edge_type_name_to_type_ids_.at(name);
+  }
+
+  /// \returns the number of node TypeIDs (including kUnknownType)
+  size_t GetNodeTypeIDsNum() const {
+    return node_type_id_to_type_names_.size();
+  }
+
+  /// \returns the number of edge TypeIDs (including kUnknownType)
+  size_t GetEdgeTypeIDsNum() const {
+    return edge_type_id_to_type_names_.size();
+  }
+
+  /// \returns the set of node type names that contain
+  /// the node TypeID @param node_type_id
+  /// (assumes that the node TypeID exists)
+  const std::unordered_set<std::string>& NodeTypeIDToTypeNames(
+      TypeID node_type_id) const {
+    return node_type_id_to_type_names_.at(node_type_id);
+  }
+
+  /// \returns the set of edge type names that contain
+  /// the edge TypeID @param edge_type_id
+  /// (assumes that the edge TypeID exists)
+  const std::unordered_set<std::string>& EdgeTypeIDToTypeNames(
+      TypeID edge_type_id) const {
+    return edge_type_id_to_type_names_.at(edge_type_id);
+  }
+
+  /// \return returns the node TypeID for @param node
+  TypeID GetNodeTypeID(Node node) const { return node_type_id_->Value(node); }
+
+  /// \return returns the edge TypeID for @param edge
+  TypeID GetEdgeTypeID(Edge edge) const { return edge_type_id_->Value(edge); }
+
   // Return type dictated by arrow
   int32_t GetNodePropertyNum() const {
     return node_properties()->num_columns();
@@ -317,6 +433,16 @@ public:
       return nullptr;
     }
     return edge_properties()->column(i);
+  }
+
+  /// \returns true if a node property/type with @param name exists
+  bool HasNodeProperty(const std::string& name) const {
+    return node_properties()->schema()->GetFieldIndex(name) != -1;
+  }
+
+  /// \returns true if an edge property/type with @param name exists
+  bool HasEdgeProperty(const std::string& name) const {
+    return edge_properties()->schema()->GetFieldIndex(name) != -1;
   }
 
   /// Get a node property by name.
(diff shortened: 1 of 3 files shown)


libgalois/include/katana/PropertyGraph.h
```diff
@@ -97,6 +97,29 @@ struct KATANA_EXPORT GraphTopology {
 /// manages the serialization of the various partitions and properties that
 /// comprise the physical representation of the logical property graph.
 class KATANA_EXPORT PropertyGraph {
+public:
+  /// TypeID uniquely identifies/contains a combination/set of types
+  /// TypeID is represented using 8 bits
+  /// TypeID for nodes is distinct from TypeID for edges
+  using TypeID = uint8_t;
```
arthurp: Would an OpaqueID be appropriate here?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -224,6 +266,11 @@ class KATANA_EXPORT PropertyGraph {
       const std::vector<std::string>& node_properties,
       const std::vector<std::string>& edge_properties) const;
 
+  /// Construct node & edge TypeIDs from node & edge properties
+  /// Also constructs metadata to convert between types and TypeIDs
+  /// Assumes all boolean or uint8 properties are types
+  Result<void> ConstructTypeIDs();
```
arthurp: Shouldn't this be done during construction? If it is not, then some `PropertyGraph`s will effectively support a different API since they will not have type data attached to them.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -114,6 +137,25 @@ class KATANA_EXPORT PropertyGraph {
   // caller of SetTopology.
   GraphTopology topology_;
 
+  /// A map from the node TypeID to
+  /// the set of the node type(-name)s it contains
+  TypeIDToTypesMap node_type_id_to_types_;
```
amberhassaan: Are these maps for nodes with multiple types?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -299,35 +346,109 @@ class KATANA_EXPORT PropertyGraph {
...
+    return node_properties()->schema();
   }
 
   std::shared_ptr<arrow::Schema> edge_schema() const {
-    return rdg_.edge_properties()->schema();
+    return edge_properties()->schema();
   }
 
+  /// \returns the number of node types
+  size_t GetNodeTypesNum() const { return node_type_to_type_ids_.size(); }
```
amberhassaan: GetNumNodeTypes sounds better to me. I personally hate Get and Set part of the name (because Getters and Setters is a horrible idea). I'd prefer NumNodeTypes() and I think Google Style Guide says something similar?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -299,35 +346,109 @@ class KATANA_EXPORT PropertyGraph {
...
+  /// \returns the set of edge TypeIDs that contain
+  /// the edge type with @param name
+  /// (assumes that the edge type exists)
+  const std::unordered_set<TypeID>& EdgeTypeToTypeIDs(
+      const std::string& name) const {
+    return edge_type_to_type_ids_.at(name);
+  }
+
+  /// \returns the number of node TypeIDs (including kUnknownType)
+  size_t GetNodeTypeIDsNum() const { return node_type_id_to_types_.size(); }
```
amberhassaan: NumNodeTypeIDs?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -299,35 +346,109 @@ class KATANA_EXPORT PropertyGraph {
...
+  /// \returns the set of edge type(-names) that contain
+  /// the edge TypeID @param edge_type_id
+  /// (assumes that the edge TypeID exists)
+  const std::unordered_set<std::string>& EdgeTypeIDToTypes(
+      TypeID edge_type_id) const {
+    return edge_type_id_to_types_.at(edge_type_id);
+  }
+
+  /// \return returns the node TypeID for @param node
+  TypeID GetNodeTypeID(Node node) const { return node_type_id_->Value(node); }
```
amberhassaan: NodeTypeID? if you agree with removing Get part.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -299,35 +346,109 @@ class KATANA_EXPORT PropertyGraph {
...
+
+  /// \return returns the edge TypeID for @param edge
+  TypeID GetEdgeTypeID(Edge edge) const { return edge_type_id_->Value(edge); }
+
   // Return type dictated by arrow
   int32_t GetNodePropertyNum() const {
```
amberhassaan, line 429: NumNodeProperties()

libgalois/include/katana/PropertyGraph.h
```diff
@@ -299,35 +346,109 @@ class KATANA_EXPORT PropertyGraph {
...
+  /// \returns the number of node TypeIDs (including kUnknownType)
+  size_t GetNodeTypeIDsNum() const { return node_type_id_to_types_.size(); }
+
+  /// \returns the number of edge TypeIDs (including kUnknownType)
+  size_t GetEdgeTypeIDsNum() const { return edge_type_id_to_types_.size(); }
+
+  /// \returns the set of node type(-names) that contain
+  /// the node TypeID @param node_type_id
+  /// (assumes that the node TypeID exists)
+  const std::unordered_set<std::string>& NodeTypeIDToTypes(
```
amberhassaan: Can we say NodeTypeIDToTypeNames?

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
       std::move(rdg_file), std::move(rdg_result.value()));
 }
 
+/// Assumes all boolean or uint8 properties are types
+katana::Result<std::shared_ptr<arrow::NumericArray<arrow::UInt8Type>>>
+GetTypeIDsFromProperties(
+    const std::shared_ptr<arrow::Table>& properties,
+    katana::PropertyGraph::TypeToTypeIDsMap* type_name_to_ids,
```
amberhassaan: See, you yourself are using names and IDs. I like it :).

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+    katana::PropertyGraph::TypeIDToTypesMap* type_id_to_names) {
+  // throw an error if each column/property has more than 1 chunk
+  for (int i = 0, n = properties->num_columns(); i < n; i++) {
+    std::shared_ptr<arrow::ChunkedArray> property = properties->column(i);
```
amberhassaan, line 159: I read a comment by Emmet that ChunkedArray is going away (gone?)?

libgalois/src/PropertyGraph.cpp
```diff
@@ -229,6 +406,76 @@ katana::PropertyGraph::Copy(
...
+          edge_properties()->num_rows(), num_edges());
+    }
+  } else if (num_edge_rows != num_edges()) {
+    return KATANA_ERROR(
+        ErrorCode::AssertionFailed,
+        "execution failed because number of rows in edge properties {} differs "
+        "from the number of edges {}",
+        edge_properties()->num_rows(), num_edges());
+  }
+  if ((node_type_to_type_ids_.size() > 0) ||
```
amberhassaan: !empty() is a better check, esp. when you change the type of data structure (computing size() may have a cost).

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+    if (current_field->type()->Equals(arrow::boolean()) ||
+        current_field->type()->Equals(arrow::uint8())) {
+      type_field_indices.push_back(i);
+    }
+  }
+
+  // assign a new ID to each type
+  // NB: cannot use unordered_map without defining a hash function for vectors;
+  // performance is not affected here because the map is very small (<=256)
+  std::map<std::vector<int>, katana::PropertyGraph::TypeID>
```
amberhassaan: Having vector as key is tricky for both map and unordered_map. In map, you use std::less comparison, which for vector does lexicographic ordering (element wise comparisons). With unordered_map, you need std::equal_to (in addition to a hash function), which does element-wise equality tests on two vectors. 

Having understood the code somewhat better: Perhaps we should keep the type IDs associated with a row (node or edge) as a bit-set . A sparse bit-set, in fact. This will lead to space savings as well as faster hash functions and perhaps equality tests as well.

libgalois/src/PropertyGraph.cpp
```diff
@@ -229,6 +406,76 @@ katana::PropertyGraph::Copy(
...
+          "the number of edge properties is {} and the number of edges is {}",
+          edge_properties()->num_rows(), num_edges());
+    }
+  } else if (num_edge_rows != num_edges()) {
+    return KATANA_ERROR(
+        ErrorCode::AssertionFailed,
+        "execution failed because number of rows in edge properties {} differs "
+        "from the number of edges {}",
+        edge_properties()->num_rows(), num_edges());
+  }
```
arthurp: These are useful checks. But I think they should be in `Make`. Having property arrays of the wrong length is an issue independently of types.

libgalois/src/PropertyGraph.cpp
```diff
@@ -229,6 +406,76 @@ katana::PropertyGraph::Copy(
...
+  if ((node_type_to_type_ids_.size() > 0) ||
+      (node_type_id_to_types_.size() > 0) ||
+      (edge_type_to_type_ids_.size() > 0) ||
+      (edge_type_id_to_types_.size() > 0)) {
+    return KATANA_ERROR(
+        ErrorCode::AssertionFailed,
+        "ConstructTypeIDs() should not called more than once");
+  }
+
+  node_type_id_to_types_.push_back({});  // for kUnknownType: assumes it is 0
```
arthurp: Probably good to static_assert `kUnknownType == 0` just in case.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+    const std::shared_ptr<arrow::Table>& properties,
+    katana::PropertyGraph::TypeToTypeIDsMap* type_name_to_ids,
+    katana::PropertyGraph::TypeIDToTypesMap* type_id_to_names) {
+  // throw an error if each column/property has more than 1 chunk
+  for (int i = 0, n = properties->num_columns(); i < n; i++) {
+    std::shared_ptr<arrow::ChunkedArray> property = properties->column(i);
+    if (property->num_chunks() != 1) {
+      return KATANA_ERROR(
+          katana::ErrorCode::NotImplemented,
+          "execution failed because property {} has {} chunks (1 chunk "
```
arthurp: You start a bunch of error messages with `execution failed because` but that's not really meaningful since they are already marked as errors from context. I would remove that phrase throughout.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+  // assert that all type IDs and 2 special type IDs (unknown and invalid)
+  // can be stored in 8 bits
+  if (type_id_to_names->size() >
+      (std::numeric_limits<katana::PropertyGraph::TypeID>::max() - size_t{2})) {
+    return KATANA_ERROR(
+        katana::ErrorCode::NotImplemented,
+        "execution failed because number of unique combination of types is {} "
+        "but only up to {} is supported currently",
+        type_id_to_names->size(),
+        std::numeric_limits<katana::PropertyGraph::TypeID>::max());
```
arthurp: We actually support this number - 2. Fix the error message.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+        std::make_pair(field_indices, new_type_id));
+
+    const std::shared_ptr<arrow::Field>& current_field = schema->field(i);
+    const std::string& field_name = current_field->name();
+    KATANA_LOG_DEBUG_ASSERT(
+        type_name_to_ids->find(field_name) == type_name_to_ids->end());
+
+    std::unordered_set<katana::PropertyGraph::TypeID> type_ids{new_type_id};
+    type_name_to_ids->emplace(std::make_pair(field_name, type_ids));
+    type_id_to_names->push_back({field_name});
```
amberhassaan: prefer emplace_back to push_back when possible.

arthurp (changes requested):
Various small things.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+  }
+
+  // collect the list of unique combination of types
+  // NB: cannot use unordered_set without defining a hash function for vectors;
+  // performance is not affected here because the set is very small (<=256)
+  std::set<std::vector<int>> type_combinations;
+  katana::PerThreadStorage<std::set<std::vector<int>>> type_combinations_pts;
+  katana::do_all(
+      katana::iterate(int64_t{0}, properties->num_rows()), [&](int64_t row) {
+        std::vector<int> field_indices;
```
amberhassaan: gstl::Vector might be a good choice.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+    std::unordered_set<katana::PropertyGraph::TypeID> type_ids{new_type_id};
+    type_name_to_ids->emplace(std::make_pair(field_name, type_ids));
+    type_id_to_names->push_back({field_name});
+  }
+
+  // collect the list of unique combination of types
+  // NB: cannot use unordered_set without defining a hash function for vectors;
+  // performance is not affected here because the set is very small (<=256)
+  std::set<std::vector<int>> type_combinations;
+  katana::PerThreadStorage<std::set<std::vector<int>>> type_combinations_pts;
```
amberhassaan: gstl::Set and gstl::Vector as these data structures (set & vector) will allocate memory in parallel section when populating them.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+
+  // allocate type IDs array
+  auto type_ids =
+      std::make_unique<katana::LargeArray<katana::PropertyGraph::TypeID>>();
+  int64_t num_rows = properties->num_rows();
+  type_ids->allocateInterleaved(num_rows);
+
+  // assign the type ID for each row
+  katana::do_all(katana::iterate(int64_t{0}, num_rows), [&](int64_t row) {
+    std::vector<int> field_indices;
```
amberhassaan: gstl::Vector.

libgalois/src/PropertyGraph.cpp
```diff
@@ -145,6 +146,182 @@ MakePropertyGraph(
...
+    for (int i : type_field_indices) {
```
amberhassaan, line 290: Doing this loop again (2nd time) over all rows, you're essentially computing a type-signature of each row (represented by the vector of field indices). I wonder if we can use bit-set (sparse?) here as well. 

Then there is the question of recomputing vs storing this info. There is re-use here, which can be exploited if each thread keeps the signatures of its share of rows, so it can access them easily 2nd time around. I'm not sure which approach (recomputing or storing) will be faster. Often, recomputing is a better choice, but these do-all loops are big (order of nodes and edges).

amberhassaan (commented):
My high-level comment to discuss is whether we should use bit-sets instead of sets of type IDs. That can lead to important space savings. Then some minor comments. Overall, looking good.

libgalois/include/katana/PropertyGraph.h
roshandathathri: Maybe but I didn't think it was necessary, so didn't attempt that in this PR.

libgalois/include/katana/PropertyGraph.h
roshandathathri: It's a map for nodes with single types as well as nodes with multiple types.

libgalois/include/katana/PropertyGraph.h
roshandathathri: Agree but not in this PR. That will require changes to the partitioner and `PartitionedGraphView`, so staging it.

libgalois/include/katana/PropertyGraph.h
roshandathathri: I agree but I wanted to keep the names in the class consistent. This class uses `Get` in the function names and uses `Num` in the end, so I stuck to that convention. We can make a separate pass/PR to change the function names of this class.

libgalois/include/katana/PropertyGraph.h
roshandathathri: Ditto

libgalois/include/katana/PropertyGraph.h
roshandathathri: Hehe, I had that in my first draft but removed it during revision to shorten the names. I'll reintroduce it.

libgalois/src/PropertyGraph.cpp
roshandathathri: Yes, hopefully soon.

libgalois/src/PropertyGraph.cpp
roshandathathri: Will do.

libgalois/src/PropertyGraph.cpp
roshandathathri: (1) Lexicographic ordering for comparison is sufficient in this case. There's an external order we're enforcing for all the vector-keys (based on the order of columns in the properties table).

(2) Dense bit-set was the old code. It would need more memory, not less. Bit-set is suitable when any combination of distinct types can occur but this is almost never the case in database schemas. With 8 bits in a bit-set, we can only support 8 distinct types and 2^8 distinct combination of these distinct types. On the other hand, an arbitrary mapping like TypeID here can support 2^8 distinct "types or combinations of distinct types" in 8 bits. This is more than sufficient for most datasets. If needed, we can use 16 bits instead of 8 bits - I don't see a need for anything more than that even for the most complex database schemas. With dynamic or sparse bit-set, 8 or 16 bits are insufficient, so we cannot use less than the memory of TypeID for all our datasets.

libgalois/src/PropertyGraph.cpp
roshandathathri: Agree but I would have to add a variable for `{field_name}` to do that (compiler throws an error otherwise) which is the same as doing `push_back`.

libgalois/src/PropertyGraph.cpp
roshandathathri: Thanks for catching these. Will do.

libgalois/src/PropertyGraph.cpp
roshandathathri: Thanks, will fix.

libgalois/src/PropertyGraph.cpp
roshandathathri: We can use bit-set for the columns but it will need more than 64 bits. I think recomputing is faster but that's debatable. I don't think the performance difference would be much.  The reason I went with recomputing was to avoid memory overhead, which is a bigger concern IMO.

libgalois/src/PropertyGraph.cpp
roshandathathri: Yes, I'll move these to `Make`. I added these more for my sanity during development.

libgalois/src/PropertyGraph.cpp
roshandathathri: Will do.

libgalois/src/PropertyGraph.cpp
roshandathathri: Thanks, will do.

roshandathathri (commented):
Thanks for your comments. Inline responses.

libgalois/include/katana/PropertyGraph.h
amberhassaan: ok. Do you want to capture this as an issue and resolve all such comments here?

libgalois/src/PropertyGraph.cpp
amberhassaan: Interesting that the compiler can't disambiguate. As another option, you can pass a temporary `emplace_back(set{field_name})`. This should invoke move constructor.

libgalois/src/PropertyGraph.cpp
amberhassaan: I agree with what you said above. My comment is more so applicable to places where you have vectors (or sets) of types and sets(and vectors) of field indices. We keep the typeIDs as you've designed them and 8 bits is enough. When it comes to represent a set of typeIDs (or field indices), we can use a bitset (sparse or dense) because these IDs are from a linear space of non-negative integers, which makes perfect sense for bit-sets. 

A related question that I have is (for my understanding primarily): what is our flow of execution when we visit a node that has multiple types. If I understand, under this design, we'll assign a 8-bit ID to a unique combination of types applicable to this node. How do we look up what those types are?

libgalois/include/katana/PropertyGraph.h
roshandathathri: That's up to you. But for such code reorgs, a TODO in the code is fine IMO.

libgalois/src/PropertyGraph.cpp
roshandathathri: Now I understand your comment. Yes, `std::vector<bool>` may be better than `std::set<uint8_t>`.  Note that we don't need atomic accesses so `DynamicBitset` is not required. Also note that the size for none of these vectors or sets are O(vertices) or O(edges), so the memory overhead is not high in either case. Nevertheless, I think using `std::vector<bool>` will make the code simpler. I'll make a pass and try to update it everywhere. If there are any issues, I'll let you know.

I'll answer your flow question in the enterprise-repo because there's more context there.

libgalois/src/PropertyGraph.cpp
roshandathathri: Ok, I'll try that.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -97,6 +97,29 @@ struct KATANA_EXPORT GraphTopology {
...
+public:
+  /// TypeID uniquely identifies/contains a combination/set of types
+  /// TypeID is represented using 8 bits
+  /// TypeID for nodes is distinct from TypeID for edges
+  using TypeID = uint8_t;
+  using ArrowTypeID = arrow::UInt8Type;
+  static constexpr TypeID kUnknownType = TypeID{0};
+  static constexpr TypeID kInvalidType = std::numeric_limits<TypeID>::max();
+  /// A map from TypeID to the set of the type(-name)s it contains
+  using TypeIDToTypesMap = std::vector<std::unordered_set<std::string>>;
```
aneeshdurg: Can this be `TypeIdToTypeNames`? And similarly `TypeNamesToTypeIDs` below?

libgalois/include/katana/PropertyGraph.h
aneeshdurg: When this is an OpaqueID, I wonder if `TypeID` is even the right name. opaquely we probably want this presented as a collection of ids right?

roshandathathri: That's interesting. TypeID represents/identifies a collection of type names. We can think of better names. Any suggestions?

libgalois/include/katana/PropertyGraph.h
ddn0: The relevant bit is: https://google.github.io/styleguide/cppguide.html#Function_Names and https://google.github.io/styleguide/cppguide.html#Exceptions_to_Naming_Rules

If a function behaves like a field access (e.g., constant time, no exceptions), it may be named like a variable `num_node_types()`. If a method is intended to mirror some other existing interface, it may also be in snake case (e.g., `.size()`, `.find_if()`).

As to `GetFoo()` vs `Foo()`, I would use `GetFoo` although there is no specific requirement in the Google style guide to do this AFAIK. To me it seems implied because if you didn't like `GetFoo` you would just name the function like a variable `foo()` and in general, people name functions as active verbs.

Let's not churn the codebase unnecessarily. The reason we have a style guide is so that people can spend their time discussing important functional requirements of code and we can delegate questions of arbitrary style to a separate document.

amberhassaan: This goes a bit beyond just stylistic issues. What I want to discourage is setters. One of the gifts of Java is mindless generation of getters and setters (IDEs automate it for you). Ideally, a setter should be a moment to pause and think: do we really need this? If the answer is yes, there's usually a better name for the setter that conveys the intent better. So my line of defense is to avoid getters so that you don't ever get to setters, but I'm OK with keeping getters (with name GetBlah) in the interest of avoiding purely stylistic changes.

ddn0: > What I want to discourage is setters.

That is fine but there is a limit to what can and should be done in the context of a particular PR, and as far as I can see, most comments have been about getters so far.

If "no setters" is important to you, there probably should be wider discussion so that we can reach some agreement and enforce it throughout the codebase; otherwise nothing will converge.

amberhassaan: Yes, the discussion is beyond the scope of this PR and we should have that discussion. I've just started looking at the code and within PropertyGraph.h, I see a bunch of setters and two conventions for getters (getters with and without get). Not sure why we have both the styles? Is it because one API is for the partitioner or underlying substack while the other is for the stack above? 
I'm for consistent and clean interfaces. We can have a discussion about cost and benefits of such a change, but to me a proliferation of getters and setters is a sign of degrading quality (degrading towards typical Java code :)).

libgalois/src/PropertyGraph.cpp
roshandathathri: No, that doesn't work either:
`/usr/include/c++/9/ext/new_allocator.h:145:20: error: no matching function for call to ‘std::unordered_set<std::__cxx11::basic_string<char> >::unordered_set(std::set<std::__cxx11::basic_string<char>, std::less<std::__cxx11::basic_string<char> >, std::allocator<std::__cxx11::basic_string<char> > >)’`

libgalois/src/PropertyGraph.cpp
roshandathathri: I decided to go with `std::bitset` for set of TypeIDs. However, for set of field indices (used only in this function), I decided to keep it as a vector of field indices (`katana::gstl::Vector<int>`) for these reasons:
(1) `std::bitset` does not support dynamic (fixed) size. The number of field indices is only known at runtime and I don't want to put any restriction on it because it includes properties.
(2) `std::vector<bool>` does not  support counting the set bits, so it would make this function more complex to do that.
(3) `katana::DynamicBitset` does not support thread-specific allocators, so it may make the function slower.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -288,35 +337,111 @@ class KATANA_EXPORT PropertyGraph {
...
+    return node_properties()->schema();
   }
 
   std::shared_ptr<arrow::Schema> edge_schema() const {
-    return rdg_.edge_properties()->schema();
+    return edge_properties()->schema();
   }
 
+  /// \returns the number of node types
+  size_t GetNodeTypesNum() const { return node_type_name_to_type_ids_.size(); }
```
scober: Does this return the number of types allowed under the schema or the number of types present in the local graph?

libgalois/include/katana/PropertyGraph.h
arthurp: Unions of types are types. So we could call this `UnionTypeID`. Another option that would make sense to non-type theorists would be `TypeSetID`.

libgalois/include/katana/PropertyGraph.h
arthurp: Please add a TODO

arthurp (approved):

libgalois/include/katana/PropertyGraph.h
roshandathathri: I like `TypeSetID` more but we have a set of current `TypeID`s called `TypeIDsSet`. If we go with that, it would be `TypeSetIDsSet`. Is that confusing?

libgalois/include/katana/PropertyGraph.h
roshandathathri: Will do.

libgalois/include/katana/PropertyGraph.h
arthurp: I think `TypeSetIDSet` (I wouldn't include the plural "s" here) is actually *less* confusing than having something called `TypeIDSet` that is actually a set of sets conceptually. We are already dealing with sets of sets, so just including it in the type name and documenting it seems like the best option. Hiding it will make things worse, since people will get farther thinking that `TypeID` represents a single type.

roshandathathri: Ok, makes sense.

libgalois/include/katana/PropertyGraph.h
roshandathathri: The former. There is no need for the latter IMO.

libgalois/src/PropertyGraph.cpp
amberhassaan: `std::bitset` is a good choice for now. We should think about whether there's benefit to using sparse bit set as we expect sparsity when the graph has different kinds of nodes. 

Secondly, I don't see why using `std::bitset` with max number of bits is any worse than using `std::vector`. Similarly, I don't see (and you can help me understand) why `std::vector<bool>` doesn't work equally well.

libgalois/src/PropertyGraph.cpp
amberhassaan: It's minor and we can defer it for a later time.

libgalois/include/katana/PropertyGraph.h
amberhassaan: I vote against TypeSetIDSet. It's a mouthful and I find it confusing. We have 2 things at hand:
1. Type IDs 
2. Sets of Type IDs (call them TypeIDSet or what you prefer), which are represented as TypeIDs themselves (just a different integer for each set of type IDs in the space of possible type IDs). But we can also represent these sets as bit sets.

libgalois/include/katana/PropertyGraph.h
amberhassaan: Is the method naming in PropertyGraph inconsistent (some are lower case and some start with uppercase) or am I missing some convention being followed here?

libgalois/include/katana/PropertyGraph.h
roshandathathri: I don't mind either name but I tend toward `TypeID` and `TypeIDsSet` because even though `TypeID` refers to a set of types, in practice it refers to a single derived class-type (the other classes are base class-types). We can revisit this name later because I don't think we have a consensus regarding the terms: type, tag, and label.

libgalois/include/katana/PropertyGraph.h
roshandathathri: Didn't Donald answer that question in this thread above? If he didn't, can you please be more specific in your question?

libgalois/src/PropertyGraph.cpp
roshandathathri: That's (2): `std::vector<bool>` does not support counting the set bits directly. It makes the code more verbose and difficult to read IMO.

libgalois/include/katana/PropertyGraph.h
arthurp: I'm pretty strongly against using `TypeID` to represent a set of `TypeID`s. I'm starting to wonder if I misunderstand what's going on here. So let me explain what I think is going on, and then people can tell me if I missed the point and maybe the text will be useful for documentation.

There are 3 things we are dealing with here and trying to create types for:

1. Actual types (also known as labels) which are usually represented as a boolean indicator property in the RDG. We don't really directly handle these in the code except during loading. Let's call these atomic types.
2. Sets of atomic types which are found by looking at the atomic types that entities have in the data set and finding sets that appear together (e.g., Doctor and Person might be a group because every Doctor is also a Person). These are used in our compressed property representation (which isn't implemented yet, IIUC, but will come along eventually). They are not used strictly as intersections (I think I said union above when I meant intersection, BTW) of their constituent types, IIRC, because entities are associated with a set of atomic types may actually have most but not all of the types in the set.
3. Sets of sets of atomic types. I don't know what these are used for, but I thought Roshan said we had these.

My understanding was that `TypeID` (which I pushed to call `TypeSetID`) was a numeric ID for 2 above (a set of atomic types). And that `TypeSetIDSet` was an actual collection (`std::vector` or whatever) of `TypeSetID`s and hense would be 3 above.

I'm starting to really doubt myself. So please tell me what the reality here is.

roshandathathri: More detailed response after the meeting but your understanding of 2 and 3 is correct.

amberhassaan: @arthurp : Your understanding is correct. I'll add this (and you may already have understood this) that 1 and 2 above are represented as integers in the same 8 bit space. I agree with the idea of 3 and it's being used in query code (the other PR), but not sure if we need to define such a type in PropertyGraph.

libgalois/include/katana/PropertyGraph.h
amberhassaan: Well, the discussion with @ddn0 was regarding (elimination of) setters, which is a separate issue that's not high priority but important to me. That aside, there's still the issue of two styles of method names. Is there a reason or is it something to be fixed? If it's something we should fix then we can create an issue.

libgalois/include/katana/PropertyGraph.h
arthurp: I didn't realize that we assigned numeric ID to both atomic types and the sets of types. I don't actually like that because from my understanding the sets of types are not always true intersection types since an entity may be assigned to a set even if it doesn't actually have every type in the set. If I'm wrong about that and entities are only assigned to a set of types if they have all the types in the set, then they are intersection types and I'm fine with having a single ID space for all types as long as we make sure we are aware of the interactions between atomic and intersection types.

libgalois/include/katana/PropertyGraph.h
ddn0: Let me summarize.

There are two ways to name a method.
- PascalCase() - most common, active verb phrase
- snake_case() - rarer, noun-phrase, should behave like a field access: O(1), no-exceptions, etc.

amberhassaan: I see. More reasons to hate GSG, I guess.

libgalois/include/katana/PropertyGraph.h
roshandathathri: @amberhassaan Two corrections:
(A) We do NOT assign numeric ID to (1). A type is recognized by its name (just like a property). 
(B) We need to define (3) in PropertyGraph because it maintains a map from a type-name to (3). 

@arthurp One correction:
(C) I'm not sure if union/intersection is an appropriate term to use here but the way you used it seemed like (2) always corresponds true intersection types. A node has a numeric ID corresponding to (2) if and only if the set of all its atomic types is strictly equal to the set identified by that numeric ID.

roshandathathri: Listening to @arthurp's comments, I'm now more inclined towards `TypeSetID` and `TypeSetIDSet` for (2) and (3) respectively.

amberhassaan: @roshandathathri : aren't you assigning 8 bit IDs to individual types as well?

roshandathathri: After looking at the code, here are the names I'm going for:
```cpp
  /// TypeSetID uniquely identifies/contains a combination/set of types
  /// TypeSetID is represented using 8 bits
  /// TypeSetID for nodes is distinct from TypeSetID for edges
  using TypeSetID = uint8_t;
  using ArrowTypeSetID = arrow::UInt8Type;
  static constexpr TypeSetID kUnknownType = TypeSetID{0};
  static constexpr TypeSetID kInvalidType =
      std::numeric_limits<TypeSetID>::max();
  /// A set of TypeSetIDs
  using SetOfTypeSetIDs =
      std::bitset<std::numeric_limits<TypeSetID>::max() + 1>;
  /// A set of type names
  using SetOfTypeNames = std::unordered_set<std::string>;
  /// A map from TypeSetID to the set of the type names it contains
  using TypeSetIDToSetOfTypeNamesMap =
      std::vector<SetOfTypeNames>;
  /// A map from the type name to the set of the TypeSetIDs that contain it
  using TypeNameToSetOfTypeSetIDsMap =
      std::unordered_map<std::string, SetOfTypeSetIDs>;
```
It's verbose but more clear IMO.

roshandathathri: @amberhassaan No, that's (2) - set of types.

arthurp: >  I'm not sure if union/intersection is an appropriate term to use here but the way you used it seemed like (2) always corresponds true intersection types. A node has a numeric ID corresponding to (2) if and only if the set of all its atomic types is strictly equal to the set identified by that numeric ID.

Yep, those are intersection types. A specific kind of intersection type, but still intersection types. https://en.wikipedia.org/wiki/Intersection_type

libgalois/include/katana/PropertyGraph.h
```diff
@@ -97,6 +98,34 @@ struct KATANA_EXPORT GraphTopology {
...
+      std::unordered_map<std::string, SetOfTypeSetIDs>;
+
+  // Pass through topology API
+  using node_iterator = GraphTopology::node_iterator;
+  using edge_iterator = GraphTopology::edge_iterator;
+  using edges_range = GraphTopology::edges_range;
+  using iterator = GraphTopology::iterator;
+  using Node = GraphTopology::Node;
+  using Edge = GraphTopology::Edge;
+
```
witchel, line 127: This is one of the few remaining places that is not using gluon's `PartitionTypes.h`.  Has the time come?

witchel (commented):
LGTM.

libgalois/include/katana/PropertyGraph.h
arthurp: This is in the open repo. So we cannot use anything from gluon. What should probably happen is that the types should be moved here and then referenced from gluon. This is more basic than gluon really.

PR #262, opened 2021-05-31
Python test and packaging improvements

arthurp:
* Improve error output from pytest runs.
* Add a small script to do correct conda package builds.


gliga (approved):
Neat

PR #264, opened 2021-06-02
conda: Accept KATANA_COMPONENTS

ddn0:


conda_recipe/build.sh
```diff
@@ -18,13 +18,19 @@ if [ -z "${CMAKE_BUILD_PARALLEL_LEVEL}" ]; then
 fi
 echo "Building with parallelism: ${CMAKE_BUILD_PARALLEL_LEVEL}"
 
+if [ -z "${KATANA_COMPONENTS}" ]; then
+  KATANA_COMPONENTS=""
+fi
+echo "Building with components: ${KATANA_COMPONENTS}"
```
arthurp, line 24: Passing environment variables into a recipe is not recommended and doesn't work without additional setup. I would really prefer to have a fixed set of components built into the Conda packages in the conda package metadata. If we want to build multiple conda packages with different component sets, we can set up conda variant builds and even have some control over their priority if someone doesn't specify which they want.

That said, this code is fine. I just don't know if we will actually want to use it.

arthurp: So I see you added the appropriate variable in enterprise. What you are doing is OK. I'll investigate how we could control this with variants and the like in the future.

ddn0: The other option I considered was just having a separate build.sh for enterprise rather than linking to this one.

arthurp: We'll have to deal with the component issue in open eventually I suspect. So this is reasonable.

arthurp (approved):

PR #270, opened 2021-06-07
Allow rdg to remove all its properties

scober:
The core issue here is that I want to modify an in-memory graph in place. And it is difficult to add a node property table that has a different number of rows than the RDG was created with.

### DIFF
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -559,6 +559,11 @@ public:
   Result<void> RemoveEdgeProperty(int i);
   Result<void> RemoveEdgeProperty(const std::string& prop_name);
 
+  /// Remove all node properties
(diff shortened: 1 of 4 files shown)


witchel (approved):
At one point in ingest, I did this by calling `RemoveProperty(0)` in a loop.  But this seems like a fine thing to do.  Should we have `drop_node_properties` and `drop_edge_properties`?  And given our conventions, shouldn't it be called `DropProperties`?

scober:
> At one point in ingest, I did this by calling `RemoveProperty(0)` in a loop. But this seems like a fine thing to do. Should we have `drop_node_properties` and `drop_edge_properties`? And given our conventions, shouldn't it be called `DropProperties`?

I started out removing columns in a loop as well. The issue was that I couldn't remove the schema and it remembered the length of the table. And I want to change the length of the table.

I will split it into node and edge variants and bring it into line with the style.

PR #271, opened 2021-06-08
changing RDGManifest/tsuba helpers to support views

yshalabi:
This PR drops the name server and transitions from RDGMeta to RDGManifest. RDGManifest supports the notion of graph views, allowing for the persistence of multiple graphs alongside each other.

A tool is provided (scripts/migrate_rdgs.py) to help migrate RDGs to the new format. Invoke it with: './scripts/migrate_rdgs.py --fs /path/to/rdg --s3 s3://bucket-with-rdg/dirs'.

The NameServer is also deprecated with this PR.


libgalois/src/PropertyGraph.cpp
```diff
@@ -626,9 +631,15 @@ katana::PropertyGraph::ReportDiff(const PropertyGraph* other) const {
 katana::Result<void>
 katana::PropertyGraph::Write(
     const std::string& rdg_name, const std::string& command_line) {
+  KATANA_LOG_DEBUG(
+      "PropertyGraph::Write(rdg_name={},command_line={}", rdg_name,
+      command_line);
```
yshalabi, line 636:
```suggestion
```

libgalois/src/PropertyGraph.cpp
```diff
@@ -626,9 +631,15 @@ katana::PropertyGraph::ReportDiff(const PropertyGraph* other) const {
 katana::Result<void>
 katana::PropertyGraph::Write(
     const std::string& rdg_name, const std::string& command_line) {
+  KATANA_LOG_DEBUG(
+      "PropertyGraph::Write(rdg_name={},command_line={}", rdg_name,
```
yshalabi, line 635:
```suggestion
```

libgalois/src/PropertyGraph.cpp
```diff
@@ -626,9 +631,15 @@ katana::PropertyGraph::ReportDiff(const PropertyGraph* other) const {
 katana::Result<void>
 katana::PropertyGraph::Write(
     const std::string& rdg_name, const std::string& command_line) {
+  KATANA_LOG_DEBUG(
```
yshalabi, line 634:
```suggestion
```

libgalois/src/PropertyGraph.cpp
```diff
@@ -626,9 +631,15 @@ katana::PropertyGraph::ReportDiff(const PropertyGraph* other) const {
...
+      "PropertyGraph::Write(rdg_name={},command_line={}", rdg_name,
+      command_line);
   if (auto res = tsuba::Create(rdg_name); !res) {
     return res.error();
   }
+  KATANA_LOG_DEBUG(
```
yshalabi, line 640:
```suggestion
```

libgalois/src/PropertyGraph.cpp
```diff
@@ -626,9 +631,15 @@ katana::PropertyGraph::ReportDiff(const PropertyGraph* other) const {
...
+      "PropertyGraph::Write(rdg_name={},command_line={}", rdg_name,
```
yshalabi, line 641:
```suggestion
```

libgalois/src/PropertyGraph.cpp
```diff
@@ -626,9 +631,15 @@ katana::PropertyGraph::ReportDiff(const PropertyGraph* other) const {
...
+      command_line);
```
yshalabi, line 642:
```suggestion
```

libtsuba/include/tsuba/tsuba.h
```diff
@@ -67,25 +67,23 @@ KATANA_EXPORT katana::Result<void> Close(RDGHandle handle);
...
-KATANA_EXPORT katana::Result<void> RegisterIfAbsent(const std::string& name);
-
-/// Forget an RDG, detaching it from the namespace
-/// \param name is storage location prefix that the RDG is stored in
-KATANA_EXPORT katana::Result<void> Forget(const std::string& name);
-
 struct KATANA_EXPORT RDGStat {
   uint64_t num_partitions{0};
   uint32_t policy_id{0};
+  //TODO(yasser): add field for view type
```
yshalabi, line 73:
```suggestion
```

libtsuba/include/tsuba/tsuba.h
```diff
@@ -67,25 +67,23 @@ KATANA_EXPORT katana::Result<void> Close(RDGHandle handle);
...
   bool transpose{false};
 };
 
 /// Get Information about the graph
 KATANA_EXPORT katana::Result<RDGStat> Stat(const std::string& rdg_name);
+//KATANA_EXPORT katana::Result<RDGStat> StatView(const std::string& rdg_name);
```
yshalabi, line 79:
```suggestion
```

libtsuba/src/MemoryNameServerClient.cpp
```diff
@@ -26,14 +26,15 @@ MemoryNameServerClient::lookup(const std::string& key) {
...
+katana::Result<RDGManifest>
 MemoryNameServerClient::Get(const katana::Uri& rdg_name) {
   return lookup(rdg_name.Encode());
 }
 
 katana::Result<void>
 MemoryNameServerClient::CreateIfAbsent(
-    const katana::Uri& rdg_name, const RDGMeta& meta) {
+    const katana::Uri& rdg_name, const RDGManifest& meta) {
+  KATANA_LOG_DEBUG("CreateIfAbsent({})", rdg_name.string());
```
yshalabi, line 37:
```suggestion
```

libtsuba/src/tsuba.cpp
```diff
@@ -122,20 +166,21 @@ tsuba::Create(const std::string& name) {
...
+  tsuba::RDGManifest meta{};
 
   katana::CommBackend* comm = Comm();
   if (comm->ID == 0) {
-    if (ContainsValidMetaFile(name)) {
-      return KATANA_ERROR(
-          ErrorCode::InvalidArgument,
-          "unable to create {}: path already contains a valid meta file", name);
-    }
+    //if (ContainsValidMetaFile(name)) {
```
yshalabi, line 175:
```suggestion
```

libtsuba/src/tsuba.cpp
```diff
@@ -122,20 +166,21 @@ tsuba::Create(const std::string& name) {
...
+    //  return KATANA_ERROR(
```
yshalabi, line 176:
```suggestion
```

libtsuba/src/tsuba.cpp
```diff
@@ -122,20 +166,21 @@ tsuba::Create(const std::string& name) {
...
+    ///      ErrorCode::InvalidArgument,
```
yshalabi, line 177:
```suggestion
```

libtsuba/src/tsuba.cpp
```diff
@@ -122,20 +166,21 @@ tsuba::Create(const std::string& name) {
...
+    //      "unable to create {}: path already contains a valid meta file", name);
```
yshalabi, line 178:
```suggestion
```

libtsuba/src/tsuba.cpp
```diff
@@ -122,20 +166,21 @@ tsuba::Create(const std::string& name) {
...
+    // }
```
yshalabi, line 179:
```suggestion
```

libtsuba/include/tsuba/RDG.h
```diff
@@ -181,15 +181,18 @@ class KATANA_EXPORT RDG {
 
   const FileView& topology_file_storage() const;
 
+  void set_view_name(const std::string& v) { view_type_ = v; }
+
```
witchel, line 221: Do we want views to be strings?  Or class enums convertible to strings?

libtsuba/include/tsuba/tsuba.h
```diff
@@ -67,16 +67,15 @@ KATANA_EXPORT katana::Result<void> Close(RDGHandle handle);
...
-KATANA_EXPORT katana::Result<void> Forget(const std::string& name);
+struct KATANA_EXPORT RDGView {
+  uint64_t view_version{0};
+  std::string view_type;
+  std::string view_args;
+  std::string view_path;
+  uint64_t num_partitions{0};
+  uint32_t policy_id{0};
+  bool transpose{false};
+};
```
witchel, line 82: Bye `RegisterIfAbsent`, I will not miss you.

witchel: This looks like an important data structure.  Do we get doxygen comments for it?

libtsuba/include/tsuba/tsuba.h
```diff
@@ -87,6 +86,9 @@ struct KATANA_EXPORT RDGStat {
 /// Get Information about the graph
 KATANA_EXPORT katana::Result<RDGStat> Stat(const std::string& rdg_name);
 
+KATANA_EXPORT katana::Result<std::vector<RDGView>> ListAvailableViews(
+    const std::string& rdg_dir);
+
```
witchel, line 95: I don't want to overdesign things, but if you look at the interfaces for listing pseudo-directories in buckets, they all return a "page" of results and a token for more.  
And/or I wonder if a limit parameter would be useful here to reduce memory use for large directories.  But then you'd need a start string (list everything equal to or later than the start string up to the given limit).

libtsuba/src/RDGManifest.h
```diff
@@ -0,0 +1,151 @@
...
+  static katana::Result<RDGManifest> Make(
+      const katana::Uri& uri, const std::string& view_type, uint64_t version);
+
+  const katana::Uri& dir() const { return dir_; }
+  uint64_t version() const { return version_; }
+  uint32_t num_hosts() const { return num_hosts_; }
+  uint32_t policy_id() const { return policy_id_; }
+  uint64_t previous_version() const { return previous_version_; }
+  const std::string& get_viewtype() const { return view_type_; }
+  void set_viewtype(std::string v) { view_type_ = v; }
```
witchel, line 112: Can we just call the getter `viewtype()`?  It is more consistent with the other getters in this class.

libtsuba/src/RDGManifest.h
```diff
@@ -0,0 +1,151 @@
...
+  std::string view_type_;
+  std::vector<std::string> view_args_;
+
+  RDGManifest(katana::Uri dir)
+      : dir_(std::move(dir)), view_type_(gDefaultRDGViewType) {}
+  RDGManifest(katana::Uri dir, const std::string& view_type)
+      : dir_(std::move(dir)) {
+    view_type_ = gDefaultRDGViewType;
+    if (view_type != "") {
+      view_type_ = view_type;
```
witchel, line 45: Is `!view_type.empty()` an equivalent test?  If so, I'd prefer it.

libtsuba/src/tsuba.cpp
```diff
@@ -1,19 +1,19 @@
 #include "tsuba/tsuba.h"
 
+#include <sstream>
+
```
witchel: Is this include necessary?

scripts/update_metadata_filenames.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import re
+import subprocess
+import sys
```
witchel: Cool conversion script.

witchel (approved):
Looks good from my end, you can see my comments for conerns.

I would want @tylershunt to sign off on the loss of all that global state.

libtsuba/include/tsuba/RDG.h
yshalabi: strings until structure emerges imo

libtsuba/include/tsuba/tsuba.h
yshalabi: I will DEFINITELY not miss it either :)
I will add some comments.

libtsuba/include/tsuba/tsuba.h
yshalabi: Probably in the future this will be pretty useful. For now, I say we stick with this as it gets us off the ground.

libtsuba/src/tsuba.cpp
yshalabi: holdover, I will remove it.

libtsuba/include/tsuba/RDG.h
tylershunt: Slight preference for enums here. It mostly serves a documentation purpose, I can look at the `enum` and know all of the view types this version of the code supports.

It also stops future refactor-ers from having to ask the question "is the view name guaranteed to be valid here".

I don't know what structure would emerge other than 'this is a list of the view types we support', but if you feel strongly I won't block this over that.

libtsuba/include/tsuba/tsuba.h
tylershunt: Pagination here seems like overkill to me as well. I think the number of views will be on the order of 10s or maybe 100s per version. (This does only return the view for the latest version right?)

libtsuba/src/RDGManifest.h
```diff
@@ -0,0 +1,151 @@
...
+  /// \returns the constructed RDGManifest and the directory of its contents
+  static katana::Result<RDGManifest> Make(
+      const katana::Uri& uri, const std::string& view_type, uint64_t version);
+
+  const katana::Uri& dir() const { return dir_; }
+  uint64_t version() const { return version_; }
+  uint32_t num_hosts() const { return num_hosts_; }
+  uint32_t policy_id() const { return policy_id_; }
+  uint64_t previous_version() const { return previous_version_; }
+  const std::string& get_viewtype() const { return view_type_; }
```
tylershunt: Small style nit. Per our guidelines this would be either `viewtype()` or `GetViewType()`. I prefer the first (in general, but here especially) because it matches the others in this class.

https://google.github.io/styleguide/cppguide.html#Function_Names

libtsuba/src/RDGManifest.h
tylershunt: oh @witchel  noticed too 😄

libtsuba/src/RDGManifest.cpp
```diff
@@ -0,0 +1,320 @@
...
+  if (auto it = j.find("transpose"); it != j.end()) {
+    it->get_to(meta.transpose_);
+  }
+  if (auto it = j.find("lineage"); it != j.end()) {
+    it->get_to(meta.lineage_);
+  }
+
+  if (magic != kRDGMagicNo) {
+    // nlohmann::json reports errors using exceptions
+    throw std::runtime_error("RDG Magic number mismatch");
```
tylershunt: Probably `RDG Manifest magic number mismatch`

libtsuba/src/RDGManifest.cpp
```diff
@@ -0,0 +1,320 @@
...
+    const katana::Uri& uri, const std::string& view_name, uint64_t version) {
+  KATANA_LOG_DEBUG_ASSERT(uri.empty() || !IsMetaUri(uri));
+  KATANA_LOG_ASSERT(!view_name.empty());
+  return uri.Join(
+      fmt::format("katana_vers{:07d}_{}.manifest", version, view_name));
+}
+
+// if it doesn't name a meta file, assume it's meant to be a managed URI
+bool
+RDGManifest::IsMetaUri(const katana::Uri& uri) {
```
tylershunt: There are a couple of references to `Meta` that are now confusing. Should probably be `Manifest`

libtsuba/src/RDGManifest.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+Result<RDGManifest>
+RDGManifest::Make(
+    const katana::Uri& uri, const std::string& view_type, uint64_t version) {
+  return MakeFromStorage(FileName(uri, view_type, version));
+}
+
+Result<RDGManifest>
+RDGManifest::Make(RDGHandle handle) {
+  return handle.impl_->rdg_meta();
```
tylershunt: another reference in a member name this time probably should be `->rdg_manifest()`

libtsuba/src/RDGManifest.h
```diff
@@ -0,0 +1,151 @@
...
+#include "katana/JSON.h"
+#include "katana/Logging.h"
+#include "katana/Uri.h"
+#include "katana/config.h"
+#include "tsuba/RDGLineage.h"
+#include "tsuba/tsuba.h"
+
+namespace tsuba {
+
+static const char* gDefaultRDGViewType = "rdg";
```
tylershunt: Should be `kDefaultRDGViewType`

https://google.github.io/styleguide/cppguide.html#Constant_Names

libtsuba/src/tsuba.cpp
```diff
@@ -27,25 +27,9 @@ FileList(const std::string& dir) {
...
-  for (const std::string& file : list_res.value()) {
-    if (auto res = tsuba::RDGMeta::ParseVersionFromName(file); res) {
-      return true;
-    }
-  }
-  return false;
-}
-
 katana::Result<katana::Uri>
 FindLatestMetaFile(const katana::Uri& name) {
```
tylershunt: Probably `FindLatestManifestFile`

libtsuba/src/tsuba.cpp
```diff
@@ -224,14 +166,60 @@ tsuba::Stat(const std::string& rdg_name) {
...
+  RDGManifest meta = rdg_res.value();
   return RDGStat{
       .num_partitions = meta.num_hosts(),
       .policy_id = meta.policy_id(),
       .transpose = meta.transpose(),
   };
 }
 
+katana::Result<std::vector<tsuba::RDGView>>
+tsuba::ListAvailableViews(const std::string& rdg_dir) {
```
tylershunt, line 176: I strongly think this should only return valid views for the latest version. Or take a version argument and return views for that version.

tylershunt (changes requested):
I think this seems very close and there's a lot of stuff here that I really like. The one thing that's missing that I think is necessary is that I don't see a way to write out a new view without updating the version of the RDG, which is what I thought was the whole point of this exercise.

If i'm missing something please let me know.

libtsuba/include/tsuba/tsuba.h
witchel: I actually missed the most important thing here which is that this is listing views of a version.  
Along the lines of OpaqueIDs preventing accidental cross-confusion, making views a typed enum will prevent auto casting from version numbers to view types (though now one is an integer and the other a string, so no present danger).

scripts/update_metadata_filenames.py
tylershunt: Can we name it something like `migrate_rdgs.py` ?

libtsuba/src/RDGManifest.h
yshalabi: Yeah, this function returns all views and makes no assumptions regarding the users intentions for requesting the available view list.

tylershunt:
Another note/question will this break the enterprise build? If it does could you link the PR that fixes it?

libtsuba/include/tsuba/tsuba.h
yshalabi: Yeah, eventually once we start to see what kinds of views we support (aside from partitioned RDG views) we will create enums and probably create some kind of plugin-based view handler support. For now, this method simply returns all the views available in the directory to the user.

yshalabi:
> Another note/question will this break the enterprise build? If it does could you link the PR that fixes it?
Yes, most definitely. Here is what keeps enterprise running. The two will be merged simultaneously to minimize disruption.
https://github.com/KatanaGraph/katana-enterprise/pull/1295

libgalois/include/katana/PropertyGraph.h
```diff
@@ -193,9 +193,11 @@ class KATANA_EXPORT PropertyGraph {
   Result<void> Validate();
 
   Result<void> DoWrite(
-      tsuba::RDGHandle handle, const std::string& command_line);
+      tsuba::RDGHandle handle, const std::string& command_line,
+      bool retain_version);
   Result<void> WriteGraph(
-      const std::string& uri, const std::string& command_line);
+      const std::string& uri, const std::string& command_line,
+      bool retain_graph);
```
witchel, line 208: I do not like the name `retain_version` when it comes to writing.  My model is every write to the graph creates a new version.  Are you adding a `WriteView`?  A `WriteView` operation would not increase the graph's version number.

yshalabi: I thought this was the cleanest way to do it. Otherwise, we will have a DoWrite, DoWriteView and a WriteGraph and a WriteGraphView. I think its too much considering the only difference is in whether or not the version number is incremented.

libtsuba/include/tsuba/RDG.h
yshalabi: Right now we have the partitioned RDG view type. In my current implementation I treat the variations in partitioned RDGs as distinct types. That means type(4-way partitioned RDG) != type(8-way partitioned RDG). Simplifies the query "is there this view available?" to constructing a string and scanning for filenames which match this string. 

I also dont mind refactoring to have the enums and view type construction if you strongly prefer. My preference is to get it in first and then make these tweaks after steady-state emerges.

libtsuba/src/RDGManifest.cpp
yshalabi: Yeah, I have been meaning to go through and sed meta==>manifest. Will do that before completion.

libgalois/include/katana/PropertyGraph.h
witchel: I disagree.  Boolean parameters are tough to use because you can't tell from the call site what they mean.  But fundamentally, a graph is read-only.  If you write the graph, the version number has to increment.  Creating a view is not writing a graph, it is changing the format of an existing graph.  `Write(false)` does not make that clear and actually hides what is going on.

I'm not worried about naming internal functions.  I suspect we will be able to merge them sensibly.

yshalabi: Boolean is definitely not the most readable way to do it. I have added PropertyGraph::WriteView and added RDG::RDGVersioningPolicy which is used by DoWrite to dictate whether the RDG Version should change with the store. Thanks.

witchel: Thank you!

tylershunt (commented):
I'm confident through our conversations that we're now on the same page (thanks you!).

tylershunt (approved):

PR #275, opened 2021-06-09
Benchmarking/perf analysis

nicelhc13:


libgalois/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ set(sources
         src/analytics/jaccard/jaccard.cpp
         src/analytics/k_core/k_core.cpp
         src/analytics/k_truss/k_truss.cpp
-        src/analytics/pagerank/pagerank-pull.cpp
+        src/analytics/pagerank/pagerank-pull-SoA.cpp
```
roshandathathri: Don't create a new file. Replace the contents of the existing file.

libgalois/include/katana/Properties.h
```diff
@@ -95,6 +98,7 @@ struct PropertyViewTuple<std::tuple<Args...>> {
   using type = std::tuple<katana::PropertyViewType<Args>...>;
 };
 
+#if 0
```
roshandathathri: Delete the code instead of doing this.

libgalois/src/analytics/pagerank/pagerank-impl.h
```diff
@@ -49,4 +49,40 @@ katana::Result<void> PagerankPushSynchronous(
     katana::PropertyGraph* pg, const std::string& output_property_name,
     katana::analytics::PagerankPlan plan);
 
+struct NodeValueOut {
+  uint32_t out;
+  PRTy value;
+};
+
+template <typename T>
+struct StructProperty {
```
roshandathathri: Move this to Properties.h

roshandathathri (changes requested):

libgalois/include/katana/TypedPropertyGraph.h
```diff
@@ -66,6 +66,12 @@ class TypedPropertyGraph {
 
   // Graph accessors
 
+  template <typename NodeIndex>
+  PropertyValueType<NodeIndex>* GetArray() {
```
roshandathathri: You can drop this if this is not used.

libgalois/src/analytics/pagerank/pagerank-impl.h
```diff
@@ -49,4 +49,12 @@ katana::Result<void> PagerankPushSynchronous(
     katana::PropertyGraph* pg, const std::string& output_property_name,
     katana::analytics::PagerankPlan plan);
 
+struct NodeValueOut {
+  uint32_t out;
+  PRTy value;
+};
+
+using NodeValueOutTy = NodeValueOut;
```
roshandathathri: Drop this. Call the struct `ValueOut` and call the property `NodeValueOut`

libgalois/src/analytics/pagerank/pagerank-pull-inlining.cpp
```diff
@@ -0,0 +1,300 @@
+/*
```
roshandathathri: Drop this file

libgalois/src/analytics/pagerank/pagerank-pull.cpp
```diff
@@ -17,14 +17,17 @@
  * Documentation, or loss or inaccuracy of data of any kind.
  */
 
+#include <arrow/type.h>
+
+#include "katana/ErrorCode.h"
+#include "katana/Properties.h"
```
roshandathathri: Are these headers required?

libgalois/src/analytics/pagerank/pagerank-pull.cpp
```diff
@@ -40,12 +43,11 @@ InitNodeDataTopological(Graph* graph) {
   katana::do_all(
       katana::iterate(*graph),
       [&](const GNode& n) {
-        auto& sdata_value = graph->GetData<NodeValue>(n);
-        auto& sdata_nout = graph->GetData<NodeNout>(n);
-        sdata_value = init_value;
-        sdata_nout = 0;
+        NodeValueOutTy& sdata = graph->GetData<NodeValueOutProp>(n);
```
roshandathathri: Use `auto` to be consistent.

libgalois/include/katana/Properties.h
```diff
@@ -182,6 +171,106 @@ ConstructPropertyViews(const std::vector<arrow::Array*>& arrays) {
...
+std::enable_if_t<
+    !HasAllocate<Prop>::value, Result<std::shared_ptr<arrow::Table>>>
+Allocate(size_t num_rows, const std::string& name) {
+  using ArrowType = typename PropertyTraits<Prop>::ArrowType;
+  using Builder = typename arrow::TypeTraits<ArrowType>::BuilderType;
+  using CType = typename arrow::TypeTraits<ArrowType>::CType;
+  Builder builder;
+
+  std::vector<CType> rows(num_rows);
+  if (auto r = builder.AppendValues(rows); !r.ok()) {
```
roshandathathri: Where is `rows` defined?

libgalois/include/katana/Properties.h
```diff
@@ -182,6 +171,106 @@ ConstructPropertyViews(const std::vector<arrow::Array*>& arrays) {
       arrays, std::make_index_sequence<std::tuple_size_v<PropTuple>>());
 }
 
+template <typename T>
```
roshandathathri, line 195: Please move all this new code to below anonymous namespace so that it is easier to compare what changed in `AllocateTable`

libgalois/include/katana/Properties.h
nicelhc13: I don't get this comment. It is defined at line 198

roshandathathri: Never mind. Thanks.

libgalois/include/katana/Properties.h
```diff
@@ -201,6 +190,107 @@ GetMutableValuesWorkAround(
...
+  auto rest =
+      AllocateTable<PropTuple>(num_rows, names, std::index_sequence<tail...>());
+  if (!rest) {
+    return rest.error();
+  }
+
+  auto rest_table = std::move(rest.value());
+
+  // TODO: tail recursion
+  // TODO: metadata
```
roshandathathri: @ddn0 TODO(ddn)?

libgalois/include/katana/Properties.h
```diff
@@ -385,23 +475,33 @@ struct LargeStringReadOnlyProperty {
...
+  static katana::Result<std::shared_ptr<arrow::Table>> Allocate(
+      size_t num_rows, const std::string& name) {
+    auto res = arrow::FixedSizeBinaryType::Make(sizeof(T));
+    if (!res.ok()) {
+      return KATANA_ERROR(katana::ErrorCode::ArrowError, "{}", res.status());
+    }
+
+    auto type = res.ValueOrDie();
+    arrow::FixedSizeBinaryBuilder builder(type);
+    std::vector<uint8_t> data(sizeof(T) * num_rows);
```
roshandathathri: Use `PODResizeableArray` instead. It will save time in constructing the elements.

libgalois/src/analytics/pagerank/pagerank-pull.cpp
```diff
@@ -228,26 +224,54 @@ ComputePRTopological(Graph* graph, katana::analytics::PagerankPlan plan) {
...
+
 katana::Result<void>
 PagerankPullTopological(
     katana::PropertyGraph* pg, const std::string& output_property_name,
     katana::analytics::PagerankPlan plan) {
   katana::EnsurePreallocated(2, 3 * pg->num_nodes() * sizeof(NodeData));
-
-  katana::analytics::TemporaryPropertyGuard temporary_property{pg};
-
+  auto compute_graph = std::move(pg->Copy({}, {}).value());
```
roshandathathri: Is this copying the edges as well? Can we avoid that? @arthurp

libgalois/src/analytics/pagerank/pagerank.cpp
```diff
@@ -58,23 +58,26 @@ katana::analytics::PagerankStatistics::Print(std::ostream& os) {
...
           pg, {property_name}, {});
   if (!graph_result) {
     return graph_result.error();
   }
+
   auto graph = graph_result.value();
-  katana::GReduceMax<PRTy> max_rank;
-  katana::GReduceMin<PRTy> min_rank;
-  katana::GAccumulator<PRTy> distance_sum;
+  num_nodes = graph.size();
```
roshandathathri: Drop. Use `graph.size()` directly.

libgalois/src/analytics/pagerank/pagerank.cpp
```diff
@@ -58,23 +58,26 @@ katana::analytics::PagerankStatistics::Print(std::ostream& os) {
 katana::Result<katana::analytics::PagerankStatistics>
 katana::analytics::PagerankStatistics::Compute(
     katana::PropertyGraph* pg, const std::string& property_name) {
+  katana::GReduceMax<PRTy> max_rank;
+  katana::GReduceMin<PRTy> min_rank;
+  katana::GAccumulator<PRTy> distance_sum;
```
roshandathathri: Not a fan of C-style early/forward declaration.

roshandathathri (approved):
LGTM

libgalois/src/analytics/pagerank/pagerank.cpp
nicelhc13: Yes, it was for another version. I will update this

libgalois/src/analytics/pagerank/pagerank-pull.cpp
arthurp: This is copying edges (and nodes, just not properties). I don't think you need this copy at all. You should just be able to create a temporary property of the correct type and use it. `TemporaryPropertyGuard` will delete the temporary property when you are done. You don't need to copy the graph unless you are sorting/relabeling the edges or nodes.

libgalois/src/analytics/pagerank/pagerank-pull.cpp
```diff
@@ -257,6 +281,10 @@ PagerankPullTopological(
   ComputePRTopological(&graph, plan);
   exec_time.stop();
 
+  if (auto r = ExtractPagerank(pg, graph, output_property_name); !r) {
+    KATANA_LOG_FATAL("Failed to extract PageRank {}", r.error());
```
arthurp: This should propagate the error not just abort.

libgalois/include/katana/Properties.h
danielmawhirter: Most efficient option is `auto st = builder.AppendEmptyValues(num_rows);` (still initializes to 0)

roshandathathri: Thanks, Daniel. Your option is better. It gets initialization at the same cost of copy.

Hochan, please use this in `AllocateTable` as well.

libgalois/include/katana/Properties.h
```diff
@@ -385,23 +475,33 @@ struct LargeStringReadOnlyProperty {
...
+template <typename T>
+struct StructProperty {
+  using ArrowType = arrow::FixedSizeBinaryType;
+  using ViewType = katana::PODPropertyView<T>;
+
+  static katana::Result<std::shared_ptr<arrow::Table>> Allocate(
+      size_t num_rows, const std::string& name) {
+    auto res = arrow::FixedSizeBinaryType::Make(sizeof(T));
+    if (!res.ok()) {
+      return KATANA_ERROR(katana::ErrorCode::ArrowError, "{}", res.status());
```
arthurp: Including extra contextual information with the errors is useful. In this case it might be `"failed to make fixed size type: {}"` or something.

Please take a look at your other error returns.

libgalois/src/analytics/pagerank/pagerank-impl.h
```diff
@@ -31,7 +31,13 @@
 #include "katana/analytics/pagerank/pagerank.h"
 
 typedef float PRTy;
+struct ValueOut {
```
arthurp: This is a very vague name. Can you rename this something that actually specifies what it means?

libgalois/src/analytics/pagerank/pagerank-pull.cpp
```diff
@@ -40,12 +41,11 @@ InitNodeDataTopological(Graph* graph) {
...
-        auto& sdata_value = graph->GetData<NodeValue>(n);
-        auto& sdata_nout = graph->GetData<NodeNout>(n);
-        sdata_value = init_value;
-        sdata_nout = 0;
+        auto& sdata = graph->GetData<NodeValueOut>(n);
+        sdata.value = init_value;
+        sdata.out = 0;
       },
-      katana::no_stats(), katana::loopname("initNodeData"));
+      katana::loopname("initNodeData"));
```
arthurp, line 56: Why are you enabling stats? I'm fine with it. I just don't want to do it by accident.

libgalois/src/analytics/pagerank/pagerank-pull.cpp
```diff
@@ -228,26 +224,54 @@ ComputePRTopological(Graph* graph, katana::analytics::PagerankPlan plan) {
 
 }  // namespace
 
+katana::Result<void>
+ExtractPagerank(
```
arthurp: This should be in the anonymous namespace since it's only used in this function. Also it should have a name that's specific to the algorithm that uses it. Because this isn't used for all of pagerank.

arthurp (changes requested):
Looking better. We're close.

libgalois/src/analytics/pagerank/pagerank-pull.cpp
roshandathathri: It's intentional. I requested Hochan to enable stats for all our `do_all`s so that we can have a better idea of where we're seeing performance differences.

libgalois/src/analytics/pagerank/pagerank-pull.cpp
nicelhc13: Thank you @arthurp . I now get more how properties and topology are stored and handled. I will fix this. It would be appreciated if I could get your evaluation later.

arthurp: Of course. Though I will be out of the office Thurs and Fri. So I may not be able to do much. But I'll try to check if I can, and if not I'm sure someone else can take a look at this. You may find it useful to look at some of the other algorithms that use intermediate/internal properties while they execute. Several do.

libgalois/include/katana/Properties.h
nicelhc13: I guess you are talking about `Allocate()`, Roshan?

roshandathathri: Yes

libgalois/src/analytics/pagerank/pagerank-pull.cpp
nicelhc13: Hmm. `ExtractScoreFromTopoGraph()` ?

roshandathathri: Aren't you extracting `value`?

nicelhc13: Value is pagerank score, ins't it. Ok then, `ExtractValueFromTopoGraph()` :-)

libgalois/src/analytics/pagerank/pagerank-impl.h
nicelhc13: Hmmmm it is very hard to think of short and clear name :-) I decided to use `PROperandsTy` and `using PROperands = katana::PODProperty<PROperandsTy>` .

arthurp: Why not just `PagerankValueAndOutputCount` (I'm totally guessing on what `out` is, so making that clear would be useful).

libgalois/include/katana/Properties.h
nicelhc13: @danielmawhirter , there is no `AppendEmptyValues()` on the repository. Compilation failed due to undefined symbol. Am I missing something?

libgalois/src/analytics/pagerank/pagerank-impl.h
nicelhc13: Out is an out degree. So `PagerankValueAndOutDegree` looks right to use!

libgalois/include/katana/Properties.h
nicelhc13: Yes, it is added to version 3.0. I am using 2.0 and am not sure about compatibility. For now, I am gonna use `PODResizeableArray` and left todo.

arthurp (approved):

uditagar (approved):
lovain changes looks fine. Onwards and upwards, Mr. Lee!

uditagar (approved):

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -307,7 +307,7 @@ struct ClusteringImplementationBase {
     return mod;
   }
 
-  template <typename EdgeWeightType>
+  template <typename EdgeWeightType, typename NodePropTy>
```
ddn0: nit: `NodePropType` or `EdgeWeightTy` or I think the predominate style in the rest of the codebase is just `EdgeWeight` `NodeProp`

arthurp:
This commit removes `PropertyArrowTuple` which is [used in `DistGraph` in the enterprise repo](https://github.com/KatanaGraph/katana-enterprise/blob/5dc55160d9e865858365e69253c89d0d88d1a76c/libgluon/include/katana/DistGraph.h#L263). Is there an enterprise PR in the works to fix that? @nicelhc13 @roshandathathri

PR #276, opened 2021-06-09
utility: add approximation function for an arrow table's memory usage

patrickkenney9801:
Add a function to wrap arrow array memory usage approximation for table memory usage approximation


tylershunt (approved):
LGTM!

PR #280, opened 2021-06-11
Cc optimizations

amberhassaan:
Trying out a formula for pre-allocing enough memory.


libgalois/src/analytics/connected_components/connected_components.cpp
```diff
@@ -688,16 +688,33 @@ struct ConnectedComponentsAfforestAlgo {
...
     });
+    //std::cout << sum << "\n";
   }
 
   void Deallocate(Graph* graph) {
     katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
       auto& sdata = graph->GetData<NodeComponent>(node);
       auto component_ptr = sdata->component();
-      delete sdata;
+      //delete sdata;
```
roshandathathri: Does this leak memory?

libgalois/src/analytics/connected_components/connected_components.cpp
```diff
@@ -688,16 +688,33 @@ struct ConnectedComponentsAfforestAlgo {
...
+    //katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
+    //  graph->GetData<NodeComponent>(node) = new NodeAfforest();
+    //});
+    // XXX
+    //int sum = 0;
     katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
-      graph->GetData<NodeComponent>(node) = new NodeAfforest();
+      auto& snode = graph->GetData<NodeComponent>(node);
+      new (&snode) NodeAfforest();
+      //snode = reinterpret_cast<ComponentType>(&snode);
```
roshandathathri, line 732: Drop commented code.

libgalois/src/analytics/connected_components/connected_components.cpp
```diff
@@ -688,16 +688,33 @@ struct ConnectedComponentsAfforestAlgo {
...
   void Initialize(Graph* graph) {
+    //katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
+    //  graph->GetData<NodeComponent>(node) = new NodeAfforest();
+    //});
+    // XXX
+    //int sum = 0;
     katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
-      graph->GetData<NodeComponent>(node) = new NodeAfforest();
+      auto& snode = graph->GetData<NodeComponent>(node);
+      new (&snode) NodeAfforest();
```
roshandathathri, line 730: Was the type of `snode` changed? I don't understand the change is equivalent.

libgalois/src/analytics/connected_components/connected_components.cpp
amberhassaan: No. NodeCompoenent was initialized with inplace new. We're not destroying or destructing it, but that's OK, because it doesn't have any members that allocate memory.

libgalois/src/analytics/connected_components/connected_components.cpp
amberhassaan: I don't think I quite understand your question

roshandathathri: I didn't know about in-place new. So this is not allocating new memory but just calling constructor on it (initializing it)?

cmake/Modules/CheckArchFlags.cmake
```diff
@@ -32,29 +32,32 @@ else()
...
-      list(APPEND ARCH_C_FLAGS "${FLAG}")
-      list(APPEND ARCH_LINK_FLAGS "${FLAG}")
-      break()
-    endif()
-  endforeach()
-endif()
+set(ARCH_CXX_FLAGS "-mtune=native")
+set(ARCH_C_FLAGS "-mtune=native")
+set(ARCH_LINK_FLAGS "-mtune=native")
+# foreach(FLAG ${ARCH_CXX_FLAGS_CANDIDATES})
```
roshandathathri, line 38: Do you want to merge this code? If so, please drop the commented code

libgalois/src/analytics/connected_components/connected_components.cpp
roshandathathri: Is this still needed? 

Please drop commented code as well.

roshandathathri (approved):
LGTM.

libgalois/src/analytics/connected_components/connected_components.cpp
```diff
@@ -643,6 +643,41 @@ approxLargestComponent(Graph* graph, uint32_t component_sample_frequency) {
...
+      std::equal_to<ComponentType>,
+      katana::gstl::Pow2Alloc<std::pair<const ComponentType, int>>>;
+  using pair_type = std::pair<ComponentType, int>;
+
+  map_type comp_freq(component_sample_frequency);
+  std::random_device rd;
+  std::mt19937 rng(rd());
+  std::uniform_int_distribution<uint32_t> dist(0, graph->size() - 1);
+  for (uint32_t i = 0; i < component_sample_frequency; i++) {
+    // ComponentType ndata = graph->template GetData<NodeIndex>(dist(rng));
```
nicelhc13: Why don't we just remove comments?

libgalois/src/analytics/connected_components/connected_components.cpp
```diff
@@ -643,6 +643,41 @@ approxLargestComponent(Graph* graph, uint32_t component_sample_frequency) {
   return most_frequent->first;
 }
 
+template <typename ComponentType, typename Graph, typename NodeIndex, typename ParentArray>
+ComponentType
+approxLargestComponent(Graph* graph, ParentArray& parent_array_, uint32_t component_sample_frequency) {
```
nicelhc13: `const ParentArray& parent_array_`?

nicelhc13: also const uint32?

nicelhc13 (approved):
Look good to me. Please remove comments. Thank you, Amber.

libgalois/src/analytics/connected_components/connected_components.cpp
amberhassaan: Yes, what you said about in-place new is correct. It calls a constructor on the pointer you provide.

I wanna keep the old code until we fix the Numa problems with our node data. Then, if we're happy, we can easily go back.

PR #283, opened 2021-06-13
Improve arch and tune build options and defaults. 

arthurp:
Adds `KATANA_USE_TUNE`. Makes the default oriented towards new Intel
 processors. Fully optimizing for one machine is now possible with:
 
` -DKATANA_USE_ARCH=native -DKATANA_USE_TUNE=native`

This removes the fixed mtune from 11d6d75efeb from #280.

### DIFF
--- a/cmake/Modules/CheckArchFlags.cmake
+++ b/cmake/Modules/CheckArchFlags.cmake
@@ -1,7 +1,17 @@
-# Find architecture-specific flags based on the value of KATANA_USE_ARCH.
-# KATANA_USE_ARCH is a cmake list of possible architecture names. Return the
-# flags corresponding to the first architecture available for the current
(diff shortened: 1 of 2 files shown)


arthurp:
This PR also makes the checking process more noisy which should make it easy to tell which options are not working during cmake configuration.

amberhassaan (approved):
Looks good. One question: Do these mtune/march flags affect linking as well? I see ARCH_LINK_FLAG being set.

arthurp:
> Do these mtune/march flags affect linking as well? I see ARCH_LINK_FLAG being set.

They do if link time optimization is turned on. We don't have LTO at the moment I don't think, but better to include it so we don't have weird issues if we do enable LTO later.

PR #292, opened 2021-06-15
Fix out of bounds error in finding edges

roshandathathri:


aneeshdurg (approved):
Maybe `mid_not_limit` should be renamed to `mid_past_limit`?

roshandathathri:
> Maybe `mid_not_limit` should be renamed to `mid_past_limit`?

That's the opposite right? Maybe `mid_within_limit`?

aneeshdurg:
Yes you're right, `mid_within_limit` makes more sense.

PR #300, opened 2021-06-16
Fix binary search for first/last edges

aneeshdurg:


libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
```diff
@@ -337,34 +337,47 @@ class LC_CSR_CSC_Labeled_Graph
...
-          } else {
+          if (adjacent_value == key) {
             if (finding_first) {
+              // We can't update the last_bounds here because mid - 1 might have
+              // value key as well.
               r = mid - 1;
             } else {
               l = mid + 1;
             }
+            continue;
```
roshandathathri: nit: prefer `else if` to `continue`.

Also, please clean-up `return`.  `else` is not needed after `return`

roshandathathri (approved):
LGTM

libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
```diff
@@ -337,34 +337,47 @@ class LC_CSR_CSC_Labeled_Graph
...
+          // Update bounds for last only if we're searching for the first
+          // element. We know mid has value == key, so the last element has to
+          // be at or after mid. The begin bound is inclusive so we inlude mid
+          // here.
           last_begin_bound = mid;
-          last_end_bound = r + 1;
         }
 
-        if (!(opt == FindEdgeWithLabelOptions::FindAny) && mid != limit) {
+        // If we're just searching for any valid edge, then we don't need to
```
ddn0: can you file an issue to make a test for this?

roshandathathri: to make unit tests*

roshandathathri: Just to be clear. This entire class will be superseded by PropertGraph soon. We should unit tests for this function then.
@amberhassaan

libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
```diff
@@ -337,34 +337,47 @@ class LC_CSR_CSC_Labeled_Graph
       GraphNode value =
           in_edges ? this->getInEdgeDst(mid) : this->getEdgeDst(mid);
       if (value == key) {
-        // Update bounds for last only if we're searching for the first element.
         if (finding_first) {
+          // Update bounds for last only if we're searching for the first
```
ddn0: the STL uses lower_bound and upper_bound to achieve the same idea. It might be more maintainable to use those (or analogues thereof) instead of merging finding first/any/last into one function.

amberhassaan: I agree. The complexity and corner cases are getting out of hand here. Better to rewrite it using STL's upper_bound and lower_bound.

roshandathathri: That's a good idea. I don't see any benefits of maintaining our own code for this. Let's use lower_bound and upper_bound directly.

libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
```diff
@@ -297,119 +297,26 @@ class LC_CSR_CSC_Labeled_Graph
...
-    auto res = FindFirstOrLastEdgeWithLabel<
-        in_edges, FindEdgeWithLabelOptions::FindAny>(node, key, data);
-    if (!res) {
-      return std::nullopt;
+  class EdgeIteratorComparator {
+  public:
+    using Graph = LC_CSR_CSC_Graph<
+        NodeTy, EdgeTy, EdgeDataByValue, HasNoLockable, UseNumaAlloc,
+        HasOutOfLineLockable, FileEdgeTy>;
+    const Graph* graph_;
```
roshandathathri: Why not `const Graph&`?

libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
```diff
@@ -422,52 +329,34 @@ class LC_CSR_CSC_Labeled_Graph
...
     // trivial check; can't be connected if degree is 0
-    if (in_edges) {
-      if (in_degrees_[node] == 0 || degrees_[key] == 0) {
-        return std::nullopt;
-      }
-    } else {
-      if (degrees_[node] == 0 || in_degrees_[key] == 0) {
-        return std::nullopt;
-      }
+    if (degrees_[node] == 0 || in_degrees_[key] == 0) {
```
roshandathathri: Why this change? This is wrong if `in_edges` is true.

roshandathathri (approved):
LGTM

libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
aneeshdurg: I just went with `*` since `this` was already a pointer type

libgalois/include/katana/LC_CSR_CSC_Labeled_Graph.h
aneeshdurg: whoops! You're right, I'll add it back

ddn0 (approved):
the STL provides... all hail our SGI overloads

amberhassaan (approved):
Looks OK.

PR #305, opened 2021-06-16
Set correct edge properties for analytic applications

nicelhc13:


scripts/bench_python_cpp_algos.py
```diff
@@ -292,10 +292,10 @@ def run_all_gap(args):
         },
     ]
 
-    def load_graph(graph_path):
+    def load_graph(graph_path, edge_properties=None):
```
arthurp, line 295: Do you want to limit `node_properties` as well? This should be done throughout if you want to.

scripts/bench_python_cpp_algos.py
```diff
@@ -292,10 +292,10 @@ def run_all_gap(args):
         },
     ]
 
-    def load_graph(graph_path):
+    def load_graph(graph_path, edge_properties=None, node_properties=None):
         print(f"Running {args.application} on graph: {graph_path}")
         with time_block("read propertyGraph"):
-            graph = PropertyGraph(graph_path)
+            graph = PropertyGraph(graph_path, edge_properties=edge_properties, node_properties=node_properties)
```
ddn0: as I understand the benchmarks, none of them actually use the node properties from the inputs, so we can just set `node_properties=[]` always.

ddn0 (approved):

scripts/bench_python_cpp_algos.py
arthurp: If this is true. I am with Donald.

arthurp (approved):

scripts/bench_python_cpp_algos.py
nicelhc13: Yes you are right

PR #308, opened 2021-06-17
Wrap "powerBurn" or "threadSpin" setting at the library level 

arthurp:
Also exposes it as an argument to a few algorithm CLIs and the python benchmarking script.

### DIFF
--- a/python/katana/cpp/libgalois/Galois.pxd
+++ b/python/katana/cpp/libgalois/Galois.pxd
@@ -3,8 +3,17 @@ from libcpp.string cimport string
 from ..libstd cimport CPPAuto
 
 
+cdef extern from "katana/ThreadPool.h" namespace "katana" nogil:
+    cppclass ThreadPool:
+        void burnPower(unsigned num)
+        void beKind()
+
+    ThreadPool& GetThreadPool()
+
+
 cdef extern from "katana/Galois.h" namespace "katana" nogil:
     unsigned int setActiveThreads(unsigned int)
+    unsigned int getActiveThreads()
 
     cppclass UserContext[T]:
         void push(...)
@@ -35,6 +44,7 @@ cdef extern from "katana/Galois.h" namespace "katana" nogil:
     cppclass disable_conflict_detection:
         disable_conflict_detection()
 
+
 cdef extern from "katana/MethodFlags.h" namespace "katana" nogil:
     cppclass MethodFlag:
         bint operator==(MethodFlag)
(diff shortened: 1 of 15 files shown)


python/katana/galois.pyx
```diff
@@ -1,10 +1,35 @@
...
+
 def set_active_threads(int n):
+    """
+    Set the number of threads Katana should use to do computation.
+
+    :return: The number of threads actually used which may be less than `n`
+    """
     return c_setActiveThreads(n)
 
+def set_busy_wait(n = None):
```
nicelhc13, line 21: I guess one more blank link should be added?

python/katana/galois.pyx
```diff
@@ -1,10 +1,35 @@
...
+    return c_getActiveTheads()
+
+
 def set_active_threads(int n):
+    """
+    Set the number of threads Katana should use to do computation.
+
+    :return: The number of threads actually used which may be less than `n`
+    """
     return c_setActiveThreads(n)
```
nicelhc13, line 23: I guess one more blank line is required?

nicelhc13 (approved):
LGTM! Thank you!

python/katana/cpp/libgalois/Galois.pxd
```diff
@@ -3,8 +3,17 @@ from libcpp.string cimport string
 from ..libstd cimport CPPAuto
 
 
+cdef extern from "katana/ThreadPool.h" namespace "katana" nogil:
+    cppclass ThreadPool:
+        void burnPower(unsigned num)
```
ddn0, line 8: do you mind changing the name of `burnPower` to `BurnPower` (as well as `beKind`)?

ddn0 (approved):

python/katana/cpp/libgalois/Galois.pxd
arthurp: If we are renaming them, I'm gonna do more than just styling. Those names are very non-descriptive. I'll do a separate PR. This is a rush PR.

ddn0: That's fine.

(for later) I don't like that this is exposed at all, but maybe if it were wrapped in a context manager it would at least allow users to express there are a bunch of loops here that I want to run together as quickly as possible.

arthurp: I agree. I actually think this also applies to `set_active_threads` and loop annotations like `steal`. I think it's worth exploring ways to hide/automate these while also providing a fallback to control them if experts need to.

PR #310, opened 2021-06-18
Consistently generate parent information for all BFS algorithms

nicelhc13:
I have done correctness check by the number of visited nodes.
This change reduces instructions and improves performance.

### DIFF
--- a/libgalois/include/katana/analytics/BfsSsspImplementationBase.h
+++ b/libgalois/include/katana/analytics/BfsSsspImplementationBase.h
@@ -336,6 +336,78 @@ struct BfsSsspImplementationBase {
 
     return true;
   }
+
+  struct BfsSrcEdgeTile {
+    GNode src;
+    GNode parent;
+    EI beg;
+    EI end;
+
+    friend bool operator<(
+        const BfsSrcEdgeTile& left, const BfsSrcEdgeTile& right) {
+      return left.parent < right.src;
+    }
+  };
+
+  struct BfsSrcEdgeTileMaker {
+    GNode src;
+    GNode parent;
+
+    BfsSrcEdgeTile operator()(const EI& beg, const EI& end) const {
+      return BfsSrcEdgeTile{src, parent, beg, end};
+    }
+  };
+
+  struct BfsSrcEdgeTilePushWrap {
+    Graph* graph;
+    BfsSsspImplementationBase& impl;
+
+    template <typename C>
+    void operator()(
+        C& cont, const GNode& n, const GNode& parent, const char* const) const {
+      impl.PushEdgeTilesParallel(
+          cont, graph, n, BfsSrcEdgeTileMaker{n, parent});
+    }
+
+    template <typename C>
+    void operator()(C& cont, const GNode& n, const GNode& parent) const {
+      impl.PushEdgeTiles(cont, graph, n, BfsSrcEdgeTileMaker{n, parent});
+    }
+  };
+
+  struct BfsUpdateRequest {
+    GNode src;
+    GNode parent;
+    BfsUpdateRequest(const GNode& s, GNode p) : src(s), parent(p) {}
+    BfsUpdateRequest() : src(), parent(0) {}
+
+    friend bool operator<(
+        const BfsUpdateRequest& left, const BfsUpdateRequest& right) {
+      return left.parent < right.src;
+    }
+  };
+
+  struct BfsReqPushWrap {
+    template <typename C>
+    void operator()(
+        C& cont, const GNode& n, const GNode& parent, const char* const) const {
+      (*this)(cont, n, parent);
+    }
+
+    template <typename C>
+    void operator()(C& cont, const GNode& n, const GNode& parent) const {
+      cont.push(BfsUpdateRequest(n, parent));
+    }
+  };
+
+  struct BfsOutEdgeRangeFn {
+    Graph* graph;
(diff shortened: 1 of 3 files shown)


libgalois/include/katana/analytics/BfsSsspImplementationBase.h
```diff
@@ -121,8 +121,9 @@ struct BfsSsspImplementationBase {
   template <typename WL, typename TileMaker>
   void PushEdgeTiles(
       WL& wl, katana::PropertyGraph* graph, GNode src, const TileMaker& f) {
-    auto beg = graph->edges(src).first;
-    const auto end = graph->edges(src).second;
+    auto edge_range = graph->topology().edge_range(src);
```
arthurp, line 124: Let's not move back to the topology API. I was hoping we could remove this API going forward. Same below.

libgalois/include/katana/analytics/BfsSsspImplementationBase.h
```diff
@@ -336,6 +311,65 @@ struct BfsSsspImplementationBase {
 
     return true;
   }
+
+  // TODO(lhc): Now BFS and SSSP use different labels: parent and distance,
+  //            respectively. Therefore, we should differentiate worklist
+  //            wrappers. For consistency, we could put `sssp` prefix later.
```
arthurp, line 317: We may want two separate subclasses of `BfsSsspImplementationBase`. One with BFS variants of the utilities, one with SSSP. No need for prefixes.

libgalois/src/analytics/bfs/bfs.cpp
```diff
@@ -147,68 +103,60 @@ AsynchronousAlgo(
...
-  using Loop = typename std::conditional<
-      CONCURRENT, katana::ForEach, katana::WhileQ<katana::SerFIFO<T>>>::type;
-
   KATANA_GCC7_IGNORE_UNUSED_BUT_SET
   constexpr bool use_CAS = CONCURRENT && !std::is_same<WL, BSWL>::value;
   KATANA_END_GCC7_IGNORE_UNUSED_BUT_SET
 
-  Loop loop;
-
+  // TODO(lhc): How about `failed_work`?
```
arthurp, line 117: Just change it.

libgalois/src/analytics/bfs/bfs.cpp
```diff
@@ -506,8 +448,11 @@ katana::analytics::BfsAssertValid(
     return pg_result.error();
   }
 
+  /*
   BfsImplementation::Graph graph = pg_result.value();
 
+  TODO(lhc): Multiple nodes could have 0 node as a parent.
+             It looks good to remove this, but let me keep this.
   GAccumulator<uint64_t> n_zeros;
   do_all(iterate(graph), [&](uint32_t node) {
```
arthurp, line 457: Just delete commented code.

libgalois/src/analytics/bfs/bfs.cpp
```diff
@@ -543,6 +492,9 @@ katana::analytics::BfsStatistics::Compute(
   BfsImplementation::Graph graph = pg_result.value();
 
   uint32_t source_node = std::numeric_limits<uint32_t>::max();
+  // TODO(lhc) max dist should be the last node
+  //           for now, keep it since it requires python modification.
+  //           The number of visited nodes is enough for verification.
```
arthurp, line 497: The stats need to be updated for the new information format. Are we still computing distances here? Or should we compute different stats?

Updating the Python wrappers is not hard.

arthurp (changes requested):
Some smaller things. But mostly we need to update statistics. They make no sense for parent information.

libgalois/include/katana/analytics/BfsSsspImplementationBase.h
amberhassaan: Agree.

libgalois/src/analytics/bfs/bfs.cpp
```diff
@@ -147,68 +103,60 @@ AsynchronousAlgo(
...
         for (auto ii : edgeRange(item)) {
           auto dest = graph.GetEdgeDest(ii);
-          Dist& ddata = (*node_data)[*dest];
+          GNode& parent = (*node_data)[*dest];
 
           while (true) {
-            Dist old_dist = ddata;
-            if (old_dist <= new_dist) {
+            GNode current_parent = parent;
+            if (current_parent != BfsImplementation::kDistanceInfinity) {
```
amberhassaan, line 143: @nicelhc13 : This is asynchronous algorithm and without checking on lower distance, it's wrong. If we're going level by level, it's fine, otherwise first CAS wins, which could be from a node with higher distance.

amberhassaan (commented):
My main comment is what I said last night: we need distance and parent for asynchronous algorithm.

libgalois/src/analytics/bfs/bfs.cpp
nicelhc13: I see. That makes sense to me. I thought that operator is visiting nodes regardless of distance, in this case. If it is the case, do you have any suggestion for asynchronous running? Should I add the second label called parent and also fill this?

amberhassaan: Let's discuss this via slack first.

nicelhc13:
Let me close this for now since we decided another direction and code is already covered by others/future commit.

PR #316, opened 2021-06-19
Use std::launch:defered in a couple more places to avoid crashes.

arthurp:
This does not fix error handling in the AWS, azure, or GS interfaces. Those WILL still cause a crash if they try to return an error when built on some compilers and c++ libraries.

### DIFF
--- a/libtsuba/src/LocalStorage.cpp
+++ b/libtsuba/src/LocalStorage.cpp
@@ -134,17 +134,19 @@ tsuba::LocalStorage::ListAsync(
   if ((dirp = opendir(dirname.c_str())) == nullptr) {
     if (errno == ENOENT) {
       // other storage backends are flat and so return an empty list here
-      return std::async(
-          []() -> katana::Result<void> { return katana::ResultSuccess(); });
+      return std::async(std::launch::deferred, []() -> katana::Result<void> {
+        return katana::ResultSuccess();
+      });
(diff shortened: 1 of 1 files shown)


witchel:
I'm curious as to why these need to be defered.

arthurp:
> I'm curious as to why these need to be defered.

The important thing isn't `deferred`, it's that there is *not* `std::launch::async`. This means that the body of the lambda is not allowed to be run in another thread which avoids breaking the error object (which cannot be moved between threads).

(https://en.cppreference.com/w/cpp/thread/launch)

witchel:
Thanks for letting me know. I was not aware of the issue.  I'm a bit afraid that I have some subtle threading bugs in the code I've been writing.  I'm not sure if it is possible to have a "thread scheduling stress mode" for testing.

arthurp:
There are known issues in several places. To fix it we will need a way to cleanly pass errors between threads. Because of how poorly C++ supports abstraction of representation, I don't know how we will do that. I think @ddn0 has some ideas. For now, we will just may have crashes if an error is returned during async operations.

One thing I should probably do at some point is figure out what about my environment triggers this issue. It doesn't happen in CI.

tylershunt (approved):
LGTM. And thanks for the reminder that the other error handling code is still broken.

PR #319, opened 2021-06-21
Update input version for mining tests

bowu:


arthurp (changes requested):
I don't think the bliss deps belong in the open repo. It's not used here.

bowu:
> I don't think the bliss deps belong in the open repo. It's not used here.

I only added it to the enterprise repo before. You asked me to follow `contribute.md` to add dependencies. Could you let me know what I should do in this case? Thanks.

bowu:
> I don't think the bliss deps belong in the open repo. It's not used here.

I guess I read the wrong document. Let me fix this later.

arthurp:
I suspect CONTRIBUTING.md isn't clear. The simple rule is, add dependencies in the repo that requires them, or just in open if they are required by both.

arthurp (approved):

PR #326, opened 2021-06-22
[KAT-716] [RFC-ENG-027] Result Propagation with Raising Macros

danielmawhirter:
[RFC-ENG-027](https://docs.google.com/document/d/1WIqCUZKik_klt1lF0Sdlixy9eUCIZi_5rZnEhtN4x60/edit?usp=sharing)
Introduces the `KATANA_CHECKED` macro, which can be used with `katana::Result` to automatically propagate the error to the caller if one occurs. This is accomplished using a Statement Expression, as suggested by @aneeshdurg in the RFC.

Http.cpp is modified using the new macro as an example.

### DIFF
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -260,6 +260,29 @@ ResultErrno() {
   return std::error_code(errno, std::system_category());
 }
 
+template <class>
+struct is_result : std::false_type {};
+
+template <class T>
+struct is_result<Result<T>> : std::true_type {};
+
+#define KATANA_CHECK_NAME(x, y) x##y
+
+#define KATANA_CHECK_IMPL(result_name, expression)                             \
+  ({                                                                           \
+    auto result_name = (expression);                                           \
+    static_assert(                                                             \
+        ::katana::is_result<decltype(result_name)>::value,                     \
+        "KATANA_CHECK requires a katana::Result");                             \
+    if (!result_name) {                                                        \
+      return result_name.error().WithContext("here");                          \
+    }                                                                          \
+    std::move(result_name);                                                    \
+  })
+
+#define KATANA_CHECK(expression)                                               \
+  KATANA_CHECK_IMPL(KATANA_CHECK_NAME(_error_or_value, __COUNTER__), expression)
+
 }  // namespace katana
 
 #endif
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -260,39 +260,39 @@ ResultErrno() {
   return std::error_code(errno, std::system_category());
 }
 
-template <class>
-struct is_result : std::false_type {};
-
-template <class T>
-struct is_result<Result<T>> : std::true_type {};
+namespace internal {
 
 template <class T>
 std::enable_if_t<!std::is_same<T, void>::value, T&&>
-extract_value(Result<T>&& result) {
+extract_result_value(Result<T>&& result) {
   return std::move(result.value());
 }
 
 inline int
-extract_value(Result<void>&&) {
+extract_result_value(Result<void>&&) {
   return 0;
 }
 
+}  // namespace internal
+
 #define KATANA_CHECK_NAME(x, y) x##y
 
-#define KATANA_CHECK_IMPL(result_name, expression)                             \
+#define KATANA_CHECK_IMPL(result_name, expression, ...)                        \
   ({                                                                           \
     auto result_name = (expression);                                           \
-    static_assert(                                                             \
-        ::katana::is_result<decltype(result_name)>::value,                     \
-        "KATANA_CHECK requires a katana::Result");                             \
     if (!result_name) {                                                        \
-      return result_name.error().WithContext("here");                          \
+      return result_name.error().WithContext(__VA_ARGS__);                     \
     }                                                                          \
-    std::move(::katana::extract_value(std::move(result_name)));                \
+    std::move(                                                                 \
+        ::katana::internal::extract_result_value(std::move(result_name)));     \
   })
 
-#define KATANA_CHECK(expression)                                               \
-  KATANA_CHECK_IMPL(KATANA_CHECK_NAME(_error_or_value, __COUNTER__), expression)
+#define KATANA_CHECK_CONTEXT(expression, ...)                                  \
+  KATANA_CHECK_IMPL(                                                           \
+      KATANA_CHECK_NAME(_error_or_value, __COUNTER__), expression,             \
+      __VA_ARGS__)
+
+#define KATANA_CHECK(expression) KATANA_CHECK_CONTEXT(expression, "backtrace")
 
 }  // namespace katana
 
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -275,9 +275,9 @@ extract_result_value(Result<void>&&) {
 
 }  // namespace internal
 
-#define KATANA_CHECK_NAME(x, y) x##y
+#define KATANA_CHECKED_NAME(x, y) x##y
 
-#define KATANA_CHECK_IMPL(result_name, expression, ...)                        \
+#define KATANA_CHECKED_IMPL(result_name, expression, ...)                      \
   ({                                                                           \
     auto result_name = (expression);                                           \
     if (!result_name) {                                                        \
@@ -287,12 +287,13 @@ extract_result_value(Result<void>&&) {
         ::katana::internal::extract_result_value(std::move(result_name)));     \
   })
 
-#define KATANA_CHECK_CONTEXT(expression, ...)                                  \
-  KATANA_CHECK_IMPL(                                                           \
-      KATANA_CHECK_NAME(_error_or_value, __COUNTER__), expression,             \
+#define KATANA_CHECKED_CONTEXT(expression, ...)                                \
+  KATANA_CHECKED_IMPL(                                                         \
+      KATANA_CHECKED_NAME(_error_or_value, __COUNTER__), expression,           \
       __VA_ARGS__)
 
-#define KATANA_CHECK(expression) KATANA_CHECK_CONTEXT(expression, "backtrace")
+#define KATANA_CHECKED(expression)                                             \
+  KATANA_CHECKED_CONTEXT(expression, "backtrace")
 
 }  // namespace katana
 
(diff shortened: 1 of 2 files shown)


libsupport/include/katana/Result.h
```diff
@@ -260,6 +260,40 @@ ResultErrno() {
...
+#define KATANA_CHECK_NAME(x, y) x##y
+
+#define KATANA_CHECK_IMPL(result_name, expression)                             \
+  ({                                                                           \
+    auto result_name = (expression);                                           \
+    static_assert(                                                             \
+        ::katana::is_result<decltype(result_name)>::value,                     \
+        "KATANA_CHECK requires a katana::Result");                             \
+    if (!result_name) {                                                        \
+      return result_name.error().WithContext("here");                          \
```
ddn0: should we take the context text ("here") as an argument?

ddn0 (commented):
seems reasonable to me

libsupport/include/katana/Result.h
```diff
@@ -260,6 +260,40 @@ ResultErrno() {
...
+template <class T>
+std::enable_if_t<!std::is_same<T, void>::value, T&&>
+extract_value(Result<T>&& result) {
+  return std::move(result.value());
+}
+
+inline int
+extract_value(Result<void>&&) {
+  return 0;
+}
```
arthurp: These names are very generic. They should be more specific to avoid conflicting with unrelated things.

libsupport/include/katana/Result.h
```diff
@@ -260,6 +260,40 @@ ResultErrno() {
...
+
+#define KATANA_CHECK_NAME(x, y) x##y
+
+#define KATANA_CHECK_IMPL(result_name, expression)                             \
+  ({                                                                           \
+    auto result_name = (expression);                                           \
+    static_assert(                                                             \
+        ::katana::is_result<decltype(result_name)>::value,                     \
+        "KATANA_CHECK requires a katana::Result");                             \
```
arthurp: Can we avoid the need for this by have a function (maybe just an identity function) that we pass the value to. That would trigger a more normal and complete type error.

libsupport/include/katana/Result.h
arthurp: Why "here"? Also how is this picking up the stack information? Doesn't there need to be a macro that adds it?

arthurp (changes requested):
Overall a really exciting start. Please request a rereview once the RFC is fully implemented here. It looks like this is very close.

libsupport/include/katana/Result.h
```diff
@@ -260,6 +260,40 @@ ResultErrno() {
...
+    if (!result_name) {                                                        \
+      return result_name.error().WithContext("here");                          \
+    }                                                                          \
+    std::move(::katana::extract_value(std::move(result_name)));                \
+  })
```
tylershunt, line 288: Neat! I thought `({....})`this was still a GCC only extension, looks like it's made its way into all of the compilers that we care about. For my education: is it part of the standard now?

libsupport/include/katana/Result.h
ddn0: `is_result` maybe is okay but `extract_value(Result<void>&&)` certainly seems too generic. At the least, you can put it in `katana::internal`?

libsupport/include/katana/Result.h
aneeshdurg: not standard

libsupport/src/Http.cpp
```diff
@@ -127,15 +114,8 @@ HttpUploadCommon(CurlHandle&& holder, const std::string& data) {
...
-  auto curl_res = CurlHandle::Make(url, response);
-  if (!curl_res) {
-    return curl_res.error();
-  }
-  CurlHandle curl(std::move(curl_res.value()));
-
-  if (auto res = curl.SetOpt(CURLOPT_HTTPGET, 1L); !res) {
-    return res.error();
-  }
+  CurlHandle curl = KATANA_CHECK(CurlHandle::Make(url, response));
```
aneeshdurg: Just out of curiosity does `auto` work with this macro/`({...})`?

aneeshdurg (approved):

libsupport/src/Http.cpp
danielmawhirter: Yes it does! For the purpose of our style guidelines though, I believe we should avoid `auto` here for the sake of clarity.

libsupport/include/katana/Result.h
danielmawhirter: I don't understand how WithContext works, but it picks up stack information itself.
I also seem to have lost a commit somewhere, "here" is certainly not final and is a parameter to the (missing) `KATANA_CHECK_CONTEXT`

libsupport/include/katana/Result.h
arthurp: It's like `__attribute__`. It's not standard, but is very common. I don't know if niche compilers like ICC and MSVC support it. But the main compilers out there do. (Yes, I called MSVC a niche compiler, fight me. ;-) )

danielmawhirter: Last I saw ICC supports it with caveats, the main one being objects with destructors cannot be initialized inside. We'll cross that bridge when we come to it

arthurp: What's the issue with destructors? I could see how this could break scope extension. Is there anything other than lacking scope extension?

libsupport/src/Http.cpp
arthurp: Other than for cases where the return value is included as an explicit non-inferable argument to the call, I would recommend.

arthurp (approved):
This is great!!!

Finally some good exception support in C++. ;-) (Sorry couldn't help myself, just trolling @tylershunt.)

libsupport/include/katana/Result.h
zhipeng-jia: Is `do {  ..... } while (0)` a solution? I frequently see this idiom when building multiple line macros.

zhipeng-jia: Oh, I see the problem here: it wants to return the last line.

arthurp: And the `do {} while()` idiom is still a statement. This needs to be an expression.

zhipeng-jia:
I may also raise my thought here:

Could we call the macro `KATANA_CHECKED` instead `KATANA_CHECK`? `CHECK` always sound to me like some assert (Google-style C++ code use this convention).

When using `KATANA_CHECKED`, the usage looks like: `value = KATANA_CHECKED(Evaluate(ctx, exp))`, reading smooth.

libsupport/include/katana/Result.h
danielmawhirter: ICC has several caveats with their statement expressions described [here](https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compatibility-and-portability/gcc-compatibility-and-interoperability.html). Since that page concerns their gcc compatibility layer, I assume those caveats spring from the fact that it's probably an ad-hoc  construction in their AST.

And yeah, `do/while` wouldn't be usable here as an outer wrapper. What the statement expression is giving us is that the last expression inside is treated as the return value for the whole block. So it's already self-contained, and wouldn't be able to work inside a `do/while`

libsupport/include/katana/Result.h
danielmawhirter: This was, in some sense, already resolved by the function extracting the value from the Result, so I've removed the static_assert

libsupport/include/katana/Result.h
danielmawhirter: IBM XLC also supports statement expressions, so MSVC is more niche than that

roshandathathri:
> Could we call the macro `KATANA_CHECKED` instead `KATANA_CHECK`?

I'd prefer `KATANA_ASSIGN_OR_RAISE`. In addition to consistency with Arrow, the semantics of the macro are more clear (at the cost of verbosity).

libsupport/include/katana/Result.h
```diff
@@ -260,6 +260,40 @@ ResultErrno() {
...
+    std::move(                                                                 \
+        ::katana::internal::extract_result_value(std::move(result_name)));     \
+  })
+
+#define KATANA_CHECK_CONTEXT(expression, ...)                                  \
+  KATANA_CHECK_IMPL(                                                           \
+      KATANA_CHECK_NAME(_error_or_value, __COUNTER__), expression,             \
+      __VA_ARGS__)
+
+#define KATANA_CHECK(expression) KATANA_CHECK_CONTEXT(expression, "backtrace")
```
roshandathathri: Does this make <Result>`.error()` obsolete? That is, we always return <Result>`.error().WithContext()`? I'm fine with that. Just wanted to confirm.

roshandathathri (approved):
LGTM

libsupport/include/katana/Result.h
arthurp: `.error()` is still useful in cases where the current call is just an internal wrapper that doesn't add any useful context.

arthurp:
> > Could we call the macro `KATANA_CHECKED` instead `KATANA_CHECK`?
> 
> I'd prefer `KATANA_ASSIGN_OR_RAISE`. In addition to consistency with Arrow, the semantics of the macro are more clear (at the cost of verbosity).

Bike shed alert. Everyone, please keep this as short as possible and be willing to accept a result that isn't your ideal.

libsupport/include/katana/Result.h
roshandathathri: But then `KATANA_CHECK` doesn't support or expose that right? My point is we either support both or support only `WithContext` - whether we use the macro or not.

witchel (commented):
Oh yeah!  I'm looking forward to it.

I'll register a vote for KATANA_CHECKED.  I think that does read better and is still short.  

Arthur, while bike shedding is bad, these low-level primitives are tough to introduce and once in place tend to proliferate into difficult to change locations.

danielmawhirter:
I'll adopt `KATANA_CHECKED`. If we ever need to revisit that, it should be a simple-ish find-and-replace pattern.
In the absence of major objections, expect a merge tomorrow!

zhipeng-jia (approved):

arthurp:
> Arthur, while bike shedding is bad, these low-level primitives are tough to introduce and once in place tend to proliferate into difficult to change locations.

Agreed. I was just trying to prevent returning to discussions of approaches we had already rejected. I should have been more explicit. https://docs.google.com/document/d/1WIqCUZKik_klt1lF0Sdlixy9eUCIZi_5rZnEhtN4x60/edit#bookmark=id.ykestgfl7890

ddn0 (approved):

PR #330, opened 2021-06-24
Add missing value in RDG WritePartArrays print

scober:
I was doing some debugging and I found that we were missing this. I usually get an error when I don't have the same number of {}s and values, I'm not sure what is going on here.


tylershunt (approved):

witchel:
This probably isn't related, but I just realized last night that my local debug build had the `NDEBUG` flag set.  This might be because I did `Release=Debug` instead of `Release=DEBUG`.  Or it might be some other user error.  I fixed it, but for a minute there it was disconcerting how K`ATANA_LOG_DEBUG_ASSERT(false) `succeeded in my debug build!

PR #334, opened 2021-06-27
Prevent reloading already loaded Katana plugin modules.

arthurp:


python/katana/plugin.py
```diff
@@ -56,13 +56,13 @@ def load_module(self, fullname):
...
-        module = importlib.import_module(self._insert_table[fullname])
+        aliased_name = self._insert_table[fullname]
         if fullname in sys.modules:
-            raise ImportError(
-                f"module {fullname} already exists while trying to insert "
-                f"{self._insert_table[fullname]} at that name."
-            )
+            module = sys.modules[aliased_name]
+        else:
+            module = importlib.import_module(aliased_name)
```
gliga, line 63: should aliasd_name,fullname be inserted here in _insert_table? (or that is done elsewhere)

gliga: ok based on that assert above, it is probably inserted somewhere else

arthurp: Yea the table is build in `_register_plugin_loader` and passed to the constructor.

python/katana/plugin.py
```diff
@@ -56,13 +56,13 @@ def load_module(self, fullname):
 
         assert fullname in self._insert_table
 
-        # This will get the module from sys.modules if it is already loaded.
-        module = importlib.import_module(self._insert_table[fullname])
+        aliased_name = self._insert_table[fullname]
         if fullname in sys.modules:
```
gliga: if fullname was in sys.modules we already returned; aliased_name?

arthurp: This code creates an alias of the module named `aliased_name` at the name `fullname`. Does that clear it up? I'm not really sure what you are trying to ask.

gliga: there is an if statement at the beginning of the method that checks if fullname in sys.modules. When does fullname change between the two if statements?

arthurp: Ah. I see. You are right. It should be `aliased_name` there. I will fix it. I have no idea how this worked then. Hmm. That's worrisome.

python/katana/plugin.py
```diff
@@ -56,13 +56,13 @@ def load_module(self, fullname):
...
         assert fullname in self._insert_table
 
-        # This will get the module from sys.modules if it is already loaded.
-        module = importlib.import_module(self._insert_table[fullname])
+        aliased_name = self._insert_table[fullname]
+        if aliased_name in sys.modules:
+            module = sys.modules[aliased_name]
+        else:
+            module = importlib.import_module(aliased_name)
         if fullname in sys.modules:
```
gliga: should this statement be only under "else"?

arthurp: It can be. I'll move it.

gliga (approved):

PR #336, opened 2021-06-28
Restrict prop graph topology interface

amberhassaan:
Move away from arrow arrays. Restrict access to internals of topology.
Create new graph instead of SetTopology

IMPORTANT: Wait for https://github.com/KatanaGraph/katana-enterprise/pull/1427 before mergin.

### DIFF
--- a/libgalois/include/katana/ParallelSTL.h
+++ b/libgalois/include/katana/ParallelSTL.h
@@ -434,7 +434,9 @@ iota(const ForwardIt& first, const ForwardIt& last, const T& start_val) {
   on_each([&](unsigned tid, unsigned total) {
     auto [begin, end] = block_range(first, last, tid, total);
     diff_type offset = std::distance(first, begin);
-    std::iota(begin, end, start_val + static_cast<T>(offset));
+    std::iota(
+        begin, end,
+        static_cast<value_type>(start_val) + static_cast<value_type>(offset));
   });
 }
 
--- a/libgalois/include/katana/ParallelSTL.h
+++ b/libgalois/include/katana/ParallelSTL.h
@@ -419,6 +419,39 @@ transform(
   return d_first + std::distance(first, last);
 }
 
+template <typename ForwardIt, typename T>
+void
+iota(const ForwardIt& first, const ForwardIt& last, const T& start_val) {
+  using diff_type = typename std::iterator_traits<ForwardIt>::difference_type;
+  using value_type = typename std::iterator_traits<ForwardIt>::value_type;
+  static_assert(
+      std::is_convertible_v<T, value_type>,
+      "Can't convert start_val to iterator's value_type");
+  static_assert(
+      std::is_arithmetic_v<T> && std::is_arithmetic_v<value_type>,
+      "iota only supported for numeric types");
+
+  on_each([&](unsigned tid, unsigned total) {
+    auto [begin, end] = block_range(first, last, tid, total);
+    diff_type offset = std::distance(first, begin);
+    std::iota(begin, end, start_val + static_cast<T>(offset));
+  });
+}
+
+template <typename ForwardIt, typename T>
+void
+fill(const ForwardIt& first, const ForwardIt& last, const T& val) {
+  using value_type = typename std::iterator_traits<ForwardIt>::value_type;
+  static_assert(
+      std::is_convertible_v<T, value_type>,
+      "Can't convert param val to iterator's value_type");
+
+  on_each([&](unsigned tid, unsigned total) {
+    auto [begin, end] = block_range(first, last, tid, total);
+    std::fill(begin, end, val);
+  });
+}
+
 template <class InputIt, class OutputIt>
 OutputIt
 copy(InputIt first, InputIt last, OutputIt d_first) {
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -18,6 +18,17 @@
 
 namespace katana {
 
+// TODO(amber): find a better place to put this
+template <
+    typename T,
+    typename __Unused = std::enable_if_t<std::is_arithmetic<T>::value>>
+auto
+ProjectAsArrowArray(const T* buf, const size_t len) noexcept {
+  using ArrowDataType = typename arrow::CTypeTraits<T>::ArrowType;
+  using ArrowArrayType = arrow::NumericArray<ArrowDataType>;
+  return std::make_shared<ArrowArrayType>(len, arrow::Buffer::Wrap(buf, len));
+}
+
 /// A graph topology represents the adjacency information for a graph in CSR
 /// format.
 class KATANA_EXPORT GraphTopology {
@@ -31,120 +42,102 @@ public:
   using iterator = node_iterator;
 
   GraphTopology() = default;
+  GraphTopology(GraphTopology&&) = default;
+  GraphTopology& operator=(GraphTopology&&) = default;
 
   GraphTopology(const GraphTopology&) = delete;
   GraphTopology& operator=(const GraphTopology&) = delete;
 
   GraphTopology(
       const Edge* adj_indices, size_t numNodes, const Node* dests,
-      size_t numEdges);
+      size_t numEdges) noexcept;
 
   GraphTopology(NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests) noexcept
-      : adj_indices_(std::move(adj_indices)),
-        dests_(std::move(dests)),
-        out_indices(LargeToArrowArray(adj_indices_)),
-        out_dests(LargeToArrowArray(dests_)) {}
+      : adj_indices_(std::move(adj_indices)), dests_(std::move(dests)) {}
 
-  static std::unique_ptr<GraphTopology> Copy(
-      const GraphTopology& that) noexcept;
+  static GraphTopology Copy(const GraphTopology& that) noexcept;
 
-  /*
-   * TODO(amber): consider re-enabling these in the near future
-  GraphTopology(GraphTopology&& that):
-    adj_indices_(std::move(that.adj_indices_)),
-    dests_(std::move(that.dests)),
-    out_indices(LargeToArrowArray(adj_indices_)),
-    out_dests(LargeToArrowArray(dests_))
-  {}
+  uint64_t num_nodes() const noexcept { return adj_indices_.size(); }
 
-  GraphTopology& operator = (GraphTopology&& that) {
-    adj_indices_ = std::move(that.adj_indices_);
-    dests_ = std::move(that.dests_);
-    out_indices = LargeToArrowArray(adj_indices_);
-    out_dests = LargeToArrowArray(dests_);
-    return *this;
-  }
-  */
+  uint64_t num_edges() const noexcept { return dests_.size(); }
 
-  uint64_t num_nodes() const { return adj_indices_.size(); }
+  const Edge* adj_data() const noexcept { return &adj_indices_[0]; }
 
-  uint64_t num_edges() const { return dests_.size(); }
+  const Node* dest_data() const noexcept { return &dests_[0]; }
 
-  bool Equals(const GraphTopology& other) const {
-    return out_indices->Equals(*other.out_indices) &&
-           out_dests->Equals(*other.out_dests);
+  /**
+   * Checks equality against another instance of GraphTopology.
+   * WARNING: Expensive operation due to element-wise checks on large arrays
+   * @param that: GraphTopology instance to compare against
+   * @returns true if topology arrays are equal
+   */
+  bool Equals(const GraphTopology& that) const noexcept {
+    if (this == &that) {
+      return true;
+    }
+    if (num_nodes() != that.num_nodes()) {
+      return false;
+    }
+    if (num_edges() != that.num_edges()) {
+      return false;
+    }
+
+    return adj_indices_ == that.adj_indices_ && dests_ == that.dests_;
   }
 
   // Edge accessors
 
-  // TODO(amp): [[deprecated("use edges(node)")]]
-  std::pair<Edge, Edge> edge_range(Node node_id) const {
-    KATANA_LOG_DEBUG_ASSERT(node_id < adj_indices_.size());
-    auto edge_start = node_id > 0 ? adj_indices_[node_id - 1] : 0;
-    auto edge_end = adj_indices_[node_id];
-    return std::make_pair(edge_start, edge_end);
+  edge_iterator edge_begin(Node node) const noexcept {
+    return edge_iterator{node > 0 ? adj_indices_[node - 1] : 0};
+  }
+
+  edge_iterator edge_end(Node node) const noexcept {
+    return edge_iterator{adj_indices_[node]};
   }
 
   /// Gets the edge range of some node.
   ///
   /// \param node an iterator pointing to the node to get the edge range of
   /// \returns iterable edge range for node.
-  edges_range edges(const node_iterator& node) const { return edges(*node); }
+  edges_range edges(const node_iterator& node) const noexcept {
+    return edges(*node);
+  }
   // TODO(amp): [[deprecated("use edges(Node node)")]]
 
   /// Gets the edge range of some node.
   ///
   /// \param node node to get the edge range of
   /// \returns iterable edge range for node.
-  edges_range edges(Node node) const {
-    auto [begin_edge, end_edge] = edge_range(node);
-    return MakeStandardRange<edge_iterator>(begin_edge, end_edge);
+  edges_range edges(Node node) const noexcept {
+    return MakeStandardRange<edge_iterator>(edge_begin(node), edge_end(node));
   }
 
-  Node edge_dest(Edge edge_id) const {
+  Node edge_dest(Edge edge_id) const noexcept {
     KATANA_LOG_DEBUG_ASSERT(edge_id < dests_.size());
     return dests_[edge_id];
   }
 
-  nodes_range nodes(Node begin, Node end) const {
+  Node edge_dest(const edge_iterator ei) const noexcept {
+    return edge_dest(*ei);
+  }
+
+  nodes_range nodes(Node begin, Node end) const noexcept {
     return MakeStandardRange<node_iterator>(begin, end);
   }
 
   // Standard container concepts
 
-  node_iterator begin() const { return node_iterator(0); }
+  node_iterator begin() const noexcept { return node_iterator(0); }
 
-  node_iterator end() const { return node_iterator(num_nodes()); }
+  node_iterator end() const noexcept { return node_iterator(num_nodes()); }
 
-  size_t size() const { return num_nodes(); }
+  size_t size() const noexcept { return num_nodes(); }
 
-  bool empty() const { return num_nodes() == 0; }
+  bool empty() const noexcept { return num_nodes() == 0; }
 
 private:
-  // TODO: generalize to typename T
-  static std::shared_ptr<arrow::UInt64Array> LargeToArrowArray(
-      NUMAArray<uint64_t>& lg_arr) noexcept {
-    return std::make_shared<arrow::UInt64Array>(
-        lg_arr.size(),
-        arrow::MutableBuffer::Wrap(lg_arr.data(), lg_arr.size()));
-  }
-
-  static std::shared_ptr<arrow::UInt32Array> LargeToArrowArray(
-      NUMAArray<uint32_t>& lg_arr) noexcept {
-    return std::make_shared<arrow::UInt32Array>(
-        lg_arr.size(),
-        arrow::MutableBuffer::Wrap(lg_arr.data(), lg_arr.size()));
-  }
-
   NUMAArray<Edge> adj_indices_;
   NUMAArray<Node> dests_;
-
-public:
-  // TODO(amber): make these private in the near future. No other class should
-  // access these directly. Instead provide access methods. Also, can't move these
-  // up due to member declaration & initialization order. out_indices depends on adj_indices_
-  std::shared_ptr<arrow::UInt64Array> out_indices;
-  std::shared_ptr<arrow::UInt32Array> out_dests;
 };
 
 /// A property graph is a graph that has properties associated with its nodes
@@ -207,10 +200,7 @@ private:
 
   tsuba::RDG rdg_;
   std::unique_ptr<tsuba::RDGFile> file_;
-
-  // The topology is either backed by rdg_ or shared with the
-  // caller of SetTopology.
-  std::unique_ptr<GraphTopology> topology_ = std::make_unique<GraphTopology>();
+  GraphTopology topology_;
 
   /// A map from the node TypeSetID to
   /// the set of the node type names it contains
@@ -325,8 +315,14 @@ public:
   PropertyGraph() = default;
 
   PropertyGraph(
-      std::unique_ptr<tsuba::RDGFile> rdg_file, tsuba::RDG&& rdg,
-      std::unique_ptr<GraphTopology> topo_to_assign);
+      std::unique_ptr<tsuba::RDGFile>&& rdg_file, tsuba::RDG&& rdg,
+      GraphTopology&& topo) noexcept
+      : rdg_(std::move(rdg)),
+        file_(std::move(rdg_file)),
+        topology_(std::move(topo)) {}
+
+  PropertyGraph(katana::GraphTopology&& topo_to_assign) noexcept
+      : rdg_(), file_(), topology_(std::move(topo_to_assign)) {}
 
   /// Make a property graph from a constructed RDG. Take ownership of the RDG
   /// and its underlying resources.
@@ -337,6 +333,10 @@ public:
   static Result<std::unique_ptr<PropertyGraph>> Make(
       const std::string& rdg_name,
       const tsuba::RDGLoadOptions& opts = tsuba::RDGLoadOptions());
+
+  /// Make a property graph from topology
+  static Result<std::unique_ptr<PropertyGraph>> Make(
+      GraphTopology&& topo_to_assign);
 
   /// \return A copy of this with the same set of properties. The copy shares no
   ///       state with this.
@@ -621,10 +621,7 @@ public:
     return rdg_.MarkEdgePropertiesPersistent(persist_edge_props);
   }
 
-  const GraphTopology& topology() const noexcept {
-    KATANA_LOG_DEBUG_ASSERT(topology_);
-    return *topology_;
-  }
+  const GraphTopology& topology() const noexcept { return topology_; }
 
   /// Add Node properties that do not exist in the current graph
   Result<void> AddNodeProperties(const std::shared_ptr<arrow::Table>& props);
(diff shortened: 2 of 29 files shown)


libgalois/include/katana/ParallelSTL.h
```diff
@@ -419,6 +419,28 @@ transform(
   return d_first + std::distance(first, last);
 }
 
+template <
+    typename ForwardIt,
+    typename T = typename std::iterator_traits<ForwardIt>::value_type,
+    typename __Unused = std::enable_if_t<std::is_arithmetic<T>::value>>
+void
+iota(const ForwardIt& first, const ForwardIt& last, const T& init_val) {
```
ddn0: can we call `init_val`, `start_val` and/or add some documentation on this function explaining what it does in terms of its arguments?

libgalois/include/katana/ParallelSTL.h
```diff
@@ -419,6 +419,28 @@ transform(
   return d_first + std::distance(first, last);
 }
 
+template <
+    typename ForwardIt,
+    typename T = typename std::iterator_traits<ForwardIt>::value_type,
+    typename __Unused = std::enable_if_t<std::is_arithmetic<T>::value>>
```
ddn0: nit: instead of using an arbitrary name for this type parameter can you use this pattern for SFINAE?

```cpp
std::enable_if_t<std::is_arithmetic_v<T>, void>
iota(const ForwardIt& first, ....) { ... }
```

putting it on the return value is the least likely to conflict with any current or future uses.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -18,6 +18,21 @@
 
 namespace katana {
 
+// TODO(amber): find a better place to put this
+template <
+    typename T,
+    typename __Unused = std::enable_if_t<std::is_arithmetic<T>::value>>
+auto
```
ddn0, line 25: nit: I'd prefer an explicit return type and SFINAE on the return type instead of on a default value for a template parameter.

Also, identifiers starting with double underscores are generally reserved: https://en.cppreference.com/w/cpp/language/identifiers and we should avoid introducing them ourselves.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -18,6 +18,21 @@
...
+ProjectAsArrowArray(const T* buf, const size_t len) noexcept {
+  using ArrowDataType = typename arrow::CTypeTraits<T>::ArrowType;
+  using ArrowArrayType = arrow::NumericArray<ArrowDataType>;
+  return std::make_shared<ArrowArrayType>(len, arrow::Buffer::Wrap(buf, len));
+}
+
+/// Provides write access to private data of GraphTopology.
+/// WARNING: Dangerous
+class KATANA_EXPORT GraphTopologyAccess;
```
ddn0: Shouldn't `KATANA_EXPORT` be on the class definition (as well)?

Also you can consider not exporting the access class, e.g.,

```cpp

class GraphTopologyAccess;

... friend class GraphTopologyAccess ...

class KATANA_EXPORT MyAccess: public GraphTopologyAccess {

};

```

libgalois/include/katana/PropertyGraph.h
```diff
@@ -41,49 +56,53 @@ class KATANA_EXPORT GraphTopology {
...
-           out_dests->Equals(*other.out_dests);
+  const Edge* adj_data() const noexcept { return &adj_indices_[0]; }
+
+  const Node* dest_data() const noexcept { return &dests_[0]; }
+
+  bool Equals(const GraphTopology& that) const {
+    if (this == &that) {
+      return true;
+    }
+    if (this->num_nodes() != that.num_nodes()) {
```
ddn0: nit: some preference to omit `this` here as we use `num_nodes()`  (without this) everywhere else in this class

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -442,61 +442,27 @@ struct ClusteringImplementationBase {
  * (read from the underlying RDG) to the in-memory
  * temporary graph (pfg_to).
  */
-  katana::Result<void> CreateDuplicateGraph(
-      katana::PropertyGraph* pfg_from, katana::PropertyGraph* pfg_to,
-      const std::string& edge_property_name,
+  template <typename NodeData>
+  katana::Result<std::unique_ptr<katana::PropertyGraph>> CreateDuplicateGraph(
```
ddn0: nit: why wouldn't we just call this function "DuplicateGraph"?

also, if this takes a template parameter, wouldn't it make more sense to return a TypedPropertyGraph instead?

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -362,7 +362,7 @@ struct LouvainClusteringImplementation
      * Construct temp property graph. This graph gets coarsened as the
      * computation proceeds.
      */
-    auto pfg_mutable = std::make_unique<katana::PropertyGraph>();
+    /*
```
ddn0: can you just delete this code?

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -380,13 +380,17 @@ struct LouvainClusteringImplementation
...
-    std::vector<std::string> temp_edge_property_names = {
-        "_katana_temporary_property_" + edge_weight_property_name};
     if (auto result = ConstructEdgeProperties<EdgeData>(
             pfg_mutable.get(), temp_edge_property_names);
         !result) {
       return result.error();
     }
+    */
+
+    std::vector<std::string> temp_edge_property_names = {
```
ddn0: There is: https://github.com/KatanaGraph/katana/blob/master/libgalois/include/katana/analytics/Utils.h#L75

tools/graph-convert/graph-properties-convert-mongodb.cpp
```diff
@@ -1157,6 +1157,6 @@ katana::ConvertMongoDB(
   if (auto r = builder.Finish(); !r) {
     KATANA_LOG_FATAL("Failed to construct graph: {}", r.error());
   } else {
-    return r.value();
+    return std::move(r.value());
```
ddn0, line 1160: I'm surprised you don't get a warning about an unnecessary move here.

libgalois/include/katana/LargeArray.h
```diff
@@ -206,6 +206,28 @@ class LargeArray {
...
+    }
+    if (left.size() != right.size()) {
+      return false;
+    }
+    // if sizes are equal and data pointers are same then arrays are equal
+    if (left.data() == right.data()) {
+      return true;
+    }
+
+    for (size_t i = 0, sz = left.size(); i < sz; ++i) {
```
roshandathathri: Isn't this an expensive operation? I think it should be documented.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -41,49 +56,53 @@ class KATANA_EXPORT GraphTopology {
...
-    adj_indices_(std::move(that.adj_indices_)),
-    dests_(std::move(that.dests)),
-    out_indices(LargeToArrowArray(adj_indices_)),
-    out_dests(LargeToArrowArray(dests_))
-  {}
-
-  GraphTopology& operator = (GraphTopology&& that) {
+      : adj_indices_(std::move(adj_indices)), dests_(std::move(dests)) {}
+  // out_indices(LargeToArrowArray(adj_indices_)),
+  // out_dests(LargeToArrowArray(dests_)) {}
```
roshandathathri: Drop

libgalois/include/katana/PropertyGraph.h
```diff
@@ -41,49 +56,53 @@ class KATANA_EXPORT GraphTopology {
...
+      return true;
+    }
+    if (this->num_nodes() != that.num_nodes()) {
+      return false;
+    }
+    if (this->num_edges() != that.num_edges()) {
+      return false;
+    }
+
+    return this->adj_indices_ == that.adj_indices_ &&
```
roshandathathri: Ditto about documentation

libgalois/include/katana/PropertyGraph.h
```diff
@@ -41,49 +56,53 @@ class KATANA_EXPORT GraphTopology {
...
   uint64_t num_nodes() const { return adj_indices_.size(); }
 
   uint64_t num_edges() const { return dests_.size(); }
 
-  bool Equals(const GraphTopology& other) const {
-    return out_indices->Equals(*other.out_indices) &&
-           out_dests->Equals(*other.out_dests);
+  const Edge* adj_data() const noexcept { return &adj_indices_[0]; }
+
+  const Node* dest_data() const noexcept { return &dests_[0]; }
```
roshandathathri: Didn't like these low-level functions. Why not provide `const` getters for the two variables instead? The user can then get its data as a pointer.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -98,15 +117,18 @@ class KATANA_EXPORT GraphTopology {
...
-    return MakeStandardRange<edge_iterator>(begin_edge, end_edge);
+    return MakeStandardRange<edge_iterator>(edge_begin(node), edge_end(node));
   }
 
   Node edge_dest(Edge edge_id) const {
     KATANA_LOG_DEBUG_ASSERT(edge_id < dests_.size());
     return dests_[edge_id];
   }
 
+  Node edge_dest(const edge_iterator ei) const noexcept {
```
roshandathathri, line 120: If this can be a `noexcept`, so can the function it calls, right?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -122,30 +144,33 @@ class KATANA_EXPORT GraphTopology {
...
+  friend class GraphTopologyAccess;
 
-  static std::shared_ptr<arrow::UInt32Array> LargeToArrowArray(
-      LargeArray<uint32_t>& lg_arr) noexcept {
-    return std::make_shared<arrow::UInt32Array>(
-        lg_arr.size(),
-        arrow::MutableBuffer::Wrap(lg_arr.data(), lg_arr.size()));
-  }
+  Edge* adj_data() noexcept { return &adj_indices_[0]; }
+  Node* dest_data() noexcept { return &dests_[0]; }
```
roshandathathri: Ditto about low-level functions. Provide direct access to the member variables and getters is better in my opinion. This seems too specific.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -122,30 +144,33 @@ class KATANA_EXPORT GraphTopology {
...
-  // up due to member declaration & initialization order. out_indices depends on adj_indices_
-  std::shared_ptr<arrow::UInt64Array> out_indices;
-  std::shared_ptr<arrow::UInt32Array> out_dests;
+  GraphTopologyAccess(GraphTopology* topo) noexcept : topo_(topo) {
+    KATANA_LOG_DEBUG_ASSERT(topo);
+  }
+
+  GraphTopology::Edge* adj_data() noexcept { return topo_->adj_data(); }
+
+  GraphTopology::Node* dest_data() noexcept { return topo_->dest_data(); }
```
roshandathathri: This is fine.

libgalois/src/BuildGraph.cpp
```diff
@@ -1866,18 +1832,51 @@ ImportData::ValueFromArrowScalar(std::shared_ptr<arrow::Scalar> scalar) {
...
+katana::Result<std::unique_ptr<katana::PropertyGraph>>
+katana::ConvertToPropertyGraph(katana::GraphComponents&& graph_comps) {
+  auto pg_result = katana::PropertyGraph::Make(std::move(graph_comps.topology));
+  if (!pg_result) {
+    return pg_result.error().WithContext("adding topology");
+  }
+  std::unique_ptr<katana::PropertyGraph> graph = std::move(pg_result.value());
+
+  if (graph_comps.nodes.properties->num_columns() > 0) {
+    auto result = graph->AddNodeProperties(graph_comps.nodes.properties);
```
roshandathathri, line 1844: @tylershunt Can the properties be moved?

libgalois/src/PropertyGraph.cpp
```diff
@@ -116,21 +116,19 @@ WriteTopology(const katana::GraphTopology& topology) {
...
     }
   }
 
   if (num_edges) {
-    const auto* raw = topology.out_dests->raw_values();
+    const auto* raw = topology.dest_data();
     static_assert(std::is_same_v<std::decay_t<decltype(*raw)>, uint32_t>);
-    auto buf = std::make_shared<arrow::Buffer>(
-        reinterpret_cast<const uint8_t*>(raw), num_edges * sizeof(uint32_t));
+    auto buf = arrow::Buffer::Wrap(raw, num_edges);
```
roshandathathri, line 131: nit: I'd prefer a similar name instead of `ProjectAsArrowArray`. Maybe `WrapAsArrowArray`?

roshandathathri (approved):
LGTM

arthurp:
@amberhassaan I pushed updates to the python wrapper. The python tests are failing though. It's possible this is a bug in the wrapper, but there is also a crash in a call to `SortAllEdgesByDest` from the k-truss implementation, so I think the issues might be in the C++ implementation.

What I'm seeing are:

* Weird results of algorithms that don't match previous results. Note that the output checks are pretty strict, so if node order changed without actually affecting correctness that would explain some, *but not all* of the failures.
* Inconsistent output in the permutation array returned by `SortAllEdgesByDest`. (This could be a Python wrapper issue, but I doubt it).
* A crash when calling KTruss which seems to be *non-deterministic*.

Note: The Python tests are more aggressive than the C++ tests in some ways because they call the algorithms in more different ways and were written by me without knowing what the algorithm "expects" so they tend to exercise paths the original algorithm implementers didn't think of. So seeing a failure in the Python tests, but not C++ can still point to a C++ issue.

Ping me if you want to have a short meeting to talk about this. Co-debugging can be really helpful for this kind of thing and we can try to determine with certainty whether the bug is in the wrapper or in the C++ code.

arthurp (commented):
Request a rereview when it's time for me to be looking at this again.

libgalois/include/katana/PropertyGraph.h
amberhassaan: Chose return type pattern of enable_if

libgalois/include/katana/ParallelSTL.h
amberhassaan: Fixed: ec837c8a1326e3e22795c6e3bf7a43627f6a0233

libgalois/include/katana/ParallelSTL.h
amberhassaan: Fixed: ec837c8a1326e3e22795c6e3bf7a43627f6a0233

libgalois/include/katana/PropertyGraph.h
amberhassaan: Can you please explain the benefit? 
Related to this, I had a discussion with Roshan and Simon and this class may not be needed, so I might get rid of it.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
amberhassaan: Agree with renaming the function. Not sure if returning a TypedPropertyGraph is worth the change in the code. The call to Duplicate is inside an if-else, each of which produce a PropertyGraph pointer. While typed property graph is constructed afterwards in an iterative loop

tools/graph-convert/graph-properties-convert-mongodb.cpp
amberhassaan: Because r.value() does not return the r-value reference. Compiler throws an error without std::move()

libgalois/include/katana/PropertyGraph.h
amberhassaan: I disagree with providing silly-old getters as they 1) expose internal type of the array 2) are prone to people (inadvertently) trying to copy, or worse, move the arrays.

amberhassaan: I'd be happier if we introduce a Buffer type and return a buffer with `const T*` and `a size`, so the user doesn't rely on `num_nodes()` or `num_edges`

arthurp: @amberhassaan you just described an arrow array pretty much. We could just wrap the underlying memory in an arrow array (that does not take ownership of the buffer) and return that.

arthurp: That said, not providing any access to the buffers would be better. Best to just provide abstract accessors so no one can observe the internal data layout.

roshandathathri: I agree with Arthur. We should either drop these or return them as an `arrow::Array`

amberhassaan: Yes, but Arrow Arrays come with minor annoyances that add up to anger. Size is int64_t. Don't have `operator[]`. Not sure but seems like I need a `shared_ptr<arrow::Buffer>` to get started with creating the array. Feels too much. I just want a 
```
template <typename T>
Buffer {
  const T* data_;
  size_t size_;
...
};

// and may be
template <typename T>
MutableBuffer {
   T* data_;
   size_t size_;
};
```

arthurp: Can we just remove these accessors then? They totally break the abstraction you were trying to create anyway.

libgalois/include/katana/PropertyGraph.h
amberhassaan: Removed GraphTopologyAccess in 6f4b8a0971e97a7fcad213f68d6e2d97359f777c

libgalois/include/katana/PropertyGraph.h
amberhassaan: Fixed: 6f4b8a0971e97a7fcad213f68d6e2d97359f777c

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
amberhassaan: Done

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
amberhassaan: Done.

libgalois/include/katana/LargeArray.h
amberhassaan: Done

libgalois/include/katana/PropertyGraph.h
amberhassaan: Done

libgalois/include/katana/PropertyGraph.h
amberhassaan: Not really. I need a back-door. I can either bring back the GraphTopologyAccess class or I can use these accessors and do a `const_cast` to make it obvious that I'm doing something bad.

libgalois/include/katana/PropertyGraph.h
amberhassaan: Yes, they all should be noexcept :)

libgalois/include/katana/PropertyGraph.h
danielmawhirter: [This RFC](https://docs.google.com/document/d/16XxEBlbaP7yBSxFPIB_XDRP4CyqxAfSdScR0Q12TFKU/edit?usp=sharing) contains my idea for how to wrap nodes and edges into arrow in the long term. It doesn't address the allocators, only the semantics, to allow us to continue using some of arrow's functionality w.r.t parquet, ipc, compute, etc. The custom classes created there could implement `operator[]`, and/or modifiability, which are not valid in arrow in general due to nullability.
This PR need not consider that if you don't want it to, but there seems to be enough of a discussion on semantics here to merit bringing it up.

arthurp: @amberhassaan Where do you need the back door? Generally needing a back door implies the rest of the interface is incomplete. Maybe there needs to be a more capable constructor or Make function in this class.

amberhassaan: @arthurp : I need it temporarily. There are functions like `SortEdgesByDest(pg_ptr)` that modify the topology arrays. Then some code in Partitioning (libgluon/Partition.cpp) that modifies the arrays. 
I share the desire of a future without such back doors.

amberhassaan: @danielmawhirter : I agree that arrow Arrays offer benefit when accessing storage. I'm not convinced yet that they offer a benefit when it comes to in-memory topologies, which I'm hoping we'll have a few more of in the near future. 

Once storage makes use of arrow arrays for topology, it may make sense for in-memory representation to follow in vein, but at least for now, LargeArray (NumaArray) offers numa-aware performance benefits and works well with various places in code, such as computing transposed or sorted topologies, even in some analytics apps, and in libgluon. Most of these places were converting LargeArrays to Arrow Arrays and leaking memory. 

Also, once we figure out num-aware allocators for Arrow Arrays, we should revisit the topology representation.

arthurp: I didn't realize it's temporary. Just leave them as returning pointers and add todos at their declarations and use. Hopefully we can update the other code to use the abstraction soon.

Re: NUMA Aware arrow arrays. I actually know how to do it. I just haven't spent the time. I assumed there was someone else tracking it. See https://arrow.apache.org/docs/cpp/memory.html#memory-pools. It's actually pretty straight forward. I don't think it matters that the allocator uses parallel sections. It's still just an allocator. Obviously we would need to use a different allocator for allocations inside parallel sections than in the main thread. But this is already true, IIUC.

roshandathathri: @danielmawhirter is aware of these differences and is looking into it. We will need different types: `arrow::Array`, `katana::gstl::ArrowArray`, and `katana::NUMAArrowArray` (analogous to `std::vector`, `katana::gstl::Vector`, and `katana::NUMAArray` respectively).

arthurp: I'll step back from this, because I have no understanding why a different class is needed for each of these cases. It seems like they should all the same interface to me. But it's not my area, so I'll leave it to y'all. I just worry that a proliferation of classes which provide extremely similar interfaces, but incompatibly, will cause maintenance problems in the future.

roshandathathri: For arrow, we will NOT need different classes; just different types. The difference b/w the types will be different allocators. But all of them will NOT provide the same semantics. While `katana::gstl::Vector` can replace all instances of `std::vector`, `katana::NUMAArray` cannot (it's not about performance but about correctness).

python/katana/_property_graph_numba.pyx.jinja
```diff
@@ -14,10 +15,10 @@ from libc.stdint cimport uint32_t, uint64_t
     return self.topology().num_edges()
 {% endcall %}
 {% call numba.method_with_body("edge_index", "uint64_t", ["uint64_t"]) %}
-    return self.topology().out_indices.get().Value(arg1)
+    return deref(self.topology().edge_end(arg1))
```
arthurp: This needs to be `self.topology().edges(arg1).end()` I didn't expose `edge_end` in python because I hope it's going away.

python/katana/_property_graph.pyx
```diff
@@ -140,13 +142,13 @@ cdef class PropertyGraphBase:
         Can be called from numba compiled code.
         """
         cdef uint64_t prev
+        #TODO(amber): need to dereference GraphTopology::edge_begin()/edge_end()
+        # here
```
arthurp: Why do you need to dereference them here? They are dereferenced below.

amberhassaan: That was a TODO for you :). You already took care of it. I'll go ahead and remove it.

python/katana/_property_graph.pyx
```diff
@@ -140,13 +142,13 @@ cdef class PropertyGraphBase:
         Can be called from numba compiled code.
         """
         cdef uint64_t prev
+        #TODO(amber): need to dereference GraphTopology::edge_begin()/edge_end()
```
amberhassaan: @arthurp : Can I remove the line above that declares a seemingly unused variable?

arthurp: Yep. My bad.

amberhassaan:
> @amberhassaan I pushed updates to the python wrapper. The python tests are failing though. It's possible this is a bug in the wrapper, but there is also a crash in a call to `SortAllEdgesByDest` from the k-truss implementation, so I think the issues might be in the C++ implementation.
> 
> What I'm seeing are:
> 
> * Weird results of algorithms that don't match previous results. Note that the output checks are pretty strict, so if node order changed without actually affecting correctness that would explain some, _but not all_ of the failures.
> * Inconsistent output in the permutation array returned by `SortAllEdgesByDest`. (This could be a Python wrapper issue, but I doubt it).
> * A crash when calling KTruss which seems to be _non-deterministic_.
> 
> Note: The Python tests are more aggressive than the C++ tests in some ways because they call the algorithms in more different ways and were written by me without knowing what the algorithm "expects" so they tend to exercise paths the original algorithm implementers didn't think of. So seeing a failure in the Python tests, but not C++ can still point to a C++ issue.
> 
> Ping me if you want to have a short meeting to talk about this. Co-debugging can be really helpful for this kind of thing and we can try to determine with certainty whether the bug is in the wrapper or in the C++ code.

@arthurp : There was a bug I introduced in SortEdgesByDest() . I've fixed it since and this should resolve most if not all test failures. I'm surprised that CPP analytics tests, including triangle_counting, ktruss and everyone else using this function, didn't fail. We must have subpar verification?

roshandathathri:
> I'm surprised that CPP analytics tests, including triangle_counting, ktruss and everyone else using this function, didn't fail. We must have subpar verification?

We do sanity checks but not full output verification for shared-memory apps.

arthurp:
> We must have subpar verification?

Yes. Yes, we do.

amberhassaan:
> > I'm surprised that CPP analytics tests, including triangle_counting, ktruss and everyone else using this function, didn't fail. We must have subpar verification?
> 
> We do sanity checks but not full output verification for shared-memory apps.

My bug should have been insane for our sanity checks.

libgalois/include/katana/ParallelSTL.h
```diff
@@ -419,6 +419,27 @@ transform(
...
+std::enable_if_t<std::is_arithmetic_v<T>>
+iota(const ForwardIt& first, const ForwardIt& last, const T& start_val) {
+  using diff_type = typename std::iterator_traits<ForwardIt>::difference_type;
+  using value_type = typename std::iterator_traits<ForwardIt>::value_type;
+  static_assert(
+      std::is_convertible_v<T, value_type>,
+      "Can't convert start_val to iterator's value_type");
+  static_assert(
+      std::is_arithmetic_v<T> && std::is_arithmetic_v<value_type>,
+      "iota only supported for numeric types");
```
roshandathathri, line 432: Why are these asserts needed here? Shouldn't line 439 throw these asserts?

roshandathathri: Sorry, I thought I submitted the above comment long ago. But apparently I hadn't!

amberhassaan: I prefer keeping these checks as we're not sure how the standard library will respond. That's implementation dependent.

PR #337, opened 2021-06-28
utility: add approximation function for an arrow table's capacity size

patrickkenney9801:
New function measures the total size in capacity allocated to arrow tables not just the amount of space used to store data


patrickkenney9801:
Early testing shows that differences between size and capacity measurements of arrow tables differ by an epsilon so adding these functions will only add unnecessary confusion

INFO: ms=262 host=0 maxMem=310.9MB mem=310.9MB node_table_bytes=1144614 edge_table_bytes=168420 node_table_capacity_bytes=1146252 edge_table_capacity_bytes=169260 | done partitioning
INFO: ms=262 host=1 maxMem=368.98MB mem=368.98MB node_table_bytes=5472096 edge_table_bytes=1466115 node_table_capacity_bytes=5472877 edge_table_capacity_bytes=1467228 | done partitioning

PR #339, opened 2021-06-29
Rename array/vector types and document

roshandathathri:
`LargeArray` is now `NUMAArray` and `PODResizeableArray` is now `PODVector`

### DIFF
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -35,7 +35,7 @@
 
 #include "katana/AtomicWrapper.h"
 #include "katana/Galois.h"
-#include "katana/PODResizeableArray.h"
+#include "katana/PODVector.h"
 #include "katana/config.h"
 
 namespace katana {
@@ -43,7 +43,7 @@ namespace katana {
  * Concurrent dynamically allocated bitset
  **/
 class KATANA_EXPORT DynamicBitset {
-  katana::PODResizeableArray<katana::CopyableAtomic<uint64_t>> bitvec_{};
+  katana::PODVector<katana::CopyableAtomic<uint64_t>> bitvec_{};
   size_t num_bits_{0};
 
 public:
--- a/libgalois/include/katana/gstl.h
+++ b/libgalois/include/katana/gstl.h
@@ -40,51 +40,152 @@ namespace katana {
 
 namespace gstl {
 
-//! [define Pow2VarSizeAlloc]
+//! [STL vector using katana Pow2BlockAllocator]
+//! specializes std::vector to use katana concurrent, scaleable allocator:
+//! the allocator is composed of thread-local allocators that allocate in
+//! multiples of (huge) pages by acquiring a global lock, and divide the pages
+//! into a power-of-2 size blocks. Each per-thread allocator maintains a pool
+//! of free blocks of different power-of-2 sizes. When an object is allocated,
+//! it gets a block from the pool of the thread that allocated it. When an
+//! object is deallocated, its block of memory is added to the pool of the
+//! thread that deallocated it.
+//!
+//! Use this when allocations and deallocations can occur in a parallel region.
+//! As the memory allocated can be reused for another allocation only by the
+//! thread that deallocated it, this is not suitable for use cases where the
+//! main thread always does the deallocation (after the parallel region).
+//!
+//! If the allocation is large and of known size, then check katana::NUMAArray.
+//! If the allocation needs to be uninitialized, then check katana::PODVector.
+//! Read CONTRIBUTING.md for a more detailed comparison between these types.
 template <typename T>
-using Pow2Alloc = Pow2BlockAllocator<T>;
-//! [define Pow2VarSizeAlloc]
+using Vector = std::vector<T, Pow2BlockAllocator<T>>;
 
+//! [STL deque using katana Pow2BlockAllocator]
+//! specializes std::deque to use katana concurrent, scaleable allocator:
+//! the allocator is composed of thread-local allocators that allocate in
+//! multiples of (huge) pages by acquiring a global lock, and divide the pages
+//! into a power-of-2 size blocks. Each per-thread allocator maintains a pool
+//! of free blocks of different power-of-2 sizes. When an object is allocated,
+//! it gets a block from the pool of the thread that allocated it. When an
+//! object is deallocated, its block of memory is added to the pool of the
+//! thread that deallocated it.
+//!
+//! Use this when allocations and deallocations can occur in a parallel region.
+//! As the memory allocated can be reused for another allocation only by the
+//! thread that deallocated it, this is not suitable for use cases where the
+//! main thread always does the deallocation (after the parallel region).
 template <typename T>
-using FixedSizeAlloc = FixedSizeAllocator<T>;
+using Deque = std::deque<T, Pow2BlockAllocator<T>>;
 
-//! [STL vector using Pow2VarSizeAlloc]
+//! [STL list using katana FixedSizeAllocator]
+//! specializes std::list to use katana concurrent, scaleable allocator:
+//! the allocator is composed of thread-local allocators that allocate in
+//! multiples of (huge) pages by acquiring a global lock, and divide the pages
+//! into fixed size blocks. Each per-thread allocator maintains a pool of free
+//! blocks. When an object is allocated, it gets a block from the pool of the
+//! thread that allocated it. When an object is deallocated, its block of
+//! memory is added to the pool of the thread that deallocated it.
+//!
+//! Use this when allocations and deallocations can occur in a parallel region.
+//! As the memory allocated can be reused for another allocation only by the
+//! thread that deallocated it, this is not suitable for use cases where the
+//! main thread always does the deallocation (after the parallel region).
 template <typename T>
-using Vector = std::vector<T, Pow2Alloc<T>>;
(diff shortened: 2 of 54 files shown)


python/katana/datastructures.pxd.jinja
```diff
@@ -44,7 +44,7 @@ cdef class {{class_name}}:
 
 {% macro wrap_large_array(inst) %}
 {% set element_type = inst.element_c_type -%}
-{% set underlying_type %}datastructures.LargeArray[{{element_type}}]{% endset -%}
+{% set underlying_type %}datastructures.NumaArray[{{element_type}}]{% endset -%}
 {% set class_name %}LargeArray_{{element_type}}{% endset -%}
```
arthurp: You missed an instance of "LargeArray".

python/katana/datastructures.pyx.jinja
```diff
@@ -135,7 +135,7 @@ class AllocationPolicy(Enum):
 
 {% macro wrap_large_array(inst) %}
 # inst.element_c_type, inst.element_py_type, inst.element_format_code
-{% set underlying_type %}datastructures.LargeArray[{{inst.element_c_type}}]{% endset -%}
+{% set underlying_type %}datastructures.NumaArray[{{inst.element_c_type}}]{% endset -%}
 {% set class_name %}LargeArray_{{inst.element_c_type}}{% endset -%}
```
arthurp: And here.

arthurp (commented):
Use the grep!

libgalois/include/katana/DynamicBitset.h
```diff
@@ -35,15 +35,15 @@
...
+#include "katana/PodVector.h"
 #include "katana/config.h"
 
 namespace katana {
 /**
  * Concurrent dynamically allocated bitset
  **/
 class KATANA_EXPORT DynamicBitset {
-  katana::PODResizeableArray<katana::CopyableAtomic<uint64_t>> bitvec_{};
+  katana::PodVector<katana::CopyableAtomic<uint64_t>> bitvec_{};
```
ddn0: slight preference for `PODVector`. We have `AtomicPODProperty` for instance already.

And for consistency, probably `NUMAArray`

arthurp:
Side question: Why is NUMAArray a type? Shouldn't it be an allocator that can be used with any number of other types?

amberhassaan:
> Side question: Why is NUMAArray a type? Shouldn't it be an allocator that can be used with any number of other types?

@arthurp : Asking the right questions. We should have a DynamicArray type, which can use a standard allocator or a Numa Allocator. Then NumArray = DynamicArray + NumaAllocator.

libgalois/include/katana/gstl.h
```diff
@@ -40,51 +40,134 @@ namespace katana {
...
+//! When destroyed, its memory is reclaimed by the thread that destroys it.
+//!
+//! Use this when allocations and deallocations can occur in a parallel region.
+//! As the memory allocated can be reused for another allocation only by the
+//! thread that deallocated it, this is not suitable for use cases where the
+//! main thread always does the deallocation (after the parallel region).
 template <typename K, typename V, typename C = std::less<K>>
-using Map = std::map<K, V, C, Pow2Alloc<std::pair<const K, V>>>;
-
+using Map = std::map<K, V, C, Pow2BlockAllocator<std::pair<const K, V>>>;
```
amberhassaan: Why are we not using FixedSizeAllocator for std::map and std::set?

roshandathathri:
> > Side question: Why is NUMAArray a type? Shouldn't it be an allocator that can be used with any number of other types?
> 
> @arthurp : Asking the right questions. We should have a DynamicArray type, which can use a standard allocator or a Numa Allocator. Then NumArray = DynamicArray + NumaAllocator.

Agree with Amber except that I'd prefer to call it an `Array` type.
However, the NumaAllocator is NOT a stand-in replacement for other allocators because it uses all threads to do the NUMA allocation. So I'd prefer the current special type instead of an `Array` or `DynamicArray` type whose behavior would change depending on whether it is using a NumaAllocator or not.

libgalois/include/katana/gstl.h
```diff
@@ -40,51 +40,134 @@ namespace katana {
 
 namespace gstl {
 
-//! [define Pow2VarSizeAlloc]
+//! [STL vector using katana Pow2BlockAllocator]
+//! specializes std::vector to use katana concurrent, scaleable allocator:
+//! the allocator is composed of thread-local allocators
```
amberhassaan: Suggesting slight change for clarity (correct me if I'm wrong as I haven't looked at that code lately). 

the allocator is composed of thread-local allocators that allocate memory in power-of-2 sized blocks. Each per-thread allocator maintain a pool of free pages it replenishes using mmap(), acquiring a global lock. Allocators use Huge pages if available [See https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt], and fall back to malloc for larger sizes, which requires a global lock.

libgalois/include/katana/gstl.h
roshandathathri: No idea. I can change it if you think that's more appropriate.

libgalois/include/katana/DynamicBitset.h
roshandathathri: Yes, will do.

libgalois/include/katana/gstl.h
roshandathathri: They always get allocation in terms of pages and they require a global lock to get pages - huge or otherwise, right?

libgalois/include/katana/gstl.h
```diff
@@ -40,51 +40,134 @@ namespace katana {
...
+//! thread that deallocated it, this is not suitable for use cases where the
+//! main thread always does the deallocation (after the parallel region).
 template <typename T>
-using Vector = std::vector<T, Pow2Alloc<T>>;
-//! [STL vector using Pow2VarSizeAlloc]
-
+using Deque = std::deque<T, Pow2BlockAllocator<T>>;
+
+//! [STL list using katana FixedSizeAllocator]
+//! specializes std::list to use katana concurrent, scaleable allocator:
```
amberhassaan, line 82: The language for List and other containers that use FixedSizeAllocator should read:

The allocator is composed of thread-local allocators that allocate by dividing pages into fixed size blocks. Each per-thread allocator maintain a pool of free pages it replenishes using mmap(), acquiring a global lock.

libgalois/include/katana/gstl.h
```diff
@@ -40,51 +40,134 @@ namespace katana {
...
+//! the allocator is composed of thread-local allocators
+//! that manage thread-local pages and only use a global lock to
+//! get allocation in chunks of (huge) pages.
+//! When destroyed, its memory is reclaimed by the thread that destroys it.
```
amberhassaan: This sentence should use the term "deallocate" instead of "destroy".

libgalois/include/katana/gstl.h
amberhassaan: Please do. My understanding is that since it's a Red-black tree, the allocations are always fixed size

libgalois/include/katana/gstl.h
amberhassaan: > > > Side question: Why is NUMAArray a type? Shouldn't it be an allocator that can be used with any number of other types?
> > 
> > 
> > @arthurp : Asking the right questions. We should have a DynamicArray type, which can use a standard allocator or a Numa Allocator. Then NumArray = DynamicArray + NumaAllocator.
> 
> Agree with Amber except that I'd prefer to call it an `Array` type.
> However, the NumaAllocator is NOT a stand-in replacement for other allocators because it uses all threads to the NUMA allocation. So I'd prefer the current special type instead of an `Array` or `DynamicArray` type whose behavior would change depending on whether it is using a NumaAllocator or not.

That is a good point. I didn't think of that.

libgalois/include/katana/gstl.h
roshandathathri: I'm changing it to "the allocator is composed of thread-local allocators that allocate in multiples of (huge) pages by acquiring a global lock and divide the pages into fixed size blocks. Each per-thread allocator maintains a pool of free blocks. When the object is deallocated, its block of memory is added to the free pool of the thread that deallocated it."

roshandathathri: I'm changing it to "the allocator is composed of thread-local allocators that allocate in multiples of (huge) pages by acquiring a global lock and divide the pages into fixed size blocks. Each per-thread allocator maintains a pool of free blocks. When an object is allocated, it gets a block from the pool of the thread that allocated it. When an object is deallocated, its block of memory is added to the pool of the thread that deallocated it."

amberhassaan: I'm happy with this text.

roshandathathri: For Pow2BlockAllocator, I'm changing it to "the allocator is composed of thread-local allocators that allocate in multiples of (huge) pages by acquiring a global lock, and divide the pages into a power-of-2 size blocks. Each per-thread allocator maintains a pool of free blocks of different power-of-2 sizes. When an object is allocated, it gets a block from the pool of the thread that allocated it. When an object is deallocated, its block of memory is added to the pool of the thread that deallocated it."

roshandathathri:
I've handled all comments so far.

arthurp (approved):

ddn0 (approved):

PR #340, opened 2021-06-30
Pin pylint

arthurp:


aneeshdurg:
Why are we pinning only pylint and nothing else?

arthurp:
> Why are we pinning only pylint and nothing else?

Because in general pinning makes the generated packages have poor compatibility with other packages built with more recent versions of packages. We have no system to automatically detect and manage upgrades. I would like to find time to write some CI jobs and detect new upstream packages and generate automatic PRs with the updated version and then setup some kind of  conda package pinning.

aneeshdurg:
Makes sense. But why pin pylint then? Is there something broken with the latest version?

arthurp:
> Makes sense. But why pin pylint then? Is there something broken with the latest version?

The lastest version added new checks it seems. We probably need to white list instead of black list the checks we support, but for now I just limit the version to get CI passing as it was before. I'll create a JIRA issues for managing the new version of pylint.

The pylint warnings are good ones, BTW. So we should change our code to pass the new version. But not tonight, not being able to merge other PRs is not something we can tolerate.

aneeshdurg (approved):
Got it, just wanted to make sure we captured the reasoning somewhere, like in this comment thread 😁

LGTM!

arthurp:
Thanks. Please review https://github.com/KatanaGraph/katana-enterprise/pull/1450 as well.

PR #343, opened 2021-06-30
Add docs about KATANA_CHECKED

ddn0:


danielmawhirter (approved):
The MakeList example looks like it could be `KATANA_CHECKED_CONTEXT`, otherwise LGTM

PR #349, opened 2021-07-06
[KAT-538] Encapsulate node and edge properties

witchel:
This change encapsulates the node and edge property tables for our graphs.
Exposing the arrow table of properties makes it impossible to change our
representation to keep a subset of properties in memory at once, so we
stop doing that.

There is a dependent PR (https://github.com/KatanaGraph/katana-enterprise/pull/1479) for enterprise that has more detail.

### DIFF
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -10,6 +10,7 @@
 #include <arrow/chunked_array.h>
 #include <arrow/type_traits.h>
 
+#include "katana/ArrowInterchange.h"
 #include "katana/Details.h"
 #include "katana/ErrorCode.h"
 #include "katana/NUMAArray.h"
@@ -262,17 +263,60 @@ private:
     rdg_.set_mirror_nodes(std::move(a));
   }
 
+  /// Return the node property table for local nodes
+  const std::shared_ptr<arrow::Table>& node_properties() const {
+    return rdg_.node_properties();
+  }
+
+  /// Return the edge property table for local edges
+  const std::shared_ptr<arrow::Table>& edge_properties() const {
+    return rdg_.edge_properties();
+  }
+
 public:
   /// PropertyView provides a uniform interface when you don't need to
   /// distinguish operating on edge or node properties
-  struct PropertyView {
-    PropertyGraph* g;
+  struct ReadOnlyPropertyView {
+    const PropertyGraph* const_g;
 
     std::shared_ptr<arrow::Schema> (PropertyGraph::*schema_fn)() const;
-    std::shared_ptr<arrow::ChunkedArray> (PropertyGraph::*property_fn)(
+    std::shared_ptr<arrow::ChunkedArray> (PropertyGraph::*property_fn_int)(
         int i) const;
-    const std::shared_ptr<arrow::Table>& (
-        PropertyGraph::*properties_fn)() const;
+    std::shared_ptr<arrow::ChunkedArray> (PropertyGraph::*property_fn_str)(
+        const std::string& str) const;
+    int32_t (PropertyGraph::*property_num_fn)() const;
+
+    std::shared_ptr<arrow::Schema> schema() const {
+      return (const_g->*schema_fn)();
+    }
+
+    std::shared_ptr<arrow::ChunkedArray> GetProperty(int i) const {
+      return (const_g->*property_fn_int)(i);
+    }
+
+    std::shared_ptr<arrow::ChunkedArray> GetProperty(
+        const std::string& str) const {
+      return (const_g->*property_fn_str)(str);
+    }
+
+    int32_t GetNumProperties() const { return (const_g->*property_num_fn)(); }
+
+    uint64_t ApproxMemUse() const {
+      uint64_t total_mem_use = 0;
+      for (int32_t i = 0; i < GetNumProperties(); ++i) {
+        const auto& chunked_array = GetProperty(i);
+        for (const auto& array : chunked_array->chunks()) {
+          total_mem_use += katana::ApproxArrayMemUse(array);
+        }
+      }
+      return total_mem_use;
+    }
+  };
+
+  struct MutablePropertyView {
+    ReadOnlyPropertyView ropv;
+    PropertyGraph* g;
+
     Result<void> (PropertyGraph::*add_properties_fn)(
         const std::shared_ptr<arrow::Table>& props);
     Result<void> (PropertyGraph::*upsert_properties_fn)(
@@ -280,19 +324,20 @@ public:
     Result<void> (PropertyGraph::*remove_property_int)(int i);
     Result<void> (PropertyGraph::*remove_property_str)(const std::string& str);
 
-    std::shared_ptr<arrow::Schema> schema() const { return (g->*schema_fn)(); }
+    std::shared_ptr<arrow::Schema> schema() const { return ropv.schema(); }
 
-    std::shared_ptr<arrow::ChunkedArray> Property(int i) const {
-      return (g->*property_fn)(i);
+    std::shared_ptr<arrow::ChunkedArray> GetProperty(int i) const {
+      return ropv.GetProperty(i);
     }
 
-    const std::shared_ptr<arrow::Table>& properties() const {
-      return (g->*properties_fn)();
+    std::shared_ptr<arrow::ChunkedArray> GetProperty(
+        const std::string& str) const {
+      return ropv.GetProperty(str);
     }
 
-    std::vector<std::string> property_names() const {
-      return properties()->ColumnNames();
-    }
+    int32_t GetNumProperties() const { return ropv.GetNumProperties(); }
+
+    uint64_t ApproxMemUse() const { return ropv.ApproxMemUse(); }
 
     Result<void> AddProperties(
         const std::shared_ptr<arrow::Table>& props) const {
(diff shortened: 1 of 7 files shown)


libgalois/include/katana/PropertyGraph.h
```diff
@@ -262,38 +262,51 @@ class KATANA_EXPORT PropertyGraph {
...
     }
 
-    std::vector<std::string> property_names() const {
-      return properties()->ColumnNames();
+    std::shared_ptr<arrow::ChunkedArray> GetProperty(
+        const std::string& str) const {
+      return (const_g->*property_fn_str)(str);
     }
 
+    int32_t GetPropertyNum() const { return (const_g->*property_num_fn)(); }
```
ddn0: idle comment: every time I read this, I always think "shouldn't this be called GetNumProperties", but I think the last time this came up, we decided the current `GetFooNum` was consistent with the current API so we have kept things as there are now (cc @roshandathathri who is the person I strongly associate with this discussion)

ddn0 (approved):

libgalois/include/katana/PropertyGraph.h
witchel: I have updated the names to `GetNumProperties()`.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -262,38 +262,51 @@ class KATANA_EXPORT PropertyGraph {
...
+  /// Return the node property table for local nodes
+  const std::shared_ptr<arrow::Table>& node_properties() const {
+    return rdg_.node_properties();
+  }
+
+  /// Return the edge property table for local edges
+  const std::shared_ptr<arrow::Table>& edge_properties() const {
+    return rdg_.edge_properties();
+  }
+
```
arthurp, line 275: Do we want to expose this? Will we always be able to construct an arrow table regardless of the internal storage of the properties?

libgalois/include/katana/PropertyGraph.h
```diff
@@ -262,38 +262,51 @@ class KATANA_EXPORT PropertyGraph {
...
 public:
   /// PropertyView provides a uniform interface when you don't need to
   /// distinguish operating on edge or node properties
   struct PropertyView {
     PropertyGraph* g;
+    const PropertyGraph* const_g;
```
arthurp, line 280: Why is this needed? I suspect there is a detail of C++ typing I don't understand.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -646,36 +660,59 @@ class KATANA_EXPORT PropertyGraph {
...
         .schema_fn = &PropertyGraph::node_schema,
-        .property_fn = &PropertyGraph::GetNodeProperty,
-        .properties_fn = &PropertyGraph::node_properties,
+        .property_fn_int = &PropertyGraph::GetNodeProperty,
+        .property_fn_str = &PropertyGraph::GetNodeProperty,
+        .property_num_fn = &PropertyGraph::GetNumNodeProperties,
         .add_properties_fn = &PropertyGraph::AddNodeProperties,
         .upsert_properties_fn = &PropertyGraph::UpsertNodeProperties,
         .remove_property_int = &PropertyGraph::RemoveNodeProperty,
         .remove_property_str = &PropertyGraph::RemoveNodeProperty,
```
arthurp: Why are we using function pointers here instead of a simple subclass? It would eliminate the need for the explicit implementations of all the silly little methods that really just implement a vtable.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -767,6 +804,9 @@ CreateSymmetricGraph(PropertyGraph* pg);
 KATANA_EXPORT Result<std::unique_ptr<PropertyGraph>>
 CreateTransposeGraphTopology(const GraphTopology& topology);
 
+/// Report approximate memory use for a property view
+KATANA_EXPORT uint64_t ApproxMemUse(const PropertyGraph::PropertyView& pview);
+
```
arthurp: Why not make this a method on PropertyView.

libgalois/src/PropertyViews.cpp
```diff
@@ -19,3 +19,24 @@ katana::internal::ExtractArrays(
...
+  std::vector<arrow::Array*> ret;
+  for (auto& property : properties) {
+    auto column = pview.GetProperty(property);
+    if (!column) {
+      return ErrorCode::PropertyNotFound;
+    }
+    if (column->num_chunks() != 1) {
+      // Katana form graphs only contain single chunk property columns.
+      return ErrorCode::TODO;
+      // TODO: Maybe we need an InvalidGraph error
```
arthurp: Username? Also add message to error code?

arthurp (approved):
Overall, this is OK. I feel like it increases the interface surface more than is required for the features you implement. I would encourage you to reduce the interface if possible. One example, is that you expose both `arrow::Table` and `PropertyView`. They do similar things. Maybe just expose one of them.

libgalois/include/katana/PropertyGraph.h
witchel: I'm not following.  I moved these functions from public to private to stop exposing them.

libgalois/include/katana/PropertyGraph.h
witchel: Sorry, I asked about this in #1479, which is the associated enterprise PR.  Here are the questions.

In PropertyGraph.h, to implement const and non-const PropertyViews I have a PropertyGraph* called g and a const PropertyGraph* called const_g. Is there a better way of doing this?

For the const PropertyView, I leave several function pointers as null. I could make functions that assert, but it seemed annoying to have several different types of dummy functions that do nothing but assert.

libgalois/include/katana/PropertyGraph.h
arthurp: My bad. I wasn't paying attention. This may also invalidate my global comment.

libgalois/include/katana/PropertyGraph.h
witchel: Donald did the PropertyView class initially.  Do you mean a nested class or subclass?  `PropertyView` isn't a subclass of any class.

libgalois/include/katana/PropertyGraph.h
arthurp: I see no connection between the two pointer types and the nulls vs asserts. But it's no big deal.

libgalois/include/katana/PropertyGraph.h
arthurp: What I mean is that you could have a pure virtual base class and then a few subclasses implementing that. Not that there is currently such a structure. That said, I just realized why this approach was probably chosen. The methods are mostly just methods that already exist on `PropertyGraph`, so this MIGHT be slightly simpler to implement.

I'll move on. :-)

libgalois/include/katana/PropertyGraph.h
witchel: Oh, good idea, thanks.

libgalois/src/PropertyViews.cpp
witchel: Ok.  For the record, I copied this code from the function above.  I guess that author did not provide a username for the TODO or an error message.

libgalois/include/katana/PropertyGraph.h
witchel: I really am interested if there is a better way to do what I want without `g` and `const_g`.

I included the second question because it seemed like the kind of thing you might know or have an opinion on and I didn't want to make you read #1479.

ddn0: `g` and `const_g` point to the same object? and `const_cast` doesn't work or a function that wraps `const_cast`?

witchel: They point to the same object.
Can you explain your question about `const_cast`?  The problem comes from creating a `PropertyView `and a `const PropertyView`.

ddn0: I see.

One pattern would be to use a template that captures the const or lack of const-ness, e.g.,

```cpp
using ConstView = View<const PropertyGraph*>;
using MutableView = View<PropertyGraph*>;
```

Though given that the const view actually has ~less~ fewer functions (the nullptr member functions), I'd suggest just making two classes and skip the templates, e.g.,
```cpp
class ReadOnlyView {

};

class MutableView: public ReadOnlyView {
public:
  Result<void> AddProperties(....) {....}
};

ReadOnlyView edge_property_view() const {....}
MutableView edge_property_view() {...}
```

whether you want to make `Mutable` and `ReadOnlyView` virtual base classes and have concrete classes for nodes and edges or want to continue using member function pointers is up to you.

I originally used member function pointers because I saw (and still see) PropertyGraph as a closed ecosystem, so how we implemented the required (limited) polymorphism was an implementation detail, so I picked what I thought was the more compact way. Virtual functions are fine too.

witchel: I have taken your advice and made two explicitly named structs.  I included the read-only version in the mutable version because it led to less duplication.  

I introduced explicit names for the "constructors" instead of const/non-const because I found myself casting a `DistPropertyGraph `pointer to a const pointer and invoking `edge_property_view()` which seemed like an unnecessarily obscure mechanism.

witchel: Thank you!

PR #352, opened 2021-07-07
Fix regression in libtsuba caused by removing code which was marked a…

yshalabi:
…s deprecated but with active users.

Add unit tests to prevent future silent regressions.

### DIFF
--- a/libtsuba/CMakeLists.txt
+++ b/libtsuba/CMakeLists.txt
@@ -38,6 +38,10 @@ set_common_katana_library_options(tsuba ALWAYS_SHARED)
 
 target_link_libraries(tsuba PUBLIC katana_support)
 
+if(KATANA_IS_MAIN_PROJECT AND BUILD_TESTING)
+  add_subdirectory(test)
+endif()
+
 install(
   DIRECTORY include/
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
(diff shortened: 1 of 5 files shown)


witchel:
While I appreciate the enthusiasm, it is not clear to me that instating support for parquet-format metadata is the right thing to do.  I want to drop support for it.  But doing so means warning people about it and providing migration tools, if necessary.  

Personally, I can stop using the RDGs on S3 that have parquet metadata.  Is anyone else using the parquet format?  If not, we should locate the parquet-based RDGs and rename their directories to something like rmat15-OLDUNSUPPORTEDFORMAT and see if anyone gets upset.  If no one says anything for two weeks or so, we can delete them.

yshalabi:
That was a quick comment, I made this PR to test whether everything is setup properly for CI (its not, need to move these tests over to enterprise).

I removed code which was marked deprecated in the prior PR and the changes caused no CI failures.

No CI tests failed. But, you were not happy because your non-CI tests were failing. Okay, that means my PR introduced a regression. NBD --  fixing is simple. Restore the functionality that was prematurely removed, add CI tests to prevent another silent regression in the future. We can remove the tests in the future.

The solution you suggest does not seem to be very practical. You want me to crawl S3 and GS and our dev file systems and detect old RDGs with parquet metadata files, move them, and wait for people to get upset?

witchel:
I have moved this conversation to #kgfs on slack.

PR #353, opened 2021-07-07
restore parquet based RDGPartHeader

yshalabi:
While deprecated, parquet encoded partition meta-data files can technically still be used to load RDGs. This PR restores this functionality which was removed from #271.


tylershunt:
I think this is no longer active; closing. Please re-open if I'm wrong!

PR #355, opened 2021-07-07
[KAT-862] Upgrade to arrow 4.0 and move arrow dependency to APT instead of conan for non-conda builds

arthurp:
With: https://github.com/KatanaGraph/katana-enterprise/pull/1496

Adds some nasty hacks for Ubuntu 16.04.

### DIFF
--- a/python/katana_setup.py
+++ b/python/katana_setup.py
@@ -351,7 +351,7 @@ def cythonize(module_list, *, source_root, **kwargs):
     require_python_module("packaging")
     require_python_module("Cython", "0.29.12")
     require_python_module("numpy", "1.10")
-    require_python_module("pyarrow", "1.0", "3.0.dev")
+    require_python_module("pyarrow", "4.0", "5.0.dev")
 
     import Cython.Build
     import numpy
--- a/python/katana_setup.py
+++ b/python/katana_setup.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import traceback
 from pathlib import Path
 from typing import Optional, Set
 
@@ -349,9 +350,22 @@ def _build_cython_extensions(pyx_files, source_root_name, extension_options):
 def cythonize(module_list, *, source_root, **kwargs):
     # TODO(amp): Dependencies are yet again repeated here. This needs to come from a central deps list.
     require_python_module("packaging")
-    require_python_module("Cython", "0.29.12")
     require_python_module("numpy", "1.10")
-    require_python_module("pyarrow", "4.0", "5.0.dev")
+    try:
+        require_python_module("Cython", "0.29.12")
+        require_python_module("pyarrow", "4.0", "5.0.dev")
+        build_extensions = True
+    except RequirementError:
+        print(
+            "WARNING: Building Katana Python without extensions! The following features will not work: katana.local, "
+            "katana.distributed, katana parallel loops. The following features will work: katana.client and "
+            "katana.remote (with limitations).",
+            file=sys.stderr,
+        )
+        build_extensions = False
+
+    if not build_extensions:
+        return []
 
     import Cython.Build
     import numpy
--- a/scripts/setup_ubuntu.sh
+++ b/scripts/setup_ubuntu.sh
@@ -2,6 +2,19 @@
 
 set -xeuo pipefail
 
+#
+# Install bootstrap requirements
+#
+
+apt update --quiet
+apt install -yq lsb-release curl software-properties-common
+
+NO_UPDATE=""
+if apt-add-repository --help | grep -q -e "--no-update"; then
+  NO_UPDATE="--no-update"
+fi
+
+
 RELEASE=$(lsb_release --codename | awk '{print $2}')
 
 SETUP_TOOLCHAIN_VARIANTS=${SETUP_TOOLCHAIN_VARIANTS:-yes}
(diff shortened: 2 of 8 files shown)


scripts/setup_ubuntu.sh
```diff
@@ -2,6 +2,19 @@
...
+# Install bootstrap requirements
+#
+
+apt update --quiet
+apt install -yq lsb-release curl software-properties-common
+
+NO_UPDATE=""
+if apt-add-repository  --help | grep -q -e "--no-update"; then
+  NO_UPDATE="--no-update"
+fi
```
danielmawhirter: I'm guessing this is the 16.04 hack you mentioned. Possibly ignorant question, would this be better gated with `$RELEASE = 'bionic'` or something similar?

danielmawhirter (approved):
Looks good to me. When we're using our own runners for enterprise CI, will we need any additional steps to nuke the old version of arrow?

scripts/setup_ubuntu.sh
arthurp: This actually isn't the hack. the hack is the ability to build python without it's extension modules and to load the `katana` module without those extensions (Handling `RequirementError` and `ModuleNotFoundError` resp).

I did a grep here because I was lazy more than anything, and apt didn't list when `--no-update` was added, so I didn't know which version to check for. So I just check for the feature. ;-)

python/katana_setup.py
```diff
@@ -349,9 +350,22 @@ def _build_cython_extensions(pyx_files, source_root_name, extension_options):
...
     require_python_module("numpy", "1.10")
-    require_python_module("pyarrow", "1.0", "3.0.dev")
+    try:
+        require_python_module("Cython", "0.29.12")
+        require_python_module("pyarrow", "4.0", "5.0.dev")
+        build_extensions = True
+    except RequirementError:
+        print(
+            "WARNING: Building Katana Python without extensions! The following features will not work: katana.local, "
+            "katana.distributed, katana parallel loops. The following features will work: katana_enterprise.client and "
```
gliga: katana_enterprise keeps appearing at more places than desired. As planned, it would be nice to hide katana_enterprise behind katana package

scripts/setup_ubuntu.sh
```diff
@@ -2,6 +2,19 @@
...
 
+#
+# Install bootstrap requirements
+#
+
+apt update --quiet
+apt install -yq lsb-release curl software-properties-common
+
+NO_UPDATE=""
+if apt-add-repository  --help | grep -q -e "--no-update"; then
```
gliga: minor: two spaces before --help

python/katana_setup.py
arthurp: I included it here because many people are still not aware of the hiding of it and this warning will be temporary. But you are right. I will change it.

scripts/setup_ubuntu.sh
```diff
@@ -108,3 +117,9 @@ if [[ -n "${SETUP_TOOLCHAIN_VARIANTS}" ]]; then
     apt install clang++-10
   fi
 fi
+
+# Install pip libraries that depend on debian packages
+
+# No binary is required to cause the pip package to use the debian package's native binaries.
```
gliga: minor: No binary =>--no-binary

scripts/setup_ubuntu.sh
```diff
@@ -85,15 +91,18 @@ run_as_original_user pip3 install --upgrade "pip$PIP_VERSION" "setuptools$SETUPT
...
+DEVELOPER_TOOLS="clang-format-10 clang-tidy-10 doxygen graphviz ccache cmake shellcheck pkg-config"
 # github actions require a more recent git
 GIT=git
 # Library dependencies
 #
 # Install llvm via apt instead of as a conan package because existing
 # conan packages do yet enable RTTI, which is required for boost
 # serialization.
-LIBRARIES="libxml2-dev llvm-10-dev"
+LIBRARIES="libxml2-dev llvm-10-dev libarrow-dev=4.0.1-1 libarrow-python-dev=4.0.1-1 libparquet-dev=4.0.1-1
```
gliga, line 107: minor: we have much longer lines, why break here

python/katana_setup.py
```diff
@@ -349,9 +350,22 @@ def _build_cython_extensions(pyx_files, source_root_name, extension_options):
...
     require_python_module("packaging")
-    require_python_module("Cython", "0.29.12")
     require_python_module("numpy", "1.10")
-    require_python_module("pyarrow", "1.0", "3.0.dev")
+    try:
+        require_python_module("Cython", "0.29.12")
+        require_python_module("pyarrow", "4.0", "5.0.dev")
+        build_extensions = True
+    except RequirementError:
+        print(
```
gliga, line 359: used "import warnings" in init and print here

gliga (approved):
LT

scripts/setup_ubuntu.sh
arthurp: I'll break the other lines. :-)

python/katana_setup.py
arthurp: I did this because `warnings` is for Python warnings (like exceptions, but non-fatal) where as this is a build warning really. That said I think you might be right. However I'm gonna leave this as is. This entire file uses `print` so I'll stick to that in this PR.

PR #362, opened 2021-07-10
python: Fix warnings in generated cython code.

arthurp:
There are still -Wmaybe-uninitialized warnings, but those are
benign and hard to avoid. They are caused because in the exception
case some functions return an uninitialized structure. However
these functions are annotated with except * so the caller always
checks for exeptions, so it will never use the return value if
an exception was thrown.

We could disable -Wmaybe-uninitialized entirely, but it is possible
to write incorrect code in Cython that has initialization issues.
So having this warning can be useful.


gliga (approved):

gliga:
Looks good

PR #372, opened 2021-07-15
Use viewtype when constructing partition header file name

yshalabi:


libtsuba/src/RDGManifest.cpp
```diff
@@ -263,15 +217,16 @@ RDGManifest::FileNames() {
...
+    fnames.emplace(PartitionFileName(view_specifier(), i, version()));
 
-    auto header_res =
-        RDGPartHeader::Make(PartitionFileName(dir(), i, version()));
+    auto header_uri = KATANA_CHECKED(katana::Uri::Make(fmt::format(
+        "{}/{}", dir(), PartitionFileName(view_specifier(), i, version()))));
+    auto header_res = RDGPartHeader::Make(header_uri);
 
     if (!header_res) {
       KATANA_LOG_DEBUG(
```
scober, line 227: If this is an error, should we be returning `KATANA_ERROR` instead of a doing a debug log?

scober (approved):
LGTM

libtsuba/src/RDGManifest.cpp
yshalabi: I think either approach is fine. I would prefer to simply log the error as the user will presumably get an error if he later attempts a RDGPartHeader::Make.

PR #379, opened 2021-07-19
Add a runtime parameter to PODVector and DynamicBitset - whether to pin memory

srogatch:
https://github.com/KatanaGraph/katana-enterprise/pull/1551 depends on this PR.
JIRA issue: https://katanagraph.atlassian.net/browse/KAT-908

### DIFF
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -49,6 +49,7 @@ class KATANA_EXPORT DynamicBitset {
 public:
   static constexpr uint32_t kNumBitsInUint64 = sizeof(uint64_t) * CHAR_BIT;
 
+  //TODO (serge): change to a polymorphic allocator to switch between pinned and swappable memory
   explicit DynamicBitset(const MemoryPinType mpt = MemoryPinType::Swappable)
       : bitvec_(mpt){};
 
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -18,9 +18,9 @@
  */
 
 /**
- * @file katana/DynamicBitsetCommon.h
+ * @file katana/DynamicBitset.h
  *
- * Contains the DynamicBitsetCommon class and most of its implementation.
+ * Contains the DynamicBitset class and most of its implementation.
  */
 
 #ifndef KATANA_LIBGALOIS_KATANA_DYNAMICBITSET_H_
@@ -39,36 +39,22 @@
 #include "katana/config.h"
 
 namespace katana {
-
-template <bool for_gpu>
-class KATANA_EXPORT DynamicBitsetCommon;
-
-template <typename integer, bool for_gpu>
-struct KATANA_EXPORT DynamicBitsetHelper {
-  static std::vector<integer> GetOffsets(
-      const DynamicBitsetCommon<for_gpu>& bitset);
-
-  static void AppendOffsets(
-      const DynamicBitsetCommon<for_gpu>& bitset, std::vector<integer>* vec);
-};
-
 /**
  * Concurrent dynamically allocated bitset
  **/
-template <bool for_gpu = false>
-class KATANA_EXPORT DynamicBitsetCommon {
-  katana::PODVector<katana::CopyableAtomic<uint64_t>, for_gpu> bitvec_{};
+class KATANA_EXPORT DynamicBitset {
+  katana::PODVector<katana::CopyableAtomic<uint64_t>> bitvec_;
   size_t num_bits_{0};
 
 public:
   static constexpr uint32_t kNumBitsInUint64 = sizeof(uint64_t) * CHAR_BIT;
 
-  DynamicBitsetCommon() = default;
+  explicit DynamicBitset(const bool pinned = false) : bitvec_(pinned){};
 
-  DynamicBitsetCommon(DynamicBitsetCommon&& bitset)
+  DynamicBitset(DynamicBitset&& bitset)
       : bitvec_(std::move(bitset.bitvec_)), num_bits_(bitset.num_bits_) {}
 
-  DynamicBitsetCommon& operator=(DynamicBitsetCommon&& bitset) {
+  DynamicBitset& operator=(DynamicBitset&& bitset) {
     bitvec_ = std::move(bitset.bitvec_);
     num_bits_ = bitset.num_bits_;
     bitset.num_bits_ = 0;
@@ -277,7 +263,7 @@ public:
   }
 
   // assumes bit_vector is not updated (set) in parallel
-  void bitwise_or(const DynamicBitsetCommon& other);
+  void bitwise_or(const DynamicBitset& other);
 
   // assumes bit_vector is not updated (set) in parallel
   void bitwise_not();
@@ -289,7 +275,7 @@ public:
    *
    * @param other Other bitset to do bitwise and with
    */
-  void bitwise_and(const DynamicBitsetCommon& other);
+  void bitwise_and(const DynamicBitset& other);
 
   /**
    * Does an IN-PLACE bitwise and of 2 passed in bitsets and saves to this
@@ -298,15 +284,14 @@ public:
    * @param other1 Bitset to and with other 2
    * @param other2 Bitset to and with other 1
    */
-  void bitwise_and(
-      const DynamicBitsetCommon& other1, const DynamicBitsetCommon& other2);
+  void bitwise_and(const DynamicBitset& other1, const DynamicBitset& other2);
 
   /**
    * Does an IN-PLACE bitwise xor of this bitset and another bitset
    *
    * @param other Other bitset to do bitwise xor with
    */
-  void bitwise_xor(const DynamicBitsetCommon& other);
+  void bitwise_xor(const DynamicBitset& other);
 
   /**
    * Does an IN-PLACE bitwise and of 2 passed in bitsets and saves to this
@@ -315,8 +300,7 @@ public:
    * @param other1 Bitset to xor with other 2
    * @param other2 Bitset to xor with other 1
    */
-  void bitwise_xor(
-      const DynamicBitsetCommon& other1, const DynamicBitsetCommon& other2);
+  void bitwise_xor(const DynamicBitset& other1, const DynamicBitset& other2);
 
   /**
    * Count how many bits are set in the bitset
@@ -333,9 +317,7 @@ public:
    * @returns vector with offsets into set bits
    */
   template <typename integer>
-  std::vector<integer> GetOffsets() const {
-    return DynamicBitsetHelper<integer, for_gpu>::GetOffsets(*this);
-  }
+  std::vector<integer> GetOffsets() const;
 
   /**
    * Given a vector, appends the set bits in this bitset in order
@@ -343,16 +325,26 @@ public:
    * Do NOT call in a parallel region as it uses katana::on_each.
    */
   template <typename integer>
-  void AppendOffsets(std::vector<integer>* vec) const {
-    return DynamicBitsetHelper<integer, for_gpu>::AppendOffsets(*this, vec);
-  }
+  void AppendOffsets(std::vector<integer>* vec) const;
 
   //! this is defined to
   using tt_is_copyable = int;
 };
 
+template <>
+std::vector<uint32_t> DynamicBitset::GetOffsets() const;
+
+template <>
+std::vector<uint64_t> DynamicBitset::GetOffsets() const;
+
+template <>
+void DynamicBitset::AppendOffsets(std::vector<uint32_t>* offsets) const;
+
+template <>
+void DynamicBitset::AppendOffsets(std::vector<uint64_t>* offsets) const;
+
 //! An empty bitset object; used mainly by InvalidBitsetFn
-extern katana::DynamicBitsetCommon<false> EmptyBitset;
+extern katana::DynamicBitset EmptyBitset;
 
 //! A structure representing an empty bitset.
 struct KATANA_EXPORT InvalidBitsetFn {
@@ -360,14 +352,10 @@ struct KATANA_EXPORT InvalidBitsetFn {
   static constexpr bool is_valid() { return false; }
 
   //! Returns the empty bitset
-  static katana::DynamicBitsetCommon<false>& get() { return EmptyBitset; }
+  static katana::DynamicBitset& get() { return EmptyBitset; }
 
   //! No-op since it's an empty bitset
   static void reset_range(size_t, size_t) {}
 };
-
-using DynamicBitset = DynamicBitsetCommon<false>;
-using DynamicBitsetForGpu = DynamicBitsetCommon<true>;
-
 }  // namespace katana
 #endif
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -18,9 +18,9 @@
  */
 
 /**
- * @file katana/DynamicBitset.h
+ * @file katana/DynamicBitsetCommon.h
  *
- * Contains the DynamicBitset class and most of its implementation.
+ * Contains the DynamicBitsetCommon class and most of its implementation.
  */
 
 #ifndef KATANA_LIBGALOIS_KATANA_DYNAMICBITSET_H_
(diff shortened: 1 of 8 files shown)


libgalois/include/katana/PODVector.h
```diff
@@ -34,6 +36,8 @@
 
 namespace katana {
 
+enum class MemoryPinType : uint8_t { Usual = 0, Pinned = 1 };
```
arthurp: I don't like `Usual`. What about `NotPinned`?

Also let's not limit the size unless we actually expect it to matter. This is a premature optimization that can result in ABI compatibility issues.

libgalois/include/katana/PODVector.h
```diff
@@ -55,6 +59,7 @@ class PODVector {
   _Tp* data_;
   size_t capacity_;
   size_t size_;
+  bool pinned_;
```
arthurp: Why not `MemoryPinType`?

libgalois/include/katana/PODVector.h
```diff
@@ -145,15 +177,25 @@ class PODVector {
...
+      }
       capacity_ = std::max(size_, kMinNonZeroCapacity);
-      _Tp* new_data_ = static_cast<_Tp*>(
-          realloc(reinterpret_cast<void*>(data_), capacity_ * sizeof(_Tp)));
+      const size_t new_bytes = capacity_ * sizeof(_Tp);
+      _Tp* new_data_ =
+          static_cast<_Tp*>(realloc(reinterpret_cast<void*>(data_), new_bytes));
       KATANA_LOG_DEBUG_ASSERT(new_data_);
+      if (pinned_) {
+        mlock(new_data_, new_bytes);
```
arthurp: Please handle errors in both `mlock` and `munlock`, throughout this code. It might be reasonable to just print a warning (to alert people of the performance issue), but the error shouldn't just be ignored. We may also need to prevent `munlock` from being called if `mlock` fails.

libgalois/include/katana/PODVector.h
```diff
@@ -163,8 +205,13 @@ class PODVector {
...
 
-    // When reallocing, don't pay for elements greater than size_
-    shrink_to_fit();
+    // The price of unpinning&pinning again exceeds the savings below
+    if (!pinned_) {
+      // When reallocing, don't pay for elements greater than size_
+      shrink_to_fit();
+    }
+
+    [[maybe_unused]] const size_t old_bytes = capacity_ * sizeof(_Tp);
```
arthurp: This is always used.

libgalois/include/katana/PODVector.h
```diff
@@ -173,9 +220,18 @@ class PODVector {
...
-        realloc(reinterpret_cast<void*>(data_), capacity_ * sizeof(_Tp)));
+    if (pinned_) {
+      if (data_ != nullptr) {
+        munlock(data_, old_bytes);
+      }
+    }
+    const size_t new_bytes = capacity_ * sizeof(_Tp);
+    _Tp* new_data_ =
+        static_cast<_Tp*>(realloc(reinterpret_cast<void*>(data_), new_bytes));
     KATANA_LOG_DEBUG_ASSERT(new_data_);
```
arthurp: I think these should be `KATANA_LOG_ASSERT` the reason is that we are already allocating which is slow and catching allocation failures quickly in some form in release builds is important. Since it can happen if the user doesn't have enough RAM in their system.

libsupport/include/katana/CpuGpuSwitch.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#ifdef KATANA_ENABLE_GPU
+#define KATANA_GPU_ENABLED true
+#define KATANA_MEMORY_PIN_TYPE (katana::MemoryPinType::Pinned)
+#else
+#define KATANA_GPU_ENABLED false
+#define KATANA_MEMORY_PIN_TYPE (katana::MemoryPinType::Usual)
+#endif  // KATANA_ENABLE_GPU
```
arthurp: This isn't used yet. I would rather add this file in a later PR that actually uses it.

libgalois/include/katana/PODVector.h
```diff
@@ -100,19 +124,27 @@ class PODVector {
...
       free(data_);
     }
     data_ = v.data_;
     capacity_ = v.capacity_;
     size_ = v.size_;
+    pinned_ = v.pinned_;
     v.data_ = NULL;
     v.capacity_ = 0;
     v.size_ = 0;
+    v.pinned_ = false;
```
arthurp: This duplicates the move constructor (and the default constructor if that exists). Maybe factor it out into a private `Clear` method.

libgalois/include/katana/PODVector.h
```diff
@@ -71,25 +76,44 @@ class PODVector {
   typedef std::reverse_iterator<iterator> reverse_iterator;
   typedef std::reverse_iterator<const_iterator> const_reverse_iterator;
 
-  PODVector() : data_(NULL), capacity_(0), size_(0) {}
```
arthurp, line 74: You are removing the default constructor. I think the compiler might be generating one, but it might be better to be explicit. I think (though I'm not sure) that an explicit constructor which can be called with zero-arguments via a default parameter is *different* from a default constructor (not marked explicit and with zero-arguments natively).

libgalois/include/katana/DynamicBitset.h
```diff
@@ -43,13 +43,14 @@ namespace katana {
...
-  katana::PODVector<katana::CopyableAtomic<uint64_t>> bitvec_{};
+  katana::PODVector<katana::CopyableAtomic<uint64_t>> bitvec_;
   size_t num_bits_{0};
 
 public:
   static constexpr uint32_t kNumBitsInUint64 = sizeof(uint64_t) * CHAR_BIT;
 
-  DynamicBitset() = default;
+  explicit DynamicBitset(const MemoryPinType mpt = MemoryPinType::Usual)
+      : bitvec_(mpt){};
```
arthurp: Same default constructor concern.

arthurp (changes requested):
Good over all. I'd like to clean some things up though.

amberhassaan (changes requested):
Hi @srogatch : 
I think we can solve this problem more cleanly with "allocators".  Just like `std::vector<T, alloctor>`, we should allow `PODVector` and `DynmaicBitSet` to take `PinnedMemoryAllocator` (and other allocators, like GPUallocator if it makes sense). This will remove need for `if(pinned_)...` type code and contain the complexity. 
The approach below is not scalable complexity wise and we should be investing into better memory management abstractions.

libgalois/include/katana/PODVector.h
amberhassaan: We can enhance it further and include GPU memory in MemoryTypes and I agree with Arthur that `Usual` should be `NonPinned`

srogatch:
> I think we can solve this problem more cleanly with "allocators". Just like `std::vector<T, alloctor>`, we should allow `PODVector` and `DynmaicBitSet` to take `PinnedMemoryAllocator` (and other allocators, like GPUallocator if it makes sense). This will remove need for `if(pinned_)...` type code and contain the complexity.
> The approach below is not scalable complexity wise and we should be investing into better memory management abstractions.

@amberhassaan , introducing an allocator is really a lot of work and changes to the code. `DynamicBitset` is not a template and it has functions with external linkage. The latter is solvable by providing externally-linked specializations, however, the former requires changes all over the code adding something like `<>` if we provide a default template parameter for the allocator. Furthermore, I've already tried to make the `pinned` flag a template parameter to `DynamicBitset` and `PODVector` and got stuck. We would need to make a lot of other functions templates, many of them are externally linked, in order to accept both `DynamicBitset<true>` and `DynamicBitset<false>` (or the equivalents with allocators) so to be able to copy data between them. The same for usages of `PODVector`: we want to enable copying between pinned and non-pinned `PODVector`s and we want functions that work with both. Finally, we want to switch between pinned and non-pinned vectors/bitsets at runtime, while a template parameter would force us to make this decision at compile-time. Perhaps I cannot foresee every problem with introducing the allocator, but it should be already clear that this change is very intrusive and requires many changes to the external code.

amberhassaan:
> > I think we can solve this problem more cleanly with "allocators". Just like `std::vector<T, alloctor>`, we should allow `PODVector` and `DynmaicBitSet` to take `PinnedMemoryAllocator` (and other allocators, like GPUallocator if it makes sense). This will remove need for `if(pinned_)...` type code and contain the complexity.
> > The approach below is not scalable complexity wise and we should be investing into better memory management abstractions.
> 
> @amberhassaan , introducing an allocator is really a lot of work and changes to the code. `DynamicBitset` is not a template and it has functions with external linkage. The latter is solvable by providing externally-linked specializations, however, the former requires changes all over the code adding something like `<>` if we provide a default template parameter for the allocator. Furthermore, I've already tried to make the `pinned` flag a template parameter to `DynamicBitset` and `PODVector` and got stuck. We would need to make a lot of other functions templates, many of them are externally linked, in order to accept both `DynamicBitset<true>` and `DynamicBitset<false>` (or the equivalents with allocators) so to be able to copy data between them. The same for usages of `PODVector`: we want to enable copying between pinned and non-pinned `PODVector`s and we want functions that work with both. Finally, we want to switch between pinned and non-pinned vectors/bitsets at runtime, while a template parameter would force us to make this decision at compile-time. Perhaps I cannot foresee every problem with introducing the allocator, but it should be already clear that this change is very intrusive and requires many changes to the external code.

Hi @srogatch , there's 1) misunderstanding 2) strong disagreement (on my side). 
Miunderstanding: I'm not asking for a `bool isPinned` as a template flag to DynamicBitSet or PODVector. Rather, I'm asking for a separation of concerns. Allocators deal with how to allocated memory. Pinned or not-pinned, CPU or GPU. PODVector or DynamicBItset (or any data structures) shouldn't have complex logic and flags of this sort. Please take a look at STL containers, all of which allow passing a custom allocator object. 
AFA the change due to adding an allocator as template parameter goes, we can limit that with typedefs (or type aliases), e.g. 
```
template <typename T, typename AllocatorType>
class PODVectorImplementation {
...
};

template <typename T>
using PODVector = PODVectorImplementation<T, SomeDefaultAllocator>;

template <typename T>
using PODVectorPinned = PODVectorImplementation<T, PinnedAllocator>;
```
And, you can add copy constructors between PODVector and PODVector Pinned. STL containers like `std::vector` already do this.

srogatch:
@amberhassaan , there is no misunderstanding on my side: I understood about the allocator, but it's a huge amount of (re)work. However, it seems you missed my point and I am not sure what to do except to ask you to re-read my comment. Whether it's allocator or a bool in a template parameter is irrelevant. The problem is that there is a huge amount of conversion in the code between pinned and non-pinned `PODVector` and `DynamicBitset`, and if they are templated, the conversion would be problematic. The details are in my previous comment. I strongly disagree with the allocator approach, and I can only suggest to create a JIRA issue so that you can look yourself into the opportunity of changing to the allocator. I tried adding a template parameter to `PODVector` and `DynamicBitset`, so I know what I am talking about. The amount of (re)work is huge.

srogatch:
Created https://katanagraph.atlassian.net/browse/KAT-949 .

libgalois/include/katana/PODVector.h
srogatch: How about `Swappable`? I try to avoid negations in the naming and the code, just think of `if(x != MemoryPinType::NotPinned)`. Pinned memory cannot go to a swap file, in contrast to swappable memory.

libsupport/include/katana/CpuGpuSwitch.h
srogatch: This is currently used in https://github.com/KatanaGraph/katana-enterprise/pull/1551 . I had to place it in `katana` rather than `katana-enterprise` repo because someone may use the constants for `DynamicBitset` and `PODVector` instances in `katana` repo too.

libgalois/include/katana/PODVector.h
srogatch: I've tested and such constructors seem to behave exactly the same. Without the no-argument constructor, the code compiles well, including the places that use `PODVector` with default construction (e.g. inside an `std::vector`). However, if I add an explicit no-argument constructor, I get compilation errors: `error: call of overloaded ‘PODVector()’ is ambiguous`. So there should be no problem here with the code as it is.

libgalois/include/katana/PODVector.h
srogatch: I've tested and such constructors seem to behave exactly the same. Without the no-argument constructor, the code compiles well, including the places that use `PODVector` with default construction (e.g. inside an `std::vector`). However, if I add an explicit no-argument constructor, I get compilation errors: `error: call of overloaded ‘PODVector()’ is ambiguous`. So there should be no problem here with the code as it is.

libgalois/include/katana/DynamicBitset.h
srogatch: Should be ok.

amberhassaan:
> @amberhassaan , there is no misunderstanding on my side: I understood about the allocator, but it's a huge amount of (re)work. However, it seems you missed my point and I am not sure what to do except to ask you to re-read my comment. Whether it's allocator or a bool in a template parameter is irrelevant. The problem is that there is a huge amount of conversion in the code between pinned and non-pinned `PODVector` and `DynamicBitset`, and if they are templated, the conversion would be problematic. The details are in my previous comment. I strongly disagree with the allocator approach, and I can only suggest to create a JIRA issue so that you can look yourself into the opportunity of changing to the allocator. I tried adding a template parameter to `PODVector` and `DynamicBitset`, so I know what I am talking about. The amount of (re)work is huge.

HI @srogatch , if you think adding a template parameter for allocator induces a lot of work, we can consider an alternative, where we define a hierarchy of allocators with virtual functions (Pinned and NonPinned allocators being derived classes). Typically, with these containers (like PODVector), allocations are not frequent. 

My main point is that we need better memory management abstractions than having flags and calling variants of malloc based on `if conditions`. Going the allocator route (or a similar abstraction), we can make it more sophisticated going forward. See https://github.com/LLNL/CHAI and https://github.com/LLNL/Umpire

srogatch:
> HI @srogatch , if you think adding a template parameter for allocator induces a lot of work, we can consider an alternative, where we define a hierarchy of allocators with virtual functions (Pinned and NonPinned allocators being derived classes). Typically, with these containers (like PODVector), allocations are not frequent.
> 
> My main point is that we need better memory management abstractions than having flags and calling variants of malloc based on `if conditions`. Going the allocator route (or a similar abstraction), we can make it more sophisticated going forward. See https://github.com/LLNL/CHAI and https://github.com/LLNL/Umpire

Hi @amberhassaan ! The polymorphism-based allocators may take a reasonable amount of effort, but I'm not sure. An investigation is needed. I would still like to keep it out of the scope of this PR and https://github.com/KatanaGraph/katana-enterprise/pull/1551 , because it's not GPU/CUDA-related and I just added pinning/unpinning of memory to the places which already use `calloc()`/`malloc()`/`realloc()`/`free()` or `new`/`delete`. I also fixed obvious memory leaks following the rule to improve along the way. But RAII and Allocators would take more work time than pinning/unpinning itself. And no customer pays for code quality: they pay for correctness, features, and performance. That is why we should keep the quality improvement efforts adequate to the number of changes we make for features, performance, and bug-fixing. I'm quite new to the project, so I could be most affected by code quality, but I wouldn't say that the code quality is so bad that we have to spend 3x time on technical debt compared to feature/performance/bug-fixing.

amberhassaan:
@srogatch : We shouldn't be cutting corners on code quality because it always comes back to bite you. The cost and effort we spend on it pays for it self through long term maintainability and robustness (avoid and catching bugs early). Bad code, weak abstractions affect every developer, so it's men * hours.

amberhassaan (approved):

amberhassaan:
I'm OK with adding TODOs along the lines of needing allocators.

srogatch:
Thanks, @amberhassaan . I've just added TODOs, also to https://github.com/KatanaGraph/katana-enterprise/pull/1551 . The RAII task seems to be the next on my list: https://katanagraph.atlassian.net/browse/KAT-952 .

arthurp:
> I wouldn't say that the code quality is so bad that we have to spend 3x time on technical debt compared to feature/performance/bug-fixing.

The CUDA code in this code base really is that bad. I agree that such improvements should be a separate PR. But I do not agree that we should limit code quality improvements so we can focus on features/performance/whatever. These code quality issues will bite us. And it may not take long as @amberhassaan said. Please try to find some time to do these refactors. They are important.

arthurp:
As a side note, I would MUCH prefer a dynamic polymorphic allocator to a template parameter. In our use cases allocation is rare and fairly slow, so the cost of a virtual function call will be low. And using a dynamic allocator parameter will make integrating with Python and potentially even the allocators of other python libraries much easier.

libsupport/include/katana/CpuGpuSwitch.h
```diff
@@ -0,0 +1,17 @@
...
+enum class MemoryPinType { Swappable = 0, Pinned = 1 };
+
+#ifdef KATANA_ENABLE_GPU
+static constexpr const bool kGpuEnabled = true;
+static constexpr const MemoryPinType kUseMemoryPinType = MemoryPinType::Pinned;
+#else
+static constexpr const bool kGpuEnabled = false;
+static constexpr const MemoryPinType kUseMemoryPinType =
+    MemoryPinType::Swappable;
+#endif  // KATANA_ENABLE_GPU
```
arthurp: Responded at: https://github.com/KatanaGraph/katana/pull/379#discussion_r673352128

In the end, I'll leave it to you. But if there is no current need for this in open, I strongly recommend not including it.

libsupport/include/katana/CpuGpuSwitch.h
arthurp: I don't really understand this argument. These values only have meaning for GPU enabled builds, and they should really be dynamic parameters or allocators anyway not build parameters, so hopefully we will be removing them in a refactor soon. So I see no reason to allow people to use them in more places that we will have to fix later.

arthurp (approved):
Seems OK. Please address my last comment before you merge. And we'll talk about refactors soon.

srogatch:
> Seems OK. Please address my last comment before you merge. And we'll talk about refactors soon.

Thanks. I've started work on the polymorphic pinned allocator: https://github.com/KatanaGraph/katana/pull/383 , https://github.com/KatanaGraph/katana-enterprise/pull/1562 .

I'll now switch to this PR to finalize.

PR #390, opened 2021-07-22
Add SIGPIPE handler

ddn0:
I debugged some SIGPIPE errors that eventually led to some configuration
options in libcurl to ignore SIGPIPE. While it is possible to enable
such options in each dependency we use if they are available, signal
handling is not reentrant or encapsulated, and it is traditionally the
responsibility of application program and not libraries to set them up.

Instead, refactor backtrace support to be the place for general signal
handling and add a "log and ignore" handler for SIGPIPE, which is a
common option for long-running programs.


libsupport/src/Signals.cpp
```diff
@@ -0,0 +1,113 @@
...
+/// process.
+///
+/// It also installs a signal handler for SIGPIPE that prints a backtrace once
+/// but otherwise ignores SIGPIPE. This is useful for SIGPIPE in particular
+/// because the default handler kills the process, but it is usually okay to
+/// ignore it.
+class SignalHandling {
+public:
+  SignalHandling() {
+    bool failed{};
```
witchel: Nit: Can we just have `failed` and `failed_` or `loaded` and `loaded_`?

witchel (approved):
Looks great.  Nice job on the SIGPIPE sleuthing.

PR #391, opened 2021-07-23
Does overread, need future optimization

nirjharm:


libtsuba/src/ParquetReader.cpp
```diff
@@ -322,7 +316,13 @@ class BlockedParquetReader {
       }
       tables.emplace_back(arrow::Table::Make(arrow::schema(fields), columns));
     }
-    return KATANA_CHECKED(arrow::ConcatenateTables(tables));
+
+    auto concatenated_tables = arrow::ConcatenateTables(tables);
```
tylershunt: This should be something like :
```
std::shared_ptr<arrow::Table> concatenated_table = KATANA_CHECKED(arrow::ConcatenateTables(tables));
```
Then just drop the `ValueOrDie`'s below. KATANA_CHECKED will check for an arrow error, translate it into one of our errors and return it if it finds one.

libtsuba/src/ParquetReader.cpp
```diff
@@ -322,7 +316,13 @@ class BlockedParquetReader {
...
+    if (slice) {
+      concatenated_tables.ValueOrDie() =
+          concatenated_tables.ValueOrDie()->Slice(slice->offset, slice->length);
+    }
+    return KATANA_CHECKED(concatenated_tables);
```
tylershunt: then if you do the above this is just `return concatenated_table;`

tylershunt (commented):
Looks good! Please just fix the error handling and this should be good to go.

tylershunt (approved):
LGTM

PR #393, opened 2021-07-24
Memoize arrow type checking and casting

roshandathathri:
Arrow type equality checks and casts to static type
are very expensive when done millions of times
(calling this for each vertex for example).
Avoid doing this in such tight loops.
Instead, memoize them once and use it many times.

### DIFF
--- a/libgalois/src/PropertyGraph.cpp
+++ b/libgalois/src/PropertyGraph.cpp
@@ -137,6 +137,15 @@ WriteTopology(const katana::GraphTopology& topology) {
   return std::unique_ptr<tsuba::FileFrame>(std::move(ff));
 }
 
+template <typename ArrowType>
+struct PropertyColumn {
+  int field_index;
+  std::shared_ptr<ArrowType> array;
+
+  PropertyColumn(int i, std::shared_ptr<ArrowType>& a)
+      : field_index(i), array(a) {}
+};
+
 /// Assumes all boolean or uint8 properties are types
 katana::Result<katana::NUMAArray<katana::PropertyGraph::TypeSetID>>
 GetTypeSetIDsFromProperties(
@@ -158,6 +167,10 @@ GetTypeSetIDsFromProperties(
 
   // collect the list of types
   std::vector<int> type_field_indices;
+  using BoolPropertyColumn = PropertyColumn<arrow::BooleanArray>;
+  std::vector<BoolPropertyColumn> bool_properties;
+  using UInt8PropertyColumn = PropertyColumn<arrow::UInt8Array>;
+  std::vector<UInt8PropertyColumn> uint8_properties;
   const std::shared_ptr<arrow::Schema>& schema = properties->schema();
   KATANA_LOG_DEBUG_ASSERT(schema->num_fields() == properties->num_columns());
   for (int i = 0, n = schema->num_fields(); i < n; i++) {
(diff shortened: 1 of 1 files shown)


libgalois/src/PropertyGraph.cpp
```diff
@@ -158,16 +158,29 @@ GetTypeSetIDsFromProperties(
 
   // collect the list of types
   std::vector<int> type_field_indices;
+  using BoolProperty = std::pair<int, std::shared_ptr<arrow::BooleanArray>>;
```
amberhassaan: For clarity, you should rename and introduce a `struct PropertyColumn { int index_; std::shared_ptr<arrow::BooleanArray> array_; };`
This will add clarity to the usage below. GSG also advises against `std::pair<>` and I hate the GSG but I agree with this clause for most cases.

amberhassaan (commented):
The other thing to think about (in the future may be) is whether we should parallelize the loops by columns instead of rows. Right now your `do_all()` goes over rows. What if we parallelize over columns. Then you don't need the frequent type checks.

libgalois/src/PropertyGraph.cpp
roshandathathri: I thought of pair vs. struct. I went with pair in this case because of the for loop iterator:
`for (auto [i, bool_property] : bool_properties)`
That's more concise code than with struct and I thought it was easier to read. But I'm not strongly in favor of it over struct.

roshandathathri:
> parallelize the loops by columns instead of rows

Yes, something to keep in mind but I don't think that's better in most cases. The number of columns are trivial compared to the number of rows, so there might not be enough parallelism. Even if there is, it can be heavily imbalanced. 

In this case, iterating over columns would require materialization of a column/array of sets, which is not a good idea.

libgalois/src/PropertyGraph.cpp
amberhassaan: while the code is concise, it's not clear with `i` is doing and why is this thing called `BoolProperty` when it's a pair.

amberhassaan:
> Yes, something to keep in mind but I don't think that's better in most cases. The number of columns are trivial compared to the number of rows, so there might not be enough parallelism. Even if there is, it can be heavily imbalanced.


Yes, I agree that load imbalance is expected so I'm not pushing for it strongly (yet).

libgalois/src/PropertyGraph.cpp
roshandathathri: Name would be the same whether it's a struct or a pair. What would you want the name to be? `BoolPropertyColumn`? `Column` is superfluous IMO.

roshandathathri: I can rename `i` to `field_index`. If you strongly want a struct, I can change it too.

roshandathathri: @amberhassaan This is urgent. Please confirm/approve.

amberhassaan (approved):
I'd prefer having the struct. You can put it as a TODO if you want.

roshandathathri:
> You can put it as a TODO if you want.

Better to do now. It's simple. What name do you want? `BoolPropertyColumn`?

libgalois/src/PropertyGraph.cpp
amberhassaan: I realize that struct needs a template type for arrow array , so you can reuse the struct for both `arrow::BooleanArray` and `arrow::UInt8Array`.

roshandathathri: That's too verbose/general IMO but I won't argue about it. Will do.

amberhassaan: Too busy and pressed to argue? :)
My suggestion is to hide the struct in anonymous namespace so it doesn't escape the .cpp file.

amberhassaan (approved):
Looks good.

libgalois/src/PropertyGraph.cpp
roshandathathri: Hehe, yes. Small stuff. 

All of this is in anonymous namespace.

PR #396, opened 2021-07-25
tsuba: Copy properties during write if they are not loaded

tylershunt:
This allows the `Write` operation to work properly when properties are out of memory. Properties are copied from the old location to the new one, we save some memory by not translating them to arrow.

### DIFF
--- a/libtsuba/src/RDG.cpp
+++ b/libtsuba/src/RDG.cpp
@@ -522,7 +522,8 @@ tsuba::RDG::Store(
       handle.impl_->rdg_manifest().policy_id(), tsuba::Comm()->Num,
       core_->part_header().metadata().policy_id_, versioning_action);
   if (handle.impl_->rdg_manifest().dir() != rdg_dir_) {
-    core_->part_header().UnbindFromStorage();
+    KATANA_CHECKED(core_->part_header().ChangeStorageLocation(
+        rdg_dir_, handle.impl_->rdg_manifest().dir()));
   }
 
   auto desc_res = WriteGroup::Make();
--- a/libtsuba/src/RDGPartHeader.cpp
+++ b/libtsuba/src/RDGPartHeader.cpp
@@ -26,6 +26,18 @@ const char* kPartProperyMetaKey = "kg.v1.part_property_meta";
 
 // special partition property names
 
+katana::Result<void>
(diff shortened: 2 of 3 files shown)


libtsuba/src/RDGPartHeader.cpp
```diff
@@ -26,6 +26,18 @@ const char* kPartProperyMetaKey = "kg.v1.part_property_meta";
...
+katana::Result<void>
+CopyProperty(
+    tsuba::PropStorageInfo* prop, const katana::Uri& old_location,
+    const katana::Uri& new_location) {
+  katana::Uri old_path = old_location.Join(prop->path());
+  katana::Uri new_path = new_location.Join(prop->path());
+  tsuba::FileView fv;
+
+  KATANA_CHECKED(fv.Bind(old_path.string(), true));
+  return tsuba::FileStore(new_path.string(), fv.ptr<uint8_t>(), fv.size());
```
witchel, line 38: This is good and property-at-a-time is probably sufficient for the long term.  It seems like there is a GCS REST API to copy a storage object without reading it.
https://cloud.google.com/storage/docs/copying-renaming-moving-objects#copy-object-json

libtsuba/src/RDG.cpp
```diff
@@ -522,7 +522,8 @@ tsuba::RDG::Store(
       handle.impl_->rdg_manifest().policy_id(), tsuba::Comm()->Num,
       core_->part_header().metadata().policy_id_, versioning_action);
   if (handle.impl_->rdg_manifest().dir() != rdg_dir_) {
-    core_->part_header().UnbindFromStorage();
+    KATANA_CHECKED(core_->part_header().ChangeStorageLocation(
+        rdg_dir_, handle.impl_->rdg_manifest().dir()));
```
witchel, line 526: I prefer the new name.

witchel (approved):
Super cool.

libtsuba/src/RDGPartHeader.cpp
tylershunt: Totally; good to keep this in mind. I was worried that this wouldn't work with the way we have it implemented via the aws sdk, so opted for something that was guaranteed to work.

PR #402, opened 2021-07-28
[KAT-985] Add fake dlclose() to retain shared library symbols in sanitized programs

srogatch:


arthurp:
What was the problematic use of `dlclose`? Our code only calls it turing shutdown. Was that the problem? Or is some library using `dlclose` at other times?

I would much rather create this library and then `LD_PRELOAD` it only when needed than link it in. Overriding a libc function should be something we scope as tightly as possible.

srogatch:
> What was the problematic use of `dlclose`? Our code only calls it turing shutdown. Was that the problem? Or is some library using `dlclose` at other times?

You can read the details at https://github.com/google/sanitizers/issues/899 and https://github.com/google/sanitizers/issues/89 . In short, our code calls some other code that loads and unloads shared libraries while running the program. Because of this, call stacks for sanitizers appear as:
```
    #0 0x4de0bd in malloc (/scratch/src/katana-enterprise/build/lonestar/analytics/distributed/bfs/bfs-cli-dist+0x4de0bd)
    #1 0x7f7bdde9d3eb  (<unknown module>)
    #1 0x7f4154f0a3eb  (<unknown module>)
    #2 0x7f7bdde9f1c1  (<unknown module>)
    #2 0x7f4154f0c1c1  (<unknown module>)
    #3 0x7f7bddebc78c  (<unknown module>)
    #4 0x7f7bddf5e05a  (<unknown module>)
    #5 0x7f7bddf5ffe8  (<unknown module>)
```
So we know nothing about the origin, except that it resulted in a `malloc()`. By disabling `dlclose()` we ensure that modules, once loaded, stay in memory till the symbol resolution phase in sanitized programs.

> 
> I would much rather create this library and then `LD_PRELOAD` it only when needed than link it in. Overriding a libc function should be something we scope as tightly as possible.

I'm not linking it in. I just use `LD_PRELOAD` here: https://github.com/KatanaGraph/katana-enterprise/pull/1626/files , see `lonestar/CMakeLists.txt`.

arthurp:
> I'm not linking it in. I just use `LD_PRELOAD` here: https://github.com/KatanaGraph/katana-enterprise/pull/1626/files , see `lonestar/CMakeLists.txt`.

Quite so. :-) Misread the CMake code.

I understand the issue. I was just trying to understand WHICH code was using `dlclose`. But in the end it probably doesn't matter.

This all makes sense.

Just one request: `libsanitize` is not specific enough. Maybe `libsanitize` should be a collection of LD_PRELOADable libraries. And this one could be called `libdisable_dlclose`. Best to be really explicit.

arthurp (approved):
LGTM.

This doesn't seem to actually set `LD_PRELOAD` anywhere. Is it only needed for GPU builds? If so, a test in this repo would be nice. Just to make sure it works before people merge changes into this repo. No need to block this PR if it's hard though.

srogatch:
> This doesn't seem to actually set `LD_PRELOAD` anywhere. Is it only needed for GPU builds?

`LD_PRELOAD` is currently set in `set_common_katana_test_options()` or `set_common_katana_test_options_no_quick()` CMake functions ( see https://github.com/KatanaGraph/katana-enterprise/pull/1626/files ), which are called for most enterprise tests which are really doing something with the codebase (rather than setting up and verifying with Python the real tests). These functions are defined in the enterprise repository too. 

> If so, a test in this repo would be nice. Just to make sure it works before people merge changes into this repo. No need to block this PR if it's hard though.

Do you mean something like loading Python, doing a leak there, unloading Python, and then checking that the leak detector printed the proper call stack because we `LD_PRELOAD`ed the `libdisable_dlclose`? If so, can you point me to the part of the codebase that facilitates interop with Python?

arthurp:
> Do you mean something like loading Python, doing a leak there, unloading Python, and then checking that the leak detector printed the proper call stack because we `LD_PRELOAD`ed the `libdisable_dlclose`? If so, can you point me to the part of the codebase that facilitates interop with Python?

I wasn't thinking anything that complex. I was just thinking of a simple C application that would just make sure `dlclose` is in fact disabled. A unit test. Not an integration test.

PR #404, opened 2021-07-28
RFC: PropertyIndex

oshofmann:
This is an RFC for a PropertyIndex structure attempting to absorb some of Daniel's comments and be closer to a 'katana-like' interface. In general, Katana's types seem to be closer to their implementation, rather than trying to expose an abstract interface. This make some sense for things that are then operated by a parallel runtime.

I'd like there to be a trivial implementation of whatever we start with. This sketches what the header of a basic implementation based on std::set might look like. A more better implementation of the same external interface would have to be a separate type (or more likely would become the one true PropertyIndex).

I don't have a strong opinion about how to distinguish indexes of nodes vs indexes of edges. Here I've provided methods more similar to those on PropertyGraph, mostly because it was different than the templated version in the doc RFC.

(n.b. as you can see this code barely does anything. this is meant to be an rfc in the form of a draft pull request)


libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,88 @@
...
+  virtual edge_iterator edge_end() = 0;
+
+private:
+  std::string column_;
+};
+
+// BasicPropertyIndex provides a PropertyIndex for anything that can be used
+// in a default std::set.
+template<typename key_type>
+class KATANA_EXPORT BasicPropertyIndex : public PropertyIndex {
```
arekay: I think it would make sense to decouple the `node` and `edge` properties. A `BasicPropertyIndex` should have just one column of values over which the index is maintained. Unless i am mistaken, we do not expect the same property to exist on both the node and edges.

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,88 @@
...
+  std::string column() { return column_; }
+
+  virtual node_iterator node_begin() = 0;
+  virtual node_iterator node_end() = 0;
+
+  virtual edge_iterator edge_begin() = 0;
+  virtual edge_iterator edge_end() = 0;
+
+private:
+  std::string column_;
```
arekay, line 44: Is this the name/alias of the index or the name of the column being indexed

arekay (commented):
Do we have some clarification on how we intend to use the API. For instance, using `match (a)-[b]-(c)  where b.id = 10 return a;` versus `match (a)-[b]-(c)  where a .id = 10 return a;` The condition `.id=10` doesn't immediately identify if it is a node or edge property - the type of `a` or `b` encodes that. Do we expect the query engine to call `index.EdgeFind` or `index.NodeFind` based on the type or should the evaluation be `index.Find`?

libgalois/include/katana/PropertyIndex.h
oshofmann: I wrote this iteration this way to avoid templating altogether. Based on some further conversation with Daniel, I've returned to a model where node/edge becomes part of the templated type of the index. So you will have either PropertyIndex<Node> or PropertyIndex<Edge>

libgalois/include/katana/PropertyIndex.h
oshofmann: Name of the column being indexed.

libgalois/include/katana/PropertyIndex.h
oshofmann: Haha, doesn't work well if you try to write template arguments in html.

`PropertyIndex<Node> or PropertyIndex<Edge>`

roshandathathri:
> Do we expect the query engine to call `index.EdgeFind` or `index.NodeFind` based on the type or should the evaluation be `index.Find`?

Index should be treated similar to properties in this regard, so node indexes and edge indexes are distinct.

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,87 @@
...
+
+// PropertyIndex provides an interface similar to an ordered container
+// over a single property.
+template <typename node_or_edge>
+class KATANA_EXPORT PropertyIndex {
+public:
+  // PropertyIndex::iterator returns a sequence of node or edge ids.
+  using iterator = typename std::set<node_or_edge>::iterator;
+
+  PropertyIndex(std::string column) : column_(std::move(column)) {}
```
roshandathathri, line 29: Make this private?

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,87 @@
...
+
+  // Create an index from a column of pg.
+  static Result<std::unique_ptr<PropertyIndex<node_or_edge>>> Make(
+      PropertyGraph* pg, const std::string& column);
```
roshandathathri, line 33: Just a suggestion.

Why not just take an `arrow::Array` (or the property)? That will have the column name. It doesn't matter whether it's a node or edge property. Then we can drop the current template parameter. Instead, we can have a template parameter to denote the type of the index-value (`LocalNodeID` or `LocalEdgeID`). 

Instead of the template, I'd prefer a `MakeNodePropertyIndex` and `MakeEdgePropertyIndex`. It simplifies the code.

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace katana {
+
+// PropertyIndex provides an interface similar to an ordered container
+// over a single property.
+template <typename node_or_edge>
+class KATANA_EXPORT PropertyIndex {
+public:
+  // PropertyIndex::iterator returns a sequence of node or edge ids.
+  using iterator = typename std::set<node_or_edge>::iterator;
```
roshandathathri, line 27: `std::multi_set`?

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,87 @@
...
+      : PropertyIndex<node_or_edge>(column),
+        set_(PropertyCompare{
+            std::static_pointer_cast<ArrowArrayType>(property)}) {}
+
+  iterator begin() override { return set_.begin(); }
+  iterator end() override { return set_.end(); }
+
+  iterator Find(key_type key) { return set_.find(key); }
+  iterator LowerBound(key_type key) { return set_.lower_bound(key); }
+  iterator UpperBound(key_type key) { return set_.upper_bound(key); }
```
roshandathathri, line 66: It might be useful to add a simple wrapper that returns a range where all items have the same key.

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,87 @@
...
+
+private:
+  struct PropertyCompare {
+    bool operator()(node_or_edge a, node_or_edge b) const {
+      // Compare the value of the indexed property for a and b using std::less.
+      (void)a;
+      (void)b;
```
roshandathathri, line 73: Instead, drop the parameter name from the signature:
`    bool operator()(node_or_edge, node_or_edge) const {`

libgalois/src/PropertyIndex.cpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  return index;
+}
+
+template <typename node_or_edge, typename key_type>
+Result<void>
+BasicPropertyIndex<node_or_edge, key_type>::BuildFromProperty(
+    std::shared_ptr<arrow::ChunkedArray> property) {
+  (void)property;
```
roshandathathri, line 60: Ditto

libgalois/include/katana/PropertyIndex.h
oshofmann: I would describe the current template parameter as a parameter that denotes the type of the index-value - is there something I am misunderstanding?

Indeed, the current pass I'm taking through the implementation just passes the property as an arrow::Array. However, I was playing a lot of ugly template tricks to have a single ::Make function. I think your suggestion to just split those cases manually does simplify things.

oshofmann: Also, does an arrow::Array have metadata somewhere that specifies the column name? Can you point me to how that is accessed?

roshandathathri: The column name is not there in the arrow::Array. An additional arrow schema/type can be passed to the `Make` function for that. 

The current template parameter is specific to node or edge type. Instead, we can just treat it as a regular POD template type and say that we only support `uintt32_t` and `uint64_t` instantiations of it. This allows definitions to be present in the CPP file.

oshofmann:
Closing this RFC PR. I've sent out a complete version in https://github.com/KatanaGraph/katana/pull/416

PR #411, opened 2021-07-30
[KAT-839] Improve and clean up numba wrapper utilities.

arthurp:
Before: https://github.com/KatanaGraph/katana-enterprise/pull/1637

These changes will enable simpler use without jinja stupidity.

### DIFF
--- a/python/katana/dynamic_bitset.pyx
+++ b/python/katana/dynamic_bitset.pyx
@@ -1,3 +1,6 @@
+from libc.stdint cimport uintptr_t
+
+
 cdef class DynamicBitset:
     def __init__(self, num_bits):
         """Initializes the bitset and sets its size to the given number of bits."""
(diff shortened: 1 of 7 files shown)


python/katana/numba_support/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+def exec_in_file(filename: str, source: str, globals_dict: Optional[dict] = None, locals_dict: Optional[dict] = None):
+    source = source.strip()
+    if not filename.endswith(".py"):
+        filename += ".py"
+    if TEMPORARY_DIRECTORY:
+        filename = str(TEMPORARY_DIRECTORY / filename)
+        with open(filename, "wt", encoding="UTF-8") as fi:
+            print(source, file=fi)
+    code = compile(source, filename, "exec")
+    exec(code, globals_dict, locals_dict)
```
arthurp: I'll add documentation to this. The basic idea is that if you write out the code to a file than the numba error reporting machinery can find it and display the generated code in the error message. This helps a lot with debugging.

python/katana/dynamic_bitset.pyx
```diff
@@ -67,3 +70,65 @@ cdef class DynamicBitset:
...
+
+
+cdef uint8_t get_item(CDynamicBitset *self, uint64_t key) nogil:
+    if key < 0 or key >= self.size():
+        raise IndexError(key)
+
+    return <uint8_t>self.test(key)
+
+DynamicBitset_numba_type_wrapper.register_method(
+    "get_item",
```
gliga, line 110: minor: we could use __name__ in case IDEs are not able to connect the string to the name of a function/method.

gliga (approved):
LG!

python/katana/dynamic_bitset.pyx
arthurp: This is an interesting idea. I will explore it later. I'm not sure it will work because cdef functions in Cython are not full objects in the sense that functions are in Python.

PR #413, opened 2021-07-31
temporary fix for graphs with 0 edges

amberhassaan:
`graph-worker` balks by triggering an assert in `EdgeTypeIndex` because there are no edges.

### DIFF
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -997,13 +997,12 @@ public:
   static EdgeTypeAwareTopology MakeFromTransposeTopology(
       const PropertyGraph* pg, const EdgeTypeIndex* edge_type_index);
 
-  // TODO(amber): logic needs fixing for the case when there are nodes but no edges
-  // This will result in edge_type_index_->num_unique_types() == 0 and allocate
-  // adj_indices_ of size 0, even when graph has non-zero number of nodes.
   uint64_t num_nodes() const noexcept {
-    if (adj_indices_.size() == 0) {
-      return 0;
-    }  // corner case: graph with 0 edges
+    // corner case: graph with 0 edges
+    if (edge_type_index_->num_unique_types() == 0) {
+      KATANA_LOG_DEBUG_ASSERT(num_edges() == 0);
+      return adj_indices_.size();
+    }
     return adj_indices_.size() / edge_type_index_->num_unique_types();
   }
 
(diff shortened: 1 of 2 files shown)


libgalois/include/katana/PropertyGraph.h
```diff
@@ -999,6 +998,8 @@ class KATANA_EXPORT EdgeTypeAwareTopology : public GraphTopologyTypes {
       const PropertyGraph* pg, const EdgeTypeIndex* edge_type_index);
 
   // TODO(amber): logic needs fixing for the case when there are nodes but no edges
+  // This will result in edge_type_index_->num_unique_types() == 0 and allocate
+  // adj_indices_ of size 0, even when graph has non-zero number of nodes.
```
arthurp: So `num_nodes` will return 0 even if there are nodes if there are no edges? If so that might be worth documenting specifically, even if this is temporary. It could be pretty surprising.

arthurp (approved):

libgalois/include/katana/PropertyGraph.h
amberhassaan: Yes, this is the situation with a corner case. As I think about it, I have a better fix in mind. I'll push it in a bit. 
For my next PR, I'll pay attention to this aspect of corner cases, e.g. graph with 0 edges, or graph with 0 nodes and 0 edges.

libgalois/src/PropertyGraph.cpp
```diff
@@ -1363,6 +1363,24 @@ katana::EdgeTypeAwareTopology::AdjIndexArray
...
     const EdgeShuffleTopology& topo) noexcept {
+  if (topo.num_nodes() == 0) {
+    KATANA_LOG_VASSERT(
+        topo.num_edges() == 0, "Found graph with edges but no nodes");
+    return AdjIndexArray{};
+  }
+
+  if (edge_type_index->num_unique_types() == 0) {
+    KATANA_LOG_VASSERT(
+        topo.num_edges() == 0, "Found graph with edges but no edge types");
```
arthurp, line 1374: Not for this PR, but if this can ever happen for real graphs (maybe because all edges have the Unknown type), then this needs to be a `Result`. That said, I think this might not actually be possible for any user input since the Unknown type would always exist and be a unique type.

arthurp:
I like this new fix better!!! And this very much makes sense. Just handle the cases.

libgalois/src/PropertyGraph.cpp
amberhassaan: Yes, that shouldn't happen, because even if input has no edge types we assign `kUnknownType` to all edges. That said, I'll review the `EdgeTypeIndex` creation to see if we should put checks there as well (probably should). 

Thanks for the quick review. I messed up a bit by being too confident and merging open-repo PR. Should be resolved in a couple of hours.

roshandathathri (commented):
LGTM

PR #415, opened 2021-08-03
[KAT-838] Wrap Barrier for Numba

arthurp:


python/katana/barrier.pyx
```diff
@@ -40,6 +40,12 @@ cdef class Barrier:
         name = self.underlying().name()
         return "<" + str(name, "ascii") + " @ " + hex(<size_t>self.underlying()) + ">"
 
+    @property
```
gliga, line 43: I think I have seen `address` at other places as well.  Maybe we should consider at some point having a parent class for them all that defines `address`.

gliga (approved):

python/katana/barrier.pyx
arthurp: Good idea. Though it will just be an interface.

Actually, I'm gonna do it in a separate PR, because I want to rename `address` to `__numba_address__` or something like that so it doesn't look public.

arthurp: https://katanagraph.atlassian.net/browse/KAT-1142

PR #416, opened 2021-08-03
Add PropertyIndex for node/edge properties.

oshofmann:
A basic implementation of PropertyIndex based on std::set.

This change implements a simple PropertyIndex based on std::multiset for primitive types and strings.

Notes/questions for reviewers:
* This was slightly complicated by the need to represent a possible search key in the same type as held by the std::multiset. Currently this is a std::variant, other strategies are of course possible.
* Q: Is there an enumeration somewhere of the supported types for node/edge properties? Currently this just instantiates a couple of primitive types and string.
* There is a very basic test here - I intend to flesh this out as part of this change but wanted to send the code sooner rather than later. Specifically, this needs some more interesting generators for node/edge properties including non-primitive property types. Also, tests for error conditions.


libgalois/include/katana/PropertyGraph.h
```diff
@@ -222,6 +223,12 @@ class KATANA_EXPORT PropertyGraph {
   /// The edge TypeSetID for each edge in the graph
   katana::NUMAArray<TypeSetID> edge_type_set_id_;
 
+  // List of node and edge indexes on this graph.
+  std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Node>>>
+      node_indexes_;
+  std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Edge>>>
+      edge_indexes_;
+
```
arekay, line 231: Would it make sense to have a map from the column name to the index instead of a vector?

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,192 @@
...
+    explicit iterator(const base_iterator& base) : base_iterator(base) {}
+    const node_or_edge& operator*() const {
+      return std::get<IndexId>(base_iterator::operator*()).id;
+    }
+  };
+
+  PropertyIndex(std::string column) : column_(std::move(column)) {}
+
+  // The name of the indexed property.
+  std::string column() { return column_; }
```
arekay: nit : Can we have something more descriptive here. To me `column()` should return the column. `ColumnName` is the laziest, alternatively `alias`...

libgalois/src/PropertyGraph.cpp
```diff
@@ -866,6 +866,88 @@ katana::PropertyGraph::InformPath(const std::string& input_path) {
...
+          column, num_nodes(), property);
+  if (!index) {
+    return index.error();
+  }
+
+  katana::Result<void> build_result = index.value()->BuildFromProperty();
+  if (!build_result) {
+    return build_result.error();
+  }
+
```
arekay: `KATANA_CHECKED`

libgalois/include/katana/PropertyIndex.h
```diff
@@ -0,0 +1,192 @@
...
+  using ArrowArrayType = typename arrow::CTypeTraits<c_type>::ArrayType;
+  using IndexId = typename PropertyIndex<node_or_edge>::IndexId;
+  using iterator = typename PropertyIndex<node_or_edge>::iterator;
+  using set_key_type = typename PropertyIndex<node_or_edge>::set_key_type;
+
+  PrimitivePropertyIndex(
+      const std::string& column, size_t num_entities,
+      std::shared_ptr<arrow::Array> property)
+      : PropertyIndex<node_or_edge>(column),
+        num_entities_(num_entities),
```
arekay, line 82: nit - `num_entities` or `num_rows`? Or maybe i am overfitting this to be an index on a column.

arekay (approved):
LGTM.

libgalois/include/katana/PropertyIndex.h
oshofmann: Both are true - I named this num_entities because it's meant to be the number of nodes/edges. This also happens to be the number of rows in the table/array, but it wasn't clear that that was an invariant.

libgalois/include/katana/PropertyGraph.h
oshofmann: A map would work. One reason I went for a vector here was looking forward: right now an index is identified just by a column name. In general, indexes could support indexing multiple columns, as well as different types of index on the same column, e.g. an ordered index like this one and one that supports substring search. So I expect that eventually, the pattern will be looking through the list of indexes to find a best match for an operation. That and I expect the number of indexes to be count-on-one-hand small.

That said if a map makes things easier here for now, can gladly change this.

libgalois/src/PropertyGraph.cpp
oshofmann: Handy macro, thanks.

libgalois/include/katana/PropertyIndex.h
oshofmann: Changed to column_name throughout.

PR #423, opened 2021-08-06
Move EntityTypeID management out of PropertyGraph

roshandathathri:


libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+    entity_type_ids.set(new_entity_type_id);
+    entity_type_id_to_atomic_entity_type_ids_.emplace_back(entity_type_ids);
+    atomic_entity_type_id_to_entity_type_ids_.emplace_back(entity_type_ids);
+
+    return new_entity_type_id;
+  }
+
+  /// adds a new entity type for the set of atomic types with names \p names
+  /// \returns the EntityTypeID for the new type
+  EntityTypeID AddNonAtomicEntityType(const std::vector<std::string>& names) {
```
roshandathathri, line 109: @witchel The above two functions provide ways to add more types

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+        atomic_entity_type_id_to_entity_type_ids_(
+            std::move(atomic_entity_type_id_to_entity_type_ids)) {}
+
+  void Reset() {
+    atomic_entity_type_id_to_type_name_.clear();
+    atomic_type_name_to_entity_type_id_.clear();
+    entity_type_id_to_atomic_entity_type_ids_.clear();
+    atomic_entity_type_id_to_entity_type_ids_.clear();
+    Init();
+  }
```
arthurp, line 84: Is there a compelling reason to provide this as part of the public API? I would generally rather create a new object than reset and reuse. It's just clearer. (I understand why this is useful as a private method to implement assignment operators and destructors.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+
+  /// adds a new entity type for the atomic type with name \p name
+  /// \returns the EntityTypeID for the new type
+  EntityTypeID AddAtomicEntityType(const std::string& name) {
```
arthurp, line 88: This creates an atomic type with a specific name. I think the documentation is the main issue, but I want to make sure this is clear. There isn't really a distinction between atomic and composite entity types. They are all in the same "namespace" (ID space I guess) and have the same supported operations. So there is no real place where you create an entity type for an atomic type. You just create an atomic type which by definition is also an entity type.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  /// \returns the number of entity types (including kUnknownEntityType)
+  size_t GetNumEntityTypes() const {
+    return entity_type_id_to_atomic_entity_type_ids_.size();
+  }
+
+  /// \returns true iff an atomic type \p name exists
+  bool HasAtomicType(const std::string& name) const {
+    return atomic_type_name_to_entity_type_id_.count(name) == 1;
```
arthurp, line 146: Isn't there a `contains` operation that is faster? Not that it really matters.

libgalois/include/katana/EntityTypeManager.h
arthurp: I think this function should take EntityTypeIDs instead of names.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+
+#include "katana/Logging.h"
+
+namespace katana {
+
+/// EntityTypeID uniquely identifies an entity (node or edge) type
+/// EntityTypeID for nodes is distinct from EntityTypeID for edges
+/// This type may either be an atomic type or an intersection of atomic types
+/// EntityTypeID is represented using 8 bits
+using EntityTypeID = uint8_t;
```
arthurp, line 16: Could this be an OpaqueID? I know it may not be worth the effort, but I think it's worth considering doing it now.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+static constexpr EntityTypeID kUnknownEntityType = EntityTypeID{0};
+static constexpr EntityTypeID kInvalidEntityType =
+    std::numeric_limits<EntityTypeID>::max();
```
arthurp, line 19: Please document these. Having two sentinels means it's extra import to make if very clear what they mean.

arthurp (commented):
This is great. I like that this is getting abstracted out.

I'm worried about error handling. There are a lot of cases where `EntityTypeManager` methods can fail due to external inputs. The most notable ones to me are:

* Conflicting type names. 
* Too many types (more that 254 at the moment I think).

It seems like both of these could happen due to weird user input (though it's possible the first actually cannot due to the distinct property name rule). If users can trigger the error, it should be handled with `Result` not an assertion, otherwise a user can crash the graph-workers which is a problem. So I think at least some of these methods need to be setup to return `Result`. In cases where we can guarantee that the error can never happen we can assert that the `Result` is a success.

I'm also worried that not all error cases are even checked. It looks like it's possible to add more types then supported and it's not clear what will happen. I think IDs will just wrap and chaos will ensue.

libgalois/include/katana/EntityTypeManager.h
scober: I think the end goal may be to have the width of this type change dynamically based on how many types are in a given graph. If that is no longer the case then I agree that we should make these opaque.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+    std::bitset<std::numeric_limits<EntityTypeID>::max() + 1>;
+/// A map from EntityTypeID to a set of EntityTypeIDs
+using EntityTypeIDToSetOfEntityTypeIDsMap = std::vector<SetOfEntityTypeIDs>;
+/// A map from the atomic type name to its EntityTypeID
+/// (that does not intersect any other atomic type)
+using AtomicTypeNameToEntityTypeIDMap =
+    std::unordered_map<std::string, EntityTypeID>;
+/// A map from the atomic type's EntityTypeID to its name
+/// (that does not intersect any other atomic type)
+using EntityTypeIDToAtomicTypeNameMap =
```
scober, line 31: +1 for moving these typedefs out of the `PropertyGraph` namespace!

libgalois/include/katana/EntityTypeManager.h
scober: I agree except that it limits the interface some. Currently, this function creates any necessary atomic types that don't exist. If we passed `EntityTypeID`s it wouldn't be able to do that. Then again, maybe it would be a good thing to explicitly force users to do that as two separate actions.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+
+    SetOfEntityTypeIDs entity_type_ids;
+    entity_type_ids.set(new_entity_type_id);
+    entity_type_id_to_atomic_entity_type_ids_.emplace_back(entity_type_ids);
+    atomic_entity_type_id_to_entity_type_ids_.emplace_back(entity_type_ids);
+
+    return new_entity_type_id;
+  }
+
+  /// adds a new entity type for the set of atomic types with names \p names
```
scober, line 107: copy-paste error: atomic -> non-atomic

Also this comment should document that any necessary atomic types will also be created.

libgalois/include/katana/EntityTypeManager.h
witchel: Agree that this seems fraught.  There is logic in this PR about sometimes subtracting 1 and sometimes 2 from entity counts.  But there is no use of `kInvalidEntityType`.  Is it absolutely necessary?

libgalois/include/katana/EntityTypeManager.h
witchel: 1. This should return a `Result<EntityTypeID>`, no?
2. I think related to what Arthur is asking, but let's say the user has a node with types A, B, C.  Then they assign a label set of C, D, E.  What path through the API do you envision?  Do I just check for the existence of atomic entity types C, D, and E, and then get the entity type of the set?

libgalois/include/katana/EntityTypeManager.h
witchel: Ah, this addresses my question above.  Strings make an easy-to-use interface for graph update.
Nit: I'd go for an `const std::unordered_set<std::string>&` input type.
`Result<EntityTypeID>` return type.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+  /// \returns the EntityTypeID for the new type
+  EntityTypeID AddNonAtomicEntityType(const std::vector<std::string>& names) {
+    for (auto& name : names) {
```
witchel, line 110: Could this be `const auto&`?

libgalois/src/PropertyGraph.cpp
```diff
@@ -201,42 +194,28 @@ GetEntityTypeIDsFromProperties(
...
-
-    katana::PropertyGraph::SetOfEntityTypeIDs entity_type_ids;
-    entity_type_ids.set(new_entity_type_id);
-    atomic_entity_type_id_to_entity_type_ids->emplace_back(entity_type_ids);
-    entity_type_id_to_atomic_entity_type_ids->emplace_back(entity_type_ids);
   }
 
   // collect the list of unique combination of types
-  using EntityType = katana::gstl::Vector<int>;
+  using FieldEntityType = katana::gstl::Vector<int>;
```
witchel, line 210: `int`?

witchel (commented):
Looks good.  I agree with previous comments, especially about Result return type and Invalid/Unknown sentinels.

libgalois/include/katana/EntityTypeManager.h
amberhassaan: Agree with Arthur.

libgalois/include/katana/EntityTypeManager.h
amberhassaan: Agree with Arthur. It is "surprising" that the function will create necessary atomic types.

libgalois/include/katana/EntityTypeManager.h
```diff
@@ -0,0 +1,227 @@
...
+
+  /// \returns true iff an atomic type \p name exists
+  bool HasAtomicType(const std::string& name) const {
+    return atomic_type_name_to_entity_type_id_.count(name) == 1;
+  }
+
+  /// \returns true iff an entity type \p entity_type_id exists
+  /// (returns true for kUnknownEntityType)
+  bool HasEntityType(EntityTypeID entity_type_id) const {
+    return entity_type_id < entity_type_id_to_atomic_entity_type_ids_.size();
```
amberhassaan, line 152: call `GetNumEntityTypes()`.

libgalois/src/PropertyGraph.cpp
```diff
@@ -148,17 +148,10 @@ struct PropertyColumn {
 };
 
 /// Assumes all boolean or uint8 properties are types
-katana::Result<katana::NUMAArray<katana::PropertyGraph::EntityTypeID>>
+katana::Result<katana::NUMAArray<katana::EntityTypeID>>
```
amberhassaan, line 151: Why not move this function as a generic API along with the EntityTypeManager. It's another way to create entity types.

amberhassaan (approved):
Overall OK. Some concerns worth discussion.

roshandathathri:
Sorry for the delay. I agree with the comments but I'll deal with them in a subsequent PR.

PR #425, opened 2021-08-06
Patch for PR #412 (Visit Arrow Support Decay Types)

rob-isaac:
These changes were supposed to be in PR #412, but somehow didn't make it (I must have either been on a detached head or messed up a rebase)

### DIFF
--- a/libsupport/include/katana/ArrowVisitor.h
+++ b/libsupport/include/katana/ArrowVisitor.h
@@ -193,14 +193,10 @@ struct VisitArrowInternalWrapper {
 
 template <typename T>
 using is_visit_arrow_base_t = typename std::disjunction<
-    std::is_same<T, const arrow::Array&>, std::is_same<T, const arrow::Scalar&>,
-    std::is_same<T, arrow::ArrayBuilder*>>;
+    std::is_same<std::decay_t<T>, arrow::Array>,
+    std::is_same<std::decay_t<T>, arrow::Scalar>,
+    std::is_same<std::decay_t<T>, arrow::ArrayBuilder*>>;
 
-template <typename T>
-using is_visit_arrow_base_sp_t = typename std::disjunction<
-    std::is_same<T, std::shared_ptr<const arrow::Array>>,
-    std::is_same<T, std::shared_ptr<const arrow::Scalar>>,
-    std::is_same<T, std::shared_ptr<arrow::ArrayBuilder*>>>;
 }  // namespace internal
 
 // VisitArrow call that supports multiple args of type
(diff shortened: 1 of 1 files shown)


libsupport/include/katana/ArrowVisitor.h
```diff
@@ -212,25 +208,13 @@ std::enable_if_t<
         internal::is_visit_arrow_base_t<Arg0>,
         internal::is_visit_arrow_base_t<Args>...>,
     katana::Result<typename std::decay_t<VisitorType>::ReturnType>>
-VisitArrow(VisitorType&& visitor, const Arg0& arg0, const Args&... args) {
+VisitArrow(VisitorType&& visitor, Arg0&& arg0, Args&&... args) {
```
witchel, line 211: Does this mean that `Arg0` and `Args` are now consumed where they were not before?

witchel (approved):
One question.

libsupport/include/katana/ArrowVisitor.h
rob-isaac: This means that the arguments are perfectly forwarded to the eventual call to visitor::Call<...>(...). There is no longer an enforcement on the arguments to visitor::Call being const. Does that answer your question?

witchel: To me, the rvalue-reference type tells me I have to `std::move` my parameter, as opposed to a `const T&`, which tells me the parameter is unchanged in the caller.  But I'm pretty naive when it comes to these C++ constructs.

rob-isaac: I am also pretty naive but my understanding is that perfect forwarding, which occurs when a function param is rval-ref and it is forwarded to another func, essentially passes the lval/rval type over to the next function that is called. You can use std::move but you dont have to - that just indicates that you are passing a reference that can be modified. Passing an l-value reference means that visitor::Call will also recieve an l-value reference

amberhassaan: @witchel : When `&&` is added to a template parameter purely ,e.g., `T&&` and not `vector<T>&&`, it becomes a "Universal reference" that captures all types of references of `T` and this enables "Perfect forwarding".

amberhassaan: @rob-isaac : do you know why we need to separate `Arg0` out? Is that we expect at least one parameter to the function? Just wondering.

amberhassaan (approved):
looks good.

libsupport/include/katana/ArrowVisitor.h
rob-isaac: @amberhassaan yes it is because we need at least one argument

PR #426, opened 2021-08-06
[KAT-1151] Re-implement C++ format checker in Python using multi-threading/multi-processing

srogatch:


scripts/check_cpp_format.py
```diff
@@ -0,0 +1,133 @@
...
+    print("Usage: check_cpp_format.py [-fix] <paths>")
+    sys.exit()
+i_arg = 1
+fix = False
+if sys.argv[i_arg] == "-fix":
+    fix = True
+    i_arg += 1
+roots = []
+for root in sys.argv[i_arg:]:
+    roots.append(root)
```
arthurp: Please use argparse or click.

scripts/check_cpp_format.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+# Create the task queue
+tasks = queue.Queue()
+
+# Launch consumer threads
+class Consumer:
+    def __init__(self):
+        self.thread = threading.Thread(target=self.entry)
+        self.check_failed = False
```
arthurp: Subclass `Thread`.

scripts/check_cpp_format.py
```diff
@@ -0,0 +1,133 @@
...
+            if not fix and (completed.returncode != 0 or ("<replacement " in str(completed.stdout))):
+                self.check_failed = True
+                print(item, " NOT OK")
+
+
+consumers = []
+for i in range(multiprocessing.cpu_count()):
+    cur_cons = Consumer()
+    consumers.append(cur_cons)
+    cur_cons.thread.start()
```
arthurp: I think you can do this much more easily with `ThreadPool` (https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.ThreadPool). Then you don't need to manage threads and queues yourself.

scripts/check_cpp_format.sh
```diff
@@ -1,41 +1,12 @@
...
+# Taken from https://stackoverflow.com/a/246128/1915854
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+
+python3 ${DIR}/check_cpp_format.py "$@"
```
arthurp, line 12: I don't think we need this wrapper. People can call `check_cpp_format.py` instead.

scripts/check_cpp_format.py
```diff
@@ -0,0 +1,133 @@
...
+
+# Finalize the processing
+tasks.put(None)
+check_failed = False
+for consumer in consumers:
+    consumer.thread.join()
+    check_failed = check_failed or consumer.check_failed
+
+# Return proper exit code
+sys.exit(1 if check_failed else 0)
```
arthurp: Please use a main function that is called using the conventional `if __name__ == "__main__":` block. It makes it easier to reuse code later and is just conventional.

The old python script style with code at the top level is just messy and hard to work with as the program grows, you can look at our other Python scripts for examples.

srogatch:
The abstraction for format/lint checkers is not yet complete. I'm just pushing to save my work remotely.

scripts/check_cpp_format.py
```diff
@@ -0,0 +1,133 @@
...
+                last_token.write(ch)
+            elif last_token.tell() != 0:
+                prune_paths.add(last_token.getvalue())
+                last_token = io.StringIO()
+            # else tokens are separated by multiple spaces
+        else:
+            last_token.write(ch)
+    if last_token.tell() != 0:
+        prune_paths.add(last_token.getvalue())
+        last_token = io.StringIO()
```
arthurp: Please don't write a parser manually. Regexs are better than this even if they are a pain in their own right. And I think you could rewrite this using `str.split` and a few other things if you really don't want to use regexs. I cannot tell what's happening here so it needs to be simplified.

arthurp (changes requested):
This needs some work. There are easy way to write this. I really recommend you read up on the Python standard library. You have reimplemented a lot of things that are already available.

I'm aware this is just a script. But a unreliable script is a pain for all of us, so I think there is value in making sure this is maintainable. I'm definitely willing to compromise on all my comments. As long as we can make this something that's at least as easy to understand as the original bash.

Some Python standard library packages that might be useful:

* argparse
* re
* shutil
* glob

And you should look again at multiprocess. It has stuff in it, I think you didn't notice.

scripts/katana_traversal/traversal.py
```diff
@@ -0,0 +1,97 @@
+from abc import ABC, abstractmethod
```
arthurp: I'm pretty sure everything you have here is already available in a functional form in the standard library. Maybe in `shutil` or `glob` I'm not sure. Regardless, I'm pretty sure we don't need visitors here. They are a real pain compared to using generators.

scripts/check_cpp_format.py
srogatch: `str.split` definitely fails, because we get on input the paths in quotes, and the paths may contain spaces. Cryptic regex would not be easier to read than the parser code.

scripts/katana_traversal/traversal.py
srogatch: The visitor is very easy to use: just inherit from the base class and override the methods. On the other hand, I have only a limited understanding of Python generators, but as I understand we are not generating anything except the `bool`s returned, and we need to pass a parameter in every visiting step, rather than initializing the generator once and then getting values from it.

scripts/check_cpp_format.sh
srogatch: Without a major reorganization of the other scripts in the codebase, we need this wrapper. The problem is that `scripts/check_all.sh` is a symlink pointing to `external/katana/scripts/check_all.sh`, so a user can call it from both directories and we need `check_cpp_format.sh` in both directories. An alternative to providing this wrapper would be to place another `check_cpp_format.py` script in the enterprise repo, that would just call the Python script from the open repo. Then we would be able to get rid of `check_cpp_format.sh` in both directories.

srogatch:
> There are easy way to write this. I really recommend you read up on the Python standard library. You have reimplemented a lot of things that are already available.

I wrote this quickly in the easiest for me way. Python standard library has a learning curve, and I'm not that familiar with it. It's much faster and easier for me to reimplement than to learn the standard library. If you are that familiar with the standard library, can you refactor the script to its usage in the next iteration? For me, it would take an unreasonable amount of time.

> I'm aware this is just a script. But a unreliable script is a pain for all of us, so I think there is value in making sure this is maintainable. I'm definitely willing to compromise on all my comments. As long as we can make this something that's at least as easy to understand as the original bash.

Why do you think that the script is unreliable? Have you found a single bug in it?
As to maintainability, indeed, standard library-based implementations may be more maintainable, however, the standard library can be studied for years, while I implemented the script in a few hours. The original bash is irrelevant because the whole team was wasting a few minutes of work time in each run of the script, which has to be run before every commit. That is more expensive than the maintainability of the Python script.

scripts/check_cpp_format.py
srogatch: It would definitely not be easier for me. I've looked at the documentation for `ThreadPool`, but can't quickly figure out what's going on.

scripts/check_cpp_format.py
srogatch: I don't want all the `Thread` members to leak into `Consumer`. Composition is better than inheritance in this case. Because I don't need e.g. to call `Consumer` methods by a reference to `Thread` object, don't need to override the methods, etc.

srogatch:
The abstraction is now complete, and the C++/CUDA formatter uses it. Please, see the latest version of the changes.

scripts/check_cpp_format.sh
arthurp: Couldn't you just put a symlink in the enterprise directory just like there are symlinks to other scripts?

arthurp (approved):
Go ahead and merge.

I'll revisit this when I have time.

srogatch:
> Go ahead and merge.
> 
> I'll revisit this when I have time.

Thanks a lot. I've created https://katanagraph.atlassian.net/browse/KAT-1162 for you.

ddn0:
> Why do you think that the script is unreliable? Have you found a single bug in it?
As to maintainability, indeed, standard library-based implementations may be more maintainable, however, the standard library can be studied for years, while I implemented the script in a few hours.

I think it is easy to think that one's experience reflects the perspective of the team, but that might not always be the case.

Some people are actually more familiar with Python than bash for instance.

And most mature code is read and executed many more times than it is written, so easy-to-write is not exactly a first-order concern.
 
> The original bash is irrelevant because the whole team was wasting a few minutes of work time in each run of the script, which has to be run before every commit. That is more expensive than the maintainability of the Python script.

Similarly, I never run `check_cpp_format.sh` myself because my IDE automatically calls clang-format on save.

It might still be the case that parallelizing cpp formatting is valuable in the long-run but that really should be weighed against the increase in complexity (software projects have a limit after which they become unmaintainable) and opportunity cost.

scripts/check_cpp_format.sh
srogatch: I would have to put a symlink to the specific Python script as well as the symlink to the directory containing the abstract package, right? If you think it's a good idea, I can do it in https://github.com/KatanaGraph/katana/pull/432 as I found some incompatibilities with Python 3.6 anyway.

arthurp: I think it's reasonable, yes. But it's up to you. Not urgent.

srogatch: Ok, I'll create symlinks in my next open PR.

PR #435, opened 2021-08-11
Default-constructed EdgeTypeAwareTopology num_nodes() Segfault

rob-isaac:
uninitialized edge_type_index unique_ptr results in segfault on deref


amberhassaan (approved):
Looks good.

PR #443, opened 2021-08-13
[KAT-1047] update CI test of metagraph-plugin

iso-p:


iso-p:
1.  The conda build should work.
2. It seems the conda package upload test is skipped for now. Is that okay?

metagraph-plugin/conda_recipe/meta.yaml
```diff
@@ -15,11 +15,14 @@ requirements:
...
     - git
     - cmake
+    - pytest
   host:
     - pip
     - python
+    - pytest
   run:
     - python
+    - pytest
```
arthurp: pytest is not needed here. The test/requires is enough. As written the user will have to install pytest to install our metagraph plugin which makes no sense.

metagraph-plugin/conda_recipe/meta.yaml
```diff
@@ -29,6 +32,14 @@ test:
   imports:
     - metagraph
     - katana
+    - pytest
+  requires:
+    - pytest
+  source_files:
+    - tests
+  script:
+    - run_pytest.sh
```
arthurp: NIT: Since the script is so simple you can just use `commands`. https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-commands

metagraph-plugin/conda_recipe/meta.yaml
```diff
@@ -29,6 +32,14 @@ test:
   imports:
     - metagraph
     - katana
+    - pytest
```
arthurp: NIT: These are things you want it to *test* by importing. It would be better to call them "import tests" not "imports". So the things you have listed here (which I think are your dependencies), should not be, but it would be reasonable to have `metagraph_katana` listed here since it's the python package that is install in this conda package and which we are testing here.

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#python-imports

.github/workflows/python.yaml
```diff
@@ -127,6 +127,12 @@ jobs:
         . .github/workflows/activate_miniconda.sh
         scripts/conda_build.sh --output-folder $HOME/build/
 
+    - name: Build conda Packages for Katana Metagraph Plugin
+      run: |
+        . .github/workflows/activate_miniconda.sh
+        cd metagraph-plugin/conda_recipe
+        conda-build . -c conda-forge -c metagraph -c katanagraph/label/dev --output-folder $HOME/build/
```
arthurp: Instead of `-c katanagraph/label/dev` which will get the katana packages that are currently on anaconda, this should use `-c file://$HOME/build/`. That will use the packages built in the previous step by using those newly built packages as a channel.

arthurp (approved):
This is very good. :-) This would be fine as is, but we can improve a lot of small things, some more important than others. I have marked the unimportant stuff with "NIT:" There are a couple of comments that are not nits as well. Please make sure to address those.

NIT: In setup.py you reference the package `katana`, that should be `katana-python`. The pip and conda packages are called `katana-python` and `katana-cpp`. https://github.com/KatanaGraph/katana/blob/fb5779916e63efc34db605a69fc7d1e650436a7c/metagraph-plugin/setup.py#L15

NIT: I think `metagraph-plugin/data` should be in `metagraph-plugin/tests/data`. Since that data will only be used for testing.

No need to get another review once you are done with these comments. This is really looking good.

.github/workflows/python.yaml
arthurp: Fixing this is *critical*. If we do not, this could fail to allow changes that affect both `katana` and `metagraph_katana` from passing CI since for this test to pass, `matagraph_katana` needs to match the anaconda.org katana package. But to actually be correct w.r.t. the changes to `katana`, `matagraph_katana` will need to be changed in ways that are not compatible with those anaconda.org packages.

arthurp (changes requested):
I'm switching to request changes. Please request another review once you fix the CI package source issue and I can check it. I don't want to create more CI problem. I have created enough this week. ;-)

metagraph-plugin/conda_recipe/meta.yaml
iso-p: Keep the pytest in "run", but remove it in "host" and "build", right?

metagraph-plugin/conda_recipe/meta.yaml
iso-p: Since the entry point of metagraph_katana is not "import metagraph_katana" but something part of "metagraph", should we add something like "metagraph.wrappers.Graph.KatanaGraph" to the import tests? 

Besides, this is for the "type" plugin of katana only. For our "algorithm" and "translator" plugin, they are written in a decorator manner, and I am not sure how to import them. (I could find them somewhere I think, but the import test might looks ugly)

metagraph-plugin/conda_recipe/meta.yaml
arthurp: No. It isn't needed in run. Normal running of this package doesn't require pytest. Only running the tests requires that, but they are not actually in the package anyway.

metagraph-plugin/conda_recipe/meta.yaml
arthurp: You don't need to have an `imports` section. It's just a quick way to test code by importing it. Your pytests will be better and more complete anyway. You can just remove this.

metagraph-plugin/conda_recipe/meta.yaml
iso-p: Got it.

metagraph-plugin/conda_recipe/meta.yaml
iso-p: Yeah, It is redundant. Agree.

metagraph-plugin/conda_recipe/meta.yaml
```diff
@@ -15,6 +17,7 @@ requirements:
   build:
     - git
     - cmake
+    - katana-python>=0.1.12.dev
```
arthurp, line 20: Next PR: This should be `katana-python=={{ KATANA_VERSION }}` to pin to the exact version.

arthurp (approved):
LGTM.

Thanks. Please merge soon. Chris is excited to see it merged.

(I have a comment, but it should be in a separate PR.)

PR #445, opened 2021-08-13
Remove metagraph-katana test since it failing.

arthurp:
It needs to be reworked anyway. Ths test as written didn't test
the repo version of katana.


arthurp:
@gliga if you need to move your PR along quickly you could just steal this commit and put it in your PR.

arthurp:
@iso-p It appears this test started failing. So I'm removing it now. https://github.com/KatanaGraph/katana/pull/443 will replace it with something better.

gliga (approved):

PR #446, opened 2021-08-13
logging: fix finishing logic for tracers

patrickkenney9801:


libsupport/include/katana/ProgressTracer.h
```diff
@@ -157,6 +156,11 @@ class KATANA_EXPORT ProgressTracer {
       const std::string& span_name, std::shared_ptr<ProgressSpan> child_of) = 0;
   ProgressScope SetActiveSpan(std::shared_ptr<ProgressSpan> span);
 
+  /// Close is called when a tracer is finished processing spans
+  /// This should be called to ensure any and all buffered spans are
```
ddn0: nit: It is probably better to write this more descriptively rather than prescriptively 

"""
Close flushes any buffered spans.
"""

libsupport/include/katana/ProgressTracer.h
```diff
@@ -157,6 +156,11 @@ class KATANA_EXPORT ProgressTracer {
...
+  /// flushed
+  virtual void Close() = 0;
```
ddn0, line 161: what does it mean to be `virtual` and `private` ?

libsupport/src/ProgressTracer.cpp
```diff
@@ -152,6 +152,21 @@ katana::ProgressTracer::SetActiveSpan(
   return ProgressScope(std::move(span));
 }
 
+void
```
ddn0, line 157: On https://github.com/KatanaGraph/katana/pull/446/files#diff-7fc4d678da093aebb69172abdffa2f315e33051b081466050d8b4497f15dfbd2R131 above:

```
    if (active_span_ != nullptr &&
        (active_span_->ScopeClosed() || active_span_->IsFinished())) {
      active_span_->Finish();
    }
```

it seems weird to me that we finish spans that are already finished.

libsupport/include/katana/ProgressTracer.h
```diff
@@ -143,10 +143,9 @@ class KATANA_EXPORT ProgressTracer {
   uint32_t GetHostID() { return host_id_; }
   uint32_t GetNumHosts() { return num_hosts_; }
 
-  /// Close is called when a tracer is finished processing spans
-  /// This should be called to ensure any and all buffered spans are
-  /// flushed
-  virtual void Close() = 0;
+  /// Finish is called when SharedMemSys/DistMemSys finish
```
ddn0: It is better to describe what this does without reference to SharedMemSys or DistMemSys.

E.g.,

"""
Finish closes the active span and its parent spans if present and resets the active span to the unnamed root span.
"""

ddn0 (approved):

libsupport/include/katana/ProgressTracer.h
patrickkenney9801: I thought virtual meant it was supposed to be overriden?

libsupport/src/ProgressTracer.cpp
patrickkenney9801: That code path is for handling someone prematurely finishing a span on the active path with an explicit call to span.Finish()
The span is not finished twice since there is a check, but it goes through the logic of setting the active_span to being the parent of the prematurely closed currently active_span

ddn0: Ideally there would a name or state associated with "finished but not really completely finished"; if there is no good way to make that evident in the code, can you at least add a comment why something may be considered "finished" but still needs "finishing"?

libsupport/include/katana/ProgressTracer.h
ddn0: The unholy trinity of `friend`, `private` and `virtual`.

I now understand what is going on but it was initially confusing. `private` typically means no one except me can call this. `private` and `virtual` is more along the lines of "this is an extension point for subclasses". I normally would make these `protected` but that is a style thing and there is some debate around this: https://stackoverflow.com/questions/2170688/private-virtual-method-in-c

Knowing this, I'd write the function documentation like this:

```cpp
/// Close is called by Finish when ProgressTracer has released all its resources.
///
/// Subclasses override this to release their specific resources.
virtual ProgressTracer::Close() = 0;

/// Close is called by Finish when a ProgressSpan has released all its resources.
///
/// Subclasses override this to release their specific resources.
virtual ProgressSpan::Close() = 0;
```

ddn0:
Also while you are here, can you add some `const`s?

```patch
diff --git a/libsupport/include/katana/ProgressTracer.h b/libsupport/include/katana/ProgressTracer.h
index 55b4a4641..6f3069e11 100644
--- a/libsupport/include/katana/ProgressTracer.h
+++ b/libsupport/include/katana/ProgressTracer.h
@@ -140,8 +140,8 @@ public:
   /// is created (in this case the program is probably not using tracing)
   virtual ProgressSpan& GetActiveSpan();
   bool HasActiveSpan() { return active_span_ != nullptr; }
-  uint32_t GetHostID() { return host_id_; }
-  uint32_t GetNumHosts() { return num_hosts_; }
+  uint32_t GetHostID() const { return host_id_; }
+  uint32_t GetNumHosts() const { return num_hosts_; }

   /// Finish is called when SharedMemSys/DistMemSys finish
   /// Handles end of program logic and calls Close
@@ -235,7 +235,7 @@ public:
   /// Primarily for internal class use only
   void MarkScopeClosed();
   virtual bool ScopeClosed();
-  bool IsFinished() { return finished_; }
+  bool IsFinished() const { return finished_; }
   const std::shared_ptr<ProgressSpan>& GetParentSpan() { return parent_; }

   /// Every ProgressSpan started must be finished
```

PR #449, opened 2021-08-13
[KAT-1280] Use panda series as the return value for properties.

arthurp:


arthurp:
After thought. I think this is a bad idea. I think a custom array type is better. Especially given that PropertyViews are coming.

PR #463, opened 2021-08-18
docs: Add guide on commit messages

ddn0:


docs/contributing/git-style.rst
```diff
@@ -46,3 +46,64 @@ features or plugins to streamline this process (e.g.,
...
+  thrust of your work, explain the rational behind any non-obvious but required
+  changes.
+
+At the same time, there is no reason to agonize over the content of a commit
+message or be overly verbose. Applying an automatic formatter changes a lot of
+files but doesn't require a lot of context, while a one-line change might merit
+several paragraphs of context if the rationale is subtle or non-obvious.
+
+The first line of a commit message should be short, generally less than 80
+characters long.
```
tylershunt: Could we add the standard language about how this line should be imperative? or do you think that's getting overly prescriptive.

Similar thought: I'm fond of starting this line with some context for what the logical center is (e.g. `csv-import:` `tsuba:` etc.) I don't want to push pedantic requirements on folks, but if other people like that it is nice for messages to have some consistent structure.

tylershunt (approved):
Well stated.

docs/contributing/git-style.rst
roshandathathri: I second both suggestions of Tyler

roshandathathri (approved):
Looks great. Thanks!

amansavani (approved):

docs/contributing/git-style.rst
ddn0: I wonder about all these Jira tags, e.g., `[KAT-XXX]`, does anyone know if the Jira linking works okay if it is just in the body of the commit rather than in the object line?

iso-p (approved):

docs/contributing/git-style.rst
ddn0: Checked and it looks like Jira reads commit bodies.

docs/contributing/git-style.rst
```diff
@@ -46,3 +46,98 @@ features or plugins to streamline this process (e.g.,
...
+  changes.
+
+At the same time, there is no reason to agonize over the content of a commit
+message or be overly verbose. Applying an automatic formatter changes a lot of
+files but doesn't require a lot of context, while a one-line change might merit
+several paragraphs of context if the rationale is subtle or non-obvious.
+
+The subject line (i.e., the first line) of a commit message should be short,
+generally less than 80 characters long, and starts with a verb in the
+imperative mood (present tense, second person). You can think of the message as
```
arthurp: imperative mode. ;-)

docs/contributing/git-style.rst
```diff
@@ -46,3 +46,98 @@ features or plugins to streamline this process (e.g.,
...
+- What plausible alternative implementations did you consider but rule out?
+
+- What general thing or abstraction is this change in furtherance of?
+
+- If this change fixes a bug, what was the original bug or its symptoms?
+
+- If you did some additional changes along the way to facilitate the main
+  thrust of your work, explain the rational behind any non-obvious but required
+  changes.
+
```
arthurp, line 95: I think this list should also include notable API changes. That makes it much easier for someone to understand when a commit may affect external code or other code they are working on.

docs/contributing/git-style.rst
```diff
@@ -46,3 +46,98 @@ features or plugins to streamline this process (e.g.,
...
+At the end of the day, remember that there are three main things we accomplish
+with commit messages:
+
+1. Facilitate review of our changes by reviewers
+
+2. Provide context to future readers about our changes
+
+3. Ground the scope of our changes to ourselves. If we struggle to write a
+   simple explanation of why we made a change, it is a sign that the underlying
+   code changes are also not as clear as they could be.
```
arthurp: I think this should be at the top not the bottom, or maybe both. Best to give people an idea of the goals immediately to have them in the right mindset.

arthurp (approved):
I'm happy with this. A few comments for you to use or ignore as needed.

docs/contributing/git-style.rst
ddn0: I'm not a linguist. I just followed wikipedia: https://en.wikipedia.org/wiki/Imperative_mood

FWIW, "imperative mode" has more hits in Google but they are mostly about non-English languages.

arthurp: .... Today I learned. That's super interesting. Carry on. And definitely check primary sources when I give you information apparently.

(I actually learned most of my grammar in the context of German, so this may have affected my thinking on this. It's not the only think, I still have a better idea of what the accusative is than what the indirect object is.)

PR #466, opened 2021-08-19
libsupport Cache provides an LRU cache 

witchel:
Support for caching properties is split into a different PR because it is messy.

### DIFF
--- a/libsupport/include/katana/Cache.h
+++ b/libsupport/include/katana/Cache.h
@@ -1,0 +1,166 @@
+#ifndef KATANA_LIBSUPPORT_KATANA_CACHE_H_
+#define KATANA_LIBSUPPORT_KATANA_CACHE_H_
+
+// This is single threaded only, it is not intended to store large objects,
+// but rather metadata (e.g., a shared_ptr to a property column).
+
+// The problem witchel had implementing a multi-threaded version using
+// parallel-hashmap is a lock ordering problem.  parallel-hashmap 1.33 allows
+// execution of code with the write lock held, only when modifying, not when adding an
+// element.  We have to modify the LRU list when eviciting an element, so the natural
+// lock ordering is parallel-hashmap write lock, then list lock.  But without a way to
+// execute insert code with the parallel-hashmap write lock held, it seemed like there
+// would be some form of race condition.
+
+#include <list>
+#include <optional>
+#include <string>
+
+#include <boost/container_hash/hash.hpp>
+
+#include "katana/Logging.h"
+
+namespace tsuba {
+
+template <typename Key, typename Value>
+class KATANA_EXPORT Cache {
+  using ListType = std::list<Key>;
+  struct MapValue {
+    Value value;
+    // This allows us to delete the old position in the LRU list without a scan
+    typename ListType::iterator lru_it;
+  };
+  using MapType = std::unordered_map<Key, MapValue>;
+  enum class ReplacementPolicy { kLRUSize, kLRUBytes };
+
+public:
+  /// Construct an LRU cache that has a fixed number of entries.
+  Cache(
+      size_t capacity,  // number of entries
+      std::function<void(const Key& key)> evict_cb = nullptr)
+      : policy_(ReplacementPolicy::kLRUSize),
+        capacity_(capacity),
+        value_to_bytes_(nullptr),
+        evict_cb_(std::move(evict_cb)) {
+    KATANA_LOG_VASSERT(capacity_ > 0, "cache requires positive capacity");
+  }
+  /// Construct an LRU cache that holds fixed number of bytes.
+  Cache(
+      size_t capacity,  // bytes of entries
+      std::function<size_t(const Value& value)> value_to_bytes,
+      std::function<void(const Key& key)> evict_cb = nullptr)
+      : policy_(ReplacementPolicy::kLRUBytes),
+        capacity_(capacity),
+        value_to_bytes_(std::move(value_to_bytes)),
+        evict_cb_(std::move(evict_cb)) {
+    KATANA_LOG_VASSERT(capacity_ > 0, "cache requires positive capacity");
+    KATANA_LOG_VASSERT(
+        value_to_bytes_ != nullptr,
+        "kLRUBytes policy requires value to bytes function");
+  }
+
+  size_t size() const {
+    if (policy_ == ReplacementPolicy::kLRUSize) {
+      return key_to_value_.size();
+    } else {
+      return total_bytes_;
+    }
+  }
+
+  size_t capacity() const { return capacity_; }
+
+  bool Empty() const { return key_to_value_.empty(); }
+
+  bool Contains(const Key& key) const {
+    return key_to_value_.find(key) != key_to_value_.end();
+  }
+
+  void Insert(const Key& key, const Value& value) {
+    auto mapit = key_to_value_.find(key);
+    if (mapit == key_to_value_.end()) {
+      lru_list_.push_front(key);
+      key_to_value_[key] = {value, lru_list_.begin()};
+      if (value_to_bytes_ != nullptr) {
+        total_bytes_ += value_to_bytes_(value);
+      }
+    } else {
+      mapit->second.value = value;
+      UpdateLRU(mapit);
+    }
+    EvictIfNecessary();
+  }
+
+  std::optional<Value> Get(const Key& key) {
+    // lookup value in the cache
+    std::optional<Value> ret;
+    auto it = key_to_value_.find(key);
+    if (it != key_to_value_.end()) {
+      ret = UpdateLRU(it);
+    }
+    return ret;
+  }
+
+private:
+  Value UpdateLRU(typename MapType::iterator mapit) {
+    auto lru_it = mapit->second.lru_it;
+    auto lru_head = lru_list_.begin();
+    if (lru_it != lru_head) {
+      // move item to the front of the most recently used list
+      lru_list_.splice(lru_head, lru_list_, lru_it);
+      mapit->second.lru_it = lru_head;
+    }
+    return mapit->second.value;
+  }
+
+  void EvictOne() {
+    // evict item from the end of most recently used list
+    auto tail = --lru_list_.end();
+    KATANA_LOG_ASSERT(tail != lru_list_.end());
+    Key evicted_key = std::move(*tail);
+    lru_list_.erase(tail);
+    auto evicted_value = std::move(key_to_value_.at(evicted_key).value);
+    key_to_value_.erase(evicted_key);
+    if (value_to_bytes_ != nullptr) {
+      total_bytes_ -= value_to_bytes_(evicted_value);
+    }
+    if (evict_cb_) {
+      evict_cb_(evicted_key);
+    }
+  }
+
+  void EvictIfNecessary() {
+    switch (policy_) {
+    case ReplacementPolicy::kLRUSize: {
+      while (size() > capacity_) {
+        EvictOne();
+      }
+    } break;
+    case ReplacementPolicy::kLRUBytes: {
+      KATANA_LOG_DEBUG_ASSERT(value_to_bytes_ != nullptr);
+      // Allow a single entry to exceed our byte capacity
+      while (size() > capacity_ && key_to_value_.size() > 1) {
+        EvictOne();
+      }
+    } break;
+    default:
+      KATANA_LOG_FATAL("bad cache replacement policy: {}", policy_);
+    }
+  }
+
+  // Map from key to value
+  MapType key_to_value_;
+  // LRU list
+  ListType lru_list_;
+
+  ReplacementPolicy policy_;
+  // for kLRUSize number of entries kLRUBytes it is byte total
+  size_t capacity_{0};
+  size_t total_bytes_{0};
+
+  std::function<size_t(const Value& value)> value_to_bytes_;
+  std::function<void(const Key& key)> evict_cb_;
+};
+
+}  // namespace tsuba
+
+#endif
--- a/libsupport/test/CMakeLists.txt
+++ b/libsupport/test/CMakeLists.txt
@@ -18,6 +18,7 @@ endfunction()
 
 add_unit_test(tracing)
 add_unit_test(bitmath)
+add_unit_test(cache)
 add_unit_test(env)
 add_unit_test(logging)
 add_unit_test(opaque-id)
--- a/libsupport/test/cache.cpp
+++ b/libsupport/test/cache.cpp
@@ -1,0 +1,210 @@
+#include "katana/Cache.h"
+
+#include <map>
+#include <random>
+
+#include "katana/Logging.h"
+#include "katana/Random.h"
+
+enum class NodeEdge { kNode, kEdge };
+
+// Basic Cache Key implementation
+struct CacheKey {
+  CacheKey(const NodeEdge& node_edge, const std::string& name)
+      : node_edge_(node_edge), name_(name) {}
+
+  bool operator==(const CacheKey& o) const {
+    return node_edge_ == o.node_edge_ && name_ == o.name_;
+  }
+
+  std::size_t hash() const noexcept {
+    using boost::hash_combine;
+    using boost::hash_value;
+
+    std::size_t seed = 0;
+    hash_combine(seed, hash_value(node_edge_));
+    hash_combine(seed, hash_value(name_));
+
+    // Return the result.
+    return seed;
+  }
+
+  NodeEdge node_edge_;
+  std::string name_;
+};
+
+// inject custom specialization of std::hash to namespace std
+namespace std {
+template <>
+struct hash<CacheKey> {
+  std::size_t operator()(CacheKey const& key) const noexcept {
+    return key.hash();
+  }
+};
+}  // namespace std
+
+struct CacheValue {
+  int64_t a;
+  int32_t b;
+  uint32_t c;
+};
+static uint64_t count_evictions{0};
+
+CacheValue
+RandomValue() {
+  std::uniform_int_distribution<int64_t> dist(
+      0, std::numeric_limits<int64_t>::max());
+  CacheValue elt;
+  elt.a = dist(katana::GetGenerator());
+  elt.b = static_cast<int32_t>(elt.a);
+  elt.c = static_cast<uint32_t>(elt.a);
+  return elt;
+}
+
+size_t
+BytesInValue(const CacheValue& value) {
+  if (value.a & 1) {
+    return 5;
+  } else {
+    return 1;
+  }
+}
+
+CacheValue
+SizeOneValue() {
+  auto v = RandomValue();
+  v.a = 0;
+  return v;
+}
+
+CacheValue
+SizeFiveValue() {
+  auto v = RandomValue();
+  v.a = 1;
+  return v;
+}
+
+void
+InsertRandom(
+    const std::vector<CacheKey>& keys,
+    tsuba::Cache<CacheKey, CacheValue>& cache) {
+  for (const auto& key : keys) {
+    cache.Insert(key, RandomValue());
+  }
+}
+
+void
+AssertLRUElements(
+    std::vector<CacheKey>::const_iterator endit, size_t num,
+    tsuba::Cache<CacheKey, CacheValue>& cache) {
+  for (auto it = endit - num; it < endit; ++it) {
+    KATANA_LOG_ASSERT(cache.Get(*it).has_value());
+  }
+  KATANA_LOG_ASSERT(!cache.Get(*(endit - num - 1)).has_value());
+}
+
+void
+TestLRUBytes(
+    const std::vector<CacheKey>& node_keys,
+    const std::vector<CacheKey>& edge_keys) {
+  count_evictions = 0;
+  size_t byte_size = 4;
+  tsuba::Cache<CacheKey, CacheValue> cache(
+      byte_size, [](const CacheValue& value) { return BytesInValue(value); },
+      [&](const CacheKey&) { count_evictions++; });
+
+  auto nodeit = --node_keys.end();
+  KATANA_LOG_ASSERT(nodeit != node_keys.begin());
+  cache.Insert(*nodeit--, SizeOneValue());
+  KATANA_LOG_ASSERT(nodeit != node_keys.begin());
+  cache.Insert(*nodeit--, SizeOneValue());
+  KATANA_LOG_ASSERT(nodeit != node_keys.begin());
+  cache.Insert(*nodeit--, SizeOneValue());
+  KATANA_LOG_ASSERT(nodeit != node_keys.begin());
+  cache.Insert(*nodeit--, SizeOneValue());
+
+  KATANA_LOG_ASSERT((byte_size + 1) < node_keys.size());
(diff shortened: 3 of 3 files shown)


libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,183 @@
...
+namespace tsuba {
+
+struct KATANA_EXPORT KeyBase {
+  enum class NodeEdge { kNode, kEdge };
+  bool operator==(const KeyBase& o) const {
+    return node_edge_ == o.node_edge_ && name_ == o.name_;
+  }
+
+  // This was supposed to suffice, but only the std::hash worked.
+  //friend size_t hash_value(const KeyBase& key)
```
witchel: It was a bummer to me that I had to use the `namespace std` trick below.

witchel:
I'm sorry, I don't like what I've written.  I used a parallel hash map even though the number of entries and their size is small, and there is not multi-threaded access.  And my lock ordering is inconsistent.  So please hold off with the reviews!

witchel:
> I'm sorry, I don't like what I've written. I used a parallel hash map even though the number of entries and their size is small, and there is not multi-threaded access. And my lock ordering is inconsistent. So please hold off with the reviews!

Ok, back in business.  Please review my single-threaded cache.

libsupport/CMakeLists.txt
```diff
@@ -27,11 +27,28 @@ set(sources
...
         src/URI.cpp
 )
 
+find_package(phmap QUIET)
+if(NOT TARGET phmap::phmap)
+  include(CheckIncludeFileCXX)
+  unset(phmap_FOUND)
+  CHECK_INCLUDE_FILE_CXX(parallel_hashmap/phmap.h phmap_FOUND)
+  if (phmap_FOUND)
+  # Since the include is directly available due to the above check, an empty target is enough.
```
ddn0: Is this comment actually true?

The documentation of

https://cmake.org/cmake/help/latest/module/CheckIncludeFileCXX.html

says that the path will be stored in `phmap_FOUND` so I'd assume we'd use this path when setting up our interface library:

https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

E.g.,

```
add_library(phmap::phmap INTERFACE IMPORTED)
target_include_directories(phmap::phmap INTERFACE ... dir of ${phmap_FOUND} ...)
```

libsupport/include/katana/CompilerSpecific.h
```diff
@@ -0,0 +1,55 @@
+/*
```
ddn0: can we delete the copy of this file in `libgalois` then?

libsupport/include/katana/ConcurrentMap.h
```diff
@@ -0,0 +1,27 @@
+#ifndef KATANA_LIBSUPPORT_KATANA_CONCURRENTMAP_H_
+#define KATANA_LIBSUPPORT_KATANA_CONCURRENTMAP_H_
+
+#include <parallel_hashmap/phmap.h>
+
+#include "katana/SimpleLock.h"
+
+namespace katana {
+
+/// Specialized parallel_flat_hash_map type that makes use of galois's
```
ddn0: nit style: "ConcurrentMap is a concurrent flat hashmap with spin locks.\n\nThe implementation will...."

- Also there is no galois's SimpleLock as it has been moved to support already

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+};
+
+template <typename Value>
+class KATANA_EXPORT Cache {
+public:
+  using ListType = std::list<KeyBase>;
+  struct MapValue {
+    Value value;
+    // This allows us to delete the old position in the LRU list without a scan
+    typename ListType::iterator lruit;
```
ddn0: nit: `lru_iterator` or just `list_it` or `ptr`?

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+    std::optional<Value> ret = std::nullopt;
+    auto it = key_to_value_.find(key);
+    if (it != key_to_value_.end()) {
+      ret = update_lru(it);
+    }
+    return ret;
+  }
+
+private:
+  Value update_lru(typename MapType::iterator mapit) {
```
ddn0: PascalCase

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+
+  size_t size() const { return key_to_value_.size(); }
+  size_t bytes() const { return total_bytes_; }
+
+  size_t LRUSizeCapacity() const { return lru_capacity_; }
+  size_t LRUBytesCapacity() const { return bytes_capacity_; }
+
+  bool Empty() const { return key_to_value_.empty(); }
+
+  bool Contains(const KeyBase& key) {
```
ddn0: `const`?

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+    } else {
+      mapit->second.value = value;
+      update_lru(mapit);
+    }
+    evict_if_necessary();
+  }
+
+  std::optional<Value> Get(const KeyBase& key) {
+    // lookup value in the cache
+    std::optional<Value> ret = std::nullopt;
```
ddn0: nit/hint: the default value of `std::optional<T>` is always `std::nullopt`

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+#include <boost/container_hash/hash.hpp>
+
+#include "katana/ArrowInterchange.h"
+#include "katana/ConcurrentMap.h"
+#include "katana/Logging.h"
+
+namespace tsuba {
+
+struct KATANA_EXPORT KeyBase {
+  enum class NodeEdge { kNode, kEdge };
```
ddn0: What is this used for?

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+      hash_combine(seed, hash_value(key.name_));
+
+      // Return the result.
+      return seed;
+    }
+  };
+};
+
+template <typename Value>
+class KATANA_EXPORT Cache {
```
ddn0: FWIW, I'd call this `LRUCache` as it doesn't really support general caching policies.

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  void evict_one() {
+    KeyBase evicted_key;
+    // evict item from the end of most recently used list
+    auto tail = --lru_list_.end();
+    KATANA_LOG_ASSERT(tail != lru_list_.end());
+    evicted_key = *tail;
+    lru_list_.erase(tail);
+    auto evicted_value = key_to_value_.at(evicted_key).value;
```
ddn0: Consider some moves here to avoid copying. Though overall, the cache interface seems to assume copying is cheap. Is this intentional?

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+      return seed;
+    }
+  };
+};
+
+template <typename Value>
+class KATANA_EXPORT Cache {
+public:
+  using ListType = std::list<KeyBase>;
+  struct MapValue {
```
ddn0: It doesn't look like `MapValue` or `MapType` need to be public. Can we make them private instead?

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,182 @@
...
+  size_t LRUSizeCapacity() const { return lru_capacity_; }
+  size_t LRUBytesCapacity() const { return bytes_capacity_; }
+
+  bool Empty() const { return key_to_value_.empty(); }
+
+  bool Contains(const KeyBase& key) {
+    return key_to_value_.find(key) != key_to_value_.end();
+  }
+
+  void Insert(const KeyBase& key, const Value& value) {
```
ddn0: Why not just template on the key type as well? and thus support any hashable thing?

ddn0 (commented):
why not move the test of the cache to this repo as well?

libsupport/include/katana/CompilerSpecific.h
witchel: Whoops, that was my intent.

libtsuba/include/tsuba/Cache.h
witchel: Donaldwhydoyoumakesuchrequests?

libtsuba/include/tsuba/Cache.h
witchel: I literally stared at `MapType `for 5 minutes thinking, "type names are supposed to be PascalCase."  It is weird how resistant my neurons are at adapting to our naming conventions.

libtsuba/include/tsuba/Cache.h
witchel: I thought about doing that, but then thought why not leave it generic.  But sure, I can make it `LRUCache`.

libtsuba/include/tsuba/Cache.h
witchel: I added some std::move's because I'm bad at knowing when those will be important.

I have this comment now at the top of the file.

```
// This is single threaded only, it is not intended to store large objects,
// but rather metadata (e.g., a shared_ptr to a property column).
```

witchel:
> why not move the test of the cache to this repo as well?

I just saw that there was no test directory in the open repo, only in enterprise.  I thought that meant we consider tests enterprise-y.  I think all of the `tsuba_` binaries make more sense in open as they use the public interfaces.  Should I at least relocate my cache test to open?

libtsuba/include/tsuba/Cache.h
witchel: What I wanted to say is, "your key needs to be able to hash itself."  I thought a pure template would not do that, so I provided a base class for any key class.  It seemed kind of like a C++-20 concept, but I'm not sure of that either.

What is the best way to say that I need to know if your key is for node or edge, it should have a name (that will be unique for nodes and unique for edges) and I need to know how to hash it.

libtsuba/include/tsuba/Cache.h
witchel: Hmm, I started doing this and I don't like it.  The replacement policy feels like an implementation detail that should stay out of the name.

libtsuba/include/tsuba/Cache.h
witchel: I think this issue is covered more completely below, but I need three things from the key, one of which is whether it is a node property or an edge property.

libsupport/CMakeLists.txt
witchel: Ok.  I just copied that code from the last `CMakeLists.txt`.  But that is no excuse.

@tylershunt might have some insight here, but if I engage my pattern matching "skills" can I simplify this whole rigamarole to this?
```
find_package(phmap REQUIRED)
target_link_libraries(katana_support PUBLIC phmap::phmap)
```

libtsuba/include/tsuba/Cache.h
```diff
@@ -0,0 +1,183 @@
+#ifndef KATANA_LIBTSUBA_TSUBA_CACHE_H_
+#define KATANA_LIBTSUBA_TSUBA_CACHE_H_
+
+// This is single threaded only, it is not intended to store large objects,
+// but rather metadata (e.g., a shared_ptr to a property column).
+
+// The problem witchel had implementing a multi-threaded version using
```
ddn0: This comment seems somewhat misplaced as the current implementation is not multithreaded, so why even mention that you were unable to make a multithreaded implementation?

witchel: It is a "notes for future me/maintainer" sort of thing.  I know in 6 months I'm going to wonder why I didn't just do this as a multi-threaded cache.

libtsuba/include/tsuba/Cache.h
ddn0: That may be true, but the implementation of different policies will lead to varying fields in the implementation class and the current map + linked list implementation may not be suitable for let's say prioritized replacement.

If you want a general interface you try a cache base class, but as you note, the current interface is only appropriate for things that are conveniently copyable. And with a base class, you'd have a hard time expressing the template Value and possible Key parameter.

With templates, you can only achieve textual/static polymorphism (i.e., if it happens to have the same name functions and the return values are interchangable, you can replace one cache class for another) versus the dynamic polymorphism of virtual. The way you indicate substitutability with templates (c.f. virtual base class) is simply to have multiple classes with the same function name and signature (C++ concepts just formalizes this). 

So, with templates the way you indicate there are bunch of interchangable cache classes is simply by convention.

```cpp

template<typename T>
class LRUCache {
public:
  T Get();
};

template<typename T>
class FancyCache {
public:
  T Get();
};

template <typename T>
using MyCache = FancyCache;

MyCache cache;
```

We can swap out cache implementations because they all have the same function names.

vs

```cpp
class Cache {
public:
  virtual Object Get() = 0;
};

class LRUCache: public Cache {
public:
  Object Get() override {};
};

Cache cache = new LRUCache();
```

it is not advisable to mix these two paradigms.

witchel: Interesting and appreciated.   I have tried to implement the template form of polymorphism for the Cache Key.

Some replacement policies might not work with the current structure, but some, like MRU or segmented LRU will.

libtsuba/include/tsuba/Cache.h
ddn0: I'm still not sure why this is here as the Cache class doesn't really care what the key is as long as it is hashable. `KeyBase` sounds like something that goes with the code making the Cache not here where we define the Cache class for everyone.

All your Key type needs to do is implement the Hash protocol: https://en.cppreference.com/w/cpp/utility/hash

witchel: It was still there because I thought it needed to be there, but you are right it doesn't.

I have removed it, but a consequence is that I have to "inject" its hash function into namespace `std`.  That has two disadvantages to my eyes: (1) the fields of the key have to be public so the hash function can read them and (2) it feels unencapsulated to me.  I wish the hash function could be defined as part of the object itself.

Neither of these issues is large, but I would be interested to hear your perspective.

ddn0: > I have removed it, but a consequence is that I have to "inject" its hash function into namespace std. That has two disadvantages to my eyes: (1) the fields of the key have to be public so the hash function can read them

You can either make your std::hash specialization a friend of your key class or just have it call a function you want on the key class itself that then returns the hash.

> and (2) it feels unencapsulated to me. I wish the hash function could be defined as part of the object itself.

You can make the hash function part of the object itself. You just need to tell the standard library what that function is called and you do that by extending `std::hash` as described.

witchel: Ah hah. I like the idea of making it a function called by `std::hash`.  That wasn't the pattern used on the cppreference website, though I should have thought of it.

ddn0:
> I just saw that there was no test directory in the open repo, only in enterprise. I thought that meant we consider tests enterprise-y. I think all of the tsuba_ binaries make more sense in open as they use the public interfaces. Should I at least relocate my cache test to open?

I don't understand. There a several tests in this repo and you even wrote one of them: https://github.com/KatanaGraph/katana/blame/master/libsupport/test/uri.cpp

witchel:
> > I just saw that there was no test directory in the open repo, only in enterprise. I thought that meant we consider tests enterprise-y. I think all of the tsuba_ binaries make more sense in open as they use the public interfaces. Should I at least relocate my cache test to open?
> 
> I don't understand. There a several tests in this repo and you even wrote one of them: https://github.com/KatanaGraph/katana/blame/master/libsupport/test/uri.cpp

Oh, I am using this cache in `libtsuba`, so I was thinking I should test it there (the test is currently in `libtsuba-fs` in enterprise and there is no test directory for `libtsuba `in open).  But my bad, I'll move it to `libsupport`.

witchel:
Can I get a Hallelujah?

witchel:
> Can I get a Hallelujah?

In this case, that means a reviewer approval.  I left the parallel hashmap in enterprise for now.

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,157 @@
...
+// execute insert code with the parallel-hashmap write lock held, it seemed like there
+// would be some form of race condition.
+
+#include <list>
+#include <optional>
+#include <string>
+
+#include <boost/container_hash/hash.hpp>
+
+#include "katana/ArrowInterchange.h"
```
ddn0: why is this included here?

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,157 @@
...
+    Value value;
+    // This allows us to delete the old position in the LRU list without a scan
+    typename ListType::iterator lru_it;
+  };
+  using MapType = std::unordered_map<Key, MapValue>;
+
+public:
+  enum class ReplacementPolicy { kLRUSize, kLRUBytes };
+
+  Cache(
```
ddn0, line 38: It is somewhat awkward to support all arguments when they are only valid for some cases.

Either:
- Create separate constructors for the Vanilla LRU and the Bytes LRU, e.g., `Cache(capacity, evict_cb)` and `Cache(capacity, evict_cb, value_to_bytes)` as the bytes policy requires `value_to_bytes`
- Create separate factory methods `template<typename Key, typename Value> static Cache<Key,Value> MakeFoo(...)`

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,157 @@
...
+        "kLRUSize policy requires non-zero LRU entries");
+    KATANA_LOG_VASSERT(
+        policy_ != ReplacementPolicy::kLRUBytes || bytes_capacity_ > 0,
+        "kLRUBytes policy requires non-zero byte capacity");
+  }
+
+  size_t size() const { return key_to_value_.size(); }
+  size_t bytes() const { return total_bytes_; }
+
+  size_t LRUSizeCapacity() const { return lru_capacity_; }
```
ddn0: I don't know how I feel about having separate `size` and `bytes` and "capacity in size" and "capacity in bytes" methods. If  the Cache policy is an actual policy, the interface of Cache objects should not change when the policy changes. If there is reason for clients to care that they have a bytes-based Cache versus a item-based Cache, that indicates that the interface itself should be different (i.e., different classes) even if the implementation is largely similar.

Either:
- Make size and capacity report values in units appropriate for the cache policy (i.e., items for `kLRUSize` or bytes in `kLRUBytes`)
- Make two separate classes `LRUCache` and `LRUBytesCache` and optionally use private inheritance or composition to pull in the common implementation

arthurp (commented):
I have nothing to add. @ddn0 hit everything I would say.

witchel:
> I have nothing to add. @ddn0 hit everything I would say.

Whew!

libsupport/include/katana/Cache.h
witchel: Mistake. It is removed.

libsupport/include/katana/Cache.h
witchel: I have done option 1.

witchel: Oh, I have not.  Hold on!

witchel: Ok, sorry, good to go.

libsupport/include/katana/Cache.h
witchel: Gone with option 1.

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,164 @@
...
+      : policy_(ReplacementPolicy::kLRUSize),
+        capacity_(capacity),
+        value_to_bytes_(nullptr),
+        evict_cb_(std::move(evict_cb)) {
+    KATANA_LOG_VASSERT(capacity_ > 0, "cache requires positive capacity");
+  }
+  Cache(
+      size_t capacity,  // bytes of entries
+      std::function<void(const Key& key)> evict_cb,
+      std::function<size_t(const Value& value)> value_to_bytes)
```
ddn0: - Perhaps swap the order of `evict_cb` and `value_to_bytes` to allow either policy/constructor to leave `evict_cb` optional.
- Add Doxygen comments on constructors indicating the reason for two different constructors.

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,164 @@
...
+      : policy_(ReplacementPolicy::kLRUBytes),
+        capacity_(capacity),
+        value_to_bytes_(std::move(value_to_bytes)),
+        evict_cb_(std::move(evict_cb)) {
+    KATANA_LOG_VASSERT(capacity_ > 0, "cache requires positive capacity");
+    KATANA_LOG_VASSERT(
```
ddn0, line 57: This assert seems partially superfluous with the new constructor

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,164 @@
...
+        policy_ != ReplacementPolicy::kLRUBytes || value_to_bytes_ != nullptr,
+        "kLRUBytes policy requires value to bytes function");
+  }
+
+  size_t size() const {
```
ddn0, line 62: `const`?

libsupport/include/katana/Cache.h
```diff
@@ -0,0 +1,164 @@
...
+    if (policy_ == ReplacementPolicy::kLRUSize) {
+      return key_to_value_.size();
+    } else {
+      return total_bytes_;
+    }
+  }
+
+  size_t capacity() const { return capacity_; }
```
ddn0, line 70: `const`?

ddn0 (approved):

libsupport/include/katana/Cache.h
witchel: Done.

libsupport/include/katana/Cache.h
witchel: Sure is, good catch.

libsupport/include/katana/Cache.h
witchel: Sorry, I don't get it.  It is a `const `method.

cmake/Modules/KatanaBuildCommon.cmake
```diff
@@ -631,3 +631,94 @@ file(CREATE_LINK
...
+# The exact paths searched by find_c_library varies based on the method and
+# version of CMake.
+#
+# If find_c_library fails to find a library, possible ways to adjust the search
+# path of find_c_library, find_package, pkg_check_modules, etc. are
+#
+# 1. Add the directory the environment variable PATH
+# 2. Add the directory to the CMake variable CMAKE_PREFIX_PATH
+# 3. Add the directory to the CMake variable ${LIBRARY_NAME}_ROOT
+function(find_c_library)
```
tylershunt: I don't think this is used anywhere in this PR (anymore).

tylershunt (approved):
LGTM

cmake/Modules/KatanaBuildCommon.cmake
witchel: It isn't, but I thought it was useful.  I'll take it out.

PR #468, opened 2021-08-19
[KAT-1316] Enforce safe use of Get*Property(string) on PropertyGraph.

arthurp:
With: https://github.com/KatanaGraph/katana-enterprise/pull/1777

They now return Result types.

This fixes bugs in some algorithms that failed to check the return value properly.


ddn0 (approved):

amberhassaan (approved):
Looks good.

PR #486, opened 2021-08-25
Rename Id to ID

ddn0:
Closes https://github.com/KatanaGraph/katana-enterprise/issues/1049

### DIFF
--- a/libgalois/include/katana/BuildGraph.h
+++ b/libgalois/include/katana/BuildGraph.h
@@ -193,7 +193,7 @@ public:
 
   bool StartNode();
   bool StartNode(const std::string& id);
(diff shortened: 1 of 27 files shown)


arthurp (approved):
May the search and replace be with you.

I didn't really check this. Did you just use a tool or was anything manual?

roshandathathri (approved):
Yay! Assuming it's not manual :P

ddn0:
> I didn't really check this. Did you just use a tool or was anything manual?

I did a series of `grep`s to find uses of `Id` and then examined each case and updated them; in cases where there was a clear pattern, I did a search-and-replace.

tylershunt (approved):
LGTM

PR #530, opened 2021-09-08
Allow the property cache to evict during RDG initialization

witchel:
The cache provides the callback function with the size of the object
being evicted.  The cache also provides an RDG pointer, which is
vital if the RDG is in the process of being constructed.

Also elminate Unload*Property(int i) because properties should
be referred to by name, not position.

Jira: KAT-855

### DIFF
--- a/libsupport/include/katana/Cache.h
+++ b/libsupport/include/katana/Cache.h
@@ -22,7 +22,7 @@
 
 namespace katana {
 
-template <typename Key, typename Value>
+template <typename Key, typename Value, typename CallerPointer = void*>
 class KATANA_EXPORT Cache {
   using ListType = std::list<Key>;
   struct MapValue {
@@ -37,7 +37,9 @@ public:
   /// Construct an LRU cache that has a fixed number of entries.
   Cache(
       size_t capacity,  // number of entries
-      std::function<void(const Key& key)> evict_cb = nullptr)
+      std::function<
+          void(const Key& key, uint64_t approx_bytes, CallerPointer rdg)>
+          evict_cb = nullptr)
       : policy_(ReplacementPolicy::kLRUSize),
         capacity_(capacity),
         value_to_bytes_(nullptr),
@@ -48,7 +50,9 @@ public:
   Cache(
       size_t capacity,  // bytes of entries
       std::function<size_t(const Value& value)> value_to_bytes,
-      std::function<void(const Key& key)> evict_cb = nullptr)
+      std::function<
+          void(const Key& key, uint64_t approx_bytes, CallerPointer rdg)>
+          evict_cb = nullptr)
       : policy_(ReplacementPolicy::kLRUBytes),
         capacity_(capacity),
         value_to_bytes_(std::move(value_to_bytes)),
@@ -75,7 +79,7 @@ public:
     return key_to_value_.find(key) != key_to_value_.end();
   }
 
-  void Insert(const Key& key, const Value& value) {
+  void Insert(const Key& key, const Value& value, CallerPointer rdg = nullptr) {
     auto mapit = key_to_value_.find(key);
     if (mapit == key_to_value_.end()) {
       lru_list_.push_front(key);
@@ -87,7 +91,7 @@ public:
       mapit->second.value = value;
       UpdateLRU(mapit);
     }
-    EvictIfNecessary();
+    EvictIfNecessary(rdg);
   }
 
   std::optional<Value> Get(const Key& key) {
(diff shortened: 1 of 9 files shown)


witchel:
I know it is deja vu all over again, but this is a result of my struggles with #1868.

This mostly covers an edge case where we experience cache misses before the RDG is fully initialized, but it makes the callback path more robust, so seems worthwhile.

libtsuba/src/AddProperties.cpp
```diff
@@ -117,24 +122,34 @@ tsuba::AddProperties(
...
-            {"type", (cache_key->node_edge == tsuba::NodeEdge::kNode) ? "node"
-                                                                      : "edge"},
-            {"name", prop->name()},
-        });
-        cache_key->name = prop->name();
-        cache->Insert(*cache_key, props);
+        // Do not put uint8 types in property cache
+        if (props->column(0)->type()->Equals(arrow::uint8())) {
+          KATANA_WARN_ONCE(
+              "deprecated graph format; type is uint8: {}",
```
tylershunt, line 133: it would be nice to support the utint8 type as a normal property now that we're not reserving it for anything.

libsupport/include/katana/Cache.h
```diff
@@ -37,7 +43,9 @@ class KATANA_EXPORT Cache {
   /// Construct an LRU cache that has a fixed number of entries.
   Cache(
       size_t capacity,  // number of entries
-      std::function<void(const Key& key)> evict_cb = nullptr)
+      std::function<
+          void(const Key& key, uint64_t approx_bytes, tsuba::RDG* rdg)>
```
tylershunt: I'm a little confused about the cardinality relationship between a `Cache` a `PropertyGraph` and an `RDG`. Is it a cache per RDG? or can a cache contain the properties of multiple RDGs? If it's the former (seems like it is) it's a little weird to expect users to keep track of both the cache and which RDG it belongs to.

If at all possible I would prefer just to capture the `rdg` pointer in a lambda since the callback would be the onlything that cares about it but I suspect you did it this way because there's a construction ordering issue.

As a separate point It's a little weird that it's an `RDG` pointer on the interface to this general class, why not make it a template variable? Something like:
`template <typename Key, typename Value, typename CallerState = void>`

tylershunt (approved):

libtsuba/src/AddProperties.cpp
witchel: @roshandathathri told me to omit uint8 properties from the property cache.  These properties are the old way we did types.  Now we have the entity type manager.  
I was seeing a bad performance effect when adding them to the cache and given that they are a legacy feature, I like Roshan's idea to not cache them.  Users can't generate uint8 type properties from Cypher.

tylershunt: That's fine I guess; just a note that users can create uint8 properties though other valid interfaces, e.g., parquet import.

libsupport/include/katana/Cache.h
witchel: The cache contains the properties for multiple AtributedGraph instances, which really should be the same RDG, though I do not have a check for this.  Now that I think about it, it has to be the same RDG.  If we want to get into caching property columns across RDGs we need to have checksums or some way of representing in storage that property values are shared.  That is not in scope for this PR and so far as I know is not on our roadmap.

The RDG pointer comes from `tsuba::RDG::DoMake`, which captures the RDG this pointer as it creates a lambda to call `AddProperties`.  I agree with you that having an RDG* parameter in the callback is ugly.  But what else can I do?  The RDG does not exist when I instantiate the cache, so how can I pass a pointer to it at that point?  I used to capture the `this `pointer for QueryClient, but I can get cache misses during RDG construction when the PropertyGraph (and hence AttributedGraph) are not yet created.

libtsuba/src/AddProperties.cpp
witchel: I added a comment to that effect, but @roshandathathri, I'd like to hear what you think.  Omitting uint8 properties "fixes" my performance issue, but that isn't a good reason to do it for anyone but me.

libsupport/include/katana/Cache.h
tylershunt: That's what I meant by
> but I suspect you did it this way because there's a construction ordering issue.

Could we make it c-style caller state with an additional class template argument instead?

```
template <typename Key, typename Value, typename CallerState = void>
class Cache {
...
  Cache(size_t, std::function(const Key& key, uint64_t aprox_bytes, CallerState* state) {...}
};
```

witchel: That was a very interesting idea, thanks!.  I had to define the template parameter as a pointer (i.e., as void* not void) because otherwise importing the RDG definition everywhere a PropertyCache is used gets awkward.

I added the rdg directory as part of the property cache key, thanks for bringing this up.  I also encapsulated the PropertyCacheKey, making it a class instead of a struct.  Such is the way of progress.

tylershunt: Nice! I think that all sounds good.

PR #538, opened 2021-09-10
[KAT-1036] Enable a CUDA synoptic application for testing CUDA build in the Open repo

srogatch:


srogatch:
This is not yet complete: I am still to determine which Open repo headers are used in the Enterprise repo CUDA files.

srogatch:
> This is not yet complete: I am still to determine which Open repo headers are used in the Enterprise repo CUDA files.

Main headers have been determined.

scripts/setup_ubuntu.sh
```diff
@@ -150,3 +150,8 @@ fi
 # --no-binary is required to cause the pip package to use the debian package's native binaries.
 # https://lists.apache.org/thread.html/r4d2e768c330b6545649e066a1d9d1846ca7a3ea1d97e265205211166%40%3Cdev.arrow.apache.org%3E
 PYARROW_WITH_PARQUET=1 run_as_original_user pip3 install --no-binary pyarrow 'pyarrow>=4.0,<5.0.0a0'
+
+# Install CUDA
+curl https://developer.download.nvidia.com/compute/cuda/11.4.2/local_installers/cuda_11.4.2_470.57.02_linux.run \
+  --output /tmp/cuda_11.4.2_470.57.02_linux.run
+sh /tmp/cuda_11.4.2_470.57.02_linux.run --silent --toolkit
```
arthurp: I don't love installing CUDA this way. The CUDA installer does such weird things to the system it's used on. Yuck. But I'm also not sure of what a better way to would be. (Other than using the conda packages in conda, but we are not using this in conda.)

Could we at least make the CUDA install optional? Then we could only turn it on in a few runs to reduce the time spent installing it and number of jobs affected by a failure in this install.

arthurp (approved):
Overall, this looks good.

I'm a bit confused by the choice of the word "synoptic". I can sort of see how it's related in that you are looking at a larger range of things in one test and "synoptic" means observing large areas or from many angles. But I think you may have meant "synthetic application". Not sure though.

@ddn0 could you weight in on this as well?I would like to hear your thoughts on how to install CUDA.

This is definitely an example where using a docker image would make things faster and smoother.

srogatch:
> I'm a bit confused by the choice of the word "synoptic". I can sort of see how it's related in that you are looking at a larger range of things in one test and "synoptic" means observing large areas or from many angles. But I think you may have meant "synthetic application". Not sure though.

Yes, I meant `synoptic` as a consolidated, summary, or combined program, that consolidates headers from multiple libraries to predict errors in the Enterprise repo. If someone comes up with a better word, I don't mind.

scripts/setup_ubuntu.sh
srogatch: I can make the CUDA installation an option to `setup_ubuntu.sh`, no problem. But we have to test that every compiler works with CUDA in every configuration (Debug/Release/Sanitizers for `clang`), because a CUDA incompatibility can pop up in any of them, while not in some other. Though we can disable Open repo CUDA checks for the combinations that are not built in the Enterprise repo CI with CUDA, i.e. `g++-9`, `clang++-10`, `clang++-12`/Debug&Release (only Sanitizers are currently built for `clang++-12`). Please, let me know what you think.

srogatch: By the way, I'm installing here CUDA without a driver, so it should be quite harmless to the destination system.

ddn0:
> Overall, this looks good.
> 
> I'm a bit confused by the choice of the word "synoptic". I can sort of see how it's related in that you are looking at a larger range of things in one test and "synoptic" means observing large areas or from many angles. But I think you may have meant "synthetic application". Not sure though.

I'd actually prefer that we call this libgpu or some variant thereof. This would make even more sense in the future if we decide to move the GPU support code here.

As this is now, it is simply a test of a very simple application (and not a library at all) so I would put this under lonestar as an "application" there. In which case, just calling this "GPU/cuda test application" would make more sense.

> 
> @ddn0 could you weight in on this as well?I would like to hear your thoughts on how to install CUDA.
> 
> This is definitely an example where using a docker image would make things faster and smoother.

If you want, later, we can move the docker-based build here (deploy/package/....) and just get rid of the VM-based CI here.

srogatch:
> I'd actually prefer that we call this libgpu or some variant thereof. This would make even more sense in the future if we decide to move the GPU support code here.

`libgpu` would collide in naming with the Enterprise repo, so at least I would propose `libgpuopen`.

> As this is now, it is simply a test of a very simple application (and not a library at all) so I would put this under lonestar as an "application" there. In which case, just calling this "GPU/cuda test application" would make more sense.

`lonestar` has quite elaborate rules for participation in the build, so it's easy to miss the CUDA test if it's under `lonestar`:
```
if(KATANA_IS_MAIN_PROJECT AND BUILD_TESTING)
  add_subdirectory(lonestar)
  # If we are testing and there is no custom input location, add rules to
  # download inputs
  if(KATANA_USE_INPUTS)
    add_subdirectory(inputs)
  endif()
else()
  # If we are not testing, only build lonestar targets if they are required by
  # targets outside of lonestar.
  add_subdirectory(lonestar EXCLUDE_FROM_ALL)
endif()
```
I propose to put the test under `tools/test` for now. I'll commit shortly to show what I mean.

arthurp:
> `lonestar` has quite elaborate rules for participation in the build, so it's easy to miss the CUDA test if it's under `lonestar`: 

We have a number of algorithm tests in lonestar that run consistently. I think avoiding it because it's complex is a terrible idea. Because by doing so we would actually be making the build MORE complex by adding more test directories. Also this is not a tool. If you really want to keep it out of lone star, one option would be to put the test in the top-level `test` directory. You would need to add it to cmake it doesn't have any cmake stuff in it at the moment.

scripts/setup_ubuntu.sh
```diff
@@ -2,6 +2,13 @@
 
 set -xeuo pipefail
 
+# Check the options
+INSTALL_CUDA=1
+if [ "${1-}" == "--no-cuda" ]; then
+  INSTALL_CUDA=
+  shift 1
+fi
```
arthurp: Can you have a `--cuda` flag instead of `--no-cuda`? Defaulting to adding cuda seems like a bad idea. Remember some people use this script to setup dev systems.

scripts/setup_ubuntu.sh
arthurp: This is true. I'm not sure where to set this. So I'm fine with whatever for the moment, we can tweak it later if needed.

srogatch:
> We have a number of algorithm tests in lonestar that run consistently. I think avoiding it because it's complex is a terrible idea. Because by doing so we would actually be making the build MORE complex by adding more test directories. Also this is not a tool. If you really want to keep it out of lone star, one option would be to put the test in the top-level `test` directory. You would need to add it to cmake it doesn't have any cmake stuff in it at the moment.

I would rather proceed with a top-level `test` directory, because the inclusion criteria for this test are different than those for `lonestar` subdirectory. Whether the CUDA test is enabled is controlled by `CMAKE_CUDA_ARCHITECTURES` flag, and may be needed independently on whether it is a build test and the project is top-level.

scripts/setup_ubuntu.sh
srogatch: Changed to `--with-cuda` flag.

ddn0:
> I would rather proceed with a top-level `test` directory, because the inclusion criteria for this test are different than those for `lonestar` subdirectory. Whether the CUDA test is enabled is controlled by `CMAKE_CUDA_ARCHITECTURES` flag, and may be needed independently on whether it is a build test and the project is top-level.

The inclusion criteria for a CUDA test should be (I hope) the same as a CUDA application. And so far, whether an application uses CUDA been considered a build detail in the same way that the libraries an application depends on is not a major organizing principle of our source tree.

If it doesn't appear to be the case, we might enjoy making them look more the same as adding more variants and more conditions under which some things build or are tested and behave differently from other parts of the codebase makes things more confusing for future developers.

My hierarchy of preference is:
- Reorg our applications directory (i.e., lonestar) to be more sane
- Add a test application under lonestar
- Create a placeholder libgpu directory here and this test goes there. Note that libgpu is a fine name now as libgpu does not `add_library`.
- Create/Use `test` for general cpp tests

srogatch:
@ddn0 I meant that if we add the CUDA test under `lonestar` now, then it may not be built even though `CMAKE_CUDA_ARCHITECTURES` flag is passed. Please, see the latest version of the sources - the CUDA test is in the top-level `test` directory there.

ddn0:
> @ddn0 I meant that if we add the CUDA test under lonestar now, then it may not be built even though CMAKE_CUDA_ARCHITECTURES flag is passed. Please, see the latest version of the sources - the CUDA test is in the top-level test directory there.

> Reorg our applications directory (i.e., lonestar) to be more sane

That might be the case now. I am suggesting we change the behavior of lonestar if that is the case.

For the specific snippet you raised: 
```
if(KATANA_IS_MAIN_PROJECT AND BUILD_TESTING)
  add_subdirectory(lonestar)
  # If we are testing and there is no custom input location, add rules to
  # download inputs
  if(KATANA_USE_INPUTS)
    add_subdirectory(inputs)
  endif()
else()
  # If we are not testing, only build lonestar targets if they are required by
  # targets outside of lonestar.
  add_subdirectory(lonestar EXCLUDE_FROM_ALL)
endif()
```

I actually don't see the problem. Either we are testing in open (`KATANA_IS_MAIN_PROJECT AND BUILD_TESTING` is true) , in which case we need this fake CUDA application or we are testing in enterprise (`KATANA_IS_MAIN_PROJECT AND BUILD_TESTING` is false) and we don't really need this fake CUDA application because we have plenty of real CUDA programs to work with.

If there are places where we conditionally include CMake directories based on build flags, we should probably remove them as that goes against the modern target-based CMake (`target_link_libraries`) versus the legacy directory based CMake style pre version 3 (`link_libraries`). 

In the future, we  may have CUDA code in this repo and for the sake of argument let's assume we package it up as a library named libgpu, in which case, we can move the fake CUDA application as a test for that library under `libgpu/test`.

PR #547, opened 2021-09-13
tsuba: RDGPartHeader fix partition file regex

pyrito:
This PR intends to fix a bug found with the deployment. There was an issue where every partition file was being copied over with the same node id. This was because there was an issue with the regex parsing which caused every partition file to be copied over with the same node id of 0, hence the copies were getting butchered. I also wrote a test to catch this edge case.

Linked PR: https://github.com/KatanaGraph/katana-enterprise/pull/1980

### DIFF
--- a/libtsuba/src/RDGPartHeader.cpp
+++ b/libtsuba/src/RDGPartHeader.cpp
@@ -74,7 +74,7 @@ namespace tsuba {
 
 // Regex for partition files
 const std::regex kPartitionFile(
-    "part_(?:(vers[0-9A-Za-z_]+))_(?:(rdg[0-9A-Za-z-]*))_(?:(node[0-9]*))$");
+    "part_vers(?:([0-9A-Za-z_]+))_(?:(rdg[0-9A-Za-z-]*))_node(?:([0-9]*))$");
 const int kPartitionMatchHostIndex = 3;
 
 katana::Result<RDGPartHeader>
--- a/libtsuba/src/RDGPartHeader.cpp
+++ b/libtsuba/src/RDGPartHeader.cpp
@@ -74,7 +74,7 @@ namespace tsuba {
 
 // Regex for partition files
 const std::regex kPartitionFile(
-    "part_vers(?:([0-9A-Za-z_]+))_(?:(rdg[0-9A-Za-z-]*))_node(?:([0-9]*))$");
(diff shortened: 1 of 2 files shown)


libtsuba/src/RDGPartHeader.cpp
```diff
@@ -74,7 +74,7 @@ namespace tsuba {
 
 // Regex for partition files
 const std::regex kPartitionFile(
-    "part_(?:(vers[0-9A-Za-z_]+))_(?:(rdg[0-9A-Za-z-]*))_(?:(node[0-9]*))$");
+    "part_vers(?:([0-9A-Za-z_]+))_(?:(rdg[0-9A-Za-z-]*))_node(?:([0-9]*))$");
```
tylershunt: this regex is still weird to me; why do we:
 * Allow arbitrary characters in the version but not in the node number?
 * why are all these non match groups used?
 * Why is the node id allowed to be empty?
 I think a clearer regex would be:
```
"part_vers([0-9]+)_(rdg[0-9A-Za-z-]*)_node([0-9]+)$"
```

tylershunt (approved):

libtsuba/src/RDGPartHeader.cpp
pyrito: I'll be honest, I'm pretty bad with regex. I kind of went with something that got the job done to some extent. I think your restrictions on the regex make sense. I'll implement this.

libtsuba/src/tsuba.cpp
```diff
@@ -220,6 +220,7 @@ tsuba::CreateSrcDestFromViewsForCopy(
     auto rdg_manifest_res =
         tsuba::RDGManifest::Make(uri, rdg_view.view_type, version);
     if (!rdg_manifest_res) {
+      KATANA_LOG_WARN("not a valid manifest file, uri: {}", uri.string());
```
witchel, line 223: Would we be grabbing a Tracer object here instead of WARN?  I'm really not sure.

witchel (approved):
Trust in regex.

libtsuba/src/tsuba.cpp
pyrito: I didn't think this was appropriate for a Tracer print so I stuck with a warning log.

PR #555, opened 2021-09-22
[KAT-1767]: Ensure all uint8 properties are loaded for type inference

scober:
Currently, we infer types from whatever properties happen to be loaded in
memory. This is not the desired behavior. Instead, we should load any
uint8 properties that aren't loaded before we begin type inference
because all uint8 properties should be treated as types.

### DIFF
--- a/libgalois/src/PropertyGraph.cpp
+++ b/libgalois/src/PropertyGraph.cpp
@@ -6,6 +6,8 @@
 #include <memory>
 #include <utility>
 
+#include <arrow/array.h>
+
 #include "katana/ArrowInterchange.h"
 #include "katana/Iterators.h"
 #include "katana/Logging.h"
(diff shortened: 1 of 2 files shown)


libgalois/src/PropertyGraph.cpp
```diff
@@ -6,6 +6,8 @@
 #include <memory>
 #include <utility>
 
+#include <arrow/array.h>
```
witchel, line 9: Is this needed?

witchel (commented):
Looks good to me.

libgalois/src/PropertyGraph.cpp
scober: Weirdly, it seems to be the home for the schema and field apis.

witchel: Ok, then.

amberhassaan (approved):
looks good.

PR #558, opened 2021-09-22
View for node and edge projections

uditagar:
This is the functionality to support node/edge projections.

### DIFF
--- a/libgalois/include/katana/GraphTopology.h
+++ b/libgalois/include/katana/GraphTopology.h
@@ -607,6 +607,9 @@ private:
   const Topo* topo_ptr_;
 };
 
+/// filter nodes and edges based on their type values
+/// and creates a new local graph based on the filtered nodes and edges
+/// also maintains mappings from old to new and new to old nodes and edges
 class KATANA_EXPORT ProjectedShuffleTopology : public GraphTopologyTypes {
 public:
   ProjectedShuffleTopology() = default;
@@ -615,25 +618,6 @@ public:
 
   ProjectedShuffleTopology(const ProjectedShuffleTopology&) = delete;
   ProjectedShuffleTopology& operator=(const ProjectedShuffleTopology&) = delete;
-
-  //ProjectedShuffleTopology(const Edge* adj_indices, size_t num_nodes, const Node* dests,
-  //size_t num_edges) noexcept;
-
-  ProjectedShuffleTopology(
-      NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests,
-      NUMAArray<Node>&& old_to_new_nodes_mapping,
-      NUMAArray<Node>&& new_to_old_nodes_mapping,
-      NUMAArray<Edge>&& old_to_new_edges_mapping,
-      NUMAArray<Edge>&& new_to_old_edges_mapping) noexcept
-      : adj_indices_(std::move(adj_indices)),
-        dests_(std::move(dests)),
-        old_to_new_nodes_mapping_(std::move(old_to_new_nodes_mapping)),
-        new_to_old_nodes_mapping_(std::move(new_to_old_nodes_mapping)),
-        old_to_new_edges_mapping_(std::move(old_to_new_edges_mapping)),
-        new_to_old_edges_mapping_(std::move(new_to_old_edges_mapping)) {}
-
-  //static ProjectedShuffleTopology Copy(
-  //  const ProjectedShuffleTopology& that) noexcept;
 
   uint64_t num_nodes() const noexcept { return adj_indices_.size(); }
 
@@ -746,6 +730,19 @@ public:
       const std::vector<std::string>& edge_types);
 
 private:
+  ProjectedShuffleTopology(
+      NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests,
+      NUMAArray<Node>&& old_to_new_nodes_mapping,
+      NUMAArray<Node>&& new_to_old_nodes_mapping,
+      NUMAArray<Edge>&& old_to_new_edges_mapping,
+      NUMAArray<Edge>&& new_to_old_edges_mapping) noexcept
+      : adj_indices_(std::move(adj_indices)),
+        dests_(std::move(dests)),
+        old_to_new_nodes_mapping_(std::move(old_to_new_nodes_mapping)),
+        new_to_old_nodes_mapping_(std::move(new_to_old_nodes_mapping)),
+        old_to_new_edges_mapping_(std::move(old_to_new_edges_mapping)),
+        new_to_old_edges_mapping_(std::move(new_to_old_edges_mapping)) {}
+
   NUMAArray<Edge> adj_indices_;
   NUMAArray<Node> dests_;
   NUMAArray<Node> old_to_new_nodes_mapping_;
--- a/libgalois/include/katana/GraphTopology.h
+++ b/libgalois/include/katana/GraphTopology.h
@@ -741,15 +741,9 @@ public:
     return node_property_index(nid);
   }
 
-  static std::unique_ptr<ProjectedShuffleTopology> MakeProjectedGraph(
-      const PropertyGraph* pg, const std::vector<std::string>& node_properties,
-      const std::vector<std::string>& edge_properties);
-
   static std::unique_ptr<ProjectedShuffleTopology> Make(
-      const PropertyGraph* pg, const std::vector<std::string>& node_properties,
-      const std::vector<std::string>& edge_properties) {
-    return MakeProjectedGraph(pg, node_properties, edge_properties);
-  }
+      const PropertyGraph* pg, const std::vector<std::string>& node_types,
+      const std::vector<std::string>& edge_types);
 
 private:
   NUMAArray<Edge> adj_indices_;
--- a/libgalois/include/katana/GraphTopology.h
+++ b/libgalois/include/katana/GraphTopology.h
@@ -6,6 +6,7 @@
 
 #include <boost/iterator/counting_iterator.hpp>
 
+#include "katana/DynamicBitset.h"
 #include "katana/Iterators.h"
 #include "katana/NUMAArray.h"
 #include "katana/config.h"
@@ -40,6 +41,7 @@ struct KATANA_EXPORT GraphTopologyTypes {
 
 class KATANA_EXPORT EdgeShuffleTopology;
 class KATANA_EXPORT EdgeTypeAwareTopology;
+class KATANA_EXPORT ProjectedShuffleTopology;
 
 /// A graph topology represents the adjacency information for a graph in CSR
 /// format.
@@ -605,6 +607,159 @@ private:
   const Topo* topo_ptr_;
 };
 
+class KATANA_EXPORT ProjectedShuffleTopology : public GraphTopologyTypes {
+public:
+  ProjectedShuffleTopology() = default;
+  ProjectedShuffleTopology(ProjectedShuffleTopology&&) = default;
+  ProjectedShuffleTopology& operator=(ProjectedShuffleTopology&&) = default;
+
+  ProjectedShuffleTopology(const ProjectedShuffleTopology&) = delete;
+  ProjectedShuffleTopology& operator=(const ProjectedShuffleTopology&) = delete;
+
+  //ProjectedShuffleTopology(const Edge* adj_indices, size_t num_nodes, const Node* dests,
+  //size_t num_edges) noexcept;
+
+  ProjectedShuffleTopology(
+      NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests,
+      NUMAArray<Node>&& old_to_new_nodes_mapping,
+      NUMAArray<Node>&& new_to_old_nodes_mapping,
+      NUMAArray<Edge>&& old_to_new_edges_mapping,
+      NUMAArray<Edge>&& new_to_old_edges_mapping) noexcept
+      : adj_indices_(std::move(adj_indices)),
+        dests_(std::move(dests)),
+        old_to_new_nodes_mapping_(std::move(old_to_new_nodes_mapping)),
+        new_to_old_nodes_mapping_(std::move(new_to_old_nodes_mapping)),
+        old_to_new_edges_mapping_(std::move(old_to_new_edges_mapping)),
+        new_to_old_edges_mapping_(std::move(new_to_old_edges_mapping)) {}
+
+  //static ProjectedShuffleTopology Copy(
+  //  const ProjectedShuffleTopology& that) noexcept;
+
+  uint64_t num_nodes() const noexcept { return adj_indices_.size(); }
+
+  uint64_t num_edges() const noexcept { return dests_.size(); }
+
+  const Edge* adj_data() const noexcept { return adj_indices_.data(); }
+
+  const Node* dest_data() const noexcept { return dests_.data(); }
+
+  /// Checks equality against another instance of ProjectedShuffleTopology.
+  /// WARNING: Expensive operation due to element-wise checks on large arrays
+  /// @param that: ProjectedShuffleTopology instance to compare against
+  /// @returns true if topology arrays are equal
+  bool Equals(const ProjectedShuffleTopology& that) const noexcept {
+    if (this == &that) {
+      return true;
+    }
+    if (num_nodes() != that.num_nodes()) {
+      return false;
+    }
+    if (num_edges() != that.num_edges()) {
+      return false;
+    }
+
+    return adj_indices_ == that.adj_indices_ && dests_ == that.dests_;
+  }
+
+  /// Gets the edge range of some node.
+  ///
+  /// \param node node to get the edge range of
+  /// \returns iterable edge range for node.
+  edges_range edges(Node node) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(node <= adj_indices_.size());
+    edge_iterator e_beg{node > 0 ? adj_indices_[node - 1] : 0};
+    edge_iterator e_end{adj_indices_[node]};
+
+    return MakeStandardRange(e_beg, e_end);
+  }
+
+  Node edge_source(const Edge& eid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(eid < num_edges());
+
+    if (eid < adj_indices_[0]) {
+      return Node{0};
+    }
+
+    auto it = std::upper_bound(adj_indices_.begin(), adj_indices_.end(), eid);
+    KATANA_LOG_DEBUG_ASSERT(it != adj_indices_.end());
+    KATANA_LOG_DEBUG_ASSERT(*it > eid);
+
+    auto d = std::distance(adj_indices_.begin(), it);
+    KATANA_LOG_DEBUG_ASSERT(static_cast<size_t>(d) < num_nodes());
+    KATANA_LOG_DEBUG_ASSERT(d > 0);
+
+    return static_cast<Node>(d);
+  }
+
+  Node edge_dest(Edge edge_id) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(edge_id < dests_.size());
+    return dests_[edge_id];
+  }
+
+  nodes_range nodes(Node begin, Node end) const noexcept {
+    return MakeStandardRange<node_iterator>(begin, end);
+  }
+
+  nodes_range all_nodes() const noexcept {
+    return nodes(Node{0}, static_cast<Node>(num_nodes()));
+  }
+
+  edges_range all_edges() const noexcept {
+    return MakeStandardRange<edge_iterator>(Edge{0}, Edge{num_edges()});
+  }
+  // Standard container concepts
+
+  node_iterator begin() const noexcept { return node_iterator(0); }
+
+  node_iterator end() const noexcept { return node_iterator(num_nodes()); }
+
+  size_t size() const noexcept { return num_nodes(); }
+
+  bool empty() const noexcept { return num_nodes() == 0; }
+
+  ///@param node node to get degree for
+  ///@returns Degree of node N
+  size_t degree(Node node) const noexcept { return edges(node).size(); }
+
+  PropertyIndex edge_property_index(const Edge& eid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(eid < num_edges());
+    return new_to_old_edges_mapping_[eid];
+  }
+
+  // Need to redefine the method here so it overrides and hides base-class
+  // version of this method
+  Edge original_edge_id(const Edge& eid) const noexcept {
+    return edge_property_index(eid);
+  }
+
+  PropertyIndex node_property_index(const Node& nid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(nid < num_nodes());
+    return new_to_old_nodes_mapping_[nid];
+  }
+
+  Node original_node_id(const Node& nid) const noexcept {
+    return node_property_index(nid);
+  }
+
+  static std::unique_ptr<ProjectedShuffleTopology> MakeProjectedGraph(
+      const PropertyGraph* pg, const std::vector<std::string>& node_properties,
+      const std::vector<std::string>& edge_properties);
+
+  static std::unique_ptr<ProjectedShuffleTopology> Make(
+      const PropertyGraph* pg, const std::vector<std::string>& node_properties,
+      const std::vector<std::string>& edge_properties) {
+    return MakeProjectedGraph(pg, node_properties, edge_properties);
+  }
+
+private:
+  NUMAArray<Edge> adj_indices_;
+  NUMAArray<Node> dests_;
+  NUMAArray<Node> old_to_new_nodes_mapping_;
+  NUMAArray<Node> new_to_old_nodes_mapping_;
+  NUMAArray<Edge> old_to_new_edges_mapping_;
+  NUMAArray<Edge> new_to_old_edges_mapping_;
+};
+
 namespace internal {
 // TODO(amber): make private
 template <typename Topo>
@@ -1032,6 +1187,8 @@ using PGViewNodesSortedByDegreeEdgesSortedByDestID =
 using PGViewBiDirectional = BasicPropGraphViewWrapper<SimpleBiDirTopology>;
 using PGViewEdgeTypeAwareBiDir =
     BasicPropGraphViewWrapper<EdgeTypeAwareBiDirTopology>;
+using PGViewProjectedGraph =
+    BasicPropGraphViewWrapper<ProjectedShuffleTopology>;
 
 template <typename PGView>
 struct PGViewBuilder {};
@@ -1095,6 +1252,17 @@ struct PGViewBuilder<PGViewEdgeTypeAwareBiDir> {
   }
 };
 
+template <>
+struct PGViewBuilder<PGViewProjectedGraph> {
+  template <typename ViewCache>
+  static PGViewProjectedGraph BuildView(
+      const PropertyGraph* pg, ViewCache& viewCache) noexcept {
+    auto topo = viewCache.BuildOrGetProjectedGraphTopo(pg);
+
+    return PGViewProjectedGraph{pg, ProjectedShuffleTopology{topo}};
+  }
+};
+
 }  // end namespace internal
 
 struct PropertyGraphViews {
@@ -1103,6 +1271,7 @@ struct PropertyGraphViews {
   using EdgeTypeAwareBiDir = internal::PGViewEdgeTypeAwareBiDir;
   using NodesSortedByDegreeEdgesSortedByDestID =
       internal::PGViewNodesSortedByDegreeEdgesSortedByDestID;
+  using ProjectedGraph = internal::PGViewProjectedGraph;
 };
 
 class KATANA_EXPORT PGViewCache {
@@ -1111,6 +1280,7 @@ class KATANA_EXPORT PGViewCache {
   std::vector<std::unique_ptr<EdgeTypeAwareTopology>> edge_type_aware_topos_;
   std::unique_ptr<CondensedTypeIDMap> edge_type_id_map_;
   // TODO(amber): define a node_type_id_map_;
+  std::unique_ptr<ProjectedShuffleTopology> projected_topos_;
 
   template <typename>
   friend struct internal::PGViewBuilder;
@@ -1148,6 +1318,10 @@ private:
   EdgeTypeAwareTopology* BuildOrGetEdgeTypeAwareTopo(
       const PropertyGraph* pg,
       const EdgeShuffleTopology::TransposeKind& tpose_kind) noexcept;
+
+  ProjectedShuffleTopology* BuildOrGetProjectedGraphTopo(
+      const PropertyGraph* pg, const std::vector<std::string>& node_properties,
+      const std::vector<std::string>& edge_properties) noexcept;
 };
 
 /// Creates a uniform-random CSR GrpahTopology instance, where each node as
--- a/libgalois/include/katana/GraphTopology.h
+++ b/libgalois/include/katana/GraphTopology.h
@@ -710,10 +710,12 @@ public:
     return new_to_old_edges_mapping_[eid];
   }
 
-  // Need to redefine the method here so it overrides and hides base-class
-  // version of this method
   Edge original_edge_id(const Edge& eid) const noexcept {
     return edge_property_index(eid);
+  }
+
+  Edge projected_edge_id(const Edge& eid) const noexcept {
+    return old_to_new_edges_mapping_[eid];
   }
 
   PropertyIndex node_property_index(const Node& nid) const noexcept {
@@ -723,6 +725,10 @@ public:
 
   Node original_node_id(const Node& nid) const noexcept {
     return node_property_index(nid);
+  }
+
+  Node projected_node_id(const Node& nid) const noexcept {
+    return old_to_new_nodes_mapping_[nid];
   }
 
   static std::unique_ptr<ProjectedShuffleTopology> Make(
--- a/libgalois/include/katana/GraphTopology.h
+++ b/libgalois/include/katana/GraphTopology.h
@@ -41,7 +41,7 @@ struct KATANA_EXPORT GraphTopologyTypes {
 
 class KATANA_EXPORT EdgeShuffleTopology;
 class KATANA_EXPORT EdgeTypeAwareTopology;
-class KATANA_EXPORT ProjectedShuffleTopology;
+class KATANA_EXPORT ProjectedSTypeTopology;
 
 /// A graph topology represents the adjacency information for a graph in CSR
 /// format.
@@ -608,16 +608,16 @@ private:
 };
 
 /// filter nodes and edges based on their type values
-/// and creates a new local graph based on the filtered nodes and edges
-/// also maintains mappings from old to new and new to old nodes and edges
-class KATANA_EXPORT ProjectedShuffleTopology : public GraphTopologyTypes {
+/// and creates a new projected graph based on the filtered nodes and edges
+/// also maintains mappings from original to projected and projected to original nodes and edges
+class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
 public:
-  ProjectedShuffleTopology() = default;
-  ProjectedShuffleTopology(ProjectedShuffleTopology&&) = default;
-  ProjectedShuffleTopology& operator=(ProjectedShuffleTopology&&) = default;
+  ProjectedTopology() = default;
+  ProjectedTopology(ProjectedTopology&&) = default;
+  ProjectedTopology& operator=(ProjectedTopology&&) = default;
 
-  ProjectedShuffleTopology(const ProjectedShuffleTopology&) = delete;
-  ProjectedShuffleTopology& operator=(const ProjectedShuffleTopology&) = delete;
+  ProjectedTopology(const ProjectedTopology&) = delete;
+  ProjectedTopology& operator=(const ProjectedTopology&) = delete;
 
   uint64_t num_nodes() const noexcept { return adj_indices_.size(); }
 
@@ -627,22 +627,24 @@ public:
 
   const Node* dest_data() const noexcept { return dests_.data(); }
 
-  /// Checks equality against another instance of ProjectedShuffleTopology.
+  /// Checks equality against another instance of ProjectedTopology.
   /// WARNING: Expensive operation due to element-wise checks on large arrays
-  /// @param that: ProjectedShuffleTopology instance to compare against
+  /// @param that: ProjectedTopology instance to compare against
   /// @returns true if topology arrays are equal
-  bool Equals(const ProjectedShuffleTopology& that) const noexcept {
-    if (this == &that) {
+  /// should take O(|V| + |E|) in the worst case
+  bool Equals(const ProjectedTopology& projected_topo_) const noexcept {
+    if (this == &projected_topo_) {
       return true;
     }
-    if (num_nodes() != that.num_nodes()) {
+    if (num_nodes() != projected_topo_.num_nodes()) {
       return false;
     }
-    if (num_edges() != that.num_edges()) {
+    if (num_edges() != projected_topo_.num_edges()) {
       return false;
     }
 
-    return adj_indices_ == that.adj_indices_ && dests_ == that.dests_;
+    return adj_indices_ == projected_topo_.adj_indices_ &&
+           dests_ == projected_topo_.dests_;
   }
 
   /// Gets the edge range of some node.
@@ -650,8 +652,8 @@ public:
   /// \param node node to get the edge range of
   /// \returns iterable edge range for node.
   edges_range edges(Node node) const noexcept {
-    KATANA_LOG_DEBUG_ASSERT(node <= adj_indices_.size());
-    edge_iterator e_beg{node > 0 ? adj_indices_[node - 1] : 0};
+    KATANA_LOG_DEBUG_ASSERT(node < adj_indices_.size());
+    edge_iterator e_beg{node != 0 ? adj_indices_[node - 1] : 0};
     edge_iterator e_end{adj_indices_[node]};
 
     return MakeStandardRange(e_beg, e_end);
@@ -664,13 +666,15 @@ public:
       return Node{0};
     }
 
+    // finds the node idx which contains the edge corresponding to eid
+    // since last entry in adj_indices corresponds to the total number of edges
+    // the value corresponding to iterator it should be greater than eid
     auto it = std::upper_bound(adj_indices_.begin(), adj_indices_.end(), eid);
     KATANA_LOG_DEBUG_ASSERT(it != adj_indices_.end());
     KATANA_LOG_DEBUG_ASSERT(*it > eid);
 
-    auto d = std::distance(adj_indices_.begin(), it);
-    KATANA_LOG_DEBUG_ASSERT(static_cast<size_t>(d) < num_nodes());
-    KATANA_LOG_DEBUG_ASSERT(d > 0);
+    const size_t d = it - adj_indices_.begin();
+    KATANA_LOG_DEBUG_ASSERT(d < num_nodes());
 
     return static_cast<Node>(d);
   }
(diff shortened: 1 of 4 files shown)


libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,159 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+
+  Node original_node_id(const Node& nid) const noexcept {
+    return node_property_index(nid);
+  }
+
+  static std::unique_ptr<ProjectedShuffleTopology> MakeProjectedGraph(
+      const PropertyGraph* pg, const std::vector<std::string>& node_properties,
+      const std::vector<std::string>& edge_properties);
+
+  static std::unique_ptr<ProjectedShuffleTopology> Make(
```
amberhassaan: Let's keep `Make` and delete `MakeProjectedGraph`

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,182 @@ katana::EdgeTypeAwareTopology::MakeFrom(
       pg, edge_type_index, e_topo, std::move(per_type_adj_indices)});
 }
 
+std::unique_ptr<katana::ProjectedShuffleTopology>
+katana::ProjectedShuffleTopology::MakeProjectedGraph(
+    const katana::PropertyGraph* pg,
+    const std::vector<std::string>& node_properties,
```
amberhassaan: This interface is not clear (and may be wrong). Below you're calling `GetNodeEntityTypeID(name)`, but that method takes the name of an atomic (singular) type, not a property. 
Further, composite types don't have names. Only atomic types do. 
Let's think more about this, but for now, you should take a vector of EntityTypeIDs, or a `StandardRange`.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,182 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+      auto entity_type_id = pg->GetNodeEntityTypeID(node_prop);
+      node_entity_type_ids.insert(entity_type_id);
+    }
+
+    katana::GAccumulator<uint32_t> accum_num_new_nodes;
+    katana::GAccumulator<uint64_t> accum_num_new_edges;
+
+    katana::do_all(katana::iterate(topology.all_nodes()), [&](auto src) {
+      for (auto type : node_entity_type_ids) {
+        if (pg->DoesNodeHaveType(src, type)) {
```
amberhassaan, line 524: This also may be wrong, because if see the implementation of `DoesNodeHaveType`, it returns true if the type you provide is a subtype of the actual node type. You should be calling `GetTypeOfNode`. 

PS: as I look at our type interface, my head is starting to spin, so no surprise that you're confused about which methods to call.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,182 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+
+  // fill old to new nodes mapping
+  NUMAArray<Node> old_to_new_nodes_mapping;
+  NUMAArray<Node> new_to_old_nodes_mapping;
+
+  old_to_new_nodes_mapping.allocateInterleaved(topology.num_nodes());
+  new_to_old_nodes_mapping.allocateInterleaved(num_new_nodes);
+
+  Node node_idx{0};
+  for (Node src = 0; src < topology.num_nodes(); src++) {
```
amberhassaan: We're gonna need a parallel implementation of this loop. See the implementation of `MakeNodeSortedTopo` to see how you can populate these mappings in parallel. 
Also, perhaps we won't need old to new mapping, but if you want it, see `MakeNodeSortedTopo`.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,182 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+  bitset_edges.resize(topology.num_edges());
+
+  NUMAArray<Edge> out_indices;
+  out_indices.allocateInterleaved(num_new_nodes);
+
+  katana::ParallelSTL::fill(out_indices.begin(), out_indices.end(), Edge{0});
+
+  if (edge_properties.empty()) {
+    // set all edges incident to projected nodes
+    katana::do_all(katana::iterate(topology.all_nodes()), [&](auto src) {
```
amberhassaan: why iterate over all nodes? You have the new nodes to iterate over, which could be a much smaller set.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,182 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+    std::set<katana::EntityTypeID> edge_entity_type_ids;
+
+    for (auto edge_prop : edge_properties) {
+      auto entity_type_id = pg->GetEdgeEntityTypeID(edge_prop);
+      edge_entity_type_ids.insert(entity_type_id);
+    }
+
+    katana::GAccumulator<uint32_t> accum_num_new_edges;
+
+    katana::do_all(katana::iterate(topology.all_nodes()), [&](auto src) {
```
amberhassaan: Same. No need to iterate over all nodes.

amberhassaan (commented):
Left first set of comments. Highlevel comments:

* If do_all is going over nodes, where inside the loop, you're iterating over edges of a node, you should use `katana::steal()`. This is needed because node degrees can vary a lot. 
* The topology building function is quite long. Let's see if we can break it up into logical pieces, but even before that, let's see if we can simplify the logic of generating the topology. I'm reminded of the subgraph_extraction code in shared analytics.

libgalois/src/GraphTopology.cpp
uditagar: yes, you are right. I am taking a vector of atomic node and edge types. Then I find their entity type ids inside the function

uditagar: I will rename it to node_types and edge_types then

libgalois/src/GraphTopology.cpp
uditagar: the `type` here is an atomic type. So I just want to check if it is a subtype of the actual node type. thanks

uditagar:
> Left first set of comments. Highlevel comments:
> 
> * If do_all is going over nodes, where inside the loop, you're iterating over edges of a node, you should use `katana::steal()`. This is needed because node degrees can vary a lot.
> * The topology building function is quite long. Let's see if we can break it up into logical pieces, but even before that, let's see if we can simplify the logic of generating the topology. I'm reminded of the subgraph_extraction code in shared analytics.

I have added stealing now. It's faster than subgraph extraction code. That code currently checks all possible node pairs for edge.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,153 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+      NUMAArray<Node>&& old_to_new_nodes_mapping,
+      NUMAArray<Node>&& new_to_old_nodes_mapping,
+      NUMAArray<Edge>&& old_to_new_edges_mapping,
+      NUMAArray<Edge>&& new_to_old_edges_mapping) noexcept
+      : adj_indices_(std::move(adj_indices)),
+        dests_(std::move(dests)),
+        old_to_new_nodes_mapping_(std::move(old_to_new_nodes_mapping)),
+        new_to_old_nodes_mapping_(std::move(new_to_old_nodes_mapping)),
+        old_to_new_edges_mapping_(std::move(old_to_new_edges_mapping)),
+        new_to_old_edges_mapping_(std::move(new_to_old_edges_mapping)) {}
```
arthurp: This shouldn't be public I don't think.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
       pg, edge_type_index, e_topo, std::move(per_type_adj_indices)});
 }
 
+std::unique_ptr<katana::ProjectedShuffleTopology>
+katana::ProjectedShuffleTopology::Make(
+    const katana::PropertyGraph* pg, const std::vector<std::string>& node_types,
```
arthurp: This appears to be shuffling based on the type specifically. But the name just says that this is a shuffled topology. If the type is supposed to always be used for type sorting, this should be called `TypeSortedTopology` or something. If other shuffles are allow (like degree order or whatever), we need to provide ways to create views for any arbitrary shuffle and make sure we can cache them all safely.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,153 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
   const Topo* topo_ptr_;
 };
 
+class KATANA_EXPORT ProjectedShuffleTopology : public GraphTopologyTypes {
```
arthurp: What does this class do? This needs to documentation because it's name implies that it supports some kind of general shuffle operation, but it appears to be type based. Also we already have `BuildOrGetEdgeTypeAwareTopo` and other ones that talk about shuffling. So it's not clear how this is distinct from those.

libgalois/src/GraphTopology.cpp
uditagar: this class is baically just projecting the node and edge types, so may be `ProjectedTypeToplogy` is a better name.

libgalois/include/katana/GraphTopology.h
uditagar: this class creates a new local graph by filtering nodes and edges according to their types. I will add documentation then

libgalois/src/GraphTopology.cpp
roshandathathri: Yes, `Shuffle` is not appropriate here.

libgalois/include/katana/GraphTopology.h
roshandathathri: Please add doxygen comments using `\\\`

roshandathathri: Also, I think this should just be `ProjectedTopology`. The projection itself can be based on types (for now) or arbitrary Cypher queries (for later).

libgalois/src/GraphTopology.cpp
uditagar: so I will make it `ProjectedTypeTopology` then

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
   const Topo* topo_ptr_;
 };
 
+/// filter nodes and edges based on their type values
+/// and creates a new local graph based on the filtered nodes and edges
```
roshandathathri: Drop the word "local". All graphs are local. It can be confusing. Instead you can use "projected"

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+/// also maintains mappings from old to new and new to old nodes and edges
```
roshandathathri: Old and new can be confusing. How about "projected" and "original" instead?

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+
+  PropertyIndex node_property_index(const Node& nid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(nid < num_nodes());
+    return new_to_old_nodes_mapping_[nid];
+  }
+
+  Node original_node_id(const Node& nid) const noexcept {
+    return node_property_index(nid);
+  }
+
```
arekay: nit : Can we rename these to represent the conversion the methods are actually doing? `projected_to_original_edge`. It helps clear the intent and since both projected and original entities have the same type, it can be unclear what is being passed.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  PropertyIndex edge_property_index(const Edge& eid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(eid < num_edges());
+    return new_to_old_edges_mapping_[eid];
+  }
+
+  Edge original_edge_id(const Edge& eid) const noexcept {
+    return edge_property_index(eid);
+  }
+
+  Edge projected_edge_id(const Edge& eid) const noexcept {
```
roshandathathri: Document that the parameter must be original and for the above, the parameter must be projected.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  PropertyIndex node_property_index(const Node& nid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(nid < num_nodes());
+    return new_to_old_nodes_mapping_[nid];
+  }
+
+  Node original_node_id(const Node& nid) const noexcept {
+    return node_property_index(nid);
+  }
+
+  Node projected_node_id(const Node& nid) const noexcept {
```
roshandathathri: Ditto

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+
+  Edge original_edge_id(const Edge& eid) const noexcept {
+    return edge_property_index(eid);
+  }
+
+  Edge projected_edge_id(const Edge& eid) const noexcept {
+    return old_to_new_edges_mapping_[eid];
+  }
+
+  PropertyIndex node_property_index(const Node& nid) const noexcept {
```
roshandathathri, line 727: @amberhassaan I dislike that we're using snake_case in many function names. It breaks our convention of CamelCase. Unless it refers to a variable, I'm not in favor of using snake_case.

Udit, this is not an action item on you. It's better to be consistent with other graph classes, which this already is.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+
+  katana::DynamicBitset bitset_nodes;
+  bitset_nodes.resize(topology.num_nodes());
+
+  NUMAArray<Node> old_to_new_nodes_mapping;
+  old_to_new_nodes_mapping.allocateInterleaved(topology.num_nodes());
+
+  katana::ParallelSTL::fill(
+      old_to_new_nodes_mapping.begin(), old_to_new_nodes_mapping.end(),
+      Node{0});
```
arekay: nit : Should this be an invalid value such as `-1` or `int_max`?

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+        old_to_new_nodes_mapping_(std::move(old_to_new_nodes_mapping)),
+        new_to_old_nodes_mapping_(std::move(new_to_old_nodes_mapping)),
+        old_to_new_edges_mapping_(std::move(old_to_new_edges_mapping)),
+        new_to_old_edges_mapping_(std::move(new_to_old_edges_mapping)) {}
+
+  NUMAArray<Edge> adj_indices_;
+  NUMAArray<Node> dests_;
+  NUMAArray<Node> old_to_new_nodes_mapping_;
+  NUMAArray<Node> new_to_old_nodes_mapping_;
+  NUMAArray<Edge> old_to_new_edges_mapping_;
```
roshandathathri: Isn't this the size of the edges (and vertices above) in the original graph?

Do we need this? This is a trade-off. We can do binary search in `new_to_old_` instead. For this PR, this is fine. Please add a TODO to revisit this (also describe the trade-off).

libgalois/include/katana/GraphTopology.h
roshandathathri: If you decide to drop `Type` from the class name (based on my previous comment), then rename this to `MakeTypeProjectedTopology`.

Also, please add documentation. Specifically, node_types means you're projecting nodes that have all the given types, right? That should be in the doxygen comments.

libgalois/src/GraphTopology.cpp
roshandathathri: This seems wrong. It goes back to my documentation question.

I think we want the projected nodes to have all the given types. If so, then get the entity type for the given set of strings/names. Then select nodes with that entity type (no for-loop in the operator).

roshandathathri (changes requested):
Except the types comment (the current code seems wrong), the rest are nits.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  uint64_t num_nodes() const noexcept { return adj_indices_.size(); }
+
+  uint64_t num_edges() const noexcept { return dests_.size(); }
+
+  const Edge* adj_data() const noexcept { return adj_indices_.data(); }
+
+  const Node* dest_data() const noexcept { return dests_.data(); }
+
+  /// Checks equality against another instance of ProjectedShuffleTopology.
+  /// WARNING: Expensive operation due to element-wise checks on large arrays
```
srogatch, line 631: Can you put here the algorithmic complexity of how expensive those operations are? Big-O should be fine.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  /// @param that: ProjectedShuffleTopology instance to compare against
+  /// @returns true if topology arrays are equal
+  bool Equals(const ProjectedShuffleTopology& that) const noexcept {
```
srogatch: Words `this` and `that` are quite similar, while here one is a pointer and the other is a const reference. Please, consider using `fellow` or `rhs`.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+      return true;
+    }
+    if (num_nodes() != that.num_nodes()) {
+      return false;
+    }
+    if (num_edges() != that.num_edges()) {
+      return false;
+    }
+
+    return adj_indices_ == that.adj_indices_ && dests_ == that.dests_;
```
srogatch: The method is marked `noexcept`, but are you sure that the vector equality comparisons here are also `noexcept`? I believe they call element-wise comparisons, and an element comparison may throw an exception.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  }
+
+  /// Gets the edge range of some node.
+  ///
+  /// \param node node to get the edge range of
+  /// \returns iterable edge range for node.
+  edges_range edges(Node node) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(node <= adj_indices_.size());
+    edge_iterator e_beg{node > 0 ? adj_indices_[node - 1] : 0};
```
srogatch: Comparison `x == 0` and `x != 0` should be slightly faster for the CPU than an open range comparison (like `x > 0`). I assume `Node` is unsigned.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+    edge_iterator e_end{adj_indices_[node]};
```
srogatch, line 657: You seem to trigger Undefined Behavior here if `node == adj_indices_.size()`. Either the `assert` above is wrong and should be `KATANA_LOG_DEBUG_ASSERT(node < adj_indices_.size());`, or the code in this line is wrong.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+
+    return MakeStandardRange(e_beg, e_end);
```
srogatch, line 659: Are you sure it's `noexcept`?

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  }
+
+  Node edge_source(const Edge& eid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(eid < num_edges());
+
+    if (eid < adj_indices_[0]) {
+      return Node{0};
+    }
+
+    auto it = std::upper_bound(adj_indices_.begin(), adj_indices_.end(), eid);
```
srogatch, line 672: `std::upper_bound` is quite error-prone: can you comment above what your intent here is? So that the reader can easily check for off-by-one and similar errors. If the array `adj_indices` is sorted, please, mention this here again.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+    KATANA_LOG_DEBUG_ASSERT(it != adj_indices_.end());
```
srogatch, line 673: This also deseves a comment why there will always be an item in `adj_indices` that is greater than `eid`.

libgalois/include/katana/GraphTopology.h
uditagar: sounds good

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+    KATANA_LOG_DEBUG_ASSERT(*it > eid);
+
+    auto d = std::distance(adj_indices_.begin(), it);
```
srogatch: I would use simply `const size_t d = it - adj_indices.begin();` to ensure it's `O(1)` or it doesn't compile. We don't want linear complexity in the number of nodes for retrieving the source node for an edge, right?

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+    KATANA_LOG_DEBUG_ASSERT(static_cast<size_t>(d) < num_nodes());
+    KATANA_LOG_DEBUG_ASSERT(d > 0);
```
srogatch: This one is too paranoid: as long as someone doesn't mess up your code, this shouldn't happen. But you can add a unit test.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+
+    return static_cast<Node>(d);
+  }
+
+  Node edge_dest(Edge edge_id) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(edge_id < dests_.size());
+    return dests_[edge_id];
```
srogatch, line 684: `operator[]` isn't `noexcept` I think. Though you check it in the line above, but there could be other reasons for `operator[]` to fail (e.g. in chunked arrays).

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  }
+
+  nodes_range nodes(Node begin, Node end) const noexcept {
+    return MakeStandardRange<node_iterator>(begin, end);
```
srogatch, line 688: `MakeStandardRange()` from `external/katana/libgalois/include/katana/Range.h` is not `noexcept`. Please, change your method accordingly.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  }
+
+  nodes_range all_nodes() const noexcept {
```
srogatch, line 691: Please, review all `noexcept` for whether what they call is also `noexcept`. I'll mostly stop commenting about `noexcept` from now on in this PR.

uditagar:
> Except the types comment (the current code seems wrong), the rest are nits.

oh it's a intersection, not union. oh man

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,156 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+      const std::vector<std::string>& edge_types);
+
+private:
+  ProjectedShuffleTopology(
+      NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests,
+      NUMAArray<Node>&& old_to_new_nodes_mapping,
+      NUMAArray<Node>&& new_to_old_nodes_mapping,
+      NUMAArray<Edge>&& old_to_new_edges_mapping,
+      NUMAArray<Edge>&& new_to_old_edges_mapping) noexcept
+      : adj_indices_(std::move(adj_indices)),
```
srogatch, line 764: I think `std::move()` should be used when we are moving something without `&&` in its declaration. For a parameter with `&&` you should definitely use `std::forward<T>()`.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+std::unique_ptr<katana::ProjectedShuffleTopology>
+katana::ProjectedShuffleTopology::Make(
+    const katana::PropertyGraph* pg, const std::vector<std::string>& node_types,
+    const std::vector<std::string>& edge_types) {
+  KATANA_LOG_DEBUG_ASSERT(pg);
+
+  const auto& topology = pg->topology();
+  if (topology.empty()) {
+    ProjectedShuffleTopology pst;
+    return std::make_unique<ProjectedShuffleTopology>(std::move(pst));
```
srogatch: Why is `std::move(pst)` needed here as a parameter to `make_unique`? It should work without any arguments, the same way the class is instantiated on the stack a line above.

uditagar:
@roshandathathri  I have another question rearding projecting nodes/edge according to their types. If a node n has 4 types, and the input node_types have only 2 type (but it is a subset of these 4 types of node n). Then do we select node n or not? Thanks

libgalois/src/GraphTopology.cpp
srogatch: It seems `0` is used as a sentinel node ID here, while it's also a legal node ID. Why not use `Node{-1}` for the sentinel?

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+  katana::ParallelSTL::fill(
+      old_to_new_nodes_mapping.begin(), old_to_new_nodes_mapping.end(),
+      Node{0});
+
+  if (node_types.empty()) {
+    num_new_nodes = topology.num_nodes();
+    // set all nodes
+    katana::do_all(katana::iterate(topology.all_nodes()), [&](auto src) {
+      bitset_nodes.set(src);
+      old_to_new_nodes_mapping[src] = 1;
```
srogatch: Or was it a sentinel indeed? Is it another sentinel, or what are the meanings of `0` and `1`? Please, comment extensively in the code. You have at least wasted the computing power by first filling everything with `0`s, then with `1`s. If filling with zeros is needed for another conditional branch, please, move the filling with zeros there.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+    }
+
+    katana::GAccumulator<uint32_t> accum_num_new_nodes;
+    katana::GAccumulator<uint64_t> accum_num_new_edges;
+
+    katana::do_all(katana::iterate(topology.all_nodes()), [&](auto src) {
+      for (auto type : node_entity_type_ids) {
+        if (pg->DoesNodeHaveType(src, type)) {
+          accum_num_new_nodes += 1;
+          accum_num_new_edges += std::distance(
```
srogatch: Again, if you never suppose it to be `O(N)` please use `-` rather than `std::distance`.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+              topology.edges(src).begin(), topology.edges(src).end());
+          bitset_nodes.set(src);
+          old_to_new_nodes_mapping[src] = 1;
```
srogatch: This was declared as having `Node` element type, but you seem to use it as a boolean. I may be very wrong, so comments would definitely help.

libgalois/src/GraphTopology.cpp
```diff
@@ -539,6 +725,19 @@ katana::PGViewCache::BuildOrGetEdgeTypeAwareTopo(
   }
 }
 
+katana::ProjectedShuffleTopology*
```
arekay: Should this be encapsulated in a shared pointer?

arekay (approved):
LGTM. Minor nits.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+      }
+    });
+    num_new_nodes = accum_num_new_nodes.reduce();
+    num_new_edges = accum_num_new_edges.reduce();
+  }
+
+  // fill old to new nodes mapping
+  katana::ParallelSTL::partial_sum(
+      old_to_new_nodes_mapping.begin(), old_to_new_nodes_mapping.end(),
+      old_to_new_nodes_mapping.begin());
```
srogatch: Only now do I see the point why there were 0s and 1s. A comment in the beginning will save readers quite some time.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+
+  NUMAArray<Node> new_to_old_nodes_mapping;
+  new_to_old_nodes_mapping.allocateInterleaved(num_new_nodes);
+
+  katana::do_all(katana::iterate(topology.all_nodes()), [&](auto src) {
+    if (bitset_nodes.test(src)) {
+      old_to_new_nodes_mapping[src]--;
+      new_to_old_nodes_mapping[old_to_new_nodes_mapping[src]] = src;
```
srogatch: I haven't thought deep enough, but I believe this yields concurrency issues. The first line will decrease an array value, separate in each thread, should be fine (but deserves a comment). The second line will assign to an array index that probably collides with `old_to_new_nodes_mapping[src+-k]` where `k` equals to the number of zeros before or after index `src` in the old `old_to_new_nodes_mapping` array before you ran `partial_sum()`. If I'm wrong, please, comprehensively describe how this parallel algorithm is correct and why there is no data race.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+  if (edge_types.empty()) {
+    // set all edges incident to projected nodes
+    katana::do_all(
+        katana::iterate(Node{0}, Node{num_new_nodes}),
+        [&](auto src) {
+          auto old_src = new_to_old_nodes_mapping[src];
+          for (Edge e : topology.edges(old_src)) {
+            auto dest = topology.edge_dest(e);
+            if (bitset_nodes.test(dest)) {
+              bitset_edges.set(e);
```
srogatch, line 586: Is `DynamicBitset` thread-safe here? Please, comment in the code.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+  });
+
+  // calculate number of new edges
+  katana::DynamicBitset bitset_edges;
+  bitset_edges.resize(topology.num_edges());
+
+  NUMAArray<Edge> out_indices;
+  out_indices.allocateInterleaved(num_new_nodes);
+
+  katana::ParallelSTL::fill(out_indices.begin(), out_indices.end(), Edge{0});
```
srogatch, line 575: Please, say in a code comment what you will be doing with this.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,192 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+          if (bitset_edges.test(e)) {
+            auto e_new = out_dests_offset[n];
+            out_dests_offset[n]++;
+
+            auto dest = topology.edge_dest(e);
+            dest = old_to_new_nodes_mapping[dest];
+            out_dests[e_new] = dest;
+
+            old_to_new_edges_mapping[e] = e_new;
+            new_to_old_edges_mapping[e_new] = e;
```
srogatch: Please, comment here about the invariants that help you avoid a race condition here.

srogatch (changes requested):
The off-by-one bug is definitely worth fixing.
For the rest, I can best describe my impression with a joke that I still don't understand what I have reviewed. Please, add comments and an overview of the algorithms in the code.

roshandathathri:
> @roshandathathri I have another question rearding projecting nodes/edge according to their types. If a node n has 4 types, and the input node_types have only 2 type (but it is a subset of these 4 types of node n). Then do we select node n or not? Thanks

Yes, it will be selected. This will be handled transparently by `graph.DoesNodeHaveType()`.

libgalois/include/katana/GraphTopology.h
roshandathathri: Agree. I also think it should be CamelCase: `ProjectedToOriginalEdge`. 

For now, let it be consistent with other classes. @amberhassaan

libgalois/src/GraphTopology.cpp
uditagar: I can't find a function in the PropertyGraph.h file that takes a list of entity type names to return a entity type id. Thanks

roshandathathri: Use `GetNodeEntityTypeManager` and then get the type from a set of strings from it.

libgalois/src/GraphTopology.cpp
uditagar: I assigned 0 here because I wanted to compute prefix sum, and then based on prefix sum values: assign new ids.

libgalois/src/GraphTopology.cpp
uditagar: I just followed a similar convention as was done for other classes in this file.

libgalois/include/katana/GraphTopology.h
uditagar: I will rename it to something else like `projected_topo_`

libgalois/include/katana/GraphTopology.h
uditagar: I will remove `noexcept` then

libgalois/include/katana/GraphTopology.h
uditagar: yeah, my mistake here

libgalois/include/katana/GraphTopology.h
uditagar: I only followed the convention used in other classes. I am sorry but I am not familiar with `std::forward`. thanks

libgalois/src/GraphTopology.cpp
uditagar: I need 0 for calculating prefix sum later. thanks

libgalois/src/GraphTopology.cpp
uditagar: you are right, I will move filling zeros to the `else` part

libgalois/src/GraphTopology.cpp
uditagar: no, it isn't boolean. I will perform prefix sum on this array.

libgalois/src/GraphTopology.cpp
uditagar: the index corresponding to src will only be updated in one thread, in which it is also used. I don't believe there is any concurrency issue here, since no two threads will either update/read the same location.

libgalois/src/GraphTopology.cpp
uditagar: those two bitsets are different: one corresponds to nodes and other to edges. And this one: `bitset_edges.set(e); ` is only set by a single thread for a given edge e. thanks

roshandathathri: Serge, `DynamicBitset` is thread-safe on both CPU and GPU. It supports concurrent reads (test) and concurrent writes (set/reset) but not concurrent reads+writes. This should be documented in the API. If it's not clear, please feel free to edit it and let me know.

libgalois/src/GraphTopology.cpp
uditagar: again there is no race condition here. locations that are updated are only written by one specific thread.

uditagar:
I have fixed most of the things as suggested by all people. Regarding `noexcept`, as per the slack discussion, I have decided to follow the convention used in other classes in this file. 
@serge @roshandathathri will need to check this to clear the bottleneck here. thank you.

libgalois/include/katana/GraphTopology.h
```diff
@@ -40,6 +41,7 @@ struct KATANA_EXPORT GraphTopologyTypes {
 
 class KATANA_EXPORT EdgeShuffleTopology;
 class KATANA_EXPORT EdgeTypeAwareTopology;
+class KATANA_EXPORT ProjectedSTypeTopology;
```
roshandathathri: `S` seems like a typo. If it isn't, why is forward declaration required?

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,174 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+
+  const Edge* adj_data() const noexcept { return adj_indices_.data(); }
+
+  const Node* dest_data() const noexcept { return dests_.data(); }
+
+  /// Checks equality against another instance of ProjectedTopology.
+  /// WARNING: Expensive operation due to element-wise checks on large arrays
+  /// @param that: ProjectedTopology instance to compare against
+  /// @returns true if topology arrays are equal
+  /// should take O(|V| + |E|) in the worst case
```
roshandathathri, line 634: nit: also it's single threaded, right?

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,174 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+  }
+
+  /// @param nid the input node id (must be original node id)
+  Node original_to_projected_node_id(const Node& nid) const noexcept {
+    return original_to_projected_nodes_mapping_[nid];
+  }
+
+  /// this function creates a topology by filtering nodes and edges
+  /// @param node_types nodes with types in node_types are filtered
+  /// @param edge_types edges with types in edge_types are filtered
```
roshandathathri: nit: filtered can mean selected or dropped. Please use "selected" instead.

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,174 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
...
+    return node_property_index(nid);
+  }
+
+  /// @param nid the input node id (must be original node id)
+  Node original_to_projected_node_id(const Node& nid) const noexcept {
+    return original_to_projected_nodes_mapping_[nid];
+  }
+
+  /// this function creates a topology by filtering nodes and edges
+  /// @param node_types nodes with types in node_types are filtered
```
roshandathathri: This seems like a union, not intersection. We are doing intersection (and that's correct).
How about this?
`  /// @param node_types the types that the selected nodes must have`

libgalois/include/katana/GraphTopology.h
roshandathathri: Serge, in this case, `std::move` is required because we're using move constructors for those variables (their type does not support copy constructors). `std::forward<t>` can return lvalue (which is not desired in this case):
https://stackoverflow.com/questions/28828159/usage-of-stdforward-vs-stdmove

libgalois/include/katana/GraphTopology.h
```diff
@@ -605,6 +607,174 @@ class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
   const Topo* topo_ptr_;
 };
 
+/// filter nodes and edges based on their type values
```
roshandathathri: Drop "based on their type values" because that's described in the `Make*` function and this class is meant to be more general.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,198 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+    });
+  } else {
+    katana::ParallelSTL::fill(
+        original_to_projected_nodes_mapping.begin(),
+        original_to_projected_nodes_mapping.end(), Node{0});
+
+    auto node_entity_type_manager = pg->GetNodeTypeManager();
+
+    auto res_node_type =
+        node_entity_type_manager.GetNonAtomicEntityTypeFromStrings(node_types);
```
roshandathathri: There's a bug here (ditto for edges): 

There is no guarantee that the vector of node types has a corresponding entity type ID. 
(1) Consider "Person" and "Message". No node has both those types, so there will no entity type ID. 
(2) Consider "XYZ". No node may have this type, so there will be no entity type ID. 

This (caller) code should check if an error is returned. An error means that no node has this type, so you can skip the code below and create a graph with no nodes and edges (if an error is returned for edge types below, create a graph with only the selected nodes and no edges).

roshandathathri (approved):
LGTM

Approving it because I don't need to review this again. Please fix the bug I mentioned before you merge.

libgalois/include/katana/GraphTopology.h
uditagar: yes, it's a bug.

libgalois/include/katana/GraphTopology.h
uditagar: yes

libgalois/include/katana/GraphTopology.h
amberhassaan: @srogatch : `&&` in parameter name is to signify that the object will be "consumed". `std::move()` is a better choice (even though `std::forward` will work with `&&`) because a named variable with `&&` may not be passed as `r-value` reference to the method/constructor being called because it has a name, thus making it `l-value` in certain situations. (`std::move()` guarantees that the parameter being passed to the constructor is always an `r-value` reference).

libgalois/include/katana/GraphTopology.h
amberhassaan: `std::upper_bound` will give predictable results because `adj_indices` is sorted.

libgalois/include/katana/GraphTopology.h
```diff
@@ -1032,6 +1202,7 @@ using PGViewNodesSortedByDegreeEdgesSortedByDestID =
 using PGViewBiDirectional = BasicPropGraphViewWrapper<SimpleBiDirTopology>;
 using PGViewEdgeTypeAwareBiDir =
     BasicPropGraphViewWrapper<EdgeTypeAwareBiDirTopology>;
+using PGViewProjectedGraph = BasicPropGraphViewWrapper<ProjectedTopology>;
```
amberhassaan, line 1287: @uditagar : you'll need to introduce a `ProjectedTopologyWrapper` which can be simply aliased to `BasicTopologyWrapper<ProjectedTopology>`. Reason being, `BasicPropGraphViewWrapper` takes topology wrappers by copy and `ProjectedTopology` cannot (and should not) be copied. `BasicTopologyWrapper` takes a topology object type by pointer and can be copied trivially.

libgalois/src/GraphTopology.cpp
```diff
@@ -435,6 +435,198 @@ katana::EdgeTypeAwareTopology::MakeFrom(
...
+std::unique_ptr<katana::ProjectedTopology>
+katana::ProjectedTopology::MakeTypeProjectedTopology(
+    const katana::PropertyGraph* pg, const std::vector<std::string>& node_types,
+    const std::vector<std::string>& edge_types) {
+  KATANA_LOG_DEBUG_ASSERT(pg);
+
+  const auto& topology = pg->topology();
+  if (topology.empty()) {
+    ProjectedTopology pst;
+    return std::make_unique<ProjectedTopology>(std::move(pst));
```
amberhassaan: Why use 2 lines to do what can be done in one line?

amberhassaan (commented):
2nd set of comments. More to come.

libgalois/include/katana/GraphTopology.h
uditagar: I get what you are saying. So should I need to do this `using PGViewProjectedGraph = BasicTopologyWrapper<ProjectedTopology>` or do I need to create a new class. And if so, what should be the contents of this new class? Thanks

srogatch (approved):
Should be fine now.

libgalois/src/GraphTopology.cpp
uditagar: fixed this now

libgalois/include/katana/GraphTopology.h
uditagar: added new wrapper now

libgalois/src/GraphTopology.cpp
uditagar: fixed now

uditagar:
So I have updated the code according to the reviews by all people. This is a list of new updates:
1. Updated node and edge projection logic: `OR` for nodes and edges; and edge is picked only if both it's endoints are also picked
2. Added a new wrapper class for `ProjectedTopology`
3. Added unit-test. For ldbc with command: `./unit-test-projection ~/udit_katana_Mar30/build/inputs/current/propertygraphs/ldbc_003/ City,Comment,Company,Continent,Country,Forum HAS_CREATOR,HAS_INTEREST,HAS_MEMBER,HAS_MODERATOR,HAS_TAG,HAS_TYPE,IS_PART_OF,IS_SUBCLASS_OF,KNOWS,LIKES`, the output is `Original Num Nodes: 29946
 Original Num Edges: 43072
 New Num Nodes: 3813
 New Num Edges: 1454`
4. Added functionality to return empty graph: returns 0 node and edge graph if no nodes are selected; and returns a graph with num_new_nodes and 0 edges (if num_new_nodes are selected and no edges)

@roshandathathri @amberhassaan Please review changes, if you want to. Otherwise, I will merge it on Monday and go ahead with my distributed projection PR. Thank you.

PR #565, opened 2021-09-24
Fix: always load schema info

scober:
One way that we populate our schemas for node and edge properties is by
loading the properties. Another way is by reading the headers of
property files that weren't loaded. We have a bug where if we don't load
any properties then we terminate a function early and skip reading the
headers. Fix that bug.

@tylershunt I decided to just move the calls higher up in `DoMake()` because my reading is that the code below the early termination deals with partition metadata and `Ensure*()` calls deal only with actual properties.

### DIFF
--- a/libtsuba/src/RDG.cpp
+++ b/libtsuba/src/RDG.cpp
@@ -556,6 +556,10 @@ tsuba::RDG::DoMake(
   std::vector<PropStorageInfo*> part_info =
       KATANA_CHECKED(core_->part_header().SelectPartitionProperties());
 
+  // these are not Node/Edge types but rather property types we are checking
+  KATANA_CHECKED(core_->EnsureNodeTypesLoaded(rdg_dir_));
+  KATANA_CHECKED(core_->EnsureEdgeTypesLoaded(rdg_dir_));
+
   if (part_info.empty()) {
     return grp.Finish();
   }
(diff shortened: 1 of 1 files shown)


tylershunt (approved):
Approve for the quick fix, but we should think about moving these after property loading to save work.

scober:
> Approve for the quick fix, but we should think about moving these after property loading to save work.

After what property loading? My reading is that the placement in this PR puts it after loading of "true" properties but before loading of partition metadata (which also uses the property loading logic). I thought that this would be fine.

libtsuba/src/RDG.cpp
```diff
@@ -556,6 +556,10 @@ tsuba::RDG::DoMake(
   std::vector<PropStorageInfo*> part_info =
       KATANA_CHECKED(core_->part_header().SelectPartitionProperties());
 
+  // these are not Node/Edge types but rather property types we are checking
+  KATANA_CHECKED(core_->EnsureNodeTypesLoaded(rdg_dir_));
+  KATANA_CHECKED(core_->EnsureEdgeTypesLoaded(rdg_dir_));
+
```
e-mcginnis, line 562: I believe it would make the most sense for these to be after the call to `AddProperties()` on line 568? (might be what @tylershunt was referring to for future work? 

approving anyway to unblobk

e-mcginnis (approved):

libtsuba/src/RDG.cpp
scober: It has to be before the `if(part_info.empty())` on line 563. And I think the call to `AddProperties()` on 568 is just using the property loading logic to load partition metadata. So it won't touch any node or edge properties, which are the only things these two calls will touch.

tylershunt: OH. Simon is totally right; apologies.

PR #568, opened 2021-09-26
[KAT-970] Prepare Katana for GPU benchmarking vs Galois

srogatch:


patrickkenney9801:
I suggest we use https://github.com/KatanaGraph/katana/pull/569 instead of this PR for now
Using Logging.h in tracing can be considered later

ddn0 (changes requested):
If you believe the tracer is too noisy, a no-op tracer would be more appropriate rather than disabling console output unilaterally.

roshandathathri:
Agree with Donald. I thought we had levels for tracing as well and we can choose one (say just "Stages") for benchmarking?

srogatch:
The problem is that that tracer is outputting everything without a level. It prints `INFO:`, but it doesn't check if the level is no higher than INFO. In this PR I tried to make it use an existing level - DEBUG. But ideally it should use a new level `TRACE`, because it prints even more info than `DEBUG`. In my conception of `DEBUG` level output, it should print what I need when debugging the program, so that I may remove the debug-prints after they are no more needed. But the tracing with `INFO` level just gets everything else lost.

roshandathathri:
IMO the logging levels (DEBUG, INFO, etc.) need to be distinct from tracing levels. They are currently. I'm not sure if the latter is configurable.

srogatch:
> IMO the logging levels (DEBUG, INFO, etc.) need to be distinct from tracing levels. They are currently. I'm not sure if the latter is configurable.

The tracing should have domains also, not only levels. See https://github.com/KatanaGraph/katana-enterprise/pull/2078#discussion_r718252176

arekay (approved):
LGTM.

scripts/check_python_format.sh
```diff
@@ -5,6 +5,10 @@ ISORT="${ISORT:-isort}"
 
 set -eu
 
+# Know the specific versions used to reproduce the formatting checks locally
+${BLACK} --version
+${ISORT} --version
```
roshandathathri, line 10: I didn't understand why this was required and/or why this is a part of this PR

roshandathathri (approved):
LGTM

Please address my comment before merging

scripts/check_python_format.sh
srogatch: I found myself again in a situation where `scripts/check_python_format.sh` says the python input file is formatted ok, but when I commit the python file to CI, CI says it needs reformatting. This happens because there are different versions of `black` and `isort` on my local laptop and on CI runners, while before this change there was no way of figuring out what exact versions are used in the CI runners. This change prints the versions to the output that is then shown in CI logs. After I got the versions from CI, I installed the exact same versions locally, ran the formatting script, it changed the python file and only then that python file was accepted by CI.

srogatch: I found myself again in a situation where `scripts/check_python_format.sh` says the python input file is formatted ok, but when I commit the python file to CI, CI says it needs reformatting. This happens because there are different versions of `black` and `isort` on my local laptop and on CI runners, while before this change there was no way of figuring out what exact versions are used in the CI runners. This change prints the versions to the output that is then shown in CI logs. After I got the versions from CI, I installed the exact same versions locally, ran the formatting script, it changed the python file and only then that python file was accepted by CI.

roshandathathri: Please do not squash them into one commit. When you rebase, please use two different commits so that this change is not combined with unrelated changes.

roshandathathri: And thanks for the clarification! Please include some context like that in the commit message.

arthurp: @srogatch Sorry about this. I have known about this issue for a long time and failed to fix it. This is a decent hack to help for the moment, until I can create a better fix that actually checks the version and fails out if it's wrong. Thanks. (https://katanagraph.atlassian.net/browse/KAT-1287)

lonestar/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ endfunction()
 
 function(add_test_scale type app)
   set(no_value_options NOT_QUICK NO_VERIFY)
-  set(single_value_options INPUT INPUT_URI)
+  set(single_value_options INPUT INPUT_URI REL_TOL MEAN_TOL)
```
arthurp, line 20: Why do these need to vary? I totally believe they do, but please document the specific reasons.

ddn0 (approved):

lonestar/CMakeLists.txt
srogatch: I've added a comment in the code.

PR #570, opened 2021-09-26
[KAT-1849] logging: add a method for suppressing logging from unnecessary spans …

patrickkenney9801:
…to tracers

This is useful for example in subgraph sampling when this function calls the query engine (which generates a lot of output that is not very important to subgraph sampling) and this becomes an issue if someone wants to call subgraph sampling thousands of times


pyrito (approved):
LGTM

patrickkenney9801:
We have not yet determined that this PR will be necessary or useful so we are closing it for now

PR #571, opened 2021-09-27
[KAT-1552] Build conda package variants for py 3.7-3.9

arthurp:


arthurp:
This would be useful and we do need it. But I'm putting it on hold. I don't know what's wrong and I don't have time to fix it right now. I will return to it: KAT-1552

PR #573, opened 2021-09-27
Fix listing of atomic edge types

roshandathathri:


witchel (approved):
If working in a graph database company has taught me one thing it is that `node != edge`.

PR #574, opened 2021-09-27
[KAT-1651] Add utility to capture execution environment information.

arthurp:
This add a utility in the python library that captures system, execution, and build information for bug reporting. It is extensible so enterprise can add to it.

The utility will be invoked using: `import katana.bug; katana.bug.capture_environment()`. This will output the the path to a zip file it creates with all the information. This will work from notebooks as well as other python environments.


python/test/test_bug.py
```diff
@@ -0,0 +1,13 @@
...
+
+def test_bug_capture_environment():
+    import katana.bug
+
+    with NamedTemporaryFile(suffix=".zip", delete=False) as fi:
+        katana.bug.capture_environment(fi)
+
+        # Check that the scripts captured version information in both environments as a general check.
+        with zipfile.ZipFile(fi) as zipin:
+            assert katana.__version__.encode("utf-8") in zipin.open("info/katana.txt").read()
```
gliga, line 12: maybe check if expected files are in the zip

python/katana/bug/environment.py
```diff
@@ -0,0 +1,195 @@
...
+        f"""
+            host: {socket.gethostname()}
+            uname: {os.uname()}
+            date: {datetime.datetime.now()}
+
+            {capture_command('lsb_release', '-a')}
+            """,
+    )
+
+    import katana  # Imported here to avoid cyclical import
```
gliga: still not excited about import (or #include) in the middle

python/katana/bug/environment.py
```diff
@@ -0,0 +1,195 @@
...
+            {capture_command('cmake', '--version')}
+                """,
+    )
+
+    capture_files(zipout, ["/etc/ld.so.conf", "/etc/ld.so.conf.d/"])
+
+
+@register_capture_routine
+def _capture_build(zipout: zipfile.ZipFile):
+    for root in [Path(__file__).parents[3], Path(__file__).parents[5], Path.cwd()]:
```
gliga: maybe a comment for [3] and [5]

python/katana/bug/environment.py
```diff
@@ -0,0 +1,195 @@
...
+def capture_command(*args, **kwargs) -> str:
+    # pylint: disable=subprocess-run-check
+    # Not using check=True because I want to capture both success and failure the same way.
+    res = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs)
+    out = res.stdout.decode("utf-8").strip("\n")
+    return out
+
+
+def capture_files(zipout: zipfile.ZipFile, paths: Iterable[Union[Path, str]]):
+    for p in paths:
```
gliga, line 131: slightly confusing that it is called paths, but then it requires Path(p)

python/katana/bug/environment.py
```diff
@@ -0,0 +1,195 @@
...
+        p = Path(p)
+        if p.is_file():
+            zipout.write(p)
+        elif p.is_dir():
+            capture_files(zipout, p.iterdir())
+        else:
+            # ignore other kinds of files
```
gliga, line 139: symlinks?

gliga (approved):
Looks good.

python/katana/bug/environment.py
arthurp: I'll put it at the top of the function. That is better. But it cannot be at the top of the file due to dependency cycles. I'll check for an easy way to remove the cycle as well.

python/katana/bug/environment.py
```diff
@@ -0,0 +1,227 @@
+import datetime
+import os
+import re
+import socket
+import subprocess
+import sys
+import tempfile
+import zipfile
```
ddn0, line 8: A small preference for using `tarfile` as our defacto bag of files format as it supports Unixy features more directly like UID and special files if we ever need them later.

ddn0 (approved):

python/katana/bug/environment.py
arthurp: I used zip because there is better support for looking at files inside it from many OSs and software. Some, notably Windows Explorer, don't have tar support (I assume this has not changed recently). I would generally prefer tar as well, but I decided zip would be easier for more people to look at and this archive does not need to be unpacked and be runnable or anything.

Is that a good enough reason for you? I can go with tar if you don't think this is a good argument.

ddn0: Is this something we expect reporters/bug submitters to unpack?

Anyway, if you can imagine a common workflow that would require windows users to unpack a report, zip is reasonable. Myself, I don't know what that would be.

arthurp: The workflow I see is this: We have a lot of developers who use windows as their main OS and use a VM or remote host with Linux. They don't generally have their web browser open in that Linux environment (especially for people without a local Linux VM), so developers working on a bug will download it on to their local windows machine and want to look at it. It would be an extra step to copy the report to a linux machine and then extract on the remote host.

ddn0: FWIW, I think most windows developers would use WSL if that matters and Windows 10 comes with tar, the command line utility.

To build or run anything in this repo, you probably need a typical Linux environment, I don't know how much extra variety we should support, especially when we dictate what constitutes a "valid developer environment".

Anyway, I don't think consistency here is a big thing one way or another.

arthurp: Makes sense. This can be changes easily if needed. I'm not wedded to one decision and changing it later wouldn't really break anything.

Now I'm curious how peoples dev environments are setup. I wonder how many people have their Linux dev work and their web browser accessing the same file system and how many do not. Hmm.

PR #575, opened 2021-09-27
Fix: tsuba::ListAvailableViews()

scober:
The ListAvailableViews() contract is that it will return all views for
the requested version and only views for the requested version. If no
version is requested, it will return all views for the latest version
and only views for the latest version.

The no-requested-view variant of this contract is being violated
because in the core loop over the
potential views there is no check for if the view under consideration is
less than the current most recent known version. So the returned list
contains views from multiple versions.

### DIFF
--- a/libtsuba/src/tsuba.cpp
+++ b/libtsuba/src/tsuba.cpp
@@ -146,6 +146,7 @@ ParseManifestName(
 
 }  // namespace
 
+// TODO (scober, KAT-1870): write a test for this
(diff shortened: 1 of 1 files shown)


yshalabi (approved):
Thanks. It looks like this was a regression caused by the removal of min_version checks in https://github.com/KatanaGraph/katana/commit/39d77fc74b05931b685a49a7b28ad137023f8d4e

I wish I added my unit tests for this ListAvailableViews code when I added this function in the first place. Thanks.

PR #576, opened 2021-09-28
Do not expose kUnknownEntityType as an atomic type

roshandathathri:
Treat kUnknownEntityType as an atomic type internally
but do not expose it as an atomic type to users of the class.
We should not return kUnknownEntityTypeName as an
atomic type name.

### DIFF
--- a/libsupport/include/katana/EntityTypeManager.h
+++ b/libsupport/include/katana/EntityTypeManager.h
@@ -237,7 +237,8 @@ public:
 
   /// \returns the number of atomic types
   size_t GetNumAtomicTypes() const {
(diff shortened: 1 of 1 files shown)


arthurp (commented):
All these special cases make me unhappy. If we need to special case it, then it doesn't fit well into out model. This makes me think that I was right in the first place when I said that "UnknownEntityType" doesn't make much sense because it's just represented by a node with no assigned type at all. Could we move towards that direction? Just have UnknownEntityType be the intersection of the empty set of types? That would make it not be atomic and should reduce the special cases. It also makes a bit more sense theoretically.

PR #577, opened 2021-09-28
Do not treat kUnknownEntityType as an atomic type

roshandathathri:
We should not return kUnknownEntityTypeName as an
atomic type name.

Do not treat kUnknownEntityType as an atomic type,
even internally within the class. Treat it as an entity type
that does not have any atomic subtypes.

### DIFF
--- a/libgalois/test/property-file-graph.cpp
+++ b/libgalois/test/property-file-graph.cpp
@@ -28,7 +28,7 @@ void
 TestTypesFromPropertiesCompareTypesFromStorage() {
   /*
   Scenario 1:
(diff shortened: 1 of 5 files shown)


witchel (approved):
Tears of joy.
(This change does deserve a comment in the code though.)

arthurp (approved):

roshandathathri:
@e-mcginnis I modified storing of entity type dictionaries. Please review.

witchel (approved):
I look forward to our grand and glorious future.

PR #578, opened 2021-09-28
Remove macOS build in CI and setup scripts.

arthurp:
Everything was unused and probably stale. Just get rid of it.


ddn0 (approved):
Can you file an issue to add  llvm-libc/libc++ (or whatever it is called) so we get some coverage on standard-compliant libc usage?

arthurp:
> Can you file an issue to add llvm-libc/libc++ (or whatever it is called) so we get some coverage on standard-compliant libc usage?

Added as KAT-1879

PR #579, opened 2021-09-28
[KAT-1885] Add support for external docs building with cmake flag

origandrew:
original: ~~`cmake -S $SRC_DIR -B $BUILD_DIR -DKATANA_LANG_BINDINGS=python -DBUILD_DOCS=ON \
      -DBUILD_EXTERNAL_DOCS=ON`~~
update:
```
cmake -S $SRC_DIR -B $BUILD_DIR -DKATANA_LANG_BINDINGS=python -DBUILD_DOCS=external

cd build
make docs
```


origandrew:
@ddn0 I couldn't figure out how to send options using a variable in cmake, they weren't getting picked up. That's why there's an ugly if-else statement

docs/conf.py
```diff
@@ -73,6 +72,12 @@
 autodoc_preserve_defaults = True
 autodoc_member_order = "groupwise"
 
+suppress_warning = []
+nitpick_ignore = []
+
+# Exclude drafts from external documentation
+if tags.has("external"):
+    exclude_patterns.extend(["**-draft*","**-internal*"])
```
ddn0, line 80: perhaps change this to `"**-draft.*"` etc, so it is clear that we are just excluding the suffix `-draft.ext`

cmake/Modules/BuildCommon.cmake
```diff
@@ -44,6 +44,7 @@ set(BUILD_SHARED_LIBS YES CACHE BOOL "Build shared libraries. Default: YES")
 # that this is a standard option.
 set(BUILD_TESTING ON CACHE BOOL "Build tests")
 set(BUILD_DOCS OFF CACHE BOOL "Build documentation with make doc")
+set(BUILD_EXTERNAL_DOCS OFF CACHE BOOL "Hide '*-draft*' documentation pages and directories when building documentation")
```
ddn0: - Preference for making the default behavior to build external docs (i.e., not build internal ones).

- Preference for a single option `BUILD_DOCS`; truthy values or `external` build the external docs; falsey values don't build any docs; `BUILD_DOCS=internal` builds the internal docs. (CMake `if` should handle truthy and falsey automatically). If you do change the meaning of `BUILD_DOCS`, grep around for previous uses of `BUILD_DOCS` and replace them with `BUILD_DOCS=external`

cmake/Modules/BuildCommon.cmake
```diff
@@ -469,18 +470,34 @@ function(add_katana_sphinx_target target_name)
   endif ()
 
   get_target_property(PYTHON_ENV_SCRIPT ${target_name} PYTHON_ENV_SCRIPT)
-
-  add_custom_target(
+  if (NOT BUILD_EXTERNAL_DOCS)
+    set(sphinx_build_flags "-W -t internal")
```
ddn0: this line seems orphaned

cmake/Modules/BuildCommon.cmake
```diff
@@ -469,18 +470,34 @@ function(add_katana_sphinx_target target_name)
...
+    add_custom_target(
       ${target_name}_sphinx_docs
       COMMAND ${CMAKE_COMMAND} -E rm -rf ${CMAKE_BINARY_DIR}/docs/${target_name}
       COMMAND env KATANA_DOXYGEN_PATH="${CMAKE_BINARY_DIR}/docs/doxygen" ${PYTHON_ENV_SCRIPT} sphinx-build
         -W
         -b html
+        -t internal
```
ddn0: Why didn't something like this work?

```cmake
set(tag_option "-t external")
if(BUILD_DOCS STREQUAL "internal")
  set(tag_option "-t internal")
endif()

add_custom_target(... ${tag_option})
```

ddn0 (commented):
Thanks for doing this!

I have some nits.

cmake/Modules/BuildCommon.cmake
origandrew: not sure. I tried that exact thing but for some reason it wasn't getting picked up

cmake/Modules/BuildCommon.cmake
origandrew: I can do that. Do we want to default to building docs? This was not the original behavior

ddn0: >> Preference for making the default behavior to build external docs (i.e., not build internal ones).

> I can do that. Do we want to default to building docs? This was not the original behavior

Sorry, clarification:

```
set(BUILD_DOCS OFF CACHE BOOL "Build documentation with make doc")
```

as before but where previously `BUILD_DOCS` could be truthy or falsy, instead: falsy, `internal` or other truthy. I.e., treat truthy as `external`

docs/conf.py
origandrew: This is to also support whole subdirectories. For example, references/http-draft/index.rst
See PR for example: https://github.com/KatanaGraph/katana-enterprise/pull/2102

cmake/Modules/BuildCommon.cmake
```diff
@@ -43,8 +43,7 @@ set(BUILD_SHARED_LIBS YES CACHE BOOL "Build shared libraries. Default: YES")
 # This option is added by include(CTest). We define it here to let people know
 # that this is a standard option.
 set(BUILD_TESTING ON CACHE BOOL "Build tests")
-set(BUILD_DOCS OFF CACHE BOOL "Build documentation with make doc")
-set(BUILD_EXTERNAL_DOCS OFF CACHE BOOL "Hide '*-draft*' documentation pages and directories when building documentation")
+set(BUILD_DOCS "" CACHE string "Build documentation with make doc. Supported values: <unset>, external, internal. external docs hid '*-draft*' and '*-internal* documentation pages and directories when building documentation")
```
ddn0: "hid" => "hide" as in "When external, hide ..."

cmake/Modules/BuildCommon.cmake
```diff
@@ -484,7 +482,7 @@ function(add_katana_sphinx_target target_name)
       COMMAND ${CMAKE_COMMAND} -E echo "${target_name} documentation at file://${CMAKE_BINARY_DIR}/docs/${target_name}/index.html"
       COMMENT "Building internal ${target_name} sphinx documentation"
     )
-  else ()
+  elseif (BUILD_DOCS STREQUAL "external")
     add_custom_target(
```
ddn0: can we dedupe this `add_custom_target` with the snippet I gave?

cmake/Modules/BuildCommon.cmake
```diff
@@ -43,7 +43,7 @@ set(BUILD_SHARED_LIBS YES CACHE BOOL "Build shared libraries. Default: YES")
 # This option is added by include(CTest). We define it here to let people know
 # that this is a standard option.
 set(BUILD_TESTING ON CACHE BOOL "Build tests")
-set(BUILD_DOCS OFF CACHE BOOL "Build documentation with make doc")
+set(BUILD_DOCS "" CACHE STRING "Build documentation with make doc. Supported values: <unset>, external, internal. external docs hid '*-draft*' and '*-internal* documentation pages and directories when building documentation")
```
ddn0: if you are going to change this, there are other uses of `BUILD_DOCS` like in `conda/build.sh` that need to be updated.

docs/conf.py
```diff
@@ -16,7 +16,7 @@
 else:
     # Tags can be defined with the -t argument to sphinx. Use not-internal to
     # override an otherwise default internal tag.
-    if not tags.has("not-internal"):
+    if not tags.has("enterprise"):
```
ddn0: can you just call this "external"? There is no enterprise in this repo.

cmake/Modules/BuildCommon.cmake
origandrew: No. Doesn't get picked up by sphinx-build.

ddn0: If we are going to write 
```cmake
if()
  add_custom_target(... a)
else()
  add_custom_target(... b)
endif()
```

instead of 

```cmake
set(var b)
if ()
  set(var a)
endif()
add_custom_target(... ${var})
```

I'm going to need a little more explanation than "it doesn't get picked up by sphinx-build".

No future developer is going to understand why this is the case and they might inadvertently revert this change if there isn't a clear explanation as to why it was written this way.

origandrew: Even the following doesn't work. Should I add a comment to make sure it doesn't get reverted?
```
set(targ_options "-t external")
add_custom_target(... sphinx-build ${tag-options})
```

cmake/Modules/BuildCommon.cmake
origandrew: added comment

cmake/Modules/BuildCommon.cmake
origandrew: added comment mentioning unexpected cmake behavior

docs/contributing/documentation.rst
```diff
@@ -5,19 +5,34 @@ Documentation
...
 
-Restructured Text and Markdown
-==============================
+Annotating Internal or Draft only Content
+=========================================
+
+Files ending in ``-draft.[rst/ipynb]`` or ``-internal.[rst/ipynb]`` will not be
+included in external facing documentation.
+
+Whole directories ending in ``-draft/`` will be omitted when building external
```
ddn0: or ``-internal``

cmake/Modules/BuildCommon.cmake
```diff
@@ -469,18 +469,36 @@ function(add_katana_sphinx_target target_name)
   endif ()
 
   get_target_property(PYTHON_ENV_SCRIPT ${target_name} PYTHON_ENV_SCRIPT)
-
-  add_custom_target(
+  #TODO(andrew): done this way since `set(tag_option "-t external") ... sphinx-build ${tag_option}...` does not work as expected
```
ddn0: BTW, have you tried

`set(tag_option "-t;external") ... sphinx-build ${tag_option}...` or  `set(tag_options "-t" "external") ... sphinx-build ${tag_options}...`

origandrew: interesting, I'll go try that

origandrew: Worked! Thanks @ddn0

ddn0 (approved):

PR #582, opened 2021-09-29
RDGManifest as param in tsuba::Open()

yige-hu:
Previously, tsuba::Open() takes a file name as input parameter. Now use RDGManifest instead. Before calling tsuba::Open(), FindManifest() is called for finding the manifest.

Test with PR #2104 in katana-enterprise.


yige-hu:
This is as what was discussed in the Monday's meeting. Not sure if it's an overly inconvenience interface change for the applications.

libtsuba/src/tsuba.cpp
```diff
@@ -91,8 +85,17 @@ tsuba::Open(const std::string& rdg_name, uint32_t flags) {
     return manifest_res.error();
   }
 
-  return RDGHandle{
-      .impl_ = new RDGHandleImpl(flags, std::move(manifest_res.value()))};
+  return std::move(manifest_res.value());
+}
+
+katana::Result<tsuba::RDGHandle>
+tsuba::Open(RDGManifest rdg_manifest, uint32_t flags) {
```
tylershunt, line 81: There's probably a discussion to be had about whether `Open` and `RDGHandle` make sense anymore. But not something that we necessarily have to address here.

libtsuba/include/tsuba/tsuba.h
```diff
@@ -15,6 +15,7 @@
 namespace tsuba {
 
 class RDGHandleImpl;
+class RDGManifest;
```
tylershunt, line 18: Just `#include "tsuba/RDGManifest.h"` rather than using this forward declaration. Users of this header will have to know about that type anyway.

libtsuba/include/tsuba/tsuba.h
```diff
@@ -53,11 +54,11 @@ OpenFlagsValid(uint32_t flags) {
...
 }
 
-KATANA_EXPORT katana::Result<RDGHandle> Open(
-    const std::string& rdg_name, uint32_t flags);
+KATANA_EXPORT katana::Result<RDGManifest> FindManifest(
+    const std::string& rdg_name);
 
 KATANA_EXPORT katana::Result<RDGHandle> Open(
-    const std::string& rdg_name, uint64_t version, uint32_t flags);
+    const RDGManifest rdg_manifest, uint32_t flags);
```
tylershunt: `const` with no `&` should probably be a const reference  no?

tylershunt (approved):
Seems like a totally reasonable change to me.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -237,6 +237,9 @@ class KATANA_EXPORT PropertyGraph {
       EntityTypeManager&& node_type_manager,
       EntityTypeManager&& edge_type_manager);
 
+  static Result<std::unique_ptr<katana::PropertyGraph>> Make(
+      tsuba::RDGManifest rdg_manifest, const tsuba::RDGLoadOptions& opts);
```
ddn0: maybe `const tsuba::RDGManifest&`?

ddn0 (approved):
probably `const Manifest&` and we don't need to `std::move` so much.

libtsuba/include/tsuba/tsuba.h
yige-hu: @tylershunt There is a circular inclusion issue between `tsuba/tsuba.h` and `tsuba/RDGHandle.h`, `tsuba/RDGManifest.h`. So I have to remain a declaration here.

tylershunt: Ah I see. Annoying; that is a mess but carry on.

libtsuba/include/tsuba/tsuba.h
yige-hu: My bad. Should be without const. Good catch!

PR #583, opened 2021-09-29
Fix listing empty sentinel views

scober:


libtsuba/include/tsuba/tsuba.h
```diff
@@ -97,11 +97,17 @@ struct KATANA_EXPORT RDGView {
...
 ///    the views for the latest version
-/// \returns a pair (RDG version, vector of RDGViews)
+/// \returns a pair (RDG version, vector of RDGViews) or ErrorCode::NotFound if
+/// rdg_dir contains no manifest files
+KATANA_EXPORT katana::Result<std::pair<uint64_t, std::vector<RDGView>>>
+ListViewsOfVersion(
+    const std::string& rdg_dir, std::optional<uint64_t> version = std::nullopt);
+
+/// deprecated; duplicate of ListViewsOfVersion maintained for compatibility
 KATANA_EXPORT katana::Result<std::pair<uint64_t, std::vector<RDGView>>>
```
witchel, line 111: How will we know when it is time to get rid of this function?

libtsuba/src/tsuba.cpp
```diff
@@ -146,21 +146,23 @@ ParseManifestName(
...
   uint64_t latest_version = 0;
+  bool any_version_found = false;
   for (const std::string& file : files) {
     std::string view_type;
     std::vector<std::string> view_args;
     uint64_t view_version{};
 
-    if (!ParseManifestName(file, &view_type, &view_args, &view_version)) {
+    if (ParseManifestName(file, &view_type, &view_args, &view_version)) {
+      any_version_found = true;
```
tylershunt: This control flow was hard for me to parse the first time, and I think it's wrong. We should also return an error if you ask for a particular version and it doesn't exist because we need to disambiguate that from the case where you ask for a particular version and it has no views. Could we instead flip this back and have: 
```c++
if (!ParseManifestName(file, &view_type, &view_args, &view_version)) {
  continue;
}

// if a specific version was requested, only return views for that version
if (version && view_version != version.value()) {
  continue;
}
any_version_found = true:
```

libtsuba/src/tsuba.cpp
```diff
@@ -202,14 +204,21 @@ tsuba::ListAvailableViews(
...
+        tsuba::ErrorCode::NotFound,
+        "{} is not a valid RDG (no manifests found)", rdg_uri);
   }
 
   return std::make_pair(latest_version, views_found);
 }
 
+katana::Result<std::pair<uint64_t, std::vector<tsuba::RDGView>>>
+tsuba::ListAvailableViews(
+    const std::string& rdg_dir, std::optional<uint64_t> version) {
```
witchel, line 221: Is this also for backward compatability?  If so, when can it go away?

witchel (approved):
Sticky wicket.

libtsuba/include/tsuba/tsuba.h
scober: There are some calls in the enterprise repo that need to be switched over

libtsuba/src/tsuba.cpp
yshalabi: Why don't we revert back to the pre #575 version or fix the empty-graph corner case in katana-enterprise?

scober: My goal was to distinguish between non-RDG directories and RDG directories where we didn't find the right view. But it doesn't seem like we care about that distinction.

tylershunt: It's different than that right? That distinction is still important, this is about distinguishing between a non-rdg directory and one were we didn't find the right version. I think if a specific version is asked for it's that last distinction that does not matter.

I.e. this function doesn't say anything about what the "right" view its. It's just telling you for some RDG dir, this is the latest version and these are the available views or if you're only interested in a specific version here are the valid views it has.

libtsuba/src/tsuba.cpp
```diff
@@ -202,14 +203,28 @@ tsuba::ListAvailableViews(
...
   }
 
-  if (views_found.empty()) {
-    return KATANA_ERROR(
-        tsuba::ErrorCode::NotFound, "no views found for that version");
+  if (!acceptable_version_found) {
+    if (version) {
+      return KATANA_ERROR(
+          tsuba::ErrorCode::NotFound,
+          "no views found for version {} in directory {}", version.value(),
```
tylershunt: +1 I like  these error messages. I would prefer to leave adding this context  that the caller is aware of up to the caller though. If we add context everwhere  we get stutters  in error messages like:
`rdg location s3://bucket_path, version 8: no views found for version 8 in directory s3://bucket_path"`

libtsuba/src/tsuba.cpp
```diff
@@ -202,14 +203,28 @@ tsuba::ListAvailableViews(
...
+          rdg_dir);
+    } else {
+      return KATANA_ERROR(
+          tsuba::ErrorCode::NotFound,
+          "no manifest files found in directory: {}", rdg_dir);
```
tylershunt: ditto here on the context.  In a bunch of other places it's up to the caller to add context when the context is one of the parameters the caller provided.

libtsuba/include/tsuba/tsuba.h
witchel: Is there a follow-on PR to switch those over?

scober: Not yet. I can open one. But in general my sense is that an interface change is best implemented as an interface addition followed by an interface deletion. This way we don't have a tightly-coupled enterprise PR.

witchel: 100% agree with this methodology.  I just worry that it can end up being interface addition followed by, "Squirrels!"

PR #586, opened 2021-09-29
Enrich RDGSlice interface

scober:
Currently, RDGSlice supports only a subset of the full RDG interface.

Move more RDG functionality into RDGCore and then expose it from RDGSlice. One distinction that I think is important to make now is that RDGSlice is still read-only, despite providing access to all the same data structures as RDG.

This code is currently untested and unused, but I decided that I had written enough that I should merge what I have before continuing.

### DIFF
--- a/libtsuba/include/tsuba/RDG.h
+++ b/libtsuba/include/tsuba/RDG.h
@@ -186,43 +186,37 @@ public:
   katana::Result<void> UnbindTopologyFileStorage();
 
   /// Inform this RDG that its topology is in storage at this location
-  /// without loading it into memory. \param new_top must exist and be in
-  /// the correct directory for this RDG
+  /// without loading it into memory.
+  /// \param new_top must exist and be in the correct directory for this RDG but
+  /// it need not be writable
   katana::Result<void> SetTopologyFile(const katana::Uri& new_top);
 
   katana::Result<void> UnbindNodeEntityTypeIDArrayFileStorage();
 
-  /// Inform this RDG that its Node Entity Type ID Array is in storage at this location
-  /// without loading it into memory. \param new_type_id_array must exist and be in
-  /// the correct directory for this RDG
+  /// Inform this RDG that its Node Entity Type ID Array is in storage at this
+  /// location without loading it into memory.
+  /// \param new_type_id_array must exist and be in the correct directory for
+  /// this RDG but it need not be writable
   katana::Result<void> SetNodeEntityTypeIDArrayFile(
       const katana::Uri& new_type_id_array);
 
   katana::Result<void> UnbindEdgeEntityTypeIDArrayFileStorage();
 
-  /// Inform this RDG that its Edge Entity Type ID Array is in storage at this location
-  /// without loading it into memory. \param new_type_id_array must exist and be in
-  /// the correct directory for this RDG
+  /// Inform this RDG that its Edge Entity Type ID Array is in storage at this
+  /// location without loading it into memory.
+  /// \param new_type_id_array must exist and be in the correct directory for
+  /// this RDG but it need not be writable
   katana::Result<void> SetEdgeEntityTypeIDArrayFile(
       const katana::Uri& new_type_id_array);
-
-  void AddMirrorNodes(std::shared_ptr<arrow::ChunkedArray>&& a) {
-    mirror_nodes_.emplace_back(std::move(a));
-  }
-
-  void AddMasterNodes(std::shared_ptr<arrow::ChunkedArray>&& a) {
-    master_nodes_.emplace_back(std::move(a));
-  }
 
   //
   // accessors and mutators
   //
 
-  const katana::Uri& rdg_dir() const { return rdg_dir_; }
-  void set_rdg_dir(const katana::Uri& rdg_dir) { rdg_dir_ = rdg_dir; }
+  const katana::Uri& rdg_dir() const;
+  void set_rdg_dir(const katana::Uri& rdg_dir);
 
-  uint32_t partition_id() const { return partition_id_; }
-  void set_partition_id(uint32_t partition_id) { partition_id_ = partition_id; }
+  uint32_t partition_id() const;
 
   /// The node properties
   const std::shared_ptr<arrow::Table>& node_properties() const;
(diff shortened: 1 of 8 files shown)


e-mcginnis (commented):
I'm not totally sure where we are trying to draw the line between RDGCore and RDG but In general I like the direction these changes are going with regards to moving lower level RDG stuffs into RDGCore. 

LGTM, but I would feel more comfortable if someone more experienced with RDG gave their input as well.

scober:
> I'm not totally sure where we are trying to draw the line between RDGCore and RDG but In general I like the direction these changes are going with regards to moving lower level RDG stuffs into RDGCore.

In general, `RDGCore` is anything that is shared between `RDG` and `RDGSlice`. That used to be a smaller set of things because `RDGSlice` didn't need to support the partition metadata stuff because it was only used with unpartitioned graphs. As we change that, more stuff is shared. Is that at all clarifying?

libtsuba/include/tsuba/RDG.h
```diff
@@ -186,43 +186,36 @@ class KATANA_EXPORT RDG {
...
 
   katana::Result<void> UnbindNodeEntityTypeIDArrayFileStorage();
 
-  /// Inform this RDG that its Node Entity Type ID Array is in storage at this location
-  /// without loading it into memory. \param new_type_id_array must exist and be in
-  /// the correct directory for this RDG
+  /// Inform this RDG that its Node Entity Type ID Array is in storage at this
+  /// location without loading it into memory.
+  /// \param new_type_id_array must exist and be in the correct directory for
+  /// this RDG
```
witchel: And it must be writable? Maybe not.

libtsuba/include/tsuba/RDG.h
```diff
@@ -186,43 +186,36 @@ class KATANA_EXPORT RDG {
...
 
-  /// Inform this RDG that its Edge Entity Type ID Array is in storage at this location
-  /// without loading it into memory. \param new_type_id_array must exist and be in
-  /// the correct directory for this RDG
+  /// Inform this RDG that its Edge Entity Type ID Array is in storage at this
+  /// location without loading it into memory.
+  /// \param new_type_id_array must exist and be in the correct directory for
+  /// this RDG
   katana::Result<void> SetEdgeEntityTypeIDArrayFile(
       const katana::Uri& new_type_id_array);
```
witchel, line 210: It isn't part of your PR, but I hate this, "here RDG, suck on this random data structure I'm feeding you" style of RDG behavior.  The RDG class truly is the Frankenstein's monster of object creation.

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -10,14 +10,15 @@
...
+#include "tsuba/RDGLineage.h"
 #include "tsuba/tsuba.h"
 
 namespace tsuba {
 
 class RDGManifest;
 class RDGCore;
 
-/// A contiguous piece of an RDG
+/// A read-only contiguous piece of an RDG
```
witchel: Tyler is often a genius at naming things, but I never really know what an RDGSlice is supposed to be.  It is a per-host section of an RDG?  a per-partition?  Does the system create slices?  Am I supposed to know how many slices there are?

Anyway, this comment is mostly me whining about my own ignorance, but it might be nice to have a comment in this file that gives programmers a bit of a heads up as to what is going on.

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -44,9 +45,47 @@ class KATANA_EXPORT RDGSlice {
       const std::optional<std::vector<std::string>>& node_props = std::nullopt,
       const std::optional<std::vector<std::string>>& edge_props = std::nullopt);
 
+  // metadata sorts of things
+  const katana::Uri& rdg_dir() const { return rdg_dir_; }
```
witchel: Random point, but why is `rdg_dir `in RDGSlice?  Should there be just one of these for the whole RDG?  It seems like something for RDGCore to me.

libtsuba/src/RDG.cpp
```diff
@@ -211,61 +154,60 @@ tsuba::RDG::WritePartArrays(const katana::Uri& dir, tsuba::WriteGroup* desc) {
...
-        StoreArrowArrayAtName(master_nodes_[i], dir, name, desc), "storing {}",
+        StoreArrowArrayAtName(master_nodes()[i], dir, name, desc), "storing {}",
         name);
     next_properties.emplace_back(tsuba::PropStorageInfo(name, path));
   }
 
-  if (host_to_owned_global_node_ids_ != nullptr) {
-    std::string name = kHostToOwnedGlobalNodeIDsPropName;
+  if (host_to_owned_global_node_ids() != nullptr) {
+    std::string name = RDGCore::HostToOwnedGlobalNodeIDsPropName;
```
witchel: If it is a constant, should it start with a `k`?

libtsuba/include/tsuba/RDGSlice.h
witchel: I'm even more confused seeing that `tsuba::RDG::rdg_dir()` returns `core_->rdg_dir();`

witchel (approved):
This was a nice size for review, thanks.

libtsuba/include/tsuba/RDG.h
scober: I agree. I think the tricky part is that it is both the reader and the writer of RDGs (among other things). I don't actually know how much use there is for one-off setting of data structures in RDG. Maybe we could reduce these setting interfaces to something more manageable like `SetTypes()`, etc.

libtsuba/include/tsuba/RDGSlice.h
scober: per-host and per-partition are the wrong way to think about it. It _should_ be any contiguous slice of a partition. It shouldn't care what the slice is or why it was loaded. I can try to improve this comment.

libtsuba/include/tsuba/RDGSlice.h
scober: This seems like an oversight. It should definitely be returning `core_->rdg_dir()`.

libtsuba/src/RDG.cpp
scober: That is an interesting question. I thought that because it was in the `RDGCore` namespace it might deserve special treatment. But I suppose prefixing with 'k' is how we indicate that something is a constant so that is probably a good idea.

libtsuba/include/tsuba/RDG.h
scober: I don't think so. I think this is just for loading. I added that to the comment.

libtsuba/include/tsuba/RDG.h
witchel: Just jawing here, but my sore spot with RDG for years now (yikes) is the mushy lifecycle.   If we always loaded RDGs from disk and wrote them back, I think the code could have a much narrower interface.  But we support in-memory creation of RDGs and there are other complications like the PropertyGraphs have their own lifecycle that interacts with RDGs and the result is a mess of individual data structures that are initialized higgledy-piggledy and passed around with no discernable rhyme or reason.

But yes, if we could group some of these data structures into a `SetTypes `that took several related data structures that would cut down the API bloat that I think makes RDGs messy and difficult to control.

libtsuba/include/tsuba/RDGSlice.h
witchel: Ok, I hear you, a slice is just a slice man.  There is no spoon.

But I still have two problems.

First, what is a slice of a complex data structure like an RDG?  A slice of an array I can understand.  Does RDGSlice mean a contiguous run of node IDs and then whatever else hangs off them?  Why isn't that an RDGBlockedSubGraph?

Second, while flexibility is great, giving the user some sense of why they might need an RDGSlice is helpful.  So if RDGSlices are not per-host or per-partition why do I want one?

libtsuba/include/tsuba/RDG.h
scober: I was also imagining that we could add `SetTopology()` or `SetProperties()`. Generally reduce the number of interfaces to a few that have a clear purpose.

libtsuba/include/tsuba/RDGSlice.h
scober: Will you take a look at it now? The reason it is called slice is that the CSR structure turns our subgraph into a series of slices. Another way of saying that would be that RDG is composed mostly of "trivially slice-able" substructures.

witchel: Great comment, thank you.

PR #588, opened 2021-09-30
 libtsuba API for optional topology storage 

e-mcginnis:
***New Classes***

PartitionTopologyMetadataEntry - This holds the basic metadata for a topology. So things like number of nodes/edge, transpose state, edge/node sort state, and topology file path. These metadata items are needed so users of RDG can quickly ask for a specific form of topology without us having to load each topology into memory. loaded/stored in json.

PartitionTopologyMetadata - Holds all of the PartitionTopologyMetadataEntries. This is managed by the RDGPartHeader

RDGTopology - This is the RDG layers in memory representation of a topology. When we load a graph, each PartitionMetadataEntry gets wrapped in an RDGTopology. It then manages, a FileView, mapping pointers to that FileView, and updating the PartitionMetadataEntry. Binding the FileView to the topology file path only occurs when a user asks for the particular topology. An RDGTopology can also be created from in memory arrays, instead of from a PartitionMetadataEntry, which is how new optional topologies are added.

RDGTopologyManager - This is how RDGCore keeps track of the various RDGTopology objects that are available.

***What the API looks like***
"Load"
1) a user constructs a minimal Shadow RDGTopology. This Shadow only has metadata fields (transpose_state, node/edge sort_state) set.
2) user passes the Shadow RDGTopology to RDG::GetTopology()
3) RDG then asks RDGCore to ask its RDGTopologyManager if it has a RDGTopology that matches the Shadow. The first match is returned.
  - if a match is found, the RDGTopologys topology file is Bind()-ed to the FileView and Mapped
  - if a match is not found, an error is returned up the call stack. The user can choose to ignore this error and move along, or fail if it depends on the topology being available. (like the CSR topology)
4) the user can then use the RDGTopology pointer to access all parts of the topology

"Store"
1) a user ensure that any new RDGTopologies it holds are passed to the RDGTopologyManager by calling RDG::AddTopology(RDGTopology topo). The user then calls RDG::Store like usual.
2) RDG::Store tells RDGTopologyManager to store all RDGTopology objects it holds.
3) a RDGTopology is written to disk if either its path is empty(indicating the rdg has moved, or has not been written) or its file store is invalid
4) RDGTopology writes the available data structures to a new file, and the metadata is passed to its PartitionTopologyMetadataEntry
5) RDGPartHeader writes out the PartitionTopologyMetadata as json

***Changes of note***

- Add PaddedWrite() to FileFrame
enables users to easily pad their writes to arbitrary byte boundries

- implemented  usage of the RDGTopology API in PropertyGraph to load/store
optional topologies

- rework DropTopology to DropAllTopology
Since we no longer have a single topology, if we want to drop the
topologies to save memory we must unmap and unbind all of the topologies

- added test for RDGTopology API in libtsuba

JIRA: KAT-574

TODO:
- ~upload the storage_format_version_2 rdg so `make input` has access to it. until then the libtsuba/topology test will only pass on my machine and not here.~ uploaded graph, test should now pass
- ~add libgalois test for optional topology creation, loading, and storing.~ done
- ~remove rdg ptr from GraphTopology.h~ done
- unify the PartitionMetadata.h transposed flag with the PartitionTopologyMetadatas transpose_state : this is done in a functional way, but more work will be placed in a future PR
- update the RFC to reflect the work done here


libtsuba/include/tsuba/FileFrame.h
```diff
@@ -78,6 +79,20 @@ class KATANA_EXPORT FileFrame : public arrow::io::OutputStream {
 
   const std::string& path() const { return path_; }
 
+  /// Calculate the number of padding bytes needed to reach nearest byte_boundry
+  /// uint64_t byte_boundry : number of bytes per boundry
+  /// uint64_t num_bytes : size of bytes
```
scober: Can you make these types match the types in the function signature? I know it is pedantic, but I like to differentiate between `uint64_t` and `size_t`.

libtsuba/src/FileFrame.cpp
```diff
@@ -187,4 +187,32 @@ FileFrame::Write(const std::shared_ptr<arrow::Buffer>& data) {
 }
 //////////// End arrow::io::BufferOutputStream method definitions ////////
 
+arrow::Status
```
scober: I think it is unnecessary to return `arrow::Status` here. We return `arrow::Status` various places in this class and in `FileView` because these classes inter-operate with the arrow library. It would probably be cleaner to add a layer of abstraction between "class that writes stuff to storage" and "class that can implements arrow storage-writing interface" but instead we just have comments marking which functions are for arrow and which are for us.

Since these are for us, I think they should return `katana::Result`. Alternatively, if you don't want to deal with this you can add a Jira issue and assign it to me. KAT-1452 is tracking a similar issue for `FileView`.

libtsuba/include/tsuba/FileFrame.h
```diff
@@ -78,6 +79,20 @@ class KATANA_EXPORT FileFrame : public arrow::io::OutputStream {
...
+  /// ex: if num_bytes = 65, and byte_boundry = 8, return = 7
+  static uint64_t calculate_padding_bytes(
+      int64_t num_bytes, size_t byte_boundry) {
+    return (~num_bytes + 1) & (byte_boundry - 1);
```
scober: Please add a comment explaining your bit-twiddling magic.

Also I think this implementation may be incorrect. `num_bytes=3` and `byte_boundary=5` gives `padding_bytes=4`.

scober (commented):
Added comments for the first commit since this is a large PR.

libtsuba/include/tsuba/FileFrame.h
scober: https://onlinegdb.com/DzQENWkI7

e-mcginnis: it assumes the byte boundary is a power of 2, which I think is reasonable. I'll add a check to ensure that is the case, and a comment declaring that assumption (and how the magic works).

the other option would be something like this https://onlinegdb.com/IaFroZfaIF which is theoretically a little slower but works in all cases.

scober: I think this function is unlikely to be a performance bottleneck so I am not opposed to a slower and more general implementation.

libgalois/include/katana/GraphTopology.h
```diff
@@ -266,6 +282,10 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
     return find_edge(src, dst) != edges(src).end();
   }
 
+  const PropertyIndex* edge_property_index_data() const noexcept {
```
scober, line 282: +1 for more indirection!

libgalois/include/katana/GraphTopology.h
```diff
@@ -778,10 +826,29 @@ class KATANA_EXPORT EdgeTypeAwareTopology
...
+
+  auto dest_data() const noexcept { return edge_shuff_topo_->dest_data(); }
+
+  auto edge_property_index_data() const noexcept {
+    return topo().edge_property_index_data();
+  }
+
+  const EntityType* edge_type_index_data() const noexcept {
+    return edge_type_index_->index_to_type_map_data();
+  }
```
scober: This seems like a good time to bring up our encapsulation problem. I am certainly guilty of perpetuating it, but I think we should stop exposing pointers to class internals. I think one fundamental roadblock we are going to have is that these classes often don't have enough context to construct themselves. So we are going to need a way for outside code to manipulate the internals. I guess we can look for a more principled way to do it? If anyone has thoughts, I would love to noodle on this a little.

libgalois/test/property-file-graph.cpp
```diff
@@ -336,6 +336,7 @@ TestRoundTrip() {
       ++value;
     }
   }
+
```
scober: delete?

libtsuba/include/tsuba/RDG.h
```diff
@@ -180,15 +177,21 @@ class KATANA_EXPORT RDG {
...
   /// Inform this RDG that its topology is in storage at this location
   /// without loading it into memory. \param new_top must exist and be in
   /// the correct directory for this RDG
-  katana::Result<void> SetTopologyFile(const katana::Uri& new_top);
+  /// *This function is only for use in graph conversion, and will deprecated*
+  katana::Result<void> SetCSRTopologyFile(const katana::Uri& new_top);
+
+  /// Ask this RDG if it has a topology matching the fields in shadow
+  /// If it does, the RDG returns the topology
+  katana::Result<tsuba::RDGTopology*> GetTopology(const RDGTopology& shadow);
```
scober, line 198: Why return a raw pointer here?

libtsuba/include/tsuba/RDG.h
```diff
@@ -180,15 +177,21 @@ class KATANA_EXPORT RDG {
...
   /// Load the RDG described by the metadata in handle into memory.
   static katana::Result<RDG> Make(RDGHandle handle, const RDGLoadOptions& opts);
 
-  katana::Result<void> UnbindTopologyFileStorage();
-
   /// Inform this RDG that its topology is in storage at this location
   /// without loading it into memory. \param new_top must exist and be in
   /// the correct directory for this RDG
-  katana::Result<void> SetTopologyFile(const katana::Uri& new_top);
+  /// *This function is only for use in graph conversion, and will deprecated*
```
scober: Could you clarify this comment a little? What do you mean by graph conversion, why does that mean it can/will be deprecated?

libtsuba/include/tsuba/RDG.h
```diff
@@ -237,7 +240,7 @@ class KATANA_EXPORT RDG {
   void DropEdgeProperties();
 
   /// Remove topology data
-  katana::Result<void> DropTopology();
+  katana::Result<void> DropAllTopology();
```
scober: nit: `DropAllTopologies()`

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -68,6 +70,7 @@ class KATANA_EXPORT RDGSlice {
   // Data
   //
 
+  tsuba::RDGTopology* topology_;
```
scober: Why can't this be part of `RDGCore`?

scober (commented):
I didn't make it all the way through the big commit, but I accumulated too many comments so I am stopping.

libgalois/include/katana/GraphTopology.h
e-mcginnis: agreed that this is gross. Unfortunately, I'm just following the existing pattern here. We don't have many options for getting data out of numarrays, and copying it needlessly would be quite painful.

libtsuba/include/tsuba/RDG.h
e-mcginnis: I'm open to suggestions for other solutions (I'm still getting my cpp sea legs, so please bare with me :) )

I originally choose to return a pointer for 3(4?) reasons:

1) RDGSlice wants consistent access to what it treats as the "main" topology, and using a reference became messy fast.
2) GraphTopology needed some way of tracking which RDGTopology it had built from. This is no longer the case and will be removed (mentioned in my TODOs)
3) the memory is completely controlled by the RDGTopologyManager, and thus the RDGCore, and needs to stick around for the lifetime of the RDGCore instance so shared_ptr and unique_ptr seemed wrong to use here (but once again I could be misunderstanding something about how to use these smart pointers)
4) when you use a non-const reference, the compiler throws a warning stating to use a pointer instead. 

I hope some of that makes some sort of sense, and provides enough context.

libtsuba/include/tsuba/RDGSlice.h
e-mcginnis: The way I see it, there is really no reason RDGCore should have to be tied to some specific topology. RDGCore doesn't need it, and RDG doesn't need it (right now RDG::DoMake loads/maps the main CSR topology, but nothing actually depends on it being loaded so I'm toying with removing the mapping step altogether)

I believe RDGSlice only actually needs this pointer for one function call, so it is very possible we could remove it altogether. I just didn't dig enough into it yet since my focus was on other parts of the code.

If all of this makes sense, I'll add a TODO to clean this pointer up.

libgalois/include/katana/GraphTopology.h
scober: What about `PropertyGraphRetractor`? It seems to be our current container for leaky abstractions, maybe we could push the RDG data copying stuff into `PGR`?

libtsuba/include/tsuba/RDG.h
scober: - What do you mean by messy?
- Why is it non-const?

libtsuba/include/tsuba/RDGSlice.h
scober: That does all make sense. Thank you.

libtsuba/include/tsuba/RDG.h
e-mcginnis: re: messy: making an RDGSlice instance hold a reference to a RDGTopology means the RDGSlice instance needs to be constructed with that reference. looking again, this wouldn't be too bad to do. That said, I would prefer to remove the need for RDGSlice to hold a reference/pointer to RDGTopology. 

re: non-const: RDG::GetTopology() needs a non const RDGTopology so it can be bound/mapped. GraphTopology currently needs to be able to call Store (which is non-const) on the RDGTopology (but like I mentioned elsewhere, this usage is getting removed.) so its very possible we can restrict RDG::GetTopology to returning a const RDGTopology instead.

libtsuba/src/FileFrame.cpp
e-mcginnis: Fixed

libtsuba/include/tsuba/FileFrame.h
e-mcginnis: fixed

libtsuba/include/tsuba/FileFrame.h
e-mcginnis: fixed

libtsuba/include/tsuba/RDG.h
e-mcginnis: This comment was a little out of date, updated comment and function name to better reflect what it does.

libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -0,0 +1,303 @@
...
+#include "tsuba/Errors.h"
+#include "tsuba/FileView.h"
+#include "tsuba/WriteGroup.h"
+#include "tsuba/tsuba.h"
+
+namespace tsuba {
+
+class PartitionTopologyMetadataEntry;
+
+// this should match what is used in PartitionTopologyMetadata.h
```
scober: Should this be defined in the header and included here?

libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -0,0 +1,303 @@
...
+
+//TODO: emcginnis the import path also defines the csr topology layout, make sure to update that as well.
+class KATANA_EXPORT RDGTopology {
+public:
+  RDGTopology();
+  RDGTopology(PartitionTopologyMetadataEntry* metadata_entry);
+  ~RDGTopology();
+  RDGTopology(RDGTopology&& other) noexcept;
+  RDGTopology(const RDGTopology& no_copy) = delete;
+  RDGTopology& operator=(const RDGTopology& other) noexcept;
```
scober: Why do you allow copy assignment but not a copy constructor?

libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -0,0 +1,303 @@
...
+  void set_topology_state(TopologyKind state) { topology_state_ = state; }
+
+  TransposeKind transpose_state() const { return transpose_state_; }
+  void set_transpose_state(TransposeKind state) { transpose_state_ = state; }
+
+  EdgeSortKind edge_sort_state() const { return edge_sort_state_; }
+  void set_edge_sort_state(EdgeSortKind state) { edge_sort_state_ = state; }
+
+  NodeSortKind node_sort_state() const { return node_sort_state_; }
+  void set_node_sort_state(NodeSortKind state) { node_sort_state_ = state; }
```
scober: Do we need all of these setters? What is `set_num_edges()` for? Why would I want to set that value without changing the topology data structures as well? What about `set_adj_indices()`? Is there a use for setting the adjacency indices without setting the dests?

libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -0,0 +1,303 @@
...
+      const uint8_t* edge_condensed_type_id_map_,
+      const uint8_t* node_condensed_type_id_map_);
+
+  // Make an RDGTopology from on storage metadata
+  static katana::Result<tsuba::RDGTopology> Make(
+      PartitionTopologyMetadataEntry* entry);
+
+private:
+  //** IMPORTANT** if you add any fields, you must add them to the assignment operator in the cpp file
+  //TODO: emcginnis, this ^ is gross. how can we make this not required?
```
scober: What if you did

```cpp
operator=(RDGTopology& no_copy_assignment) = delete;
operator=(RDGTopology&& other)=default;
```
Do you need to allow copy assignment?

libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -0,0 +1,303 @@
...
+
+  // valid immediately, stored in metadata
+  uint64_t num_edges_{0};
+  uint64_t num_nodes_{0};
+  //TODO: emcginnis PartitionMetadata.h has a transposed flag, see if that is the same thing and replace it?
```
scober, line 274: It should be the same thing

libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -0,0 +1,303 @@
...
+  // rather than purely created in memory
+  PartitionTopologyMetadataEntry* metadata_entry_{nullptr};
+
+  // must be loaded from file store or set
+  const uint64_t* adj_indices_{nullptr};
+  const uint32_t* dests_{nullptr};
+  const uint64_t* edge_index_to_property_index_map_{nullptr};
+  const uint64_t* node_index_to_property_index_map_{nullptr};
+  const uint8_t* edge_condensed_type_id_map_{nullptr};
+  const uint8_t* node_condensed_type_id_map_{nullptr};
```
scober, line 302: What is a condensed type id map?

libtsuba/src/PartitionTopologyMetadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+  if (entries_.empty()) {
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument, "no topology metadata entries present");
+  }
+  for (size_t i = 0; i < num_entries_; i++) {
+    auto res = ValidateEntry(entries_.at(i));
+    if (!res) {
+      return res;
+    }
+  }
```
scober: It looks like this could be shortened to
```cpp
KATANA_CHECKED(ValidateEntry(entries_.at(i));
```

libtsuba/src/PartitionTopologyMetadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+katana::Result<void>
+PartitionTopologyMetadata::ValidateEntry(
+    const PartitionTopologyMetadataEntry& entry) const {
+  if (entry.path_.empty()) {
+    std::string name = EntryToString(entry);
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument, "no topology file path: topology: {} ",
+        name);
+  }
+  if (entry.path_.find('/') != std::string::npos) {
```
scober, line 70: It looks like this will return an error if '/' _is_ found?

libtsuba/src/RDG.cpp
```diff
@@ -988,25 +958,27 @@ tsuba::RDG::full_edge_schema() const {
...
+  return core_->RegisterCSRTopologyFile(new_top.BaseName(), rdg_dir_);
+}
+
+katana::Result<tsuba::RDGTopology*>
+tsuba::RDG::GetTopology(const tsuba::RDGTopology& shadow) {
+  auto res = core_->topology_manager().GetTopology(shadow);
+  if (!res) {
+    return res.error();
+  }
+  RDGTopology* topology = res.value();
```
scober: `KATANA_CHECKED()`?

scober (commented):
Still on commit #2, thank you for doing this.

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: this is probably a result of of my unfamiliarity with c++, but I'll try to explain my thought process.
1) I would prefer to not allow any copying of RDGTopology, since it holds the `FileView file_storage` memory that can be quite large
2) I want RDGTopologyManager to hold the instances of RDGTopology, so I need to be able to insert an RDGTopology into the RDGTopologyManager https://github.com/KatanaGraph/katana/blob/d6a00838f4250fe8cbb42353de37d1aacbfb219f/libtsuba/src/RDGTopologyManager.h#L32

With how RDGTopologyManager and RDGTopology are defined and constructed right now, these two goal conflict. Its not clear to me how best to resolve this without allowing copy assignment, but disallowing the copy constructor. 
Would be more than happy to hear suggestions on how to re-work RDGTopology(Manager) to better reach those two goals.

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: It will be used in GraphTopology.h to construct EdgeTypeAware topologies in a future PR

libtsuba/src/PartitionTopologyMetadata.h
```diff
@@ -0,0 +1,124 @@
...
+#define KATANA_LIBTSUBA_PARTITIONTOPOLOGYMETADATA_H_
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+#include <utility>
+
+#include <bits/stdint-uintn.h>
```
tylershunt: the `bits` header here scares me. I think `<cstdint>` makes it unnecessary no?

libtsuba/src/PartitionTopologyMetadata.h
```diff
@@ -0,0 +1,124 @@
+#ifndef KATANA_LIBTSUBA_PARTITIONTOPOLOGYMETADATA_H_
+#define KATANA_LIBTSUBA_PARTITIONTOPOLOGYMETADATA_H_
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <string_view>
+#include <unordered_map>
```
tylershunt, line 8: I don't think `unordered_map` is used anywhere

libtsuba/src/PartitionTopologyMetadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return katana::ResultSuccess();
+}
+
+std::string
+PartitionTopologyMetadata::EntryToString(
+    const PartitionTopologyMetadataEntry& entry) const {
+  boost::format fmt = boost::format(
+                          "transpose_%1%_node_sort_%2%_edge_sort_%3%_num_nodes_"
+                          "%4%_num_edges_%5%") %
```
tylershunt: This is fine, but in general I'd prefer to limit our dependence on boost though maybe to get rid of it #someday, this function could be something like:
```c++
return fmt::format("transpose_{}_node_sort_{}_edge_sort_{}_num_nodes_{}_num_edges_{}", entry.transpose_state_,
    entry.node_sort_state_, entry.edge_sort_state_, entry.num_nodes_, entry.num_edges_);
```
instead

libtsuba/src/FileFrame.cpp
```diff
@@ -187,4 +187,32 @@ FileFrame::Write(const std::shared_ptr<arrow::Buffer>& data) {
 }
 //////////// End arrow::io::BufferOutputStream method definitions ////////
 
+katana::Result<void>
+FileFrame::PaddedWrite(
```
tylershunt, line 200: Didn't see uses of these functions, In general we tend to write arrow arrays out as parquet. That said, I don't object to this being here.

libtsuba/src/RDGTopology.cpp
```diff
@@ -0,0 +1,611 @@
...
+    : metadata_entry_(metadata_entry) {}
+
+RDGTopology::RDGTopology() = default;
+
+RDGTopology::~RDGTopology() = default;
+
+RDGTopology::RDGTopology(RDGTopology&& other) noexcept = default;
+
+RDGTopology&
+RDGTopology::operator=(const RDGTopology& other) noexcept {
```
tylershunt: Better just to define a move assignment operator rather than override the default behavior of copy.

libtsuba/include/tsuba/RDGTopology.h
tylershunt: I think the example you gave is already using the implicitly defined move assignment operator. I also think the copy assignement and copy constructor are already implicitly deleted because `FileView` itself is not copyable and is a class member.

This class seems like a good candidate for "the rule of zero". https://en.cppreference.com/w/cpp/language/rule_of_three

That is, I would expect the code you posted to work even if you delete the above 4 lines.

libtsuba/include/tsuba/RDGTopology.h
scober: Is it one `EntityTypeID` per node/edge?

libtsuba/src/PartitionTopologyMetadata.h
e-mcginnis: bah, this is emacs lsp client trying to be *super clever* and add includes automatically for me. I thought I turned that off...
will remove these

libtsuba/src/FileFrame.cpp
e-mcginnis: These get used in https://github.com/KatanaGraph/katana/blob/d6a00838f4250fe8cbb42353de37d1aacbfb219f/libtsuba/src/RDGTopology.cpp#L287
added by the next commit

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: yes, but unfortunately I can't do that as `PartitionTopologyMetadataEntry` is an incomplete type at this point. 
```
error: incomplete type ‘tsuba::PartitionTopologyMetadataEntry’ used in nested name specifier
  289 |   using RDGTopologySet = std::array<RDGTopology, tsuba::PartitionTopologyMetadataEntry::max_num_topologies>;
```

Since RDGTopology.h is externally accessible, and PartitionTopologyMetadata.h is internally accessible, I have to forward declare PartitionTopologyMetadataEntry.

I think the correct solution here would be for me to move `max_num_topologies` to `tsuba.h`? Not sure though.

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: This condensed map is simply a mapping of EntityTypeIDs to a smaller space
a graph can have 255 unique EntityTypeID, but in practice there are only a few, so we create this map from 255 to say 5 things. (paraphrased from conversation with @amberhassaan)

the `adj_indices` for an EdgeTypeAwareTopology is actually `per_type_adj_indices`, which is generated from some base topologys  `adj_indices` + the `condensed_type_id_map` generated at the time.

The layout of the `per_type_adj_indices` is strongly dependent on the layout of the `condensed_type_id_map`, so we must store both.

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: good point, we definitely do not need most of these setters. removing all besides `set_path()` and `set_metadata_entry()`

libtsuba/src/PartitionTopologyMetadata.cpp
e-mcginnis: Definitely, this was copy-pasted directly from:
https://github.com/KatanaGraph/katana/blob/e950b86809e6fb69223286670c5811b4904d7923/libtsuba/src/RDGPartHeader.cpp#L219
where the meaning seems to have gotten mixed up a bit. Looking at some on disk graphs, it seems like we explicitly *don't* want to see a leading `/`, and the error message that was added to that check by a later commit flipped the meaning.
Will fix the error message. 


cc @tylershunt in case I am misunderstanding something here

libgalois/include/katana/GraphTopology.h
```diff
@@ -167,6 +171,14 @@ class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
     return edge_property_index(eid);
   }
 
+  //TODO: emcginnis, this sort of glue belongs somewhere outside this class.
+  // move it to PropertyGraph property or some sort of GraphTopologyManager class?
+  // or at least make it part of the constructor if it is going to stay in here?
+  tsuba::RDGTopology* rdg_topology_ptr() const { return rdg_topology_ptr_; }
+  void set_rdg_topology_ptr(tsuba::RDGTopology* ptr) {
```
amberhassaan: I agree wholeheartedly and I'd like to avoid getters and setters (setters in particular). `libtsuba` seems to be rife with these.

libgalois/include/katana/GraphTopology.h
```diff
@@ -244,22 +258,62 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
...
+        &(rdg_topo->edge_index_to_property_index_map()[rdg_topo->num_edges()]),
+        edge_prop_indices.begin());
+
+    std::unique_ptr<EdgeShuffleTopology> shuffle =
+        std::make_unique<EdgeShuffleTopology>(EdgeShuffleTopology{
+            rdg_topo->transpose_state(), rdg_topo->edge_sort_state(),
+            std::move(adj_indices_copy), std::move(dests_copy),
+            std::move(edge_prop_indices)});
+
+    shuffle->set_rdg_topology_ptr(rdg_topo);
```
amberhassaan: why do I need to set this pointer?

libgalois/include/katana/GraphTopology.h
amberhassaan: sad

libgalois/include/katana/GraphTopology.h
amberhassaan: my first reaction is : let's not overload that donkey (or mule if you prefer).
My question is: why do I need these pointer returning functions? Why can't I have the class populate a struct (RDGTopology?) where these pointers will end up in?

libgalois/include/katana/GraphTopology.h
```diff
@@ -1390,30 +1532,87 @@ class KATANA_EXPORT PGViewCache {
...
   }
 
+  katana::Result<std::vector<tsuba::RDGTopology>> ToRDGTopology() {
+    std::vector<tsuba::RDGTopology> rdg_topos;
+
+    for (size_t i = 0; i < edge_shuff_topos_.size(); i++) {
+      // only need to store if we haven't before, or if we have updates
+      if (edge_shuff_topos_[i]->rdg_topology_ptr() == nullptr ||
+          !edge_shuff_topos_[i]->rdg_topology_ptr()->file_storage().Valid()) {
+        tsuba::RDGTopology topo = KATANA_CHECKED(tsuba::RDGTopology::Make(
```
amberhassaan: IMO, this `RDGTopology` creating code belongs in `EdgeShuffleTopology`.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -290,6 +296,19 @@ class KATANA_EXPORT PropertyGraph {
     return edge_entity_type_manager_;
   }
 
+  katana::Result<tsuba::RDGTopology*> LoadTopology(
```
amberhassaan: Why is this method needed here?  topology should be loaded as part of `BuildView` when possible.

libtsuba/include/tsuba/RDGTopology.h
amberhassaan: Your description is correct.

libtsuba/src/PartitionTopologyMetadata.h
```diff
@@ -0,0 +1,122 @@
...
+  std::string path_{""};
+  uint64_t num_edges_{0};
+  uint64_t num_nodes_{0};
+  bool edge_index_to_property_index_map_present_{false};
+  bool node_index_to_property_index_map_present_{false};
+  bool edge_condensed_type_id_map_present_{false};
+  bool node_condensed_type_id_map_present_{false};
+  uint8_t topology_state_{0};
+  uint8_t transpose_state_{0};
+  uint8_t edge_sort_state_{0};
```
amberhassaan: Can we use the same enums here instead of ints?

libtsuba/src/RDGTopology.cpp
```diff
@@ -0,0 +1,579 @@
...
+
+constexpr uint64_t
+tsuba::RDGTopology::GetGraphSize() const {
+  /// version, sizeof_edge_data, num_nodes, num_edges
+  constexpr int mandatory_fields = 4;
+  int graphsize = (mandatory_fields + num_nodes_) * sizeof(uint64_t) +
+                  (num_edges_ * sizeof(uint32_t));
+
+  if (metadata_entry_->edge_index_to_property_index_map_present_) {
+    // 1 is for the magic number
```
amberhassaan, line 667: what do you mean?

amberhassaan (commented):
Partial set of comments. Looks OK overall.

libtsuba/src/PartitionTopologyMetadata.cpp
tylershunt: Agree it's a bad error, not a bad check.

libgalois/include/katana/GraphTopology.h
e-mcginnis: This is the pointer I am removing as one of my TODOs

e-mcginnis: Main reason I haven't removed it: I want to get the libgalois test working as-is first, I'm quite confident the pointer can be removed completely, but having something to test against makes things easier.

libtsuba/src/RDG.cpp
```diff
@@ -528,11 +486,19 @@ tsuba::RDG::DoMake(
...
+      core_->MakeTopologyManager(metadata_dir), "populating topologies");
+
+  // find & bind the default csr topology
+  RDGTopology shadow_csr = RDGTopology::MakeShadowCSR();
+  RDGTopology* csr = KATANA_CHECKED_CONTEXT(
+      core_->topology_manager().GetTopology(shadow_csr),
+      "unable to find csr topology, must have csr topology");
+
+  KATANA_LOG_VASSERT(csr != nullptr, "csr topology is null");
+  //TODO: emcginnis is there any reason we should be binding the csr topology here?
```
amberhassaan, line 443: minor: todo style should be: 
// TODO(emcginnis): do blah blah blah

libtsuba/src/RDGTopology.cpp
```diff
@@ -0,0 +1,579 @@
...
+  }
+
+  file_store_bound_ = true;
+  storage_valid_ = true;
+
+  return katana::ResultSuccess();
+}
+
+katana::Result<void>
+RDGTopology::Bind(
```
amberhassaan, line 88: why is this needed? Is it part of public API?

libtsuba/src/RDGTopology.cpp
```diff
@@ -0,0 +1,579 @@
...
+  // ensure the data file matches the metadata
+  KATANA_LOG_VASSERT(
+      num_nodes_ == data[2], "expected {} nodes, found {} nodes", num_nodes_,
+      data[2]);
+  KATANA_LOG_VASSERT(
+      num_edges_ == data[3], "expected {} edges, found {} edges", num_edges_,
+      data[3]);
+
+  adj_indices_ = &data[4];
+  const uint64_t* cursor = adj_indices_;
```
amberhassaan, line 146: This code with `cursor` and its math is a bit low-level and error prone. My recommendation is that we introduce a "byte data iterator". This iterator supports usual input iterator stuff, as well as things like `AdvanceBy<T>()`, which advances the cursor inside the iterator by `sizeof(T)`. All this math of `/2 + %2` will be hidden in the iterator and life will be better :).

libtsuba/src/RDGTopology.cpp
amberhassaan: OK, I get the idea of magic (somewhat) now.

amberhassaan (commented):
Remaining comments.

libtsuba/include/tsuba/RDGTopology.h
scober: I don't totally understand the point of the map, but I think that is okay. Should it be `EntityTypeID*`?

libtsuba/src/RDGTopology.cpp
e-mcginnis: Apologies, the doc comment on the Map function is a bit out of date. Will update it. 

The idea of the magic number here is to provide some sanity check between these of uint64_t, uint32_t, uint8_t, etc, arrays we are writing out.  The magic is simply the sum of the number of nodes and edges. Maybe a better name would be `sanity_value`. I could also see the argument for calculating a proper checksum on the data.

libtsuba/src/RDGTopology.cpp
e-mcginnis: Fully in support of this suggestion, but it might have to wait for a future PR.

libtsuba/src/RDGTopology.cpp
e-mcginnis: "rule of zero"d this away

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: I believe you are right @scober 
@amberhassaan is there a reason GraphTopology.h defines its own `GraphTopologyTypes::Entitytype` instead of using `katana::EntityTypeID`?

libgalois/include/katana/GraphTopology.h
e-mcginnis: Bringing the RDGTopology creation into each topology helped a bit with this, but we can't completely remove these because of how the different GraphTopology Classes are defined.

`ShuffleTopology` inherits from `EdgeShuffleTopology` so here we could simply make the data pointers that `ShuffleTopology` needs from `EdgeShuffleTopology` protected and we would be okay.

But: `EdgeTypeAwareTopology` is different, it inherits from `BasicTopologyWrapper<EdgeShuffleTopology>` which means `EdgeTypeAwareTopology` can't get access to the data pointers of the `EdgeShuffleTopology` it needs to create an RDGTopology.

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: From talking with Amber, it seems `GraphTopologyTypes::Entitytype` came about before `katana::EntityTypeID` was fully realized.
Going to convert to using `katana::EntityTypeID`

libgalois/include/katana/GraphTopology.h
e-mcginnis: rdg_topology_ptr is now no more

libgalois/include/katana/GraphTopology.h
e-mcginnis: done

libtsuba/src/RDGTopology.cpp
e-mcginnis: yes, other users of RDGTopology need more specific control over the file_storage (like RDGSlics)

libgalois/include/katana/PropertyGraph.h
e-mcginnis: In order for the `PGViewCache::BuildOrGet*` functions to get RDGTopologies, they need access to the `PG::rdg_`, so this function takes care of that.

libtsuba/include/tsuba/RDG.h
e-mcginnis: alright, the picture here has changed a bit. GraphTopology now no longer holds on to a copy of the `RDGTopology*` for Storing, so that use case is gone. But RDGSlice still needs non-const access to the RDGTopology so it can call RDGTopology::Bind()

libtsuba/include/tsuba/RDGSlice.h
e-mcginnis: this pointer is now gone. The one use case has been replaced with another call to GetTopology().

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -213,17 +228,17 @@ RDGPartHeader::Validate() const {
...
-    return KATANA_ERROR(ErrorCode::InvalidArgument, "topology_path is empty");
-  }
-  if (topology_path_.find('/') != std::string::npos) {
-    return KATANA_ERROR(
-        ErrorCode::InvalidArgument,
-        "topology_path doesn't contain a slash (/): {}", topology_path_);
+
+  auto res = topology_metadata_.Validate();
+  if (!res) {
+    return res;
```
tylershunt: If you like, this could just be `KATANA_CHECKED(topology_metadata_.Validate());`.

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -213,17 +228,17 @@ RDGPartHeader::Validate() const {
...
   }
 
   if (IsEntityTypeIDsOutsideProperties()) {
-    return ValidateEntityTypeIDStructures();
+    auto res = ValidateEntityTypeIDStructures();
+    if (!res) {
+      return res;
+    }
```
tylershunt: Another `KATANA_CHECKED` opportunity.

tylershunt (approved):
I'm fine with all of the libtsuba stuff.

libgalois/test/property-graph-optional-topologies.cpp
```diff
@@ -0,0 +1,188 @@
...
+  SortedGraphView loaded_sorted_view = pg2.BuildView<SortedGraphView>();
+
+  //TODO: emcginnis need some way to verify we loaded this view, vs just generating it again
+
+  KATANA_LOG_ASSERT(
+      generated_sorted_view.num_edges() == loaded_sorted_view.num_edges());
+  KATANA_LOG_ASSERT(
+      generated_sorted_view.num_nodes() == loaded_sorted_view.num_nodes());
+
+  for (std::pair<edge_iterator, edge_iterator> i(
```
amberhassaan: We have `make_zip_iterator` in `libsupport` that may simplify life a bit.

e-mcginnis:
@tylershunt dropped the extra bind mentioned here https://github.com/KatanaGraph/katana/pull/618#discussion_r728332400 since we now have a flag to mark whether a topology should be stored. In the case of graph-convert, that path marks the topology to be stored when we extract the metadata.

e-mcginnis:
@scober made some minor changes to RDGSlice.cpp, found that the RDGTopologyManager wasn't getting initialized since I was relying on the RDG to ask the core to initialize it.

libtsuba/src/PartitionTopologyMetadata.h
e-mcginnis: now using the enums here

libtsuba/include/tsuba/RDGTopology.h
e-mcginnis: Will clean up the duplicate EntityTypeId/uint8_ts in a future PR. This one has gone on long enough.

e-mcginnis:
I believe I have addressed all major comments folks have had, thanks again for everyones time reviewing this

e-mcginnis:
Enterprise PR passed all tests: https://github.com/KatanaGraph/katana-enterprise/pull/2208

libgalois/include/katana/GraphTopology.h
```diff
@@ -244,22 +245,59 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
...
-      KATANA_LOG_DEBUG_ASSERT(ret->has_transpose_state(TransposeKind::kNo));
+      KATANA_LOG_DEBUG_ASSERT(
+          ret->has_transpose_state(tsuba::RDGTopology::TransposeKind::kNo));
     }
 
     ret->sortEdges(pg, edge_sort_todo);
     return ret;
   }
 
+  static std::unique_ptr<EdgeShuffleTopology> Make(
```
amberhassaan, line 266: You should move the body of this method to .cpp

libgalois/include/katana/GraphTopology.h
```diff
@@ -377,11 +427,52 @@ class KATANA_EXPORT ShuffleTopology : public EdgeShuffleTopology {
     return ret;
   }
 
+  static std::unique_ptr<ShuffleTopology> Make(tsuba::RDGTopology* rdg_topo) {
+    KATANA_LOG_DEBUG_ASSERT(rdg_topo);
```
amberhassaan: Move body to .cpp

libgalois/include/katana/GraphTopology.h
```diff
@@ -1392,30 +1516,52 @@ class KATANA_EXPORT PGViewCache {
   PGViewCache& operator=(const PGViewCache&) = delete;
 
   template <typename PGView>
-  PGView BuildView(const PropertyGraph* pg) noexcept {
+  PGView BuildView(PropertyGraph* pg) noexcept {
     return internal::PGViewBuilder<PGView>::BuildView(pg, *this);
   }
 
+  katana::Result<std::vector<tsuba::RDGTopology>> ToRDGTopology() {
+    std::vector<tsuba::RDGTopology> rdg_topos;
```
amberhassaan: move body to .cpp.

libgalois/include/katana/GraphTopology.h
```diff
@@ -1324,11 +1448,11 @@ template <>
 struct PGViewBuilder<PGViewNodesSortedByDegreeEdgesSortedByDestID> {
   template <typename ViewCache>
   static PGViewNodesSortedByDegreeEdgesSortedByDestID BuildView(
-      const PropertyGraph* pg, ViewCache& viewCache) noexcept {
+      PropertyGraph* pg, ViewCache& viewCache) noexcept {
     auto sorted_topo = viewCache.BuildOrGetShuffTopo(
-        pg, EdgeShuffleTopology::TransposeKind::kNo,
-        ShuffleTopology::NodeSortKind::kSortedByDegree,
-        EdgeShuffleTopology::EdgeSortKind::kSortedByDestID);
+        pg, tsuba::RDGTopology::TransposeKind::kYes,
```
amberhassaan: `s/kYes/kNo/ ` here because this was one of the bugs in triangle-counting.

libgalois/include/katana/PropertyGraph.h
amberhassaan: I don't like the fact that this is part of public API if it's only needed by the PGViewCache.

libgalois/src/GraphTopology.cpp
```diff
@@ -130,11 +132,21 @@ katana::EdgeShuffleTopology::MakeOriginalCopy(const katana::PropertyGraph* pg) {
...
+      tsuba::RDGTopology::TransposeKind::kNo,
+      tsuba::RDGTopology::EdgeSortKind::kAny,
       std::move(copy_topo.GetAdjIndices()), std::move(copy_topo.GetDests()),
       std::move(edge_prop_indices)});
 }
 
+katana::Result<tsuba::RDGTopology>
+katana::EdgeShuffleTopology::ToRDGTopology() const {
+  tsuba::RDGTopology topo = KATANA_CHECKED(tsuba::RDGTopology::Make(
+      num_edges(), adj_data(), num_nodes(), dest_data(),
```
amberhassaan: This is a weird order of parameters. Why this order?

libgalois/test/property-graph-optional-topologies.cpp
amberhassaan: Consider using zip_iterator (from libsupport). Also, consider moving this into a common function as you're repeating the same pattern multiple times.

libtsuba/src/RDGTopology.cpp
amberhassaan: Add a TODO please (assuming you've decided to postpone this).

amberhassaan (approved):
Approving, but please address the comments before merging.

libgalois/include/katana/PropertyGraph.h
e-mcginnis: I renamed it to "GetTopology" since that seemed more accurate. 
Would you prefer PGViewCache was made a friend of PropertyGraph, or something else?

libtsuba/src/RDGTopology.cpp
e-mcginnis: TODO added.

libgalois/include/katana/PropertyGraph.h
amberhassaan: I'd prefer that PGViewCache be a friend. And I like `LoadTopology` better than `Get`. Get/Set names are lazy.

libgalois/src/GraphTopology.cpp
e-mcginnis: it now matches GraphTopology() ordering
`      adj_data(), num_nodes(), dest_data(), num_edges(),`

libgalois/include/katana/PropertyGraph.h
e-mcginnis: done

PR #592, opened 2021-10-01
Balance block_range

ddn0:
commit 4cc50701c4649fa4b553bc23a2083c26dc83ca78 (HEAD -> feature/block, ddn0/feature/block)
Author: Donald Nguyen <ddn0@users.noreply.github.com>
Date:   Fri Oct 1 13:40:48 2021 -0700

    libgalois: block_range more evenly

    Balance block ranges more evenly. The previous approach of taking the
    ceiling can leave the last block much smaller than all others. This is
    has not been significant in practice but it is easy enough to fix with
    integers by taking the floor and then adding 1 to the first remainder
    blocks.

    Also, move range functions to Range.h.

### DIFF
--- a/libgalois/include/katana/Range.h
+++ b/libgalois/include/katana/Range.h
@@ -33,6 +33,73 @@
 namespace katana {
 
 /**
+ * Finds the midpoint of a range.  The first half is always be bigger than
+ * the second half if the range has an odd length.
+ */
+template <typename Iterator>
+Iterator
+split_range(Iterator b, Iterator e) {
+  std::advance(b, (std::distance(b, e) + 1) / 2);
+  return b;
+}
+
+/**
+ * Returns a contiguous block from the range based on the number of divisions
+ * and the index of the block requested.
+ */
+template <typename Iterator>
+std::enable_if_t<
+    !std::is_integral<Iterator>::value, std::pair<Iterator, Iterator>>
+block_range(Iterator begin, Iterator end, unsigned idx, unsigned num) {
+  if (num == 0 || idx >= num) {
+    return std::make_pair(end, end);
+  }
+
+  size_t dist = std::distance(begin, end);
+
+  size_t block_size = dist / num;
+  size_t remaining = dist % num;
+
+  // Each block contains at least floor(dist / num) elements and may also have
+  // one more element to account for the k remaining elements (i.e., dist % num
+  // or dist - floor(dist / num)).
+  //
+  // Assign an extra element to the first k blocks (size is block_size + 1), and
+  // when idx >= k, account for the extra elements by adding k.
+  size_t b =
+      idx < remaining ? idx * (block_size + 1) : idx * block_size + remaining;
+  size_t e = b + block_size + (idx < remaining ? 1 : 0);
+
+  Iterator block_begin = begin;
+  std::advance(block_begin, b);
+
+  Iterator block_end = block_begin;
+  if (e == dist) {
+    // A micro-optimization to avoid advancing a non-random access iterator
+    // when idx == num - 1.
+    block_end = end;
+  } else {
+    std::advance(block_end, e - b);
+  }
+
+  return std::make_pair(block_begin, block_end);
+}
+
+/**
+ * Returns an integral range (a, b] that divides the input range (A, B] into
+ * contiguous blocks based on the number of divisions and the index of the block
+ * requested.
+ */
+template <typename Integer>
+std::enable_if_t<std::is_integral<Integer>::value, std::pair<Integer, Integer>>
+block_range(Integer begin, Integer end, unsigned idx, unsigned num) {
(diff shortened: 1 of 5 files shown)


libgalois/include/katana/Range.h
```diff
@@ -32,6 +32,66 @@
...
+
+  size_t block_size = dist / num;
+  size_t remaining = dist % num;
+
+  size_t b =
+      id < remaining ? id * (block_size + 1) : id * block_size + remaining;
+  size_t e = b + block_size + (id < remaining ? 1 : 0);
+
+  b = std::min(b, dist);
+  e = std::min(e, dist);
```
rob-isaac: are these necessary?

libgalois/include/katana/Range.h
```diff
@@ -32,6 +32,66 @@
...
+  }
+
+  size_t dist = std::distance(begin, end);
+
+  size_t block_size = dist / num;
+  size_t remaining = dist % num;
+
+  size_t b =
+      id < remaining ? id * (block_size + 1) : id * block_size + remaining;
+  size_t e = b + block_size + (id < remaining ? 1 : 0);
```
rob-isaac: can replace `(id < remaining ? 1 : 0)` with `(id < remaining)`

rob-isaac (approved):
Looks good

libgalois/include/katana/Range.h
roshandathathri: I think the current code is more clear to read. The compiler will optimize it anyway.

libgalois/include/katana/Range.h
```diff
@@ -32,6 +32,66 @@
...
+    end = begin;
+    std::advance(end, e - b);
+  }
+
+  return std::make_pair(begin, end);
+}
+
+template <typename Integer>
+std::enable_if_t<std::is_integral<Integer>::value, std::pair<Integer, Integer>>
+block_range(Integer begin, Integer end, unsigned id, unsigned num) {
```
roshandathathri: nit:: paste the same doxygen comments for this function too?

roshandathathri (approved):
+1

LGTM

libgalois/include/katana/Range.h
amberhassaan: @roshandathathri : The issue @ddn0 is trying to tackle is of imbalanced assignment of work, which will only be mitigated by work-stealing. 
For example: let's say we have 1000 elements and 30 threads.
Previously each thread would get `ceiling(1000/30) == 34, except last thread would get 1000 - 29*34 = 16 elements`
Now, @ddn0 is deciding that some threads get 33 and some get 34 based on their id.

@ddn0 : The code is certainly harder to read, and I'm not able to follow all of it. I had to do some pencil on paper for the example above :). So, I suggest adding comments . For example, I don't understand the micro-optimizations.

One style that I personally like is making the parameters const values, and computing new values to return. It improves readability IMO.

amberhassaan (approved):
Needs explanation for mere mortals (like me).

libgalois/include/katana/Range.h
roshandathathri: Amber, I understand the code. My comment was a response to Rob's comment, not to the code itself. 

Also, we had faced this issue of imbalance before (when dealing blocks to distributed machines). We couldn't steal work then. But we were not using this block distribution. We were using edge-balanced node-blocking and we wrote that such that it avoiding this imbalance.

libgalois/include/katana/Range.h
```diff
@@ -32,6 +32,73 @@
...
+}
+
+/**
+ * Returns a contiguous block from the range based on the number of divisions
+ * and the index of the block requested.
+ */
+template <typename Iterator>
+std::enable_if_t<
+    !std::is_integral<Iterator>::value, std::pair<Iterator, Iterator>>
+block_range(Iterator begin, Iterator end, unsigned idx, unsigned num) {
```
amberhassaan, line 53: minor: making these parameters consts would make me happy but may be I'm the only one who thinks like that.

amberhassaan (approved):
I find the code easier to understand now.

libgalois/include/katana/Range.h
ddn0: I'm personally not a fan of `const` on values, though I understand why we might want it.

Philosophically, it feels weird because it leaks an implementation detail to callers.

It also looks very similar to `const T&`, and it is a common error for someone to write `const T` when they really meant `const T&`. And on balance, I think the benefit for single-assigment semantics over just value semantics isn't that huge. Though, I would agree for references or pointers having the single-assignment property is very useful.

amberhassaan: OK. I'm not insisting on it. I just wish parameters would be const unless specified otherwise, but that's not how C++ is.

PR #598, opened 2021-10-04
print error classes

micah-white:
The `ErrorInfo` class now prints the error message supplied from the error category along with the error message supplied by the code.

### DIFF
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -267,14 +267,16 @@ public:
     fmt::format_to(std::back_inserter(out), " ({}:{})", base_name, line_no);
 
     std::string code = ec.message();
-    std::string msg(out.data());
+    char* data = out.data();
+    data[out.size()] = '\0';
+
+    std::string msg(data);
 
     std::string combined = "\n" + code + "\n" + msg;
     const char* output = combined.c_str();
 
     ErrorInfo ei(ec);
-    // TODO(micah) sometimes random characters get printed at the end.
-    ei.Prepend(output, output + combined.length());
+    ei.Prepend(output, output + combined.length() + 1);
 
     return ei;
   }
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -266,8 +266,15 @@ public:
 
     fmt::format_to(std::back_inserter(out), " ({}:{})", base_name, line_no);
 
+    std::string code = ec.message();
+    std::string msg(out.data());
+
+    std::string combined = "\n" + code + "\n" + msg;
+    const char* output = combined.c_str();
+
     ErrorInfo ei(ec);
-    ei.Prepend(out.data(), out.data() + out.size());
+    // TODO(micah) sometimes random characters get printed at the end.
+    ei.Prepend(output, output + combined.length());
 
     return ei;
   }


libsupport/include/katana/Result.h
```diff
@@ -266,8 +266,15 @@ class KATANA_EXPORT [[nodiscard]] ErrorInfo {
...
 
+    std::string code = ec.message();
+    std::string msg(out.data());
+
+    std::string combined = "\n" + code + "\n" + msg;
+    const char* output = combined.c_str();
+
     ErrorInfo ei(ec);
-    ei.Prepend(out.data(), out.data() + out.size());
+    // TODO(micah) sometimes random characters get printed at the end.
```
ddn0: > sometimes random characters get printed at the end.

This seems problematic. We should probably understand why before inflicting this change to everyone.

At a higher level, if someone needs a specific error output, they should just write the output they want rather than changing the format for everyone. Generally, the message output is arbitrary and subject to change.

micah-white: Okay, I was having a hard time figuring out why, but I'll keep looking into it. It's only up to 4 characters at the very end of the message so it's not a huge deal.

Sure, but I need the TCK and QueryErrorCode to print for every error reported, and making a new macro or expecting everyone to remember to do that seems unnecessary.

micah-white: Actually it seems the random characters are causing some python tests to fail so I guess I have to fix it

arthurp: Some string copy function copy in 32-bit chunks and even over copy by as much as 3 bytes. Is one of the inputs not null terminated, so when you over copy the junk at the end is included in the string?

arthurp: Actually I'm pretty sure null termination is the issue. c_str( ) adds a null terminator, but length does not include that null, so when you are prepending below, you are dropping the null which might be messing things up.

arthurp (commented):
I like this. Once the string termination issue is fixed. I'm with @ddn0, it needs to be fixed. It could cause buffer overruns.

libsupport/include/katana/Result.h
micah-white: ^ I was thinking something like that was the problem. I added a null terminator and it seems to have fixed it. I really don't know how this wasn't a problem before...

micah-white: Actually I do know why. Didn't convert to a C++ string before.

ddn0: > Sure, but I need the TCK and QueryErrorCode to print for every error reported, and making a new macro or expecting everyone to remember to do that seems unnecessary.

What is wrong with making whatever is actually printing the error message to call something like `FormatForTCK(err)`?

micah-white: Well by the time that it gets to that point the string is the full call stack, extracting that is messier than just adding it in the beginning. If you want me to make it print the error code only when it's a `QueryErrorCode`, I could do that. I do want both errors to print though, because the TCK errors can be unhelpful and I want people to be familiar with the new error codes so they can use them properly and understand them better when they get thrown.

ddn0 (changes requested):
I do not agree with this approach.

If someone cares about the format of error messages, they should express it explicitly by constructing the string they want from an ErrorInfo and not rely on ErrorInfo's default printing.

If ErrorInfo is missing the required accessors, we should just add the missing accessors (e.g., stack of error codes, stack of context messages), rather than changing how we append context information.

For anyone who cares about this issue, react to this post and I'll schedule a meeting to discuss.

micah-white:
Decided to handle this at the top of the call stack instead of the bottom

PR #600, opened 2021-10-04
Missing command is not an error for capture_command

denish44:
Missing command is not an error for capture_command

capture_command is to aid debugging possibly broken build environments so it should complete executing even if the build environment is missing expected programs.

jira: KAT-1651

### DIFF
--- a/python/katana/bug/environment.py
+++ b/python/katana/bug/environment.py
@@ -116,8 +116,14 @@ def _capture_build(zipout: zipfile.ZipFile):
 def capture_command(*args, **kwargs) -> str:
     # pylint: disable=subprocess-run-check
     # Not using check=True because I want to capture both success and failure the same way.
-    res = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs)
-    out = res.stdout.decode("utf-8").strip("\n")
+    out = ""
+
+    try:
+        res = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs)
+        out = res.stdout.decode("utf-8").strip("\n")
+    except OSError as err:
+        out = f"errno: {err.errno}, filename: {err.filename}, strerror: {err.strerror}"
+
     return out
 
 
--- a/python/katana/bug/environment.py
+++ b/python/katana/bug/environment.py
@@ -61,21 +61,21 @@ def _capture_system(zipout: zipfile.ZipFile):
     )
 
     capture_string(zipout, "env.txt", "\n".join(f"{k}={v}" for k, v in get_filtered_environ().items()))
-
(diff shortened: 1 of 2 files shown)


python/test/test_bug.py
```diff
@@ -1,6 +1,8 @@
 import zipfile
 from tempfile import NamedTemporaryFile
 
+from pandas.core.algorithms import isin
```
ddn0: what is `isin` used for?

python/test/test_bug.py
```diff
@@ -13,3 +15,11 @@ def test_bug_capture_environment():
             files = {f.filename for f in zipin.filelist}
             for fn in ["info/os.txt", "info/python.txt", "info/conda.txt", "info/cmake.txt", "etc/ld.so.conf"]:
                 assert fn in files
+
+
+def test_bug_capture_command_pass_invalid_command():
+    from katana.bug.environment import capture_command
+
+    # Make sure it doesn't fail for non existent command
+    result = capture_command("notreal", "xyz")
```
ddn0: nit: just change the command to "non-existent-executable"

python/katana/bug/environment.py
```diff
@@ -116,8 +116,14 @@ def _capture_build(zipout: zipfile.ZipFile):
...
     # Not using check=True because I want to capture both success and failure the same way.
-    res = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs)
-    out = res.stdout.decode("utf-8").strip("\n")
+    out = ""
+
+    try:
+        res = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs)
+        out = res.stdout.decode("utf-8").strip("\n")
+    except OSError as err:
+        out = f"errno: {err.errno}, filename: {err.filename}, strerror: {err.strerror}"
```
ddn0: nit: it would probably be better to just use the original error message as much as possible as that is what most people would be familiar with, e.g., ```f"error executing {args}: {err}"```

ddn0 (commented):
I made some changes to the commit message and PR description, please make sure they are reflected in the merged commit.

1. Try to describe why you are doing something rather than what you did. We have access to the diff, so simply summarizing what text edits you did isn't particularly valuable. It is better to describe why the change is made as this is not directly captured in the diff.
2. Try to keep the commit messages self-contained. People reading this repo do not have access to the enterprise repo, so generally reference from enterprise to open (if at all).

python/test/test_bug.py
denish44: oops did not see that pull in. must of been some auto import on something. will take it out.

ddn0 (approved):

arthurp (approved):

PR #603, opened 2021-10-06
Testing/analytics benchmarking

GioLomia:
Setting up a testing environment for the analytics benchmarking script.


python/katana/benchmarking/bench_python_cpp_algos.py
```diff
@@ -17,11 +17,17 @@
...
-
 Routine = namedtuple("Routine", ["func", "args"])
+OutputTuple = namedtuple("OutputTuple", ["write_success", "write_data"])
+
+
+def initialize_global_vars():
+    APPS = ["bfs", "sssp", "cc", "bc", "pagerank", "tc", "jaccard", "kcore", "louvain", "all"]
+    GRAPHS = ["GAP-road", "GAP-kron", "GAP-twitter", "GAP-web", "GAP-urand", "rmat15"]
+
+    return (APPS, GRAPHS)
```
arthurp: Why not just make these globals? I don't see any reason to have them in a function like this. Also this does NOT initialize the globals, it just returns some hard codes configuration.

python/katana/benchmarking/bench_python_cpp_algos.py
```diff
@@ -452,12 +458,20 @@ def load_graph(graph_path, edge_properties=None):
...
+
+def main(parsed_args):
+
+    run_all_gap(parsed_args)
 
 
 if __name__ == "__main__":
+    options = initialize_global_vars()
+    all_apps = options[0]
+    all_graphs = options[1]
```
arthurp: Use multiple assignment. https://www.w3schools.com/python/gloss_python_assign_value_to_multiple_variables.asp

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+        "--graph", default=graph, choices=options[0], help="Graph name (default: %(default)s)",
+    )
+    parser.add_argument(
+        "--application", default=app, choices=options[1], help="Application to run (default: %(default)s)",
+    )
+    parser.add_argument("--source-nodes", default=source_nodes, help="Source nodes file(default: %(default)s)")
+    parser.add_argument("--trials", type=int, default=trails, help="Number of trials (default: %(default)s)")
+    parser.add_argument("--num-sources", type=int, default=num_sources, help="Number of sources (default: %(default)s)")
+
+    parsed_args = parser.parse_args()
```
arthurp: You don't need to use a full ArgumentParser here. You can just create an argparse.Namespace:

```
parsed_args = argparse.Namespace()
parsed_args.threads = whatever
.
.
.
```
That reduces duplication with the script code and simplifies the test.

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if not os.path.isdir(parsed_args.input_dir):
+        print(f"input directory : {parsed_args.input_dir} doesn't exist")
+        sys.exit(1)
+    if not parsed_args.threads:
+        parsed_args.threads = int(os.cpu_count())
+    print(f"Using input directory: {parsed_args.input_dir} and Threads: {parsed_args.threads}")
```
arthurp: The tests should always provide full values I would think, so this stuff shouldn't be needed.

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+
+def assert_routine_output(routine_output, max_time=1_000_000):
+    for subproccess in routine_output:
+        assert (
+            0 <= routine_output[subproccess] <= max_time
+        ), f"Invalid time for subproccess {subproccess}: took {routine_output[subproccess]} ms"
+
+
+def assert_types_match(ground_truth, outp):
+    for sec in ground_truth:
```
arthurp: What does `sec` stand for? If it's section or something please say it fully. To me "sec" is "second".

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+        out_type = type(outp[sec])
+        assert truth_type == out_type, f"Expected types for {sec} do not match - Expected:{truth_type}, Got:{out_type}"
+
+    return True
+
+
+def test_single_trial_gaps():
+    arguments = {
+        "json_output": "../../../bench_statistics.json",
+        "input_dir": "../../inputs/v24/propertygraphs",
```
arthurp: These relative paths are probably not correct since the tests can be called from various different paths. Use `__file__` and make the paths relative to this file. Also use `pathlib` if appropriate.

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+        "graph": "rmat15",
+        "app": "all",
+        "source_nodes": "",
+        "trails": np.random.randint(1, 15),
+        "num_sources": np.random.randint(4, 64),
```
arthurp: Don't randomize in tests unless we are explicitly doing fuzzing. Just set an arbitrary value here. Then we can scan across multiple values using pytest if needed.

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+import time
+from collections import namedtuple
+from datetime import datetime
+
+import numpy as np
+import pytest
+import pytz
+from pyarrow import Schema
+
+import katana.benchmarking.bench_python_cpp_algos as katbench
```
arthurp: Don't use `as`. It hides what you are referring to. You can either say `bench_python_cpp_algos` below or import the specific values you need from that module. (`np` is an exception since it's so well known, though honestly I personally don't like it either.)

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+def generate_args(
+    json_output,
+    input_dir="./",
+    graph="GAP-road",
+    app="bfs",
+    source_nodes="",
+    trails=1,
+    num_sources=4,
+    thread_spin=False,
+    threads=1,
```
arthurp: Since this is used in testing, defaults will hide bugs, so don't provide them.

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,122 @@
...
+    ground_truth = katbench.create_empty_statistics(args)
+    output_tuple = katbench.run_all_gap(args)
+    assert output_tuple.write_success, "Writing JSON statistics to disc failed!"
+    assert_types_match(ground_truth, output_tuple.write_data)
+    for subroutine in output_tuple.write_data["routines"]:
+        assert_routine_output(output_tuple.write_data["routines"][subroutine])
+
+
+if __name__ == "__main__":
+    test_single_trial_gaps()
```
arthurp: I don't see any need to include this. In fact, it's probably a bad idea. The reason is that if you want to run this you should always use pytest (`pytest -k single_trial_gaps`). Calling it directly my have a slightly different run environment and produce confusion. So please remove this.

(You can replace it with a comment of how to run it using pytest if you would like. It might be useful to others.)

python/katana/benchmarking/bench_python_cpp_algos.py
```diff
@@ -17,11 +17,17 @@
 # TODO(giorgi): This script needs to be tested in CI.
 PathExt = namedtuple("PathExt", ["warn_prefix", "path_ext"])
```
arthurp: I'm not sold of having this in `katana.benchmarking` and hence installed. This is a quite special purpose and hard to use script. I think I would rather have this in `python/tests/benchmarking`. Is that reasonable? It would prevent it from being installed, which I think is good.

arthurp (changes requested):
I don't see why the permissions on python/test/test_cpp_algos.py where changed. Is there a reason? If not, please revert.

Let's not have `scripts/__init__.py` It will confuse python path stuff. I want to fix up script python paths at some point soon. I'll make an issue for it. (edit: KAT-1943)

python/katana/benchmarking/bench_python_cpp_algos.py
GioLomia: When I initialized them as global our lint checker failed because of that.

python/katana/benchmarking/bench_python_cpp_algos.py
GioLomia: @gliga Let me know if this change would still work for you?

GioLomia:
> I don't see why the permissions on python/test/test_cpp_algos.py where changed. Is there a reason? If not, please revert.


Not sure how this happened, what's the safest way to revert this @arthurp ?

python/katana/benchmarking/bench_python_cpp_algos.py
gliga: Yes, I like the suggested change.

python/katana/benchmarking/bench_python_cpp_algos.py
```diff
@@ -469,9 +468,8 @@ def main(parsed_args):
 
 
 if __name__ == "__main__":
-    options = initialize_global_vars()
-    all_apps = options[0]
-    all_graphs = options[1]
+    all_apps, all_graphs = initialize_global_vars()
```
gliga: Maybe move the entire body of this block into a function (for example called `main`), so we can call it nicely from tests; we do  not want to start the script.

python/test/test_algo_bench.py
```diff
@@ -1,102 +1,96 @@
...
-    return parsed_args
+    return parser
 
 
 def assert_routine_output(routine_output, max_time=1_000_000):
     for subproccess in routine_output:
         assert (
             0 <= routine_output[subproccess] <= max_time
-        ), f"Invalid time for subproccess {subproccess}: took {routine_output[subproccess]} ms"
+        ), f"Invalid completion time for subproccess {subproccess}: took {routine_output[subproccess]} ms"
```
gliga, line 35: Potentially blocking our CI for 16min seems too much.

python/test/test_algo_bench.py
```diff
@@ -1,102 +1,96 @@
...
+    return arguments
+
+
+def test_single_trail_gaps():
+    arguments = get_default_args()
+    run_on_all_graphs(arguments)
+
 
-    options = katbench.initialize_global_vars()
+def test_multi_trail_gaps():
```
gliga, line 71: My worry here is that we seem to focus on test completion time rather than on correctness (or I might be missing something about the tests).

I suggest we change the focus on correctness. 

How long does it take to run these tests?

python/test/test_algo_bench.py
```diff
@@ -1,102 +1,96 @@
...
+    for i in range(6):
+        arguments["num_sources"] = pow(2, i)
+        run_on_all_graphs(arguments)
+
+
+def test_thread_gaps():
+    arguments = get_default_args()
+    for i in range(6):
+        arguments["threads"] = pow(2, i)
+        run_on_all_graphs(arguments)
```
gliga, line 88: Same comments for each test: do we have assertions?`

python/test/test_algo_bench.py
```diff
@@ -110,13 +104,9 @@ def test_single_trial_gaps():
 
 def run_single_t(arguments):
```
gliga: is _t our convention?

python/test/test_algo_bench.py
```diff
@@ -110,13 +104,9 @@ def test_single_trial_gaps():
...
     args = generate_args(**arguments)
-    ground_truth = katbench.create_empty_statistics(args)
-    output_tuple = katbench.run_all_gap(args)
+    ground_truth = katana.benchmarking.bench_python_cpp_algos.create_empty_statistics(args)
+    output_tuple = katana.benchmarking.bench_python_cpp_algos.run_all_gap(args)
     assert output_tuple.write_success, "Writing JSON statistics to disc failed!"
```
gliga, line 180: If possible we should avoid writing to disk in our tests.

python/test/test_algo_bench.py
GioLomia: I'm open to suggestions on how to test for correctness. At the moment I check if JSON was written successfully as well as all the types in the dictionaries match the expected output types. I'm not sure how else can I test for correctness.

gliga: what are all the results that run of one of the algos is supposed to return/write?

python/test/test_algo_bench.py
GioLomia: Yes. Its abstracted 
```
for element in ground_truth:
assert element in outp, f"Output missing an element: {element}"
truth_type = type(ground_truth[element])
out_type = type(outp[element])
assert (
    truth_type == out_type
), f"Expected types for {element} do not match - Expected:{truth_type}, Got:{out_type}"

return True
```

python/test/test_algo_bench.py
arthurp: No. I think you mean "run single test" so the name should be `run_single_test`.

python/katana/benchmarking/bench_python_cpp_algos.py
arthurp: Why? Mutable globals? If so use a tuple instead of a list. If it's something else, I'm happy to help if you send me a full error.

arthurp (commented):
> Not sure how this happened, what's the safest way to revert this @arthurp ?

You can just use `git restore` or `git reset` on the file to put it back to the old permissions. Or you can just set them yourself with `chmod`.

python/test/test_algo_bench.py
GioLomia: Wouldn't Pytest catch that as a test just because it has a word test in it?

arthurp: Pytest only picks up functions that begin with `test` not those that end with it. https://docs.pytest.org/en/6.2.x/reference.html#confval-python_functions

python/test/test_algo_bench.py
```diff
@@ -0,0 +1,187 @@
...
+    ground_truth = test.benchmarking.bench_python_cpp_algos.create_empty_statistics(args)
+    output_tuple = test.benchmarking.bench_python_cpp_algos.run_all_gap(args)
+    assert output_tuple.write_success, "Writing JSON statistics to disc failed!"
+    assert_types_match(ground_truth, output_tuple.time_write_data)
+    for subroutine in output_tuple.time_write_data["routines"]:
+        assert_routine_output(output_tuple.time_write_data["routines"][subroutine])
+    return output_tuple.analytics_write_data
+
+
+test_bfs()
```
arthurp: This shouldn't be here. It will cause a test to run while pytest tries to collect the test information which is bad.

arthurp (approved):
The one comment is quite important as it could cause weird test failures. 

Other than that this looks good. No need to check back with me after you fix the erroneous test call.

PR #608, opened 2021-10-07
Fix the unmet dependencies error for conan builds

yan-zaretskiy:


arthurp (commented):
Huh. Why is this needed? Can you document what's going on in more detail? This should not be needed given that we have llvm-12-dev above. And I was not aware of us using libclang at all.

yan-zaretskiy:
This was prompted by an error from `apt` that it detected these unmet dependencies and would halt the installation. Strangely enough, I just tried rerunning the `setup_ubuntu.sh` script on a fresh installation of Ubuntu and it went just fine. So it must have been an issue with my system.

PR #610, opened 2021-10-09
[KAT-1982] Build GPU packages

srogatch:
It's a prerequisite for https://github.com/KatanaGraph/katana-enterprise/pull/2431 : allow the GPU feature in a conda build.

### DIFF
--- a/conda_recipe/build.sh
+++ b/conda_recipe/build.sh
@@ -23,6 +23,13 @@ if [ -z "${KATANA_COMPONENTS}" ]; then
 fi
 echo "Building with components: ${KATANA_COMPONENTS}"
 
+if [ -z "${KATANA_GPU_OPTIONS}" ]; then
+  KATANA_GPU_OPTIONS=""
+  echo "Building without GPU support"
+else
+  echo "Building with GPU support"
+fi
+
 # Useful debugging addition to the below: -DCMAKE_VERBOSE_MAKEFILE=ON
 cmake \
   $CMAKE_CCACHE_OPTION \
@@ -32,6 +39,7 @@ cmake \
   -DKATANA_LANG_BINDINGS=python \
   -DCMAKE_BUILD_TYPE=Release \
   -DKATANA_COMPONENTS="${KATANA_COMPONENTS}" \
+  ${KATANA_GPU_OPTIONS} \
   -Wno-dev \
   -S "$SRC_DIR"
 cmake --build . --parallel ${CMAKE_BUILD_PARALLEL_LEVEL}


arthurp (commented):
I don't think this will work the way you think. You need to add `KATANA_GPU_OPTIONS` to the variables allowed through my conda build.

BTW if you are changing this here solely for use from the enterprise packaging, it might be time to split the open and enterprise build scripts and call the open build script from the enterprise one. That would avoid the need for adding GPU code here since I'm guessing GPU support will not be in open packages any time soon.

I also think we should start including these build option in the build string of the package so that we can upload multiple variants of the same package version so the installation can select the appropriate package for a given environment. That's something I can do. I'll create an issue for it.

srogatch:
> I don't think this will work the way you think. You need to add `KATANA_GPU_OPTIONS` to the variables allowed through my conda build.

Could you please provide more details on what will not work? How to add it to the variables allowed? I searched for the other variable from the build script, `KATANA_COMPONENTS`, and it seems there is no special handling for it in the codebase. You just set it in `scripts/dev` and pass to conda.
 
> BTW if you are changing this here solely for use from the enterprise packaging, it might be time to split the open and enterprise build scripts and call the open build script from the enterprise one. That would avoid the need for adding GPU code here since I'm guessing GPU support will not be in open packages any time soon.

Yes, it's for Enterprise packaging. The build script in the Enterprise repo is a symlink to the Open repo build script.

> I also think we should start including these build option in the build string of the package so that we can upload multiple variants of the same package version so the installation can select the appropriate package for a given environment. That's something I can do. I'll create an issue for it.

I would much appreciate it. But can we do it incrementally and start with using the same package names for GPU and non-GPU builds?

arthurp:
> I would much appreciate it. But can we do it incrementally and start with using the same package names for GPU and non-GPU builds?

You can continue as is for now. This can be done incrementally.

I should explain. Build strings are not part of the package name or the version. They are a label of sorts that allows there to be multiple variants of the same package and version. They can be used for things like cuda and non-cuda versions of packages (like we might use it) or MPI and non-MPI versions and the like.

arthurp:
> Could you please provide more details on what will not work? How to add it to the variables allowed? I searched for the other variable from the build script, `KATANA_COMPONENTS`, and it seems there is no special handling for it in the codebase. You just set it in `scripts/dev` and pass to conda.

https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#inherited-environment-variables

Also why not include GPU in `KATANA_COMPONENTS`? It seems like calling it a component would be very reasonable.

ddn0:
> https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#inherited-environment-variables

An example of the special handling of `KATANA_COMPONENTS`: https://github.com/KatanaGraph/katana-enterprise/blob/master/conda_recipe/meta.yaml#L12

srogatch:
Thanks, @arthurp and @ddn0 . I'm then changing the GPU feature to be part of `KATANA_COMPONENTS`.

PR #611, opened 2021-10-11
Add an override flag to version release_branch, to allow arbitrary branching.

arthurp:
…anches.


arthurp:
@origandrew This is not really tested. I think it will work correctly based on dry runs. But I'm gonna leave more testing to you, since I have too many other things to do and I hope we will not need this feature too often.

bosconi (approved):
Looks good to me as far as this makes sense abstractly as a patch to a script I have never read before. The only way is forward, as you pointed out in the comments already, so let's merge this down and deal with any issues are they arise?

arthurp:
I have tested some. I'm going to stop though. Testing this is such a pain because the reset process for a different test is tricky and not automated.

I think everything is working. I ran into some issues, but they were due to how I was testing not the real situation.

This script will NOT work correctly in a dirty repository, so you cannot just copy the new script to an old commit and use it.

PR #612, opened 2021-10-11
Make entity type id assignment warning more explicit

e-mcginnis:
Goal: make the user aware that there is extra overhead on this run from extracting types from properties, and that storing this RDG would avoid this overhead in the future.


tylershunt (approved):

roshandathathri (approved):
LGTM

GioLomia (approved):
LGTM. Nitpick: I feel like it would be better to word it the other way around: "In order to avoid overhead from assigning entity types, please store the RDG."

PR #613, opened 2021-10-11
Persist node and edge indexes in partition header

oshofmann:
This takes Nirhjar's work to add index persistence to RDGPartHeader and:
- Rebase and resolve conflicts on current head
- Tweaks for some shorter names, better types (don't return or pass non-const references), and data flow
- Adds what I believe are the necessary bits to to_json/from_json to actually persist the index list

### DIFF
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -74,6 +74,9 @@ private:
   Result<void> WriteView(
       const std::string& uri, const std::string& command_line);
 
+  // Recreate indexes from json
+  katana::Result<void> RecreatePropertyIndexes();
+
   tsuba::RDG rdg_;
   std::unique_ptr<tsuba::RDGFile> file_;
   GraphTopology topology_;
@@ -91,42 +94,14 @@ private:
   // List of node indexes on this graph.
   std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Node>>>
       node_indexes_;
-  //And the columns that created them to persist in json
-  std::vector<std::string> node_property_indexes_column_name_;
 
   // List of edge indexes on this graph.
   std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Edge>>>
       edge_indexes_;
-  //And the columns that created them to persist in json
-  std::vector<std::string> edge_property_indexes_column_name_;
 
   PGViewCache pg_view_cache_;
 
   friend class PropertyGraphRetractor;
-
-  // recreate indexes from json
-  katana::Result<void> recreate_node_property_indexes() {
-    node_property_indexes_column_name_ =
-        rdg_.node_property_indexes_column_name();
-    for (const std::string& column_name : node_property_indexes_column_name_) {
-      auto result = MakeNodeIndex(column_name);
-      if (!result) {
-        return result.error();
-      }
-    }
-    return katana::ResultSuccess();
-  }
-  katana::Result<void> recreate_edge_property_indexes() {
-    edge_property_indexes_column_name_ =
-        rdg_.edge_property_indexes_column_name();
-    for (const std::string& column_name : edge_property_indexes_column_name_) {
-      auto result = MakeEdgeIndex(column_name);
-      if (!result) {
-        return result.error();
-      }
-    }
-    return katana::ResultSuccess();
-  }
 
 public:
   /// PropertyView provides a uniform interface when you don't need to
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -74,7 +74,7 @@ private:
   Result<void> WriteView(
       const std::string& uri, const std::string& command_line);
 
-  // Recreate indexes from json
+  /// Recreate indexes listed in RDG metadata.
   katana::Result<void> RecreatePropertyIndexes();
 
   tsuba::RDG rdg_;
@@ -91,11 +91,11 @@ private:
   /// The edge EntityTypeID for each edge's most specific type
   EntityTypeIDArray edge_entity_type_ids_;
 
-  // List of node indexes on this graph.
+  /// List of node indexes on this graph.
   std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Node>>>
       node_indexes_;
 
-  // List of edge indexes on this graph.
+  /// List of edge indexes on this graph.
   std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Edge>>>
       edge_indexes_;
 
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -88,15 +88,45 @@ private:
   /// The edge EntityTypeID for each edge's most specific type
   EntityTypeIDArray edge_entity_type_ids_;
 
-  // List of node and edge indexes on this graph.
+  // List of node indexes on this graph.
   std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Node>>>
       node_indexes_;
+  //And the columns that created them to persist in json
+  std::vector<std::string> node_property_indexes_column_name_;
+
+  // List of edge indexes on this graph.
   std::vector<std::unique_ptr<PropertyIndex<GraphTopology::Edge>>>
       edge_indexes_;
+  //And the columns that created them to persist in json
+  std::vector<std::string> edge_property_indexes_column_name_;
 
   PGViewCache pg_view_cache_;
 
   friend class PropertyGraphRetractor;
+
+  // recreate indexes from json
+  katana::Result<void> recreate_node_property_indexes() {
+    node_property_indexes_column_name_ =
+        rdg_.node_property_indexes_column_name();
+    for (const std::string& column_name : node_property_indexes_column_name_) {
+      auto result = MakeNodeIndex(column_name);
+      if (!result) {
+        return result.error();
+      }
+    }
+    return katana::ResultSuccess();
+  }
+  katana::Result<void> recreate_edge_property_indexes() {
+    edge_property_indexes_column_name_ =
+        rdg_.edge_property_indexes_column_name();
+    for (const std::string& column_name : edge_property_indexes_column_name_) {
+      auto result = MakeEdgeIndex(column_name);
+      if (!result) {
+        return result.error();
+      }
+    }
+    return katana::ResultSuccess();
+  }
 
 public:
   /// PropertyView provides a uniform interface when you don't need to
--- a/libgalois/include/katana/PropertyGraph.h
+++ b/libgalois/include/katana/PropertyGraph.h
@@ -74,7 +74,7 @@ private:
   Result<void> WriteView(
       const std::string& uri, const std::string& command_line);
 
-  /// Recreate indexes listed in RDG metadata.
+  /// Recreate indexes from column names in RDG metadata.
   katana::Result<void> RecreatePropertyIndexes();
 
   tsuba::RDG rdg_;
--- a/libgalois/src/PropertyGraph.cpp
+++ b/libgalois/src/PropertyGraph.cpp
@@ -1310,17 +1310,11 @@ katana::PropertyGraph::GetNodePropertyIndex(
 katana::Result<void>
 katana::PropertyGraph::RecreatePropertyIndexes() {
   for (const std::string& column_name : rdg_.node_property_index_columns()) {
-    auto result = MakeNodeIndex(column_name);
-    if (!result) {
-      return result.error();
-    }
+    KATANA_CHECKED(MakeNodeIndex(column_name));
   }
 
   for (const std::string& column_name : rdg_.edge_property_index_columns()) {
-    auto result = MakeEdgeIndex(column_name);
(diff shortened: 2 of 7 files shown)


tylershunt (approved):
LGTM

libgalois/src/PropertyGraph.cpp
```diff
@@ -236,24 +238,26 @@ katana::PropertyGraph::Make(
...
         MakeDefaultEntityTypeIDArray(topo.num_edges()), EntityTypeManager{},
         EntityTypeManager{});
 
-    KATANA_CHECKED(pg->ConstructEntityTypeIDs());
-
-    return MakeResult(std::move(pg));
+    KATANA_CHECKED(property_graph->ConstructEntityTypeIDs());
   }
+
+  KATANA_CHECKED(property_graph->RecreatePropertyIndexes());
```
arekay, line 258: nit: Do we want to put this behind a config option - might make it easier to debug/test.

libgalois/src/PropertyGraph.cpp
```diff
@@ -1289,3 +1306,16 @@ katana::PropertyGraph::GetNodePropertyIndex(
...
+  for (const std::string& column_name : rdg_.node_property_index_columns()) {
+    KATANA_CHECKED(MakeNodeIndex(column_name));
+  }
+
+  for (const std::string& column_name : rdg_.edge_property_index_columns()) {
+    KATANA_CHECKED(MakeEdgeIndex(column_name));
+  }
+
+  return katana::ResultSuccess();
+}
```
arekay: Do we want to check the property type to ensure that we create indexes for types supported.

arekay (approved):
LGTM.

libgalois/src/PropertyGraph.cpp
oshofmann: That sounds like a great idea - what do we use for that kind of config?

libgalois/src/PropertyGraph.cpp
oshofmann: I think that MakeNodeIndex should do the right thing and return an error for an unsupported type (if not it's a bug). The larger question is what to do when that happens, or some other error in index-building. Should we try to create all of the indexes possible and proceed, or treat any failure as a failure in Make? At this point I think I'd prefer the former since a problem will manifest as a clear failure to load the graph, rather than just future slowness. Thoughts?

libgalois/src/PropertyGraph.cpp
arekay: Thats a great question - currently we have at least a couple of `option` classes. But that approach is not scalable as it requires the object to be passed around. I guess we can leave it for now and then revisit once we have better config management in place. @witchel might have better insight into this.

libgalois/src/PropertyGraph.cpp
arekay: I encountered some issues when i was testing the code. For instance, a column of type `int32_t` would cause the indexing code to crash - which we don't want. But there might be cases where we would like it to crash - for instance specifying an index for a column that doesn't exist. I don't have any concrete suggestions for here - probably some points to think about.

oshofmann: Crash or return an error? If it's crashing then that's bad, and I'd like to fix that. But returning an error here seems like the right thing to do when asking to index a column for an as-yet-unsupported data type, or a column that does not exist.

libgalois/src/PropertyGraph.cpp
witchel: The two main option structures are `BlockedReadOptions` and `RDGLoadOptions`.  The latter is passed to `PropertyGraph` for one of the Make routines.  While I dream of cleaning up these classes, I have no current plan, so if you want to add a field to `RDGLoadOptions` that would probably be ok.

I did intend to ask @ddn0 if there is a reason to avoid a per-thread ambient options structure, like we have for the ProgressTracer object.  I think Galois used to have this for stats, but something about it was bad.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -74,6 +74,9 @@ class KATANA_EXPORT PropertyGraph {
   Result<void> WriteView(
       const std::string& uri, const std::string& command_line);
 
+  // Recreate indexes from json
```
witchel: Indexes or indices?

More substantively, our serialization format for indexes is json??

libgalois/include/katana/PropertyGraph.h
```diff
@@ -88,9 +91,11 @@ class KATANA_EXPORT PropertyGraph {
   /// The edge EntityTypeID for each edge's most specific type
   EntityTypeIDArray edge_entity_type_ids_;
 
-  // List of node and edge indexes on this graph.
+  // List of node indexes on this graph.
```
witchel: Three slashes for a doxygen/sphynx/autogenerated documentation.  I think that is what you intend here.

libgalois/src/PropertyGraph.cpp
```diff
@@ -1289,3 +1306,16 @@ katana::PropertyGraph::GetNodePropertyIndex(
   }
   return KATANA_ERROR(katana::ErrorCode::NotFound, "node index not found");
 }
+
+katana::Result<void>
+katana::PropertyGraph::RecreatePropertyIndexes() {
+  for (const std::string& column_name : rdg_.node_property_index_columns()) {
+    KATANA_CHECKED(MakeNodeIndex(column_name));
+  }
```
witchel: Are we creating property indices for all properties, or only properties loaded into memory?

libgalois/src/PropertyGraph.cpp
witchel: Yeah, we can leave it as return an error for now, given that we should be checking for validity when the user creates an index (or we do).

libgalois/test/property-index.cpp
```diff
@@ -21,12 +26,7 @@ using Edge = NodeOrEdge<katana::GraphTopology::Edge>;
 
 template <>
 katana::Result<katana::PropertyIndex<katana::GraphTopology::Node>*>
```
witchel, line 28: You need these empty template declarations to make the `NodeOrEdge` trick work?  They can't just be overloads?

libgalois/test/property-index.cpp
```diff
@@ -263,6 +311,31 @@ TestStringIndex(size_t num_nodes, size_t line_width) {
...
+  auto write_result = g->Write(rdg_dir, "test command line");
+
+  if (!write_result) {
+    boost::filesystem::remove_all(rdg_dir);
+    KATANA_LOG_FATAL("writing result: {}", write_result.error());
+  }
+
+  katana::Result<std::unique_ptr<katana::PropertyGraph>> make_result =
+      katana::PropertyGraph::Make(rdg_dir, tsuba::RDGLoadOptions());
+  boost::filesystem::remove_all(rdg_dir);
```
witchel, line 333: I'm not crazy about our dependence on `boost::filesystem`, but I've used `remove_all`.  So nice.

libtsuba/include/tsuba/RDG.h
```diff
@@ -230,6 +230,17 @@ class KATANA_EXPORT RDG {
   /// Remove all edge properties
   void DropEdgeProperties();
 
+  // Set the list of node and edge column names to persist. Consumes the
+  // provided parameters.
+  void set_node_property_index_columns(
```
witchel, line 235: Why a `move` interface?

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -33,6 +33,9 @@ const char* kEdgeEntityTypeIDDictionaryKey =
 const char* kNodeEntityTypeIDNameKey = "kg.v1.node_entity_type_id_name";
 // Name maps from Atomic Edge Entity Type ID to set of string names for the Edge Entity Type ID
 const char* kEdgeEntityTypeIDNameKey = "kg.v1.edge_entity_type_id_name";
+// List of node and edge indexed columns
+const char* kNodePropertyIndexColumnsKey = "kg.v1.node_property_index_columns";
+const char* kEdgePropertyIndexColumnsKey = "kg.v1.edge_property_index_columns";
```
witchel, line 38: Hey @tylershunt, given that these key names seem to have no bearing on the RDG version, is it wise to keep calling them v1?

witchel (commented):
Looking good.

libgalois/include/katana/PropertyGraph.h
oshofmann: I've been using indexes, both are grammatically correct. The top search result suggests 'indices' are used in the mathematical sense, while 'indexes' for e.g. written documents, and I think this use is from the idea of a written index of a document. As an added bonus, text search for 'index' will find 'indexes'.

Currently we're not persisting anything about the contents of the index, just the metadata - list of columns indexed. If we have an index which has some representation in storage that helps build it faster than just rebuilding from the property, then yes json would be silly.

libgalois/src/PropertyGraph.cpp
oshofmann: Currently a request to make an index only looks at the PropertyGraph, so will fail unless the property is loaded into memory. I had intended to defer any questions of in/out-of-core properties, however it would be simple here to only try to make the index if the column is loaded (otherwise this will fail the entire graph load) - I'll make that change.

There will still exist the bugaboo which sounds like something you were describing a while ago - we load indexes only for in-core properties, then only persist those, losing indexes for out-of-core properties. That I would like to defer for this change - having a basic index flow in place, as well as me understanding more, will help do that correctly in the future.

libgalois/test/property-index.cpp
oshofmann: I think these are required for template specializations like these, unless I'm mistaken. I think it would be hard to do this with normal overloading, especially since it's.just the return type. I'm totally open to something else, this was just the limit of my imagination.

libgalois/test/property-index.cpp
oshofmann: Fair, I'll also say that in tests I tend to be much more permissive for my own coding style and grabbing arbitrary dependencies to just get the job done.

libtsuba/include/tsuba/RDG.h
oshofmann: In part this is from the code Nirjhar wrote - it did the same move but had these just as vector&. Moving out of a non-const reference seems like extremely bad form so I changed them to rvalue references. It technically saves some work, because these are called with a temporary vector of strings created from the list of PropertyIndex-es. But it's an unnecessary optimization for something that only happens on writing a graph, I'd be just as happy to have these be a const reference and make a copy.

libgalois/include/katana/PropertyGraph.h
oshofmann: Ahh, that's why there are so many three-slash comments. I thought maybe someone had a heavy pinky finger. Will go through and update as appropriate.

libgalois/src/PropertyGraph.cpp
ddn0: I actually prefer explicitly passing things around over global variables (or thread-local variables).

(old) Galois stats are bad because it predates and does not interop with tracing (and future metrics).

I think tracing, logging and metrics is a little different than configuration for loading files. In the former, the policy and behavior is supposed to orthogonal to the code being instrumented.

libtsuba/src/RDGPartHeader.cpp
tylershunt: These are versions of the key namespace. I don't see a reason to change them at this time, we're just adding keys.

libgalois/include/katana/PropertyGraph.h
witchel: I'd maybe say index names from json in the comment.

libgalois/include/katana/PropertyGraph.h
witchel: There are other symbols extracted by markup.  You can group methods, define parameters, refer to parameters, etc.

libgalois/src/PropertyGraph.cpp
witchel: Yes, the in-core/out-of-core logic does get tricky.

I like your plan--only make the index if the column is loaded.  

For the grand and glorious future, you can look at `RDGPartHeader.h` for the state machine for properties.  We also have `full_node|edge_schema` and` loaded_node|edge_schema`, so you do have access to all properties all the time, even if they aren't loaded.

libtsuba/include/tsuba/RDG.h
witchel: Yeah, unless it is an important space optimization, I think move-only interfaces are error-prone.

libtsuba/src/RDGPartHeader.cpp
witchel: Everything is an IDspace to you these days.  It makes me feel a bit better about it though.

libgalois/include/katana/PropertyGraph.h
oshofmann: Updated these - I'll send out a small PR for any that should have been part of an earlier change.

libtsuba/include/tsuba/RDG.h
oshofmann: Done.

PR #615, opened 2021-10-12
tsuba: Include property split files in RDGManifest::FilesNames()

pyrito:
This PR fixes the bug where split property files were not being copied over during a `CopyRDGOperation`. This was occurring because the RDGManifest only had information for the property file, but excluded the split property files if they existed. This PR tackles this issue by checking to see whether a given property file contains offset mappings or not and then save those file names to be returned later on. 

JIRA: https://katanagraph.atlassian.net/browse/KAT-1892

### DIFF
--- a/libtsuba/include/tsuba/FileView.h
+++ b/libtsuba/include/tsuba/FileView.h
@@ -110,7 +110,7 @@ public:
   }
 
   uint64_t size() const { return file_size_; }
-  std::string filename() const { return filename_; }
+  const std::string& filename() const { return filename_; }
 
   // support iterating through characters
   const char* begin() const { return ptr<char>(); }
--- a/libtsuba/include/tsuba/FileView.h
+++ b/libtsuba/include/tsuba/FileView.h
@@ -110,6 +110,7 @@ public:
   }
 
   uint64_t size() const { return file_size_; }
+  std::string filename() const { return filename_; }
 
   // support iterating through characters
   const char* begin() const { return ptr<char>(); }
--- a/libtsuba/include/tsuba/ParquetReader.h
+++ b/libtsuba/include/tsuba/ParquetReader.h
@@ -75,7 +75,10 @@ public:
   /// Get the number of rows for the table stored in a parquet file
   ///   \param uri an identifier for a parquet file
   katana::Result<int64_t> NumRows(const katana::Uri& uri);
-
+  
+  /// Get the number of offset files for the parquet file
+  ///   \param uri an identifier for a parquet file
+  katana::Result<uint64_t> NumOffsetFiles(const katana::Uri& uri);
 private:
   ParquetReader(std::optional<Slice> slice, bool make_cannonical)
       : slice_(slice), make_cannonical_{make_cannonical} {}
--- a/libtsuba/include/tsuba/ParquetReader.h
+++ b/libtsuba/include/tsuba/ParquetReader.h
@@ -76,10 +76,9 @@ public:
   ///   \param uri an identifier for a parquet file
   katana::Result<int64_t> NumRows(const katana::Uri& uri);
 
-  /// Get the possible sub files for the logical parquet file, will return
-  /// a original parquet file if there are no sub files
+  /// Get the files for the logical parquet table
   ///   \param uri an identifier for a parquet file
-  katana::Result<std::vector<std::string>> GetSubFiles(const katana::Uri& uri);
+  katana::Result<std::vector<std::string>> GetFiles(const katana::Uri& uri);
 
 private:
   ParquetReader(std::optional<Slice> slice, bool make_cannonical)
--- a/libtsuba/include/tsuba/ParquetReader.h
+++ b/libtsuba/include/tsuba/ParquetReader.h
@@ -75,10 +75,11 @@ public:
   /// Get the number of rows for the table stored in a parquet file
   ///   \param uri an identifier for a parquet file
   katana::Result<int64_t> NumRows(const katana::Uri& uri);
-  
-  /// Get the number of offset files for the parquet file
+
+  /// Get the number of files for the logical parquet file
   ///   \param uri an identifier for a parquet file
-  katana::Result<uint64_t> NumOffsetFiles(const katana::Uri& uri);
+  katana::Result<uint64_t> NumFiles(const katana::Uri& uri);
+
 private:
   ParquetReader(std::optional<Slice> slice, bool make_cannonical)
       : slice_(slice), make_cannonical_{make_cannonical} {}
--- a/libtsuba/include/tsuba/ParquetReader.h
+++ b/libtsuba/include/tsuba/ParquetReader.h
@@ -76,9 +76,10 @@ public:
   ///   \param uri an identifier for a parquet file
   katana::Result<int64_t> NumRows(const katana::Uri& uri);
 
-  /// Get the number of files for the logical parquet file
+  /// Get the possible sub files for the logical parquet file, will return
+  /// a original parquet file if there are no sub files
   ///   \param uri an identifier for a parquet file
-  katana::Result<uint64_t> NumFiles(const katana::Uri& uri);
+  katana::Result<std::vector<std::string>> GetSubFiles(const katana::Uri& uri);
 
 private:
   ParquetReader(std::optional<Slice> slice, bool make_cannonical)
--- a/libtsuba/src/ParquetReader.cpp
+++ b/libtsuba/src/ParquetReader.cpp
@@ -317,6 +317,10 @@ public:
     return concatenated_table;
   }
 
+  Result<uint64_t> NumOffsetFiles() {
+    return fvs_.size();
+  }
+
 private:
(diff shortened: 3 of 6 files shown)


libtsuba/src/RDGPartHeader.h
```diff
@@ -472,6 +498,29 @@ class KATANA_EXPORT RDGPartHeader {
...
+    for (const auto& prop : storage_info) {
+      auto full_path = katana::Uri::JoinPath(prop_dir_name, prop.path());
+      auto uri_path = KATANA_CHECKED(katana::Uri::Make(full_path));
+      auto num_files = KATANA_CHECKED(reader->NumFiles(uri_path));
+      // num_files can be 1, which means there are no split property files
+      if (num_files == 1) {
+        continue;
+      }
+      for (uint64_t i = 0; i < num_files; i++) {
+        auto file_name = fmt::format("{}.part_{:09}", prop.path(), i);
```
witchel: I do wish we can centralize all RDG naming logic to one area, rather than having individual` fmt::format` calls.

witchel (commented):
PropStorageInfo has a path and the thing I'm a little unclear on is how that interacts with these offset files.  It looks like you are not changing the PropStorageInfo, so these offset files are more of an implementation detail.

I just wonder if these offset files are as "hidden" as they could be.  It looks like we are adding one more piece of ad hoc, not logically structured state to an already messy struct.

pyrito:
@witchel , that's right, I'm not explicitly modifying `PropStorageInfo` because I didn't want to mess anything up there, but it's possible for there to be some of these split files. I'm leveraging the `ParquetReader` to give me this information since it already handles some of this logic. 

I'm not a fan of adding more state, so it's possible that I could pull this out of `RDGPartHeader` and just determine these file names somewhere else. However, I do think that the cleanest abstraction would be within `RDGPartHeader` since it is information that should be kept track of. Albeit, the way I've approached it is probably not the best, so suggestions would be appreciated.

witchel:
> @witchel , that's right, I'm not explicitly modifying `PropStorageInfo` because I didn't want to mess anything up there, but it's possible for there to be some of these split files. I'm leveraging the `ParquetReader` to give me this information since it already handles some of this logic.
> 
> I'm not a fan of adding more state, so it's possible that I could pull this out of `RDGPartHeader` and just determine these file names somewhere else. However, I do think that the cleanest abstraction would be within `RDGPartHeader` since it is information that should be kept track of. Albeit, the way I've approached it is probably not the best, so suggestions would be appreciated.

My comment isn't exactly fair because I don't know what to do.  Looking more closely, it does seem like you've kept most of this state out of the public interface, only exposing the information via `FileNames`, which I think is the right thing to do.

I like that you are not modifying `PropStorageInfo`, but I guess in my mind the cleaner thing is to have a `StorageGroup `abstraction, where each property is written to a storage group that consists of some number of files.  So let's stick with what you've got.

tylershunt:
> I like that you are not modifying `PropStorageInfo`, but I guess in my mind the cleaner thing is to have a `StorageGroup `abstraction, where each property is written to a storage group that consists of some number of files. So let's stick with what you've got.

These split files are a hack. If your arrow array  has too many nulls you'll create a parquet file that you can't read. My hope is that it will be addressed soon and this will transparently become unnecessary.

It's worth thinking about promoting this abstraction, but for now I'd like to keep it  hidden behind `Parquet{Reader,Writer}` except for in bummer leaky cases like this.

witchel:
> It's worth thinking about promoting this abstraction, but for now I'd like to keep it hidden behind `Parquet{Reader,Writer}` except for in bummer leaky cases like this.

Tyler, you promised you wouldn't talk about my leaky bummer cases at work.

tylershunt (commented):
I thought about this for a while and its a bummer to make all of this other code to reason about blocked parquet. Why not add a `Copy` method to the `ParquetReader` class instead? It can check if the file is blocked and copy over sub files keeping that implementation detail hidden.

pyrito:
@tylershunt I feel like that approach doesn't make sense because `FileNames` would still be technically incomplete. I understand that the blocked files are technically an implementation detail but `FileNames` should return all the files that are linked to it based on the manifest.

pyrito:
@tylershunt what I could envision is removing a lot of the new functions I added in `RDGPartHeader` and only make these checks when FileNames is called. I added the original data structures because I didn't want to duplicate work when `FileNames` was called. I don't think it's overly performance critical, so we could encapsulate the logic of determining these sub files in `ParquetReader` and just have `FileNames` call something like `ParquetReader::GetSubFiles`.

tylershunt:
I don't think `FileNames` is useful for anything other than copying an RDG, which is something I think libtsuba should be doing anyway. What about just allowing `ParquetReader` to return a list of files? and `FileNames` uses it to get the full list of files for any parquet file?

I also do not like putting more state in RDGPartHeader which is not persisted. I think it's confusing.

libtsuba/src/ParquetReader.cpp
```diff
@@ -317,6 +317,8 @@ class BlockedParquetReader {
     return concatenated_table;
   }
 
+  Result<uint64_t> NumFiles() { return fvs_.size(); }
```
tylershunt: Why not a call that just returns the list of the files involved? That seems less brittle than having to reconstruct them in another part of the code.

pyrito: Yeah, that would be a good idea. I think the only reason why I avoided it in the first place was because I didn't want to modify FileView to return its filename since it doesn't do that right now. It might make sense to add that now however.

tylershunt: In general when I write things I just don't add interfaces that aren't needed. So if you end up needing an interface to something later, it's almost always fine to add it. In this case it totally seems reasonable to be able to ask a FileView what file it's bound to.

pyrito: Agreed. Thanks

libtsuba/include/tsuba/FileView.h
```diff
@@ -110,6 +110,7 @@ class KATANA_EXPORT FileView : public arrow::io::RandomAccessFile {
   }
 
   uint64_t size() const { return file_size_; }
+  std::string filename() const { return filename_; }
```
tylershunt: This kind of accessor would usually return `const std::string&`; that way it's up to the caller whether an actual copy is needed or not.

libtsuba/include/tsuba/ParquetReader.h
```diff
@@ -76,6 +76,11 @@ class KATANA_EXPORT ParquetReader {
   ///   \param uri an identifier for a parquet file
   katana::Result<int64_t> NumRows(const katana::Uri& uri);
 
+  /// Get the possible sub files for the logical parquet file, will return
+  /// a original parquet file if there are no sub files
+  ///   \param uri an identifier for a parquet file
+  katana::Result<std::vector<std::string>> GetSubFiles(const katana::Uri& uri);
```
tylershunt: Probably simpler just to make this `GetFiles` an have it always return the "top level" parquet file. (maybe it already does that in which case the comment is confusing :D)

pyrito: Yeah that is what the function does actually. I went back and forth on it and forgot to update the comment.

libtsuba/src/RDGPartHeader.h
```diff
@@ -17,6 +17,7 @@
 #include "katana/Result.h"
 #include "katana/URI.h"
 #include "tsuba/Errors.h"
+#include "tsuba/ParquetReader.h"
```
tylershunt: no longer needed.

tylershunt (approved):
Sweet, much more self contained. Thanks!

libtsuba/src/RDGPartHeader.h
pyrito: Good catch

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -97,11 +97,8 @@ RDGPartHeader::MakeJson(const katana::Uri& partition_path) {
 
 katana::Result<RDGPartHeader>
 RDGPartHeader::Make(const katana::Uri& partition_path) {
-  katana::Result<RDGPartHeader> res = MakeJson(partition_path);
-  if (!res) {
-    return res.error();
-  }
-  return res;
+  auto part_header = KATANA_CHECKED(MakeJson(partition_path));
+  return part_header;
```
tylershunt: actually this could just be:
`return MakeJson(partition_path);`
or if you still want the call trace
`return KATANA_CHECKED(MakeJson(partition_path));`

pyrito: Yeah I was thinking about that. I wasn't sure if I could return KATANA_CHECKED like that.

pyrito: Ideally I refactor the rest of the class to use KATANA_CHECKED, but that'll be for another PR

PR #618, opened 2021-10-12
tsuba: make sure new single-partion rdgs are readable by enterprise

tylershunt:
Tweaks to tsuba and graph-convert to prove that we can successfully
store a single partition RDG here that is readable when distributed.

This requires tweaks in Enterprise (and there is a test added there to
prove this works/prevent regressions).

### DIFF
--- a/libtsuba/include/tsuba/FileFrame.h
+++ b/libtsuba/include/tsuba/FileFrame.h
@@ -76,6 +76,11 @@ public:
     return reinterpret_cast<T*>(map_start_); /* NOLINT */
   }
 
+  /// only data up to cursor is written out, flavors of Write
+  /// automatically track this, so SetCursor is only useful when
+  /// treating the FileFrame like a buffer (e.g., using ptr())
+  katana::Result<void> SetCursor(uint64_t new_cursor);
+
   const std::string& path() const { return path_; }
 
   ///// Begin arrow::io::BufferOutputStream methods ///////
--- a/libtsuba/src/RDGCore.cpp
+++ b/libtsuba/src/RDGCore.cpp
@@ -205,15 +205,6 @@ RDGCore::EnsureEdgeTypesLoaded() {
 }
 
 void
-tsuba::RDGCore::InitArrowVectors() {
-  // Create an empty array, accessed by Distribution during loading
-  host_to_owned_global_node_ids_ = katana::NullChunkedArray(arrow::uint64(), 0);
-  host_to_owned_global_edge_ids_ = katana::NullChunkedArray(arrow::uint64(), 0);
-  local_to_user_id_ = katana::NullChunkedArray(arrow::uint64(), 0);
-  local_to_global_id_ = katana::NullChunkedArray(arrow::uint64(), 0);
-}
-
-void
 RDGCore::InitEmptyProperties() {
   std::vector<std::shared_ptr<arrow::Array>> empty;
   node_properties_ = arrow::Table::Make(arrow::schema({}), empty, 0);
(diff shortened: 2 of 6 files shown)


tylershunt:
Tested in https://github.com/KatanaGraph/katana-enterprise/pull/2197

tylershunt:
My tests pass, just waiting on a review to merge.

libtsuba/src/RDGCore.cpp
```diff
@@ -204,15 +204,6 @@ RDGCore::EnsureEdgeTypesLoaded() {
   return katana::ResultSuccess();
 }
 
-void
-tsuba::RDGCore::InitArrowVectors() {
-  // Create an empty array, accessed by Distribution during loading
```
witchel, line 209: So this PR eliminates `InitArrowVectors`?  I thought there was code that assumes these vectors are not null.

libtsuba/src/RDGCore.h
```diff
@@ -206,7 +202,8 @@ class KATANA_EXPORT RDGCore {
 
   katana::Result<void> RegisterTopologyFile(const std::string& new_top) {
     part_header_.set_topology_path(new_top);
-    return topology_file_storage_.Unbind();
+    return topology_file_storage_.Bind(
+        rdg_dir_.Join(new_top).string(), /*resolve =*/false);
```
witchel, line 206: Just curious what is going on here.  This function went from calling `Unbind` to calling `Bind`.  Seems weird.

libtsuba/include/tsuba/FileFrame.h
```diff
@@ -76,6 +76,11 @@ class KATANA_EXPORT FileFrame : public arrow::io::OutputStream {
     return reinterpret_cast<T*>(map_start_); /* NOLINT */
   }
 
+  /// only data up to cursor is written out, flavors of Write
+  /// automatically track this, so SetCursor is only useful when
+  /// treating the FileFrame like a buffer (e.g., using ptr())
+  katana::Result<void> SetCursor(uint64_t new_cursor);
```
witchel, line 82: I'm not sure why I always find FileFrames confusing but is this new interface only for testing?  If so, it should be commented as a test only interface.

witchel (approved):
Some confusing parts, but this looks like important functionality.

libtsuba/src/RDGCore.cpp
tylershunt: This was part of the problem problem. Some code assumed that there were always arrays here, some code checks if these are null. There was an issue because while `distribution` in enterprise fixes them up correctly, but if you don't touch them you end up storing empty arrays which we would then barf on because the loading code assumes their length is not zero.

libtsuba/src/RDGCore.h
tylershunt: There's an invariant that this `FileView` is  "bound" when you write things out and do not provide a new topology. I think we'll probably converge on better tracking of topology objects (a la properties) but for now this point fix satisfies the invariant.

libtsuba/include/tsuba/FileFrame.h
tylershunt: It's not only for testing. It's an interface that lets you use `FIleFrame` as a buffer. Until this patch it's only been used as a stream.

Not saying it should be this way but currently: `FileFrame` has some mapped memory, and copies things into it as you call `Wrtie`.  When you call write it bumps the cursor and when things are persisted they are only persisted up to the cursor.

libtsuba/src/RDGCore.cpp
tylershunt: To be clear: Open was fine with this because it doesn't touch those arrays, but when you would try to load that RDG in enterprise things would blow up.

witchel: Ok. It sounds like the right thing to do was to allow them to be null, and now all the code does that?  Sorry for initializing them, that looks like it was the wrong path.

PR #619, opened 2021-10-13
Install and prefer version 12 of clang tools

witchel:
If this PR works, I'll announce on #code-announce to make sure people realize we are upgrading to clang-12.


witchel:
If you guys want me to eliminate clang-10, this is the first step.

ddn0 (approved):

PR #621, opened 2021-10-13
A tool for analyzing git repo (with focus on changes)

GioLomia:
Creating a tool to analyze file changes in our repos by collecting the following two mapping:

1. A mapping (in json format) to include number of times each file has changed in the past; the mapping can be from file name (full path) to shas when the changes happened
2. A mapping (in json format) to show how many files changed in each commit; keep things sorted based on the order of commits on master


arthurp (commented):
Some large changes. ;-) Please rerequest a review once there is reviewing to be done.

PR #623, opened 2021-10-13
hide AppendScalarToBuilderVisitor and use 2-argument visitor

danielmawhirter:
`AppendScalarToBuilder` is an implementation detail, hide it behind `AppendToBuilder` API. This frees up the ability to modify its internals without breaking things, as will be necessary for Aman's work


libsupport/src/ArrowVisitor.cpp
```diff
@@ -7,6 +7,137 @@
...
+      arrow::is_struct_type<ScalarArrowType>::value&&
+          arrow::is_struct_type<BuilderArrowType>::value;
+
+  template <typename ScalarArrowType, typename BuilderArrowType>
+  static constexpr bool MatchFailed =
+      !(MatchNull<ScalarArrowType> ||
+        MatchPrimitive<ScalarArrowType, BuilderArrowType> ||
+        MatchString<ScalarArrowType, BuilderArrowType> ||
+        MatchList<ScalarArrowType, BuilderArrowType> ||
+        MatchStruct<ScalarArrowType, BuilderArrowType>);
```
danielmawhirter, line 48: Types don't have to strictly match, they just need to be compatible. This is the cleanest way I could think of to express the conditions above, in a way that is extensible for @amansavani to generalize. I'm all ears if there's a better way

libsupport/src/ArrowVisitor.cpp
```diff
@@ -102,8 +225,8 @@ struct ToArrayVisitor : public katana::ArrowVisitor {
 katana::Result<void>
 katana::AppendToBuilder(
     const arrow::Scalar& scalar, arrow::ArrayBuilder* builder) {
-  katana::AppendScalarToBuilder visitor(builder);
```
micah-white, line 105: I've never had to do it, but I'm assuming there are parts of the code where it is necessary/convenient to use the same visitor over multiple builders?

micah-white (approved):
LGTM

libsupport/src/ArrowVisitor.cpp
danielmawhirter: A visitor can only be used for one builder. The main reason it might make sense to reuse it is if there's some initialization cost. Since the constructor is now trivial, there's no cost to avoid, and AppendScalarToBuilderVisitor is just an implementation detail

PR #624, opened 2021-10-14
More missing packages for GPU build under conda

srogatch:


conda_recipe/environment.yml
```diff
@@ -48,6 +48,8 @@ dependencies:
  - gxx_linux-64
  - gcc_linux-64
  - cudatoolkit
```
yan-zaretskiy: I don't think you need `cudatoolkit` with `cudatoolkit-dev`.

yan-zaretskiy (approved):

PR #625, opened 2021-10-15
Adding functionality for projecting properties and creating typed projected property graph

uditagar:


libgalois/include/katana/GraphTopology.h
```diff
@@ -741,6 +741,10 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
     return original_to_projected_nodes_mapping_[nid];
   }
 
+  const uint8_t* node_bitmask() const noexcept { return node_bitmask_.data(); }
+
+  const uint8_t* edge_bitmask() const noexcept { return edge_bitmask_.data(); }
```
arthurp: These seem like bad methods to have in the public API. In general I don't think we should expose this kind of thing, but if we need to could we expose it as typed representation like a boolean arrow array or something similar? That way there is no urge for people to directly work on the bytes.

libgalois/include/katana/GraphTopology.h
```diff
@@ -852,10 +865,16 @@ class KATANA_EXPORT ProjectedTopologyWrapper : public GraphTopologyTypes {
     return topo().original_to_projected_edge_id(eid);
   }
 
+  const PropertyGraph& property_graph() const noexcept { return *prop_graph_; }
+
+  auto node_bitmask() const noexcept { return topo().node_bitmask(); }
+  auto edge_bitmask() const noexcept { return topo().edge_bitmask(); }
```
arthurp: I definitely don't want to expose these on the graph if it can be avoided.

libgalois/include/katana/GraphTopology.h
```diff
@@ -783,12 +790,18 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
...
 };
 
 template <typename Topo>
 class KATANA_EXPORT ProjectedTopologyWrapper : public GraphTopologyTypes {
+  using Base = Topo;
+
 public:
-  explicit ProjectedTopologyWrapper(const Topo* t) noexcept : topo_ptr_(t) {
+  explicit ProjectedTopologyWrapper(
+      const PropertyGraph* pg, const Topo* t) noexcept
```
arthurp: Will having a topology that depends on a graph create any weird cyclical references or anything? It may not. I just want to check.

libgalois/include/katana/TypedPropertyGraph.h
```diff
@@ -270,6 +270,12 @@ class TypedPropertyGraphView : public PGView {
       const std::vector<std::string>& edge_properties);
   static Result<TypedPropertyGraphView<PGView, NodeProps, EdgeProps>> Make(
       PropertyGraph* pg);
+  static Result<TypedPropertyGraphView<PGView, NodeProps, EdgeProps>> Make(
+      PropertyGraph* pg, const PGView& pg_view,
+      const std::vector<std::string>& node_properties,
+      const std::vector<std::string>& edge_properties);
+  static Result<TypedPropertyGraphView<PGView, NodeProps, EdgeProps>> Make(
+      PropertyGraph* pg, const PGView& pg_view);
```
arthurp: Is passing the `pg` here required? It would make sense for the view to have all the information needed to wrap it as a type graph view.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -59,6 +61,76 @@ ConstructNodeProperties(
   return pg->AddNodeProperties(res_table.value());
 }
 
+inline std::shared_ptr<arrow::ChunkedArray>
+ApplyBitMask(
+    const std::shared_ptr<arrow::ChunkedArray>& chunked_array,
+    const uint8_t* bit_mask) {
```
arthurp: I really think we need a BitMask type. Passing them around untyped in this way is dangerous.

arthurp (commented):
I'm worried about the bare uint8_t* as a bitset. Can we get that wrapped up in some type. Even if it's just a container for the pointer. I want to have some static protection to prevent stupid things from happening.

libgalois/include/katana/analytics/Utils.h
uditagar: I will try to figure out how to use the `Bitmap` class in the `arrow` library then. thanks

roshandathathri: No, that's not the point. You can pass `const NUMAArray<uint8_t>&` everywhere and use it as a pointer in line 72

libgalois/include/katana/GraphTopology.h
roshandathathri: Agree with Arthur. 

Return the const arrays here directly. We should NOT be passing around pointers. Arrow requires pointers, so we can call `.data()` when we call Arrow but we don't need to deal with pointers in our code. 

Please add doxygen comments to the functions saying what set and unset in the mask represents. 

Also maybe add `projected_` prefix to the names?

libgalois/include/katana/GraphTopology.h
```diff
@@ -783,12 +790,18 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
   NUMAArray<Node> projected_to_original_nodes_mapping_;
   NUMAArray<Edge> original_to_projected_edges_mapping_;
   NUMAArray<Edge> projected_to_original_edges_mapping_;
+  NUMAArray<uint8_t> node_bitmask_;
+  NUMAArray<uint8_t> edge_bitmask_;
```
roshandathathri: Can we avoid storing this and generating this on-demand (when requested) from the mappings above? 

On that note, taking dependent arguments in the constructor is a risk. What if these variables are passed such that they are not agree-ing with each other?

If we need to store them, I'd suggest storing `DynamicBitset` and generating `NUMAAray<uint8_t>` on-demand (before calling Arrow) in the caller code (not in this call).

libgalois/include/katana/GraphTopology.h
roshandathathri: Agree. Please move this class out of this file and change the class name to `ProjectedPropertyGraphWrapper`.  

Why are `PropertyGraph` and `Topo` passed and stored? Doesn't `PropertyGraph` have the `Topo`? This also leads to the dependent arguments problem I mentioned. What if the topology in `PropertyGraph` is NOT the same as the `Topo` passed?

libgalois/include/katana/TypedPropertyGraph.h
roshandathathri: Agree with Arthur. Ditto about dependent arguments. Doesn't `PGView` have the `PropertyGraph`? What if the one it has is different from the one passed?

libgalois/include/katana/analytics/Utils.h
```diff
@@ -59,6 +61,76 @@ ConstructNodeProperties(
...
+    columns.emplace_back(std::move(new_col));
+  }
+
+  return arrow::Table::Make(arrow::schema(fields), columns);
+}
+
+template <typename PGView, typename NodeProps>
+inline katana::Result<void>
+ConstructNodeProperties(
+    PropertyGraph* pg, const PGView& pg_view,
```
roshandathathri: Ditto about dependent arguments.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -72,6 +144,26 @@ ConstructEdgeProperties(
   return pg->AddEdgeProperties(res_table.value());
 }
 
+template <typename PGView, typename EdgeProps>
+inline katana::Result<void>
+ConstructEdgeProperties(
+    PropertyGraph* pg, const PGView& pg_view,
```
roshandathathri: Ditto

libgalois/src/GraphTopology.cpp
```diff
@@ -569,6 +582,27 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
     }
   });
 
+  katana::do_all(
+      katana::iterate(
+          (uint32_t)0, (uint32_t)std::ceil((double)topology.num_nodes() / 8.0)),
```
roshandathathri: Store the num of node bytes as a variable instead of using the same expression.

libgalois/src/GraphTopology.cpp
```diff
@@ -569,6 +582,27 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
...
+      [&](uint32_t i) {
+        auto node_start = i * 8;
+        auto node_end = (i + 1) * 8;
+        node_end =
+            (node_end > topology.num_nodes()) ? topology.num_nodes() : node_end;
+        uint8_t val{0};
+        while (node_start != node_end) {
+          if (bitset_nodes.test(node_start)) {
```
roshandathathri: My point is that we can just store this bitset in the class if we really need to store it.

libgalois/src/GraphTopology.cpp
```diff
@@ -685,12 +722,34 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
...
+      katana::iterate(
+          (uint32_t)0, (uint32_t)std::ceil((double)topology.num_edges() / 8.0)),
+      [&](uint32_t i) {
+        auto edge_start = i * 8;
+        auto edge_end = (i + 1) * 8;
+        edge_end =
+            (edge_end > topology.num_edges()) ? topology.num_edges() : edge_end;
+        uint8_t val{0};
+        while (edge_start != edge_end) {
+          if (bitset_edges.test(edge_start)) {
```
roshandathathri: Ditto about storing this bitset.

roshandathathri (changes requested):
Thanks, Udit. The logic is there but this needs some code reorganization.

libgalois/include/katana/GraphTopology.h
uditagar: so which class should I put this into? `PropertyGraph.h` one? I did the same thign as Amber had for SortedTopology thing, after consulting him.

uditagar: and here `topo` corresponds to projected topology

libgalois/include/katana/GraphTopology.h
uditagar: I have changed this now by using arrow's Bitmap class

libgalois/include/katana/TypedPropertyGraph.h
uditagar: yeah, will update this

libgalois/include/katana/analytics/Utils.h
uditagar: I feel that the arrow's Bitmap class is more clean for this. So I have updated the code with that

libgalois/include/katana/GraphTopology.h
uditagar: the bitmask I thought were required to store, to even create the node properties.

libgalois/src/GraphTopology.cpp
uditagar: I thought that the bitmask is required throughout as it stays with the node properties (same for edge properties). Maybe I am wrong

uditagar:
I have done some updates. Will need to discuss this more in the Monday analytics's meeting. Seems everyone has their own ideas. Better to update after we reach some consensus. thanks

libgalois/include/katana/GraphTopology.h
roshandathathri: Ok. Hadn't realized that. That makes sense.

libgalois/include/katana/GraphTopology.h
roshandathathri: If it has properties, it is no longer a simple `*Topology`. 

@amberhassaan Where should this go? I'm fine if it's in a new header.

libgalois/src/GraphTopology.cpp
roshandathathri: Yes, you are right. I hadn't realized that.

libgalois/include/katana/analytics/Utils.h
uditagar: but the problem here is that `pg_view` returns a `const` property graph and hence I will not be able to update the property graph then

libgalois/include/katana/analytics/Utils.h
uditagar: ditto

uditagar:
I have taken care of things that we have discussed here as well as during our meeting today. @roshandathathri please check to clear the bottleneck, once you get time. thanks

libgalois/include/katana/GraphTopology.h
```diff
@@ -6,6 +6,7 @@
 
 #include <boost/iterator/counting_iterator.hpp>
 
+#include "arrow/util/bitmap.h"
```
amberhassaan, line 10: separate this include from katana includes

libgalois/include/katana/GraphTopology.h
```diff
@@ -772,7 +783,15 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
         original_to_projected_edges_mapping_(
             std::move(original_to_projected_edges_mapping)),
         projected_to_original_edges_mapping_(
-            std::move(projected_to_original_edges_mapping)) {}
+            std::move(projected_to_original_edges_mapping)),
+        node_bitmask_data_(std::move(node_bitmask_data)),
+        edge_bitmask_data_(std::move(edge_bitmask_data)),
+        node_bitmask_(
+            static_cast<void*>(node_bitmask_data_.data()), 0,
+            (int64_t)original_to_projected_nodes_mapping_.size()),
```
amberhassaan: Use `static_cast`. Never (ever) use C style cast.

libgalois/test/projection.cpp
```diff
@@ -6,6 +6,7 @@
 #include "katana/Logging.h"
 #include "katana/PropertyGraph.h"
 #include "katana/SharedMemSys.h"
+#include "katana/TypedPropertyGraph.h"
 #include "katana/analytics/Utils.h"
 #include "llvm/Support/CommandLine.h"
 #include "stdio.h"
```
amberhassaan: why do you need `stdio.h`. Please remove as we have `fmt` library. Also, in C++, you should include `<cstdio>` etc.

libgalois/src/GraphTopology.cpp
```diff
@@ -559,6 +568,11 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
   NUMAArray<Node> projected_to_original_nodes_mapping;
   projected_to_original_nodes_mapping.allocateInterleaved(num_new_nodes);
 
+  uint32_t num_nodes_bytes =
```
amberhassaan: This is bad. Never use C-style casts. Here's the integer ceil trick:
`(x + y - 1) / y` computes the ceil of `x/y`.

libgalois/src/GraphTopology.cpp
```diff
@@ -467,12 +469,19 @@ katana::ProjectedTopology::CreateEmptyEdgeProjectedTopology(
       original_to_projected_edges_mapping.begin(),
       original_to_projected_edges_mapping.end(), Edge{topology.num_edges()});
 
+  NUMAArray<uint8_t> node_bitmask;
+  node_bitmask.allocateInterleaved(topology.num_nodes());
```
amberhassaan: These lengths seem wrong for a bit mask. It should be `num_nodes/8` rounded up.

libgalois/src/GraphTopology.cpp
```diff
@@ -569,6 +583,26 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
     }
   });
 
+  katana::do_all(
```
amberhassaan: I don't like this code at all. We need to either use `std::bitset` or `vector<bool>` (both of which I believe pack bits into bytes). I don't like the divide by 8 here, multiply by 8 there, and take a remainder over there. This is all recipe for future subtle bugs. If none of the STL classes fit the bill, we should write or modify our own class, e.g. DynamicBitset. cc: @roshandathathri

libgalois/include/katana/analytics/Utils.h
```diff
@@ -59,6 +62,76 @@ ConstructNodeProperties(
...
+
+  data->buffers[0] = bit_mask;
+  data->null_count = arrow::kUnknownNullCount;
+
+  auto array_ptr = arrow::MakeArray(data);
+
+  auto status = array_ptr->ValidateFull();
+
+  KATANA_LOG_ASSERT(status.ok());
+  new_chunks.emplace_back(array_ptr);
```
amberhassaan: move declaration of new_chunks here, as close to use as possible.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -59,6 +62,76 @@ ConstructNodeProperties(
...
+}
+
+inline katana::Result<std::shared_ptr<arrow::Table>>
+AddBitMaskToTable(
+    std::shared_ptr<arrow::Table> table,
+    const std::shared_ptr<arrow::Buffer>& bit_mask) {
+  std::vector<std::shared_ptr<arrow::Field>> fields;
+  std::vector<std::shared_ptr<arrow::ChunkedArray>> columns;
+
+  for (const auto& field : table->fields()) {
```
amberhassaan: what's the need for this loop. Why can't you do `arrow::schema(table->fields())`.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -59,6 +62,76 @@ ConstructNodeProperties(
...
+  return arrow::Table::Make(arrow::schema(fields), columns);
+}
+
+template <typename PGView, typename NodeProps>
+inline katana::Result<void>
+ConstructNodeProperties(
+    const PGView& pg_view,
+    const std::vector<std::string>& names = DefaultPropertyNames<NodeProps>()) {
+  auto pg = const_cast<PropertyGraph*>(pg_view.get_property_graph());
+  auto res_table = katana::AllocateTable<NodeProps>(pg->num_nodes(), names);
```
amberhassaan: We are allocating twice here (unless I'm wrong). First `AllocateTable` allocates the arrays, then we recreate them with the right bit-mask. We should overload `AllocateTable` to take a bitmask. Please add a TODO here for now and address it soon.

amberhassaan (commented):
Left some comments.

libgalois/include/katana/analytics/Utils.h
uditagar: I will try to do that then in `Properties.h ` file.

libgalois/include/katana/analytics/Utils.h
uditagar: will remove it

libgalois/src/GraphTopology.cpp
uditagar: I am not sure of how to do this as I also want to be able to convert the data to the format of `const uint8_t*`, which I am currently able to do with the help of `NUMAArray`. But yes if there is a solution that allows parallel setting of bits plus also a conversion to` const uint8_t*`, then it's fine. thank you

uditagar: I am going to leave this as TODO in the code for now (as we have done for the last summer deadline), as I don't have much time left to complete this project. thanks

uditagar:
fixed more comments and left one as TODO. thanks

libgalois/src/GraphTopology.cpp
amberhassaan: ok, for now, you can at least refactor this code into a common function and then put a TODO. You're doing the same thing twice for nodes and edges.

uditagar: sure. I did it now

libgalois/include/katana/GraphTopology.h
```diff
@@ -819,7 +835,16 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
...
         projected_to_original_edges_mapping_(
-            std::move(projected_to_original_edges_mapping)) {}
+            std::move(projected_to_original_edges_mapping)),
+        node_bitmask_data_(std::move(node_bitmask_data)),
+        edge_bitmask_data_(std::move(edge_bitmask_data)),
+        node_bitmask_(
+            static_cast<void*>(node_bitmask_data_.data()), 0,
+            static_cast<int64_t>(original_to_projected_nodes_mapping_.size())),
+        edge_bitmask_(
+            edge_bitmask_data_.data(), 0,
```
roshandathathri: Why static cast for one but not the other?

libgalois/include/katana/GraphTopology.h
```diff
@@ -830,12 +855,20 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
   NUMAArray<Node> projected_to_original_nodes_mapping_;
   NUMAArray<Edge> original_to_projected_edges_mapping_;
   NUMAArray<Edge> projected_to_original_edges_mapping_;
+  NUMAArray<uint8_t> node_bitmask_data_;
+  NUMAArray<uint8_t> edge_bitmask_data_;
+  arrow::internal::Bitmap node_bitmask_;
+  arrow::internal::Bitmap edge_bitmask_;
 };
 
 template <typename Topo>
```
roshandathathri: nit: call this `ProjectedTopo` and rename the variables accordingly

libgalois/include/katana/GraphTopology.h
```diff
@@ -899,10 +932,24 @@ class KATANA_EXPORT ProjectedTopologyWrapper : public GraphTopologyTypes {
     return topo().original_to_projected_edge_id(eid);
   }
 
+  const PropertyGraph& property_graph() const noexcept { return *prop_graph_; }
+
+  const std::shared_ptr<arrow::Buffer>& node_bitmask() const noexcept {
+    return topo().node_bitmask();
+  }
+  const std::shared_ptr<arrow::Buffer>& edge_bitmask() const noexcept {
+    return topo().edge_bitmask();
```
roshandathathri, line 939: I'm NOT sure why we are taking a template parameter for topology when these functions are only supported/defined by `ProjectedTopology`. Should we drop the template parameter and replace it with `ProjectedTopology`?

libgalois/include/katana/Properties.h
```diff
@@ -249,6 +249,86 @@ AllocateTable(uint64_t num_rows, const std::vector<std::string>& names) {
       std::make_index_sequence<std::tuple_size_v<PropTuple>>());
 }
 
+inline std::shared_ptr<arrow::ChunkedArray>
```
roshandathathri, line 252: Why inline? This is a reasonably big function and it's better to let the compiler decide whether to inline it or not.

libgalois/include/katana/Properties.h
```diff
@@ -249,6 +249,86 @@ AllocateTable(uint64_t num_rows, const std::vector<std::string>& names) {
...
+ApplyBitMask(
+    const std::shared_ptr<arrow::ChunkedArray>& chunked_array,
+    const std::shared_ptr<arrow::Buffer>& bit_mask) {
+  std::shared_ptr<arrow::Array> array = chunked_array->chunk(0);
```
roshandathathri, line 256: If you're assuming only one chunk, then accept and return an `arrow::Array`

libgalois/include/katana/Properties.h
```diff
@@ -249,6 +249,86 @@ AllocateTable(uint64_t num_rows, const std::vector<std::string>& names) {
...
+  std::shared_ptr<arrow::ArrayData> data = array->data()->Copy();
+
+  data->buffers[0] = bit_mask;
+  data->null_count = arrow::kUnknownNullCount;
+
+  auto array_ptr = arrow::MakeArray(data);
+
+  auto status = array_ptr->ValidateFull();
+
+  KATANA_LOG_ASSERT(status.ok());
```
roshandathathri: Is this validation and assert required? If not:
`KATANA_LOG_DEBUG_ASSERT(array_ptr->ValidateFull().ok());`

libgalois/include/katana/Properties.h
```diff
@@ -249,6 +249,86 @@ AllocateTable(uint64_t num_rows, const std::vector<std::string>& names) {
...
+  std::vector<std::shared_ptr<arrow::Array>> new_chunks;
+  new_chunks.emplace_back(array_ptr);
+
+  auto res_new_chunked_array = arrow::ChunkedArray::Make(new_chunks);
+  if (res_new_chunked_array.ok()) {
```
roshandathathri, line 270: Isn't this handled by `KATANA_CHECKED`?

libgalois/include/katana/Properties.h
```diff
@@ -249,6 +249,86 @@ AllocateTable(uint64_t num_rows, const std::vector<std::string>& names) {
...
+  } else {
+    return nullptr;
+  }
+}
+
+template <typename PropTuple, size_t head, size_t... tail>
+Result<std::shared_ptr<arrow::Table>>
+AllocateTable(
+    uint64_t num_rows, const std::vector<std::string>& names,
+    const std::shared_ptr<arrow::Buffer>& bit_mask,
```
roshandathathri, line 283: This needs documentation: specifically that the bit_mask is used for all columns.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -59,6 +62,22 @@ ConstructNodeProperties(
   return pg->AddNodeProperties(res_table.value());
 }
 
+template <typename PGView, typename NodeProps>
+inline katana::Result<void>
+ConstructNodeProperties(
+    const PGView& pg_view,
+    const std::vector<std::string>& names = DefaultPropertyNames<NodeProps>()) {
+  auto pg = const_cast<PropertyGraph*>(pg_view.get_property_graph());
```
roshandathathri, line 72: This is NOT ideal. Please add a TODO and mention in the doxygen for these functions that pg_view is modified to add properties.

libgalois/src/GraphTopology.cpp
```diff
@@ -611,12 +613,19 @@ katana::ProjectedTopology::CreateEmptyEdgeProjectedTopology(
       original_to_projected_edges_mapping.begin(),
       original_to_projected_edges_mapping.end(), Edge{topology.num_edges()});
 
+  NUMAArray<uint8_t> node_bitmask;
+  node_bitmask.allocateInterleaved(std::ceil(topology.num_nodes() / 8));
+
+  NUMAArray<uint8_t> edge_bitmask;
+  edge_bitmask.allocateInterleaved(std::ceil(topology.num_edges() / 8));
```
roshandathathri: Use `sizeof(uint8_t)` instead of 8 to make the reason clear. Also avoid `ceil` as Amber pointed out earlier. I'd suggest a simple helper-function defined in this class for that purpose.

libgalois/src/GraphTopology.cpp
```diff
@@ -625,6 +634,31 @@ katana::ProjectedTopology::CreateEmptyProjectedTopology(
   return CreateEmptyEdgeProjectedTopology(pg, 0);
 }
 
+void
+katana::ProjectedTopology::FillBitMask(
```
roshandathathri: Add doxygen comments

libgalois/src/GraphTopology.cpp
```diff
@@ -625,6 +634,31 @@ katana::ProjectedTopology::CreateEmptyProjectedTopology(
...
+    uint32_t num_bytes, uint32_t num_elements,
```
roshandathathri: (1) Typically use `size_t` for variables that represent the size. 

(2) Aren't these dependent variables? What if the passed arguments are such `num_bytes * 8 != num_elements`. Take one argument and have a local varaible for the other.

libgalois/src/GraphTopology.cpp
```diff
@@ -625,6 +634,31 @@ katana::ProjectedTopology::CreateEmptyProjectedTopology(
...
+    const katana::DynamicBitset& bitset, katana::NUMAArray<uint8_t>* bitmask) {
+  // TODO(udit) find another way to do the following
+  katana::do_all(
+      katana::iterate(static_cast<uint32_t>(0), num_bytes),
```
roshandathathri: `uint32_t{0}` is shorter and semantically the same.

libgalois/src/GraphTopology.cpp
```diff
@@ -625,6 +634,31 @@ katana::ProjectedTopology::CreateEmptyProjectedTopology(
...
+      [&](uint32_t i) {
+        auto start = i * 8;
+        auto end = (i + 1) * 8;
+        end = (end > num_elements) ? num_elements : end;
+        uint8_t val{0};
+        while (start != end) {
+          if (bitset.test(start)) {
+            uint8_t bit_offset = static_cast<uint8_t>(1);
```
roshandathathri: `uint8_t bit_offset{1};`

libgalois/src/GraphTopology.cpp
```diff
@@ -625,6 +634,31 @@ katana::ProjectedTopology::CreateEmptyProjectedTopology(
...
+            bit_offset <<= (start % 8);
+            val = val | bit_offset;
+          }
+          start++;
+        }
+        (*bitmask)[i] = val;
+      },
+      katana::chunk_size<1>());
```
roshandathathri: Why change the chunk size? This seems too small a chunk size. Such small chunk sizes make sense only if each iteration of the do_all is very heavy-weight - executing in ms.

libgalois/src/GraphTopology.cpp
```diff
@@ -703,6 +737,11 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
   NUMAArray<Node> projected_to_original_nodes_mapping;
   projected_to_original_nodes_mapping.allocateInterleaved(num_new_nodes);
 
+  uint32_t num_nodes_bytes = (topology.num_nodes() + 7) / 8;
```
roshandathathri, line 744: Replace 7 with `sizeof(uint8_t) - 1` and 8 with `sizeof(uint8_t)`

libgalois/src/GraphTopology.cpp
```diff
@@ -625,6 +634,31 @@ katana::ProjectedTopology::CreateEmptyProjectedTopology(
   return CreateEmptyEdgeProjectedTopology(pg, 0);
 }
 
+void
+katana::ProjectedTopology::FillBitMask(
+    uint32_t num_bytes, uint32_t num_elements,
+    const katana::DynamicBitset& bitset, katana::NUMAArray<uint8_t>* bitmask) {
+  // TODO(udit) find another way to do the following
```
roshandathathri: Why does this need to be changed?

libgalois/src/GraphTopology.cpp
```diff
@@ -713,6 +752,10 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
     }
   });
 
+  FillBitMask(
+      num_nodes_bytes, static_cast<uint32_t>(topology.num_nodes()),
```
roshandathathri: Don't type-cast

roshandathathri (approved):
LGTM

libgalois/src/GraphTopology.cpp
uditagar: Amber said that this is inefficient or prone to errors

libgalois/include/katana/GraphTopology.h
uditagar: you are right. I will do that then

libgalois/src/GraphTopology.cpp
roshandathathri: That needs to be in comments :)
Not that Amber said it but that it is inefficient or prone to errors

libgalois/src/GraphTopology.cpp
uditagar: then how else should I convert it to `uint32_t` format

roshandathathri: Related to my point above about using `size_t`. Why do you need to convert?

uditagar: yeah you are right. I didn't see that. sorry about that

libgalois/include/katana/Properties.h
uditagar: if I remove inline, then I get the error of multiple definition of this fucntion. like these: `CMakeFiles/katana_galois.dir/src/PropertyGraph.cpp.o: In function `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)':
/home/udit/udit-Jul19/katana/libgalois/include/katana/Properties.h:255: multiple definition of `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)'
CMakeFiles/katana_galois.dir/src/Properties.cpp.o:/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1900: first defined here
CMakeFiles/katana_galois.dir/src/PropertyViews.cpp.o: In function `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)':
/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1459: multiple definition of `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)'
CMakeFiles/katana_galois.dir/src/Properties.cpp.o:/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1900: first defined here
CMakeFiles/katana_galois.dir/src/analytics/Utils.cpp.o: In function `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)':
/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1465: multiple definition of `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)'
CMakeFiles/katana_galois.dir/src/Properties.cpp.o:/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1900: first defined here
CMakeFiles/katana_galois.dir/src/analytics/betweenness_centrality/betweenness_centrality.cpp.o: In function `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)':
/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1222: multiple definition of `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)'
CMakeFiles/katana_galois.dir/src/Properties.cpp.o:/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1900: first defined here
CMakeFiles/katana_galois.dir/src/analytics/betweenness_centrality/level.cpp.o: In function `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)':
/home/udit/udit-Jul19/katana/libgalois/include/katana/Properties.h:255: multiple definition of `katana::ApplyBitMask(std::shared_ptr<arrow::ChunkedArray> const&, std::shared_ptr<arrow::Buffer> const&)'
CMakeFiles/katana_galois.dir/src/Properties.cpp.o:/home/udit/.conan/data/fmt/6.2.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/fmt/format.h:1900: first defined here
`

uditagar: Also the `ConstructNodeProperties` from where `AllocTable` is called (which in turn calls this fucntion) is also `inline`.

libgalois/src/GraphTopology.cpp
uditagar: you are right. Don't know what was I thinking

libgalois/include/katana/Properties.h
uditagar: seems not

libgalois/include/katana/Properties.h
uditagar: I think this is better.  Otherwise it will clutter the other fucntion from where it is called. thanks

libgalois/src/GraphTopology.cpp
uditagar: `sizeof(uint8_t)` return 1. So I am reverting back to my previous formulation

libgalois/src/GraphTopology.cpp
uditagar: `sizeof(uint8_t)` returns 1

uditagar:
I fixed most of the comments. I left the comment about `sizeof(uint8_t)` since it returns 1. I also left the comment about `inline`, since it was giving compilation error. I will push it to the repo now so that I can run tests for the distributed PR. thanks

libgalois/src/GraphTopology.cpp
roshandathathri: Sorry! My bad. Check DynamicBitset.h for a better solution. Do something similar:
`static constexpr size_t kNumBitsInUint8 = sizeof(uin8_t) * CHAR_BIT;`
Use `kNumBitsInUint8` instead of 8.

libgalois/include/katana/Properties.h
roshandathathri: Then throw an error if the array has more than 1 chunk and document the behavior in doxygen comments for the function. That should not be that much cluttered.

libgalois/include/katana/Properties.h
roshandathathri: `ConstructNodeProperties` is templated right?

Can you move the definition of this function to .cpp file?

libgalois/src/GraphTopology.cpp
uditagar: oh I merged this one. I will create another PR to take care of this then. thanks

roshandathathri: No worries. Thanks!

PR #626, opened 2021-10-16
Enable easy download of environment capture file, add CLI for it.

arthurp:
Also tweaks a few things in the environment capture code.

### DIFF
--- a/python/katana/bug/__main__.py
+++ b/python/katana/bug/__main__.py
@@ -14,15 +14,19 @@ if __name__ == "__main__":
 
     parser.add_argument(
         "destination",
-        help="Output the binary to the given path or stdout if '-' is given",
+        help="Output the binary to the given path, if this is not provided a temporary file name is chosen.",
         type=str,
         nargs="?",
         default=None,
     )
 
+    parser.add_argument("--stdout", help="Send output to stdout instead of a file.", action="store_true")
+
     args = parser.parse_args()
 
-    if args.destination == "-":
+    if args.stdout:
+        if args.destination:
+            print("WARNING: Ignoring filename since --stdout was given.", file=sys.stderr)
         destination = sys.stdout.buffer
     else:
         destination = args.destination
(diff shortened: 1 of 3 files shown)


arthurp:
@prphntm63 Can you check that what I'm doing with `HTML` in Jupyter is sane? I'm not confident and you have done a bunch of stuff with it.

python/katana/bug/__main__.py
```diff
@@ -0,0 +1,32 @@
...
+        "destination",
+        help="Output the binary to the given path or stdout if '-' is given",
+        type=str,
+        nargs="?",
+        default=None,
+    )
+
+    args = parser.parse_args()
+
+    if args.destination == "-":
```
denish44: This feels a bit odd, i would say make it explicit "stdout" or default

denish44 (approved):
LGTM

python/katana/bug/__main__.py
arthurp: `-` meaning stdout is reasonably common in Unix tools. Which is why I used it. "stdout" isn't great because it's something that could be a file name. I don't want it to be the default because sending to stdout is an expert feature. I'll use an `--stdout` option instead though. The old Unix conventions are not well known any more.

denish44: ah if it's common than it's fine. I haven't seen it before so lack of knowledge there.

prphntm63:
> @prphntm63 Can you check that what I'm doing with `HTML` in Jupyter is sane? I'm not confident and you have done a bunch of stuff with it.

Sorry I'm late on this, but you're good there. I'm a firm believer if it looks and works like you want it to without breaking anything else, it's fine ;)

arthurp:
> Sorry I'm late on this, but you're good there. I'm a firm believer if it looks and works like you want it to without breaking anything else, it's fine ;)

Well then. I'll remember to use you as a reviewer when I'm trying to sneak bad code into the codebase. ;-)

PR #627, opened 2021-10-18
tsuba: Remove stray KATANA_LOG_ERROR

ddn0:


witchel (approved):
I totally noticed this.
I was going to turn it into a DEBUG, but that is probably also too verbose.

tylershunt (approved):

PR #628, opened 2021-10-18
Update AppendScalarToBuilderVisitor

amansavani:
MatchPrimitive in AppendScalarToBuilderVisitor checked that 
the builder type and scalar type matched, but it is not necessary 
for the types to match, its ok if they are compatible


danielmawhirter (approved):
That's great if this is all it takes!

PR #629, opened 2021-10-18
Fix: RDGSlice accounts for header in entity type storage

scober:
RDGSlice treats the entity type id array storage as a binary array on
disk (i.e. mmap file, cast to pointer, now you have an array). In fact,
the on disk format contains a 8 byte header that contains the size
of the array that follows (i.e. mmap file, offset by 8, cast to pointer,
now you have an array of length <first 8 bytes of file interpreted as
uint64_t>). Modify RDGSlice to ignore that header and offset all of its
math by 8 bytes.

It would be nice to have a more consistent treatment of this offset than
the hard-coded one-off ignoring in this commit. But RDG doesn't
do any accounting for the header and PropertyGraph does, so it isn't
clear where we could centralize our handling.

JIRA: KAT-2029

### DIFF
--- a/libgalois/src/PropertyGraph.cpp
+++ b/libgalois/src/PropertyGraph.cpp
@@ -21,6 +21,7 @@
 #include "tsuba/FileFrame.h"
 #include "tsuba/RDG.h"
 #include "tsuba/RDGManifest.h"
+#include "tsuba/RDGPrefix.h"
 #include "tsuba/tsuba.h"
 
 namespace {
@@ -152,8 +153,8 @@ WriteTopology(const katana::GraphTopology& topology) {
 /// favor of this method.
 katana::Result<katana::PropertyGraph::EntityTypeIDArray>
 MapEntityTypeIDsArray(const tsuba::FileView& file_view) {
-  const auto* data = file_view.ptr<uint64_t>();
-  const int64_t type_ID_array_size = data[0];
+  const auto* data = file_view.ptr<tsuba::EntityTypeIDArrayHeader>();
+  const auto header = data[0];
 
   if (file_view.size() == 0) {
     return katana::ErrorCode::InvalidArgument;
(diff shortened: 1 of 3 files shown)


witchel:
I mean, I know you generally don't overlook the obvious, but can I make you go one more round on why this is the best workaround?  'Cause somebody described it as "puke-a-licious" but that wasn't me because I'm a nice guy and I'm your friend and I would never say anything like that.

e-mcginnis:
>It would be nice to have a more consistent treatment of this offset than
the hard-coded one-off ignoring in this commit. But RDG doesn't
do any accounting for the header and PropertyGraph does, so it isn't
clear where we could centralize our handling.

agreed. I think its something that could be moved into the RDG. I want to get a standard format put together for how libtsuba handles fileframes that it dumps to disk.

e-mcginnis (approved):
lgtm!

witchel:
> lgtm!

But does it?  Does it look good?

e-mcginnis:
I didn't give proper thought to rdgslice when I made the executive decision to include the uint64_t length of the array in the fileframe (going to blame that on me being a noob at the time :D ) Now Simon has to deal with it. This seems like a *fine* option. 

I'm not seeing a more obvious, nice way to work around this to be honest. But maybe I'm missing something obvious that the "puke-a-licious" person sees :) 
Every option I can think of has some grossness. 

EDIT: all of that said, I do like the idea of having the array size at the top of the file as a sanity check. But maybe its not worth it for the pain here.

scober:
This can't be moved into the the RDG because there is no RDG. `RDG` and `RDGSlice` are logically siblings. Their shared logic goes in `RDGCore`. We could put this in `RDGCore`, but it isn't obvious what that would look like. We could have `RDGCore` expose a `type_array()` and a `type_array_slice()` and then we would at least have everything in one class. But that sort of defeats the purpose of `RDGCore`, because then we are just implementing the two different classes in `RDGCore`.

Another option is to have `RDGCore` expose only `type_array_slice()` and `type_array_length()` and then `RDG` could request the degenerate slice. But this feels a little awkward.

The fundamental issue with combining these is that `RDG` and `RDGSlice` need to know about this header for different reasons (`RDG` arguably doesn't need to know about it at all).

We could also move the current handling of this in `PropertyGraph` into `RDG`. Then we would at least always handle it at the `RDG`/`RDGSlice` level of the code.

We could also make a little
```cpp
struct EntityTypeIDArrayHeader {
  uint64_t size;
}
```
Which would allow us to centralize our definition of what this header is and how large it is without trying to actually centralize the logic for it.

EDIT: the last option is the closest to how we handle the topology file format and I think it is my preference.

witchel:
> We could also make a little
> 
> ```c++
> struct EntityTypeIDArrayHeader {
>   uint64_t size;
> }
> ```
> 
> Which would allow us to centralize our definition of what this header is and how large it is without trying to actually centralize the logic for it.
> 
> EDIT: the last option is the closest to how we handle the topology file format and I think it is my preference.

Actually, that sounds like a good compromise.

That RDG and RDGSlice are siblings is just something I'm going to have to meditate on.

And I'll tell the "puke-a-licious" person, who really is very rude, to put a sock in it.

PR #631, opened 2021-10-19
graph-convert: fix generated types

tylershunt:
Simon helped me realize that this was wrong as well. If we get a
better way to update this we should change this to use that. What's here
at least makes it correct.


scober (commented):
Would you be willing to match your implementation more closely with https://github.com/KatanaGraph/katana/pull/629?

e-mcginnis (approved):
lgtm!
just another reason for me to better represent the layout of the node/edge type arrays in the code :/ 
Sorry for the mess.

PR #634, opened 2021-10-19
Allow any column name in property Parquet files

darabos:
Hi,
With this change I'm able to use the same Parquet files in a new RDG version under a new name.
I'm not sure about the C++. `ValueOrDie` sounds scary, but it's after the `ok()` check. Is there a better way?

### DIFF
--- a/libtsuba/src/AddProperties.cpp
+++ b/libtsuba/src/AddProperties.cpp
@@ -38,7 +38,8 @@ DoLoadProperties(
   auto renamed = out->RenameColumns({expected_name});
   if (!renamed.ok()) {
     std::shared_ptr<arrow::Schema> schema = out->schema();
-    return KATANA_ERROR(tsuba::ErrorCode::InvalidArgument, "{}", renamed.status().ToString());
+    return KATANA_ERROR(
+        tsuba::ErrorCode::InvalidArgument, "{}", renamed.status().ToString());
   }
   return renamed.ValueOrDie();
 }
--- a/libtsuba/src/AddProperties.cpp
+++ b/libtsuba/src/AddProperties.cpp
@@ -35,19 +35,12 @@ DoLoadProperties(
 
   std::shared_ptr<arrow::Table> out = std::move(out_res.value());
 
-  std::shared_ptr<arrow::Schema> schema = out->schema();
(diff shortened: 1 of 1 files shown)


libtsuba/src/AddProperties.cpp
```diff
@@ -35,19 +35,12 @@ DoLoadProperties(
 
   std::shared_ptr<arrow::Table> out = std::move(out_res.value());
 
-  std::shared_ptr<arrow::Schema> schema = out->schema();
-  if (schema->num_fields() != 1) {
-    return KATANA_ERROR(
-        tsuba::ErrorCode::InvalidArgument, "expected 1 field found {} instead",
```
darabos, line 41: `RenameColumns()` performs this same check, so we can just rely on that.

libtsuba/src/AddProperties.cpp
```diff
@@ -35,19 +35,12 @@ DoLoadProperties(
...
-        schema->num_fields());
-  }
-
-  if (schema->field(0)->name() != expected_name) {
-    return KATANA_ERROR(
-        tsuba::ErrorCode::InvalidArgument, "expected {} found {} instead",
-        expected_name, schema->field(0)->name());
+  auto renamed = out->RenameColumns({expected_name});
```
darabos, line 38: I'm not sure renaming is necessary. This way nobody outside of this function can even tell that the name in the Parquet file is not the same as in the metadata. Perhaps nobody looks at it anyway?

tylershunt: Yeah, renaming is important since we rely on the arrow schema to name properties. thanks!

tylershunt (commented):
thanks for this! I'll open a PR for katana-enterprise to make sure CI is ok with this and then we should be able to merge.

darabos:
> thanks for this! I'll open a PR for katana-enterprise to make sure CI is ok with this and then we should be able to merge.

Sorry, I totally forgot about it! Let me know if you have an update on this. This is still kind of required for the LynxKite integration. Thanks!

wkyu2kg:
@tylershunt Daniel mentioned about this PR in today's meeting. Would you review and approve this?

PR #635, opened 2021-10-19
Cache adds clear method

witchel:
Cache should be able to clear itself, and eventually evict a given entry.

This is needed when the graph is updated because we have to evict updated properties.

This is just part of what is needed to fix the property cache.
Jira: KAT-2085

### DIFF
--- a/libsupport/include/katana/Cache.h
+++ b/libsupport/include/katana/Cache.h
@@ -73,7 +73,13 @@ public:
 
   size_t capacity() const { return capacity_; }
 
-  bool Empty() const { return key_to_value_.empty(); }
+  void clear() {
+    key_to_value_.clear();
+    lru_list_.clear();
+    total_bytes_ = 0;
+  }
+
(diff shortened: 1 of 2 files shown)


roshandathathri (approved):
LGTM

Where is/would-be the cache cleared from? I didn't see code for that, so checking.

witchel:
> LGTM
> 
> Where is/would-be the cache cleared from? I didn't see code for that, so checking.

That is in my soon-to-come enterprise PR, in QueryClient.cpp.

It comes with a passive-aggressive comment about how we need to know which properties were written so we don't have to invalidate the entire cache.  Of course, it fails to mention that I spent over an hour today fruitlessly trying to evict a particular entry from my cache and keep the LRU list straight.  So easy an undergrad could do it.  I should find an undergrad.

Finally, I update KAT-1472 to mention that graph-worker and graph-query create query clients with different options and different behavior (e.g., the property cache is disabled from graph-worker).

PR #640, opened 2021-10-21
python: Reset busy wait / burnPower value

ddn0:
Without resetting, the leaked value interferes with manual changes of
the number of active threads, which is quite common in tests. Promote
the debug build assert to an always on assertion as inconsistencies in
the number of threads is dangerous.

Along the way, be aggressive about hiding private state of ThreadPool.

Context: I am running tests with debug and sanitizer on and need fixes like this.

### DIFF
--- a/python/test/benchmarking/bench_python_cpp_algos.py
+++ b/python/test/benchmarking/bench_python_cpp_algos.py
@@ -35,6 +35,15 @@ def time_block(run_name, time_data):
     time_data[run_name] = round(1000 * (timer_algo_end - timer_algo_start))
 
 
+@contextlib.contextmanager
+def busy_wait(enabled):
+    if enabled:
+        katana.set_busy_wait()
+    yield
+    if enabled:
+        katana.set_busy_wait(0)
+
+
 def check_schema(graph: Graph, property_name):
     node_schema: Schema = graph.loaded_node_schema()
     num_node_properties = len(node_schema)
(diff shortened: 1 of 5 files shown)


python/test/benchmarking/bench_python_cpp_algos.py
```diff
@@ -35,6 +35,15 @@ def time_block(run_name, time_data):
     time_data[run_name] = round(1000 * (timer_algo_end - timer_algo_start))
 
 
+@contextlib.contextmanager
+def busy_wait(enabled):
+    if enabled:
+        katana.set_busy_wait()
+    yield
+    if enabled:
+        katana.set_busy_wait(0)
```
arthurp, line 44: I wonder if we should get ride of the other API and only provide a context manager like this. Hmm.

arthurp (approved):
Are the C++ changes just random clean ups? Or do they make some real semantic change?

ddn0:
> Are the C++ changes just random clean ups? Or do they make some real semantic change?

Per the commit message, the private/protected stuff is to minimize the API surface and one the variables exposed is actually used as a weakly atomic flag, so it is not great to expose it.

The only semantic change in the C++ code is to move an assert from LOG_DEBUG to just LOG. I think the cost of proceeding when the invariant is not true is too high. It should only affect code that uses burnPower/set_busy_wait though.

PR #641, opened 2021-10-21
PropertyView: fix IsValid() bug

scober:
An arrow array can omit its null bitmap in two situations: when the
entire array is null and when the entire array is non-null. The current
IsValid() logic for PODPropertyView does not distinguish between these two
cases. It always treats a non-existent bitmap as meaning the entire
array is valid (non-null).

Track the original array's null count in PODPropertyView in order to
distinguish between these two cases.

EDIT: after writing a test, I don't know whether it is possible to construct an arrow array with all null values and no null bitmap. But I think it makes sense to bring our implementation more in line with arrow's anyway.


ddn0:
I don't suppose there is a simple regression test we could add?

scober:
> I don't suppose there is a simple regression test we could add?

We looked into it and it would be NP-hard to add a regression test for this bug. It's a shame, really. Maybe next time.

ddn0:
> We looked into it and it would be NP-hard to add a regression test for this bug. It's a shame, really. Maybe next time.

When something is NP-hard, it has a polynomial time verifier, so it should in principle be simpler to write a test (i.e., check that the result is correct or incorrect) than it is to write the solution to the problem.

Unless you are saying the verification itself is NP-hard. In which case, that means the original problem is strictly harder than NP.

libgalois/test/property-view.cpp
```diff
@@ -0,0 +1,65 @@
+#include "katana/Properties.h"
+#include "katana/Result.h"
+
+namespace {
+const int NUM_ARRAY_ENTRIES = 10;
+
+katana::Result<std::shared_ptr<arrow::Int8Array>>
+AllValid() {
+  arrow::NumericBuilder<arrow::Int8Type> builder;
+  KATANA_LOG_ASSERT(builder.AppendEmptyValues(NUM_ARRAY_ENTRIES).ok());
```
danielmawhirter, line 10: Since you're already returning Result, these could be KATANA_CHECKED

scober:
> > We looked into it and it would be NP-hard to add a regression test for this bug. It's a shame, really. Maybe next time.
> 
> When something is NP-hard, it has a polynomial time verifier, so it should in principle be simpler to write a test (i.e., check that the result is correct or incorrect) than it is to write the solution to the problem.
> 
> Unless you are saying the verification itself is NP-hard. In which case, that means the original problem is strictly harder than NP.

The writing of the test is NP-hard, the complexity of the test itself is irrelevant. We could of course verify that the test is correct in polynomial time. The original problem is likely deserving of a paper if we can find some time for it.

libgalois/test/property-view.cpp
```diff
@@ -0,0 +1,65 @@
+#include "katana/Properties.h"
+#include "katana/Result.h"
+
+namespace {
+const int NUM_ARRAY_ENTRIES = 10;
```
ddn0: nit: `constexpr int`

libgalois/test/property-view.cpp
```diff
@@ -0,0 +1,65 @@
...
+      katana::PODPropertyView<int8_t>::Make<arrow::Int8Type>(*valid_array));
+  for (size_t i = 0; i < NUM_ARRAY_ENTRIES; ++i) {
+    KATANA_LOG_ASSERT(valid_view.IsValid(i));
+  }
+
+  auto null_array = KATANA_CHECKED(NoValid());
+  KATANA_LOG_ASSERT(null_array->length() == NUM_ARRAY_ENTRIES);
+  KATANA_LOG_ASSERT(null_array->null_count() == NUM_ARRAY_ENTRIES);
+  // it may not be possible to obtain an all-nulls array with no bitmap
+  //KATANA_LOG_ASSERT(null_array->null_bitmap_data() == nullptr);
```
ddn0: Daniel may know more but you should be able to just construct an array from arrow::ArrayData directly.

danielmawhirter (approved):
LGTM

libgalois/test/property-view.cpp
scober: That is a good point. But I think the test functions okay as is so I am going to prioritize speedy merging over elegance here.

PR #642, opened 2021-10-21
Entity type id 16 bit support

e-mcginnis:
use EntityTypeID = uint16_t
- remove duplicate EntityType in GraphTopology
- Make RDGTopology use EntityTypeID instead of uint8_t
- Added test for uint16_t EntityTypeID, basic ldbc_003 conversion and full 2^16 EntityTypeIDs 

Refine storage format version testing organization and coverage
- Better label & describe test cases.
- Check generated EntityTypeManager information for ldbc_003 against known-good values

test files renamed:
libgalois/test/property-file-graph-rdg-conversion.cpp -> storage-format-version-v2-entity-type-ids.cpp
libgalois/test/property-graph-optional-topologies.cpp -> storage-format-version-v3-optional-topologies.cpp
libtsuba/test/topology.cpp -> storage-format-version/v3-optional-topologies.cpp 

Add optional topology generation tests
- currently this gets tested as part of the storage-format-version bump tests (storage-format-version-v3-optional-topologies.cpp), but its nice to have unit tests for just the generation as well


tylershunt (approved):
Thanks! This looks really great.

libgalois/include/katana/GraphTopology.h
```diff
@@ -40,7 +39,7 @@ struct KATANA_EXPORT GraphTopologyTypes {
   using AdjIndexVec = NUMAArray<Edge>;
   using EdgeDestVec = NUMAArray<Node>;
   using PropIndexVec = NUMAArray<PropertyIndex>;
-  using EntityTypeVec = NUMAArray<EntityType>;
+  using EntityTypeIDVec = NUMAArray<katana::EntityTypeID>;
```
roshandathathri: `Drop katana::` throughout this file. This is already in the katana namespace.

Also, please don't mix "array" and "vec". Arrays are not resizeable. Let's stick to that convention. So `EntityTypeIDArray`

libgalois/include/katana/GraphTopology.h
```diff
@@ -555,11 +554,12 @@ class KATANA_EXPORT CondensedTypeIDMap : public GraphTopologyTypes {
...
   void invalidate() noexcept { is_valid_ = false; };
 
-  const EntityType* index_to_type_map_data() const noexcept {
+  const katana::EntityTypeID* index_to_type_map_data() const noexcept {
     return &index_to_type_map_[0];
   }
 
-  bool index_to_type_map_matches(size_t size, const EntityType* other) const {
+  bool index_to_type_map_matches(
+      size_t size, const katana::EntityTypeID* other) const {
```
roshandathathri: Why `const Type *?

libtsuba/src/RDGTopology.cpp
```diff
@@ -613,9 +618,9 @@ tsuba::RDGTopology::Make(
     const uint64_t* edge_index_to_property_index_map,
     const uint64_t* node_index_to_property_index_map,
     uint64_t edge_condensed_type_id_map_size,
-    const uint8_t* edge_condensed_type_id_map,
+    const katana::EntityTypeID* edge_condensed_type_id_map,
     uint64_t node_condensed_type_id_map_size,
-    const uint8_t* node_condensed_type_id_map) {
+    const katana::EntityTypeID* node_condensed_type_id_map) {
```
roshandathathri, line 644: Ditto

roshandathathri (approved):
LGTM

libgalois/include/katana/GraphTopology.h
e-mcginnis: @amberhassaan do you have some insight on why `const EntityType* other` was chosen?

libgalois/include/katana/GraphTopology.h
e-mcginnis: renaming these *Vec types to *Array is outside the scope of this PR. No point muddying the waters.  Added a TODO note.

e-mcginnis:
Making `SetOfEntityTypeIDs` dynamically sized vs the fixed 2^16 bit width drastically improved the performance of the EntityTypeManager.Equals() function. It was pushing over 40 seconds on small sets of EntityTypeIDs. Now it takes much less than a second.

libgalois/include/katana/GraphTopology.h
amberhassaan: I think you wrote this method, but it's clear why you're choosing a pointer. Pointer + size here refers to a buffer whose contents are most likely loaded from disk and are being compared against existing contents (that may have been generated). We should replace this with an upcoming `ArrayView` class. Please put a TODO here.

e-mcginnis: hmm, I must have misread the git blame then. My bad. Thanks, will do.

libsupport/include/katana/EntityTypeManager.h
```diff
@@ -19,13 +21,35 @@ namespace katana {
...
+using EntityTypeID = uint16_t;
 static constexpr EntityTypeID kUnknownEntityType = EntityTypeID{0};
 static constexpr EntityTypeID kInvalidEntityType =
     std::numeric_limits<EntityTypeID>::max();
-/// A set of EntityTypeIDs
-using SetOfEntityTypeIDs =
-    std::bitset<std::numeric_limits<EntityTypeID>::max() + 1>;
+
+/// A dynamically sized set of EntityTypeIDs
+class KATANA_EXPORT SetOfEntityTypeIDs : public boost::dynamic_bitset<> {
```
roshandathathri: Why boost `dynamic_bitset` when we have our own `DynamicBitset` that supports concurrent and parallel operations?

libsupport/src/EntityTypeManager.cpp
```diff
@@ -1,4 +1,60 @@
...
+      // new_size = kDefaultSize * ceiling(max_id/kDefaultSize);
+      new_size = kDefaultSetOfEntityTypeIDsSize *
+                        ((max_id + kDefaultSetOfEntityTypeIDsSize - 1) /
+                         kDefaultSetOfEntityTypeIDsSize);
+    }
+
+    // When large
+    // Resizing all of the sets in EntityTypeManager becomes time expensive
+    // Resize in larger chunks to limit the number of resizes we must perform
+    // at the cost of a bit more memory overhead
```
roshandathathri: I'm not sure what this function is trying to achieve. `DynamicBitset` already has the flexibility to grow its capacity so as to amortize frequent `resize`s; similar to `std::vector`, it only doubles it capacity. Why is more complex logic required?

libsupport/src/EntityTypeManager.cpp
```diff
@@ -190,17 +252,64 @@ katana::EntityTypeManager::AddAtomicEntityType(const std::string& name) {
...
+
   entity_type_ids.set(new_entity_type_id);
   entity_type_id_to_atomic_entity_type_ids_.emplace_back(entity_type_ids);
   atomic_entity_type_id_to_entity_type_ids_.emplace_back(entity_type_ids);
 
   return Result<EntityTypeID>(new_entity_type_id);
 }
 
+void
+katana::EntityTypeManager::CheckResizeSetOfEntityTypeIDsMaps(katana::EntityTypeID new_entity_type_id){
```
roshandathathri: The utility/necessity of this function is not clear to me.

roshandathathri (changes requested):

libsupport/include/katana/EntityTypeManager.h
e-mcginnis: EntityTypeManager is in libsupport, DynamicBitset is in libgalois. libtsuba needs access to EntityTypeManager, which is why it was moved out of libgalois in the first place.

roshandathathri: Oops. Okay, thanks. Reorganizing the libraries is orthogonal. But please simplify the resizing. Keep it simple. Just double the capacity each time (min of 256 is fine).

libsupport/src/EntityTypeManager.cpp
e-mcginnis: When an EntityTypeID is added, every bitset we store in 
```
  EntityTypeIDToSetOfEntityTypeIDsMap entity_type_id_to_atomic_entity_type_ids_;
```
and 
```
  EntityTypeIDToSetOfEntityTypeIDsMap atomic_entity_type_id_to_entity_type_ids_;
```
must now have enough bits to support setting that EntityTypeID
so we have to go through and make sure all of the bitsets are big enough

an argument could be made for only resizing the bitsets when we set/check a bit in them, but then all of our bitsets can be different sizes, depending on the most significant bit which is set.

roshandathathri: Thanks. Can we store a bitset-size variable on the type-manager and use that for checking? The invariant should be that all bitsets in the manager are of the same size as that variable. I'm trying to keep the logic simple.

e-mcginnis: this is implemented in the latest commit

libsupport/include/katana/EntityTypeManager.h
e-mcginnis: done

libsupport/src/EntityTypeManager.cpp
```diff
@@ -1,5 +1,40 @@
...
+katana::EntityTypeManager::CalculateSetOfEntityTypeIDsSize(
+    EntityTypeID max_id) {
+  // min number of bits to fit bitset[max_id] is max_id +1
+  size_t min_size = max_id + 1;
+
+  KATANA_LOG_VASSERT(
+      max_id < kInvalidEntityType, "Katana only supports {} entity types",
+      kInvalidEntityType);
+
+  size_t new_size = kDefaultSetOfEntityTypeIDsSize;
```
yan-zaretskiy, line 20: Why do we start with `kDefaultSetOfEntityTypeIDsSize` and not the actual current size?

e-mcginnis: We don't really care about the current size, we really only care about how big we need the set to be.
I could see the argument for passing in the current size to ensure the calculated size is larger than the current size, but at this point that is on the caller. 
This function is really only for calculating how big of a set we should use to hold some arbitrary number of EntityTypeIDs

libsupport/src/EntityTypeManager.cpp
```diff
@@ -190,17 +235,70 @@ katana::EntityTypeManager::AddAtomicEntityType(const std::string& name) {
         kInvalidEntityType);
   }
 
+  KATANA_LOG_DEBUG("adding new_entity_type id = {}", new_entity_type_id);
```
roshandathathri: Please drop the debug prints

libsupport/src/EntityTypeManager.cpp
```diff
@@ -282,20 +380,47 @@ katana::EntityTypeManager::Equals(
     const katana::EntityTypeManager& other) const {
   if (entity_type_id_to_atomic_entity_type_ids_ !=
       other.entity_type_id_to_atomic_entity_type_ids_) {
+    KATANA_LOG_DEBUG(
+        "this.entity_type_id_to_atomic_entity_type_ids_.size() = {}, "
```
roshandathathri, line 482: These debug prints are fine because they are NOT in the common/expected case.

roshandathathri (approved):
LGTM. Thanks!

e-mcginnis:
Will merge once https://github.com/KatanaGraph/katana-enterprise/pull/2391 passes

libsupport/include/katana/EntityTypeManager.h
```diff
@@ -9,6 +11,7 @@
 
 #include <arrow/array/array_primitive.h>
 #include <arrow/table.h>
+#include <boost/dynamic_bitset.hpp>
```
roshandathathri: Drop

libsupport/include/katana/EntityTypeManager.h
```diff
@@ -19,13 +22,76 @@ namespace katana {
...
+  bool all() const {
+    for (auto it : *this) {
+      if (!it) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  // these operators are super inefficient
```
roshandathathri: Even the above function. Please add this as doxygen to all inefficient functions. 

nit: It's better to not use operator overloading for inefficient functions.

roshandathathri (approved):
LGTM

e-mcginnis:
Seems libgluon is having some trouble serializing/deserializing the `std::vector<bool>` version of `SetOfEntityTypeIDs `
working through why that is the case

roshandathathri:
> Seems libgluon is having some trouble serializing/deserializing the `std::vector<bool>` version of `SetOfEntityTypeIDs `
> working through why that is the case

Let's talk on slack. We have specializations to support `vector<bool>`. Its inefficient.

e-mcginnis:
Made a ripoff version of `DynamicBitset`, called `DyanmicBitsetSlow` which removes the hard to move libgalos things, and adds some methods to match what users of `SetOfEntiyTypeIDs` expected from `std::bitset`
Intention is to drop `DynamicBitsetSlow` once we get `DynamicBitset` in a location that libsupport can use. 

Ripped `libgalois/include/katana/AtomicWrapper.h` and `libgalois/include/katana/PODVector.h` out of libgalois, and into libsupport to support `DyanmicBitsetSlow`

Added serializer to enterprise katana for DynamicBitsetSlow
Enterprise PR: https://github.com/KatanaGraph/katana-enterprise/pull/2391

libsupport/include/katana/DynamicBitsetSlow.h
```diff
@@ -0,0 +1,376 @@
...
+
+#include "katana/AtomicWrapper.h"
+#include "katana/PODVector.h"
+#include "katana/config.h"
+
+namespace katana {
+
+//TODO(emcginnis): Remove this class entirely when DynamicBitset is available to libsupport
+/**
+ * Non-Concurrent, Slow dynamically allocated bitset
```
roshandathathri: nit: this is concurrent and thread-safe. Just say "serial implementation" instead.

libsupport/include/katana/DynamicBitsetSlow.h
```diff
@@ -0,0 +1,376 @@
...
+
+  //! this is defined to
+  using tt_is_copyable = int;
+};
+
+//! An empty bitset object; used mainly by InvalidBitsetFn
+extern katana::DynamicBitsetSlow EmptySlowBitset;
+
+//! A structure representing an empty bitset.
+struct KATANA_EXPORT InvalidBitsetSlowFn {
```
roshandathathri: Is this required?

libsupport/include/katana/EntityTypeManager.h
```diff
@@ -420,10 +469,18 @@ class KATANA_EXPORT EntityTypeManager {
...
     // treat it as an entity type that does not have any atomic subtypes
-    auto id = AddNonAtomicEntityType(SetOfEntityTypeIDs());
+    SetOfEntityTypeIDs empty_type_id_set;
+    empty_type_id_set.resize(kDefaultSetOfEntityTypeIDsSize);
+    auto id = AddNonAtomicEntityType(empty_type_id_set);
     KATANA_LOG_ASSERT(id.value() == kUnknownEntityType);
   }
 
+  /// The current size of the SetEntityTypeIDs bitsets
+  size_t SetOfEntityTypeIDsSize_ = kDefaultSetOfEntityTypeIDsSize;
```
roshandathathri, line 482: constants should start `k`. Why isn't this a static constexpr?

libsupport/src/EntityTypeManager.cpp
```diff
@@ -1,5 +1,40 @@
...
+  KATANA_LOG_VASSERT(
+      max_id < kInvalidEntityType, "Katana only supports {} entity types",
+      kInvalidEntityType);
+
+  size_t new_size = kDefaultSetOfEntityTypeIDsSize;
+  // Double the default size of the set until it is bigger than min_size.
+  // There are faster, more clever ways to do this, like using a DeBruijn sequence but
+  // 1) we don't use this function too often
+  // 2) this is easier to understand
+  while (new_size < min_size) {
```
roshandathathri, line 25: This logic is redundant. `PODVector` in `DynamicBitsetSlow` handles this (by using capacity and size similar to std::vector)

roshandathathri (approved):
LGTM

libsupport/include/katana/EntityTypeManager.h
e-mcginnis: its not constant, otherwise our bitsets wouldn't be dynamic!

roshandathathri: Then CamelCase is confusing. Use snake_case_. But just `size_` is fine right?

libsupport/src/EntityTypeManager.cpp
e-mcginnis: While it will reserve the extra space ahead of time, making the call to resize faster, it won't actually resize larger than you ask. 
PodVector.h
```
  void resize(size_t n) {
    reserve(n);
    size_ = n;
  }
```
The slow part that I'm trying to avoid here is calling resize in the first place. 
When we add an EntityTypeID, we have to go through at most 2x (number of EntityTypeID) bitsets and resize each one.
Having to call resize on 130k bitsets would be much slower than resizing early, and avoiding the calls.

libsupport/include/katana/DynamicBitsetSlow.h
e-mcginnis: probably not, I'll remove it.

libsupport/src/EntityTypeManager.cpp
e-mcginnis: That said, maybe the optimization isn't worth the extra code/work?

roshandathathri: Fair point. But setting the size should be reasonably cheap. Yes, I don't think the extra code is worth. However, I'd prefer we merge this soon, so it's fine to keep the code around. If and only if you push further changes, add a TODO/note in the code regarding this.

libtsuba/src/RDGPartHeader.h
```diff
@@ -151,6 +151,7 @@ class KATANA_EXPORT RDGPartHeader {
       const std::string& file);
 
   bool IsEntityTypeIDsOutsideProperties() const;
+  bool IsUint16tEntityTypeIDs() const;
```
roshandathathri, line 154: nit: I'd prefer an enum because it's not clear what false implies. At least add doxygen comment to say that false implies uint8 type-ids.

It's fine to do this in a subsequent PR.

roshandathathri (approved):
LGTM

PR #643, opened 2021-10-22
ToArrayVisitor: fix null bug

scober:
If the DataType passed to ToArrayVisitor is null, it
finishes the empty array builder and returns. It should first append the
correct number of nulls because part of its contract is that the output
array will have the same length as the input vector.


danielmawhirter (approved):
LGTM

PR #644, opened 2021-10-22
Force safe uses of fmt

arthurp:
With: https://github.com/KatanaGraph/katana-enterprise/pull/2277

This forces the use of `FMT_STRING` in all uses of `KATANA_CHECKED`. There is a catch: `KATANA_CHECKED` was effectively overloaded since the function it called with `__VA_ARGS__` is overloaded (`WithContext`). Because of this, safety forced me to add a new macro `KATANA_CHECKED_ERROR_CODE` which handles the overload where the error code is replaced. This macro is not used in the open repo, but is used in enterprise (in libquery).

### DIFF
--- a/libsupport/include/katana/Logging.h
+++ b/libsupport/include/katana/Logging.h
@@ -121,7 +121,7 @@ KATANA_EXPORT void AbortApplication [[noreturn]] ();
     if (!(cond)) {                                                             \
       ::katana::LogLine(                                                       \
           ::katana::LogLevel::Error, __FILE__, __LINE__,                       \
-          "assertion not true: {}", #cond);                                    \
+          FMT_STRING("assertion not true: {}"), #cond);                        \
       ::katana::AbortApplication();                                            \
     }                                                                          \
   } while (0)
--- a/libsupport/include/katana/Logging.h
+++ b/libsupport/include/katana/Logging.h
@@ -121,7 +121,7 @@ KATANA_EXPORT void AbortApplication [[noreturn]] ();
     if (!(cond)) {                                                             \
       ::katana::LogLine(                                                       \
           ::katana::LogLevel::Error, __FILE__, __LINE__,                       \
-          FMT_STRING("assertion not true: {}"), #cond);                        \
+          "assertion not true: {}", #cond);                                    \
       ::katana::AbortApplication();                                            \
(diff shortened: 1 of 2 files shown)


arthurp:
It would be great to use some template magic to apply FMT_STRING to either the first or second argument depending on whether the first argument is a string. But I was not able to figure out how to do that since my C++-fu is not great, and C++ templates are the worst logic programming language ever.

arthurp:
The explicit overload change is shown in: https://github.com/KatanaGraph/katana-enterprise/pull/2277

ddn0:
> It would be great to use some template magic to apply FMT_STRING to either the first or second argument depending on whether the first argument is a string. 

I don't know if templates are the issue. It is more that macros are annoying and we would need to support "overloading" at the macro level (which is not possible). Once we are in the world of C++, we could easily do typed-based things, but that is too late do decide the argument to `FMT_STRING`.

arthurp:
> > It would be great to use some template magic to apply FMT_STRING to either the first or second argument depending on whether the first argument is a string.
> 
> I don't know if templates are the issue. It is more that macros are annoying and we would need to support "overloading" at the macro level (which is not possible). Once we are in the world of C++, we could easily do typed-based things, but that is too late do decide the argument to `FMT_STRING`.

That may be true. But the expansion of FMT_STRING is actually very simple and just a few templates (as far as I can tell). So I thought it might be possible to setup a SFINAE situation, so that if FMT_STRING failed, the other "overload" would be chosen.

ddn0:
> That may be true. But the expansion of FMT_STRING is actually very simple and just a few templates (as far as I can tell). So I thought it might be possible to setup a SFINAE situation, so that if FMT_STRING failed, the other "overload" would be chosen.

My worry is that since FMT_STRING is actually a macro, non-string instantiations are just straight-up wrong C++, e.g. https://godbolt.org/z/Tbnx7aqrd . I think I've spent my allotted C++-puzzler time for the day on this. There may still be a path forward, but also having more explicit macros (as you have in this PR) seems ok to me.

ddn0 (approved):
LGTM but provisional on @danielmawhirter as he did the last round of this interface.

danielmawhirter (approved):
Makes sense to me, looks good

PR #645, opened 2021-10-22
make projected topology copyable

uditagar:
This PR makes the projected topolgy copyable. This is required for distributed graph views.


uditagar:
This is not the right way, so closing the PR. Thanks

PR #648, opened 2021-10-25
Add support for applying boolean lambda over edges.

johnbowen42:


libgalois/include/katana/GraphTopology.h
```diff
@@ -1105,6 +1105,30 @@ class KATANA_EXPORT EdgeTypeAwareTopology
     return std::binary_search(e_range.begin(), e_range.end(), dst, comp);
   }
 
+  /// Search over
+  ///
+  /// @param src source node of the edge
+  /// @param dst destination node of the edge
+  /// @returns true iff the edge exists
+  katana::Result<bool> HasEdgeSatisfyingFunction(Node src, Node dst, std::function<katana::Result<bool>(Edge)> func) const {
```
amberhassaan: I'd like a few important changes here: 
0) Rename to a clearer name. Something like `FindFirstMatchingEdge` or `FindFirstSatisfyingEdge`
1) This method should take a function object as a template parameter. No need for `Result<bool>` either in the function object's return type or method's return type. 
2) Reuse existing methods instead of repeating their implementation. For example, you can directly use `FindAllEdgesSingleType` inside your outer for loop.

amberhassaan: Here's how I'd implement this method:
```
template <typename TestFunc>
bool HasEdgeSatisfyingTest(Node src, Node dst, const TestFunc& func) const noexcept {
  using RetTy = decltype(func(Edge{}));
  static_assert(std::is_same_v(RetTy, bool)); // ensure that return type is bool.

  for (const auto& edge_type: GetDistinctEdgeTypes()) {
    for (auto e: FindAllEdgesSingleType(src, dst, edge_type)) {
       if (func(e)) {
         return true;
       }
    }
  }
  return false;
}
```

libgalois/include/katana/GraphTopology.h
```diff
@@ -1326,6 +1350,27 @@ class KATANA_EXPORT EdgeTypeAwareBiDirTopology
     return Base::out().FindAllEdgesSingleType(src, dst);
   }
 
+  /// Search over
+  ///
+  /// @param src source node of the edge
+  /// @param dst destination node of the edge
+  /// @returns true iff the edge exists
+  katana::Result<bool> HasEdgeSatisfyingFunction(Node src, Node dst, std::function<katana::Result<bool>(Edge)> func) const {
```
amberhassaan: Update after changing the signature of the method above.

amberhassaan (changes requested):
I've suggested some changes that will simplify life. Please point me to your enterprise PR as well, just so I can see how the method is being used. I think you did send me something earlier. If that PR is still valid, I can take a look at that.

libgalois/include/katana/GraphTopology.h
amberhassaan: I just remembered that we can use `std::is_invocable_v` for the `static_assert` above.

libgalois/include/katana/GraphTopology.h
```diff
@@ -1347,6 +1370,18 @@ class KATANA_EXPORT EdgeTypeAwareBiDirTopology
...
+  /// Search over all edges of each type between src and dst until an edge satisfying func is
+  /// found.
+  ///
+  /// @param src source node of the edge
+  /// @param dst destination node of the edge
+  /// @returns true iff the edge exists
+  template <typename TestFunc>
+  bool HasEdgeSatisfyingPredicate(
+      Node src, Node dst, const TestFunc& func) const {
+    return Base::out().HasEdgeSatisfyingPredicate(src, dst, func);
```
amberhassaan, line 1414: why did you take out the degree checks?

libgalois/include/katana/GraphTopology.h
```diff
@@ -1105,6 +1105,29 @@ class KATANA_EXPORT EdgeTypeAwareTopology
...
+  /// found.
+  ///
+  /// @param src source node of the edge
+  /// @param dst destination node of the edge
+  /// @returns true iff an edge satisfying func exists
+  template <typename TestFunc>
+  bool HasEdgeSatisfyingPredicate(
+      Node src, Node dst, const TestFunc& func) const noexcept {
+    using RetTy = decltype(func(Edge{}));
+    static_assert(
```
amberhassaan, line 1119: you can simplify this assert to `static_assert(std::is_invocable_v<bool, TestFunc, Edge>); // order goes like return type, function name, args`

libgalois/include/katana/GraphTopology.h
```diff
@@ -1347,6 +1394,62 @@ class KATANA_EXPORT EdgeTypeAwareBiDirTopology
...
+    return Base::out().HasEdgeSatisfyingPredicate(src, dst, func);
+  }
+
+  /// Search over all out edges of src of each type until an edge satisfying func is
+  /// found.
+  ///
+  /// @param src source node of the edge
+  /// @returns true iff the edge exists
+  template <typename TestFunc>
+  bool HasEdgeSatisfyingPredicate(Node src, const TestFunc& func) const {
```
amberhassaan: I'd rename this to `HasOutEdge...`. Although we don't do that consistently throughout yet. 
Also, degree check in this case doesn't give us much because the loop over edges would be empty any way.

libgalois/include/katana/GraphTopology.h
amberhassaan: Your earlier version had comparison of degrees and ran the function on smaller degree node. I think that's still useful.

amberhassaan (approved):
Looks good.

PR #649, opened 2021-10-26
Adding a fix for projections (Counting total number of new edges)

uditagar:
This PR provides a fix to the projection code. Earlier the number of new edges were not computed if the edge types are not provided (or basically all edges are allowed to be projected irrespective of their types). Verified this fix using the unit-test.

### DIFF
--- a/libgalois/src/GraphTopology.cpp
+++ b/libgalois/src/GraphTopology.cpp
@@ -769,6 +769,7 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
   katana::ParallelSTL::fill(out_indices.begin(), out_indices.end(), Edge{0});
 
   if (edge_types.empty()) {
+    katana::GAccumulator<uint32_t> accum_num_new_edges;
     // set all edges incident to projected nodes
     katana::do_all(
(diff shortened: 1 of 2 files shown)


arthurp (approved):

libgalois/src/GraphTopology.cpp
```diff
@@ -769,6 +769,7 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
   katana::ParallelSTL::fill(out_indices.begin(), out_indices.end(), Edge{0});
 
   if (edge_types.empty()) {
+    katana::GAccumulator<uint32_t> accum_num_new_edges;
```
amberhassaan, line 766: I noticed that you're also counting edges up in lines 687-733. Please remove those as you're going to count again here any way, so it doesn't confuse others as to why you're doing it twice.

amberhassaan (approved):
Left a comment to cleanup the logic a bit.

libgalois/src/GraphTopology.cpp
uditagar: ok

roshandathathri (approved):
LGTM

PR #650, opened 2021-10-26
[KAT-1779] Progress on running tests with NUMBA_DISABLE_JIT

arthurp:
This makes it so the core library works correctly when the Numba JIT is disabled. However, tests will fail because the APIs of some types is different in JIT and non-JIT mode. Which was a big mistake on my part. :-/

The goal of this is to eventually enable better coverage, which is important, but not critical, so I'm going to commit this and stop trying to make it better. We can finish it later.

JIRA: KAT-1779

### DIFF
--- a/python/test/test_loops.py
+++ b/python/test/test_loops.py
@@ -1,6 +1,7 @@
 import weakref
 from functools import partial
 
+import numba
 import numpy as np
 import pytest
 from numba import from_dtype
@@ -71,6 +72,7 @@ def test_for_each_conflict_detection_unsupported():
         for_each(range(10), f, disable_conflict_detection=False)
 
 
+@pytest.mark.xfail(numba.config.DISABLE_JIT, reason="Closures are not type checked in non-jit runs.")
 def test_for_each_wrong_closure():
     @do_all_operator()
     def f(out, i):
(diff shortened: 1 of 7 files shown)


python/test/test_loops.py
```diff
@@ -138,8 +141,9 @@ def f(out, i):
     out = np.zeros(1000, dtype=typ)
     do_all(data, f(out), **modes)
     assert np.allclose(out, np.array(range(1000)))
-    # Check that the operator was actually compiled for the correct type
-    assert list(f.inspect_llvm().keys())[0][1][0] == from_dtype(np.dtype(typ))
+    if not numba.config.DISABLE_JIT:
+        # Check that the operator was actually compiled for the correct type
+        assert list(f.inspect_llvm().keys())[0][1][0] == from_dtype(np.dtype(typ))
```
gliga, line 148: [0][1][0]?

gliga (approved):
Looks good

python/test/test_loops.py
arthurp: I cannot even remember honestly. The result of `inspect_llvm` is a dict from type signatures to LLVM code strings. Oh. I think it's selecting the first overload (`[0]`) then the second argument (`[1]`) and then something else. I'll go figure it out really quick.

PR #651, opened 2021-10-27
tsuba: fix missing header

tylershunt:
When debugging a missing header I found that `PropertyCache` was missing
an arrow inclusion.

Also these fields in PartitionMetadata can be easily inferred from the
policy so mark them as deprecated because I intend to change enterprise
to do that.


witchel (approved):
Sure thing.

PR #653, opened 2021-10-28
Add function to return path for csv datasets

bhargavivadali:
Updated example_data.py to add a function to return path for csv datasets


python/katana/example_data.py
```diff
@@ -66,3 +66,10 @@ def get_input_as_url(rel_path) -> str:
     """
     path = get_input(rel_path).resolve()
     return f"file://{path}"
+
+def get_csv_input() -> Path:
+    """
+    Return the location of ldbc dir
+    """
+    path = Path(__file__).parent.parent.parent.parent.parent.parent.parent.parent / "katana-enterprise/external/csv-datasets"
```
gliga: can you leverage `get_inputs_directory`?  This parent.parent.parent is very fragile (if we start moving files around).
Can you use `get_input` in this case too?  get_input should not be specific to RDGs

arthurp: This assumes got is running in a build directory. This is not a valid assumption for this code. The other code here will actually automatically download the data from S3 if it isn't available and cache it in the users home directory. And this will work even in user installs.

If you want to have this function here, I would recommend you do as @gliga says and put your data in the same inputs blob as the RDGs. If that's not possible or practical, then this function needs to be used to a non-user visible part of the system or at least hidden in an obviously internal package like a `katana.testing.data` package or something.

Whichever you think is appropriate is fine.

arthurp (changes requested):

python/katana/example_data.py
bhargavivadali: CSV data is already present in katana-enterprise/external dir. I was trying to fetch its path. I removed the new function and using get_input and get_inputs_director.

bhargavivadali:
Closing this PR. I'll create another PR with all the changes.

PR #654, opened 2021-10-28
Add a mechanism to bypass system random number generator

aneeshdurg:
This will be used in tests to be able to "replay" a test from CI using the same seed.

JIRA: KAT-2158

### DIFF
--- a/libsupport/include/katana/Random.h
+++ b/libsupport/include/katana/Random.h
@@ -20,17 +20,14 @@ KATANA_EXPORT std::string RandomAlphanumericString(
     uint64_t len, RandGenerator* gen = nullptr);
 
 /// \returns a random number generator seeded with a user provided seed, or
-/// randomness from the platform. The generator is local to the calling thread
-/// so uses of it are thread safe. Useful for things like
-/// `std::uniform_int_distribution`
+/// randomness from the platform.
 KATANA_EXPORT std::pair<katana::RandGenerator, katana::Seed> CreateGenerator(
     const std::optional<Seed>& seed_in);
 
-/// \returns a random number generator obtained from CreateGenerator. This
-/// method will store the generator in a thread local variable so multiple calls
-/// from the same thread will reuse a previously created generator.
-KATANA_EXPORT RandGenerator& GetGenerator(
-    const std::optional<Seed>& seed = std::nullopt);
+/// \returns a random number generator. The generator is local to the calling
+/// thread so uses of it are thread safe. Useful for things like
+/// `std::uniform_int_distribution`
+KATANA_EXPORT RandGenerator& GetGenerator();
 
 /// Fills the iterator range with  a uniform random sequence of numbers from
 /// interval [min_val, max_val]
--- a/libsupport/include/katana/Random.h
+++ b/libsupport/include/katana/Random.h
@@ -2,6 +2,7 @@
 #define KATANA_LIBSUPPORT_KATANA_RANDOM_H_
 
 #include <algorithm>
+#include <optional>
 #include <random>
 #include <string>
 
@@ -10,6 +11,7 @@
 namespace katana {
 
 using RandGenerator = std::mt19937;
+using Seed = std::vector<std::seed_seq::result_type>;
 
 /// Generate a random alphanumeric string of length \param len using
 /// \param gen if provided. If no generator is specified, use the output of
@@ -17,10 +19,18 @@ using RandGenerator = std::mt19937;
 KATANA_EXPORT std::string RandomAlphanumericString(
     uint64_t len, RandGenerator* gen = nullptr);
 
-/// \returns a random number generator seeded with randomness from the platform.
-/// The generator is local to the calling thread so uses of it are thread safe.
-/// Useful for things like `std::uniform_int_distribution`
-KATANA_EXPORT RandGenerator& GetGenerator();
+/// \returns a random number generator seeded with a user provided seed, or
+/// randomness from the platform. The generator is local to the calling thread
+/// so uses of it are thread safe. Useful for things like
+/// `std::uniform_int_distribution`
+KATANA_EXPORT std::pair<katana::RandGenerator, katana::Seed> CreateGenerator(
+    const std::optional<Seed>& seed_in);
+
+/// \returns a random number generator obtained from CreateGenerator. This
+/// method will store the generator in a thread local variable so multiple calls
+/// from the same thread will reuse a previously created generator.
+KATANA_EXPORT RandGenerator& GetGenerator(
+    const std::optional<Seed>& seed = std::nullopt);
 
 /// Fills the iterator range with  a uniform random sequence of numbers from
 /// interval [min_val, max_val]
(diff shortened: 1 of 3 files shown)


libsupport/test/random.cpp
```diff
@@ -48,5 +48,14 @@ main() {
         first_val, val);
   }
 
+  // Test that CreateGenerator returns a seed that can be used to
+  // get deterministic results
+  auto [gen1, seed1] = katana::CreateGenerator(std::nullopt);
+  auto val1 = katana::RandomAlphanumericString(12, &gen1);
+  auto [gen2, seed2] = katana::CreateGenerator(seed1);
+  auto val2 = katana::RandomAlphanumericString(12, &gen2);
```
scober, line 56: Would it be correct to `KATANA_LOG_ASSERT(seed1 == seed2)`?

libsupport/include/katana/Random.h
```diff
@@ -10,17 +11,26 @@
...
+/// randomness from the platform. The generator is local to the calling thread
+/// so uses of it are thread safe. Useful for things like
+/// `std::uniform_int_distribution`
+KATANA_EXPORT std::pair<katana::RandGenerator, katana::Seed> CreateGenerator(
+    const std::optional<Seed>& seed_in);
+
+/// \returns a random number generator obtained from CreateGenerator. This
+/// method will store the generator in a thread local variable so multiple calls
+/// from the same thread will reuse a previously created generator.
+KATANA_EXPORT RandGenerator& GetGenerator(
```
scober: Is this mostly for backwards compatiblity? It seems safe to always return the seed.

libsupport/test/random.cpp
aneeshdurg: good point, yes

libsupport/src/Random.cpp
```diff
@@ -44,41 +44,54 @@ MakeRandomGenerator(S& source) {
...
 
+  auto [gen, unused_seed] = katana::CreateGenerator(seed);
+  kRNG = std::make_unique<RandGenerator>(gen);
+  return *kRNG;
+}
+
+std::pair<katana::RandGenerator, katana::Seed>
+katana::CreateGenerator(const std::optional<Seed>& seed_in) {
+  if (seed_in) {
+    std::seed_seq seq(std::begin(*seed_in), std::end(*seed_in));
```
scober, line 89: I didn't know you could dereference optionals. Very fun.

scober (approved):

libsupport/include/katana/Random.h
aneeshdurg: This method might return the previously created generator, which may have used a different seed. I can just have this method not take in a seed at all if you feel like that might be clearer.

libsupport/src/Random.cpp
```diff
@@ -32,7 +32,7 @@ namespace {
 
 // https://stackoverflow.com/questions/440133
 template <typename S, typename T = katana::RandGenerator>
-T
+std::pair<T, katana::Seed>
```
amberhassaan, line 35: why are we returning a seed?

libsupport/src/Random.cpp
```diff
@@ -44,41 +44,54 @@ MakeRandomGenerator(S& source) {
...
+  auto [gen, unused_seed] = katana::CreateGenerator(seed);
+  kRNG = std::make_unique<RandGenerator>(gen);
+  return *kRNG;
+}
+
+std::pair<katana::RandGenerator, katana::Seed>
+katana::CreateGenerator(const std::optional<Seed>& seed_in) {
+  if (seed_in) {
+    std::seed_seq seq(std::begin(*seed_in), std::end(*seed_in));
+    return std::make_pair(katana::RandGenerator(seq), *seed_in);
```
amberhassaan, line 90: This path does not use a lock to protect the update to thread_local variable `kRNG`. See the lock being used on line 92. 

@ddn0  : I don't understand why the lock isn't around line 79 (instead of line 94). Can you help clarify the reasoning? I believe it should be at line 79 where `kRNG` is being initialized for the first time.

amberhassaan (commented):
Important question regarding atomic initialization needs a resolution.

libsupport/src/Random.cpp
aneeshdurg: We're currently seeing some non-deterministic tests fail in CI, if we return the seed, we could log the seed used by random-oec, and the reproduce the failures

amberhassaan: ok, got it.

libsupport/src/Random.cpp
amberhassaan: Oh, OK, I wasn't looking at the right code. The previous state of the code does protect `kRNG`'s initialization. @aneeshdurg : please make sure that this invariant stays true here as well and move the lock to line 79.

aneeshdurg: my mistake, I moved things around without reading.

libsupport/src/Random.cpp
```diff
@@ -44,43 +44,57 @@ MakeRandomGenerator(S& source) {
...
     return *kRNG;
   }
 
+  auto result = katana::CreateGenerator(seed);
   static std::mutex lock;
-  static RandGenerator system_gen = MakeSystemGenerator();
   std::lock_guard guard(lock);
-
-  kRNG = std::make_unique<RandGenerator>(MakeRandomGenerator(system_gen));
+  kRNG = std::make_unique<RandGenerator>(std::get<0>(result));
```
tylershunt, line 81: Overriding this with a seeded generator has the potential create problems with parts of the code that actually rely on randomness, e.g., coming up with different filenames for properties.  I think now if every host provides the same seed filenames will collide and invariants will break.

Can we have a solution that doesn't assume that everyone wants determinism when some part of the code base seeds this generator?

Fwiw. We made it thread local and persistent because we were exhausting system randomness in CI, there's no reason to restrict the creation of deterministically seeded generators.

tylershunt (changes requested):

libsupport/src/Random.cpp
aneeshdurg: yeah, we can have this method not allow the user to pass in a seed, and only use CreateGenerator for seeded randomness

tylershunt: Thanks!

libsupport/include/katana/Random.h
```diff
@@ -10,16 +11,22 @@
...
 
-/// \returns a random number generator seeded with randomness from the platform.
-/// The generator is local to the calling thread so uses of it are thread safe.
-/// Useful for things like `std::uniform_int_distribution`
+/// \returns a random number generator seeded with a user provided seed, or
+/// randomness from the platform.
+KATANA_EXPORT std::pair<katana::RandGenerator, katana::Seed> CreateGenerator(
+    const std::optional<Seed>& seed_in);
+
+/// \returns a random number generator. The generator is local to the calling
```
tylershunt, line 27: Sorry to be a pain but could you update this comment to restore the statement about `GetGenerator()` always returning a generator seeded with system randomness.

tylershunt (approved):

PR #655, opened 2021-10-28
[KAT-1090] Global requirements file

arthurp:
This does *not* complete the work on this. It starts it. The main missing elements are using the requirements data in more places. Notably, none of the docker containers directly use the data.


arthurp:
I really really tried. But I cannot make this work. There is an unending parade of weird issues. I'll come back to this in M5.

arthurp:
This is now ready for review. The largest change that isn't obvious is replacing Ubuntu 18.04 CI with 20.04 because 20.04 provides the command `apt-get satisfy` that enables installing packages based on a version specification instead of a specific version. I don't think we have any need to do builds on 18.04 any more anyway. But this may require delaying this PR.

It would be possible to build our own simple solver to select appropriate versions for packages so we could use `apt-get install` or even generate a meta package with the requirements we have as dependencies. But both of these would be a lot more effort than just dropping 18.04 support. The problem will be reduced though not eliminated once KAT-830 is fixed.

katana_requirements.yaml
```diff
@@ -0,0 +1,517 @@
+# Katana package requirements
+#
+# For schema documentation see the scripts/requirements tool. Remember that YAML support user defined aliases. They are
```
ddn0: YAML /supports/

ddn0 (approved):
I don't have anything deep to say about this.

Structurally it seems fine to me, and I'm sure we can do improvements if/when we need to

tylershunt (approved):
LGTM

PR #658, opened 2021-11-01
Adds PyTorch to docs

jdent23:


docs/contributing/building.rst
```diff
@@ -153,6 +153,22 @@ or source it into your shell,
...
 
+PyTorch
+=======
+
+To install PyTorch, follow the commands below. The first 2 lines install dependencies. The subsequent steps install PyTorch from source.
+
+.. code-block:: bash
+
+   mamba install -c conda-forge --override-channels astunparse numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing_extensions future six requests dataclasses
+   mamba install -c pytorch magma-cuda110
```
arthurp: Can you create an environment file for this instead of having the list here? Just like `conda_recipe/environment.yml`, but you can put it somewhere more appropriate or rename it.

docs/contributing/building.rst
```diff
@@ -153,6 +153,22 @@ or source it into your shell,
...
+   git clone --recursive --depth 1 --branch v1.10.0 https://github.com/pytorch/pytorch
+   cd pytorch
+   export CMAKE_PREFIX_PATH=${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
```
arthurp: If you activate the conda environment, this should not be needed. WE should eliminate it if we can. You might need to include instructions here to deactivate and reactivate the environment.

docs/contributing/building.rst
```diff
@@ -153,6 +153,22 @@ or source it into your shell,
...
+   python3 setup.py install
+   cd ..
+   rm -r -f pytorch
```
arthurp: Leave this out. People who follow these instructions will know how to clean up their file system.

docs/contributing/building.rst
arthurp: The environment file can also configure the channels as needed.

arthurp (changes requested):

docs/contributing/building.rst
pyrito: My understanding was that we were still not sure about advocating the use of mamba. @arthurp is this still correct?

jdent23: Sure thing, changes reflected

docs/contributing/building.rst
jdent23: Removed, and added a `cd <workspace>` just so people don't clone PyTorch into a commit-able location at least

docs/contributing/building.rst
jdent23: Sorry, I changed all conda calls to mamba in my notes. For this documentation, it's now conda.

conda_recipe/pytorch_deps_environment.yaml
```diff
@@ -0,0 +1,22 @@
+# This file may be used to create an environment using:
+# $ conda env create --name <env> --file conda_recipe/environment.yml
```
ddn0: - `pytorch_deps_environment.yaml`
- and I thought these files were all called `<something>.yml`

docs/contributing/building.rst
```diff
@@ -153,6 +153,20 @@ or source it into your shell,
...
    . $BUILD_DIR/python_env.sh
 
+PyTorch
+=======
+
+To install PyTorch, follow the commands below. The first 2 lines install dependencies. The subsequent steps install PyTorch from source.
+
+.. code-block:: bash
+
+   conda env update --name katana-dev --file $SRC_DIR/external/katana/conda_recipe/pytorch_deps_environment.yml
```
ddn0: Is this path correct? I see ` conda_recipe/pytorch_deps_environment.yaml` but no `external/.../pytorch_deps_environment.yml` (since this is in the open repo).

jdent23: Sure thing, corrected both.

conda_recipe/pytorch_deps_environment.yml
```diff
@@ -0,0 +1,22 @@
+# This file may be used to create an environment using:
+# $ conda env create --name <env> --file conda_recipe/environment.yml
```
ddn0: probably should be `... conda_recipe/pytorch_deps_environment.yml`

jdent23: Missed this, should be correct now.

scripts/install_pytorch.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+mamba env update --file $(dirname "$0")/../conda_recipe/pytorch_deps_environment.yml
```
pyrito: Did you still want to use mamba or conda?

jdent23: If I actually have to run it, I'd prefer mamba. Can change it if conda is still preferred. @arthurp

jdent23:
Adding a note that `conda env update` behaves differently with
```
conda activate <env>
conda update env --file <file>
```
when compared to
```
conda update env --name <env> --file <file>
```
Where the first might fail to solve, the second might have no issue.

pyrito (approved):
This seems pretty clean to me. Thanks for all the cleanup!

conda_recipe/pytorch_environment.yml
```diff
@@ -0,0 +1,23 @@
...
+ - katanagraph::pytorch
+ - katanagraph::pytorch-spline-conv
+ - katanagraph::pytorch-sparse
+ - katanagraph::pytorch-scatter
+ - katanagraph::pytorch-cluster
+ - katanagraph::pyg
+ - dglteam::dgl
+ - pip
+ - pip:
+    - dgllife
```
arthurp, line 23: Why the use of pip here? I would really rather avoid it.

jdent23:
This PR is no longer needed. Closing.

PR #660, opened 2021-11-01
[KAT-2183] Protect potentially racy python operations in tests.

arthurp:
Fixes KAT-2183 and it's several dups.

Should eliminate a few spurious CI failure. I don't think this happened that often though.

### DIFF
--- a/python/test/test_loops.py
+++ b/python/test/test_loops.py
@@ -1,5 +1,6 @@
 import weakref
 from functools import partial
+from threading import Lock
 
 import numba
 import numpy as np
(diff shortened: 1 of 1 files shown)


python/test/test_loops.py
```diff
@@ -34,39 +35,47 @@
...
 
 
+@pytest.fixture
+def lock():
+    return Lock()
+
+
 @pytest.mark.parametrize("modes", simple_modes)
-def test_do_all_python(modes):
+def test_do_all_python(modes, lock):
```
denish44, line 44: From the little that I have read my understanding is that a new "Lock" instance would be passed per call? 

https://docs.pytest.org/en/6.2.x/fixture.html#requesting-fixtures

Not too familiar with pytest but just wanted to make sure each thread is accessing the same lock instance.

denish44: ahh nvm actually i see the method f() called inside this method in parallel

denish44 (approved):

PR #661, opened 2021-11-02
Kat-2129 Update get_input, get_input_as_url, and get_inputs_directory to return correct URIs when testing katana cluster deployed in GCP

bhargavivadali:
Updated functions in example_data.py to return correct path when testing against deployed katana.

Currently, all the tests under katana-enterprise/python/tests/ are run against a locally running katana server. We want to run these tests against a deployed katana cluster in GCP (and later in AWS).

When launching the tests, user can specify where to run the tests in form of a input parameter  "datasource". When the parameter is set to "gcp", the tests use the csv, rdg files in google storage to generate graphs and test on katana cluster in GCP. If the "datasource" parameter is not specified, then tests are run on local server.

Example: 
./build/python_env.sh pytest --datasource gcp
./build/python_env.sh pytest --datasource  local
./build/python_env.sh pytest --datasource aws


python/katana/example_data.py
```diff
@@ -11,8 +11,15 @@
...
 
-def get_inputs_directory(*, invalidate=False) -> Path:
+def get_inputs_directory(*, invalidate=False, rel_path=None) -> Path:
     inputs_dir = None
+    if rel_path == "external/csv-datasets":
+        location = os.environ['HOME']
+        for root, dirs, files in os.walk(location):
+            for d in dirs:
+                if d == "katana-enterprise":
+                    return Path(root) / d
```
arthurp: We really cannot do this. If it needs to be local, then you need to download it somewhere and return that URL, we cannot assume the source is checked out somewhere in the home directory.

python/katana/example_data.py
```diff
@@ -54,15 +61,24 @@ def get_input(rel_path) -> Path:
...
     """
-    path = get_inputs_directory() / rel_path
+    path = get_inputs_directory(rel_path=rel_path) / rel_path
     if path.exists():
         return path
     return get_inputs_directory(invalidate=True) / rel_path
 
 
-def get_input_as_url(rel_path) -> str:
+def get_input_as_url(rel_path, datasource="local") -> str:
```
arthurp: It looks like datasource is an enumeration. Make a real enum. https://docs.python.org/3/library/enum.html#creating-an-enum

python/katana/example_data.py
```diff
@@ -11,8 +11,15 @@
 __all__ = ["get_input", "get_input_as_url"]
 
 
-def get_inputs_directory(*, invalidate=False) -> Path:
+def get_inputs_directory(*, invalidate=False, rel_path=None) -> Path:
     inputs_dir = None
+    if rel_path == "external/csv-datasets":
```
arthurp: I don't like special casing for specific relative paths. If we need to so that we might want to add a "data set name" parameter or something to select a specific data set that needs to be downloaded.

python/katana/example_data.py
```diff
@@ -54,15 +61,24 @@ def get_input(rel_path) -> Path:
...
+def get_input_as_url(rel_path, datasource="local") -> str:
     """
     Similar to get_input, but return the graph as file:// URL.
     """
+    
+    if datasource == "gcp":
+        urls = {"gcp": {"propertygraphs/ldbc_003": "gs://katana-demo-datasets/rdg-datasets/ldbc/v1/ldbc_003_",
+                "external/csv-datasets": "gs://katana-demo-datasets/csv-datasets",
+                "propertygraphs/rmat10": "gs://katana-demo-datasets/rdg-datasets/rmat/rmat10",
+                "propertygraphs/rmat10_symmetric": "gs://katana-demo-datasets/rdg-datasets/rmat/rmat10_symmetric"}
```
arthurp: Is `gs://katana-demo-datasets` public? If it's listed here it needs to be.

arthurp (changes requested):
This is still doing some weird stuff. See my comments.

It would also help me if you included an actual description of the Pr in the title, so I know what the PR is. I don't know JIRA issue numbers. :-)

python/katana/example_data.py
bhargavivadali: I updated the description. Hope it helps.
These csv files are already present in katana-enterprise/external directory and I'm trying to fetch that path. I'm following the current implementation seen in katana-enterprise/python/test/remote_test.py and other files:

dataset_dir = pathlib.Path(__file__).parent.parent.parent / "external/csv-datasets"

python/katana/example_data.py
bhargavivadali: The tests will be run against a katana cluster deployed in cloud and these csv/rdg datasets are used for generating graphs and queries.

python/katana/example_data.py
gliga: Note that the plan is not to download data, but to use data available on gs

python/katana/example_data.py
```diff
@@ -54,15 +61,27 @@ def get_input(rel_path) -> Path:
...
     return get_inputs_directory(invalidate=True) / rel_path
 
 
-def get_input_as_url(rel_path) -> str:
+def get_input_as_url(rel_path, datasource="local") -> str:
     """
     Similar to get_input, but return the graph as file:// URL.
     """
+    
+    if datasource == "gcp":
```
gliga: I would move this to enterprise and pass it as an argument or some other magical Python mechanism

python/katana/example_data.py
gliga: since there is mapping for the cloud storage, maybe simply create a mapping for the local dirs and then pick one of the two

python/katana/example_data.py
bhargavivadali: Done.

python/katana/example_data.py
bhargavivadali: Moved them to enterprise.

python/katana/example_data.py
```diff
@@ -11,8 +11,15 @@
...
 
-def get_inputs_directory(*, invalidate=False) -> Path:
+def get_inputs_directory(*, invalidate=False, rel_path=None) -> Path:
     inputs_dir = None
+    if rel_path == "csv-datasets":
+        paths_to_check = list(Path(__file__).parents) + list(Path.cwd().parents)
+        for path in paths_to_check:
+            csv_path = (path / "katana-enterprise" / "external").resolve()
+            if csv_path.exists():
+                return csv_path
```
arthurp, line 21: Why do you need this special case? I thought there was gonna be some way to register data sources like this so special cases are not needed.

bhargavivadali: I moved the datasources needed for running tests against deployed katana to different repo. This is to fetch csv datasets when testing locally.

arthurp: My point is that I don't like having a special case like this. It's already really odd that the open code looks for the enterprise directory. I don't see why this isn't handled in the enterprise repo. I assumed you would setup some way to register "data sets" or something so the enterprise repo could register the `csv-datasets` directory and the open code would not need to be aware of it.

arthurp (approved):
This can be merged if needed, but please do make sure to come back and clean up the special case and enterprise mentions in the code.

PR #663, opened 2021-11-03
Collect read and write set from operations

yige-hu:
See detailed subtasks and progress in [KAT-1688](https://katanagraph.atlassian.net/browse/KAT-1688).

* Read set: from `EnsureProperties`
Done

* Write set: from `RDGCore::UpsertProperties`...
Done
- `UpsertProperties`
TODO:
- `AddProperties`
- `RemovePoperties`


libgalois/include/katana/PropertyGraph.h
```diff
@@ -202,7 +203,7 @@ class KATANA_EXPORT PropertyGraph {
       return (g->*add_properties_fn)(props);
     }
 
-    Result<void> UpsertProperties(
+    katana::Result<std::vector<katana::fbs::RDGAccessT>> UpsertProperties(
```
arthurp: I would strongly advise against using flatbuffer types in the API. The reason is that if we want to change storage format we would be forced to change the API too. I think there should be a pure C++ wrapper that can be updated to work with a different backing data store if we change storage formats and stop using flatbuffer.

This applies throughout.

python/katana/cpp/libgalois/graphs/Graph.pxd
```diff
@@ -20,6 +20,9 @@ cdef extern from "tsuba/RDG.h" namespace "tsuba" nogil:
         optional[vector[string]] node_properties
         optional[vector[string]] edge_properties
 
+cdef extern from "rdg_generated.h" namespace "katana::fbs" nogil:
+    cdef struct RDGAccessT:
+        pass
```
arthurp: Add a TODO to add fields here and expose them to python as a named tuple. There isn't much value in having these if we cannot use them from Python.

fbs/rdg.fbs
```diff
@@ -0,0 +1,145 @@
...
+  rdg_version : uint64;
+  metadata_version : uint64;
+
+  mode : RDGAccessMode;
+
+  type : RDGAccessType;
+  // property_name is empty unless type is NamedNodeProperty or
+  // NamedEdgeProperty.
+  property_name : string;
+}
```
arthurp: I don't like the name "access" here. What about "RDGAccessEvent" or something like that. I initially assumed these were about controlling access or permissions like ACLs.

libgalois/include/katana/PropertyGraph.h
arthurp: I would really rather hit the conflict detection machinery. I would prefer passing a mutable transaction object into the function that is used to log the accesses. Returning something here makes it very easy to just throw away the results by accident and not add them to the actual read/write set causing transaction soundness bugs.

python/katana/local/_graph.pyx
```diff
@@ -237,7 +243,7 @@ cdef class GraphBase:
             length ``self.num_nodes()``. (Optional)
         :param kwargs: Properties to add. The values must be arrays or sequences of length ``self.num_nodes()``. (Optional)
         """
-        handle_result_void(self.underlying_property_graph().UpsertNodeProperties(GraphBase._convert_table(table, kwargs)))
+        handle_result_RDGAccessT(self.underlying_property_graph().UpsertNodeProperties(GraphBase._convert_table(table, kwargs)))
```
arthurp: If we are discarding the access object, doesn't that mean we will be making any transaction generated from Python unsafe?

fbs/rdg.fbs
ddn0: If you think "access" is confusing, I don't think "access event" is any better.

Originally, these were called "locations" but no one liked that name either: https://github.com/KatanaGraph/katana-enterprise/pull/2142#discussion_r723373632

And access information could be used to control access to properties.

libgalois/include/katana/PropertyGraph.h
ddn0: @yige-hu can you make a `tsuba::Access` class to hold the information for `katana::fbs::RDGAccessT`? If you do that, you don't even need to put the flatbuffer definitions in this repo.

> I would prefer passing a mutable transaction object into the function that is used to log the accesses. Returning something here makes it very easy to just throw away the results by accident

It would be good to make sure the access sets are updated consistently, but I don't see how `void fn(txn_log)` would be any more safe or hard to misuse than `access fn()`.

ddn0: We might prefer `void fn(txn_context)` not because it is safer but because it is easier to extend without changing the callees. For instance if we wanted to add more information about a transaction, we can have the leaf functions update the `txn_context` and all the intervening functions don't have to change their API.

arthurp: The reason I say it's safer is that it's not hard to store access information into the wrong log or not store it at all. And those would create silent unhandled transaction conflicts that could corrupt data. Unless I misunderstand.

ddn0: Aside: with `[[nodiscard]]` you can force users to at least do something with return values, while no one is expected to do anything with their parameters.

I think having a "transaction context" would be good for other reasons (easier to iterate the API). Having an explicit transaction context is a pretty big (initial) change to the API though. The alternative would be some thread-local variable, but that seems worse.

Thoughts?

fbs/rdg.fbs
```diff
@@ -0,0 +1,145 @@
+namespace katana.fbs;
+
+enum RDGAccessType : byte {
```
witchel: I know we are bikeshedding on terms, but it does seem like an important one.  Here are some ideas.

```
RDGAccessScope
RDGAccessGranularity
RDGReadWriteScope
RDGReadWriteLocation
```

libgalois/include/katana/PropertyGraph.h
arthurp: I wasn't implying you would have a parameter you have to do something with. I was saying you would have a transaction context of some kind that the transaction engine would be aware of and would look at when it performed commits or abort points. The normal OO approach to this is to have the transaction object have the mutating operations on it, or pass the transaction object to the mutating operations.

`[[nodiscard]]` helps, for sure. But there are two issues: it doesn't exist in python, and it's still easy to store it in a variable and then drop it, or fail to merge the different access sets together correctly or what not. I think in python we should have an implicit context using `with`. It's conventional in Python. We could implement this based on returned access sets, but it would be much easier to implement using a context object.

arthurp: This is clearly pretty complex, so we can definitely push this off to another PR. Though I don't love planning to have breaking API changes.

fbs/rdg.fbs
witchel: Good terms here are maddeningly difficult.  I guess one dimension is whether we prefer `Access `to `ReadWrite`.  I am sympathetic to Arthur's confusion as to the subject and object of RDGAccess.  I think `RDGDataReadWritten` is too ridiculous.

libgalois/include/katana/PropertyGraph.h
witchel: The standard is to have a transaction context object (`void fn(txn_context)`), which gets rid of the flatbuffer type from the API.
My first instinct would be to insist on that as early as possible, but I'm not sure how that plays with the alpha deadline.
Yige, how long would it take you to add a transaction object?

witchel (commented):
I think we need a transaction context class for this change.

libgalois/include/katana/PropertyGraph.h
yige-hu: @arthurp @ddn0 I agree with the idea of a transaction context while I was in the middle of coding this PR. I still sent out the PR to collect opinions. I also considered enforcing all related classes to inherit a transactional object base class. But that might be less intuitive compared to passing a transaction context.

@witchel I think it should take around 2-3 days to modify the code changes covered by this PR.

I would support modifying this PR instead of doing it in other PR, so that I don't need to merge all the API changes twice.

fbs/rdg.fbs
yige-hu: Or maybe `RDGrwset`, `RdgRWSet`?

ddn0: Another option: `AccessSet`

witchel: I like `RDGAccessSet` and `RDGRWSet`, but given how we have to capitalize the latter, it doesn't scan well.

ddn0: RDGAccessSet seems viable.

yige-hu:
I tried to use rebase to remove my previous commits and add new ones, but it seems it messed up the git log. I'll try to fix it later. If I'm not able to fix this issue, I'll start a new PR.

arthurp (changes requested):
I think your git history is messed up, a bunch of changes from master are appearing in the PR. Can you rebase or merge this with master to correct the branch point and diff?

yige-hu:
> I think your git history is messed up, a bunch of changes from master are appearing in the PR. Can you rebase or merge this with master to correct the branch point and diff?

@arthurp Yea, I noticed this last night. I think `git rebase` messed up after I dropped my earlier commits. It's now fixed by dropping the rebased upstream commits and merge from upstream again. And it's ready for review, thanks!

libgalois/include/katana/PropertyGraph.h
```diff
@@ -170,7 +170,7 @@ class KATANA_EXPORT PropertyGraph {
     Result<void> (PropertyGraph::*add_properties_fn)(
         const std::shared_ptr<arrow::Table>& props);
     Result<void> (PropertyGraph::*upsert_properties_fn)(
-        const std::shared_ptr<arrow::Table>& props);
+        const std::shared_ptr<arrow::Table>& props, tsuba::TxnContext& ctx);
```
witchel: We prefer C pointers to mutable references.  Is there a compelling case for a mutable reference here?

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,70 @@
+#ifndef KATANA_LIBTSUBA_TSUBA_TXNCONTEXT_H_
+#define KATANA_LIBTSUBA_TSUBA_TXNCONTEXT_H_
+
+#include <string>
```
witchel: Is this include necessary?

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,70 @@
...
+  void SetAllPropertiesRead() { all_properties_read = true; }
+
+  void SetAllPropertiesWrite() { all_properties_write = true; }
+
+  void SetTopologyRead() { topology_read = true; }
+
+  void SetTopologyWrite() { topology_write = true; }
+
+private:
+  std::vector<std::string> node_properties_read{};
```
witchel: You only need the bracket initializers for primitive types.

libtsuba/src/RDGCore.cpp
```diff
@@ -98,7 +100,11 @@ AddProperties(
           tsuba::ErrorCode::Exists, "column names are not distinct");
     }
   }
-  return UpsertProperties(props, to_update, prop_state);
+  auto written_prop_names =
+      KATANA_CHECKED(UpsertProperties(props, to_update, prop_state));
+
+  // TODO (yige): return write set to TxnContext
```
witchel, line 107: Why leave this as a `TODO`?

witchel (commented):
I'd like to see the mutable reference turned into a C pointer and the other nits addressed.

Is the transactional context required?  If so, we should assert that the C pointer is not null.

python/katana/local/_graph.pyx
```diff
@@ -229,15 +229,16 @@ cdef class GraphBase:
         """
         handle_result_void(self.underlying_property_graph().AddNodeProperties(GraphBase._convert_table(table, kwargs)))
 
-    def upsert_node_property(self, table=None, **kwargs):
+    def upsert_node_property(self, PyTxnContext ctx, table=None, **kwargs):
```
arthurp: Is there a safe default value for `ctx`? We should support non-transactional calls for simple uses.

python/katana/local/_graph.pxd
```diff
@@ -44,3 +44,6 @@ cdef class Graph(GraphBase):
 
     @staticmethod
     cdef Graph make(shared_ptr[_PropertyGraph] u)
+
+cdef class PyTxnContext:
```
arthurp: This should be called `TxnContext`. It's obviously python.

Also, how is this expected to be created and used? It's not clear how transactions are really created with this. Is there some `StartTransaction` operation?

python/katana/local/_graph.pyx
arthurp: In the long run I think we should track the transaction in a thread local at the python level. It's much more natural for the programmer to write `with transaction(): ...` than to have to explicitly manage the context object. This shouldn't happen in the PR. But we should make sure we have a reasonable path to it.

python/katana/local/_graph.pxd
arthurp: To avoid the name conflict you can import `TxnContext as CTxnContext` above.

arthurp:
Oh another question: Why do only the upserts take a transaction context? Is it because they are non-atomic? However if we want to have multiple inserts or updates we want to have atomic we still need to be able to include them in transactions. Am I missing something?

libtsuba/src/RDGCore.cpp
yige-hu: I'll add it later in this PR if people think the current way of doing it looks good. Or it can be pushed in a follow-up PR.

yige-hu: @witchel I'll ei

> Oh another question: Why do only the upserts take a transaction context? Is it because they are non-atomic? However if we want to have multiple inserts or updates we want to have atomic we still need to be able to include them in transactions. Am I missing something?

@arthurp This will go in in the following commits or a followup PR if people think the current way of handling `RDGCore::UpsertProperties` looks fine. It should be very easy since a lot of presumable changes overlap with the current code.

yige-hu:
> Oh another question: Why do only the upserts take a transaction context? Is it because they are non-atomic? However if we want to have multiple inserts or updates we want to have atomic we still need to be able to include them in transactions. Am I missing something?

@arthurp  This will go in in the following commits or a followup PR if people think the current way of handling RDGCore::UpsertProperties looks fine. It should be very easy since a lot of presumable changes overlap with the current code.

python/katana/local/_graph.pxd
yige-hu: There is a set of python APIs for transactions, but was implemented and tested long time ago and is probably broken. I'll take a look at it, and see whether TxnContext can be fit into it, but in another PR.

arthurp: I think the API you are thinking of is in `katana.remote`. And it's not broken. It's tested in CI.

Please at least rename the class. And we need defaults so things keep working in the non-transactional mode until this stuff is completed.

python/katana/local/_graph.pyx
yige-hu: The `TxnContext` right now only stores read and write set information. The ctx passed into the entry point contains empty read and write sets. If the users does not need transactions, they can simply ignore the content of ctx. But I think we assume single operations outside transactions to be single operation transactions as well. 

I agree on the thread local python transaction. But do you think it'll affect c++ implementation?

arthurp: I don't think it will affect C++. Python can keep the object around as needed without needing to involve C++. As long as C++ doesn't make assumptions about the threading model.

libgalois/include/katana/PropertyGraph.h
yige-hu: Hmm... I saw other contexts mostly all uses references. e.g. `ThreadContext` is also mutable and uses a mutable reference. So I thought it's a convention.

After a second thought, I think pointer may make more sense. So I'm going to change all of them into pointers if you agree on this?

yige-hu: I read that some people tend to use references unless NULL value of the pointer is allowed. "Since references can’t be NULL, they are safer to use." I think we don't allow a NULL pointer to `TxnContext` in our case, and operations outside of transactions are also treated as single operation transactions, hence it's not necessary to do NULL checking on the potential `TxnContext` pointer. Maybe pass by reference is better?

witchel: @ddn0, language feature consult please.

yige-hu: @ddn0 any opinion on this? Thanks!

ddn0: The traditional arguments are:
1. As Yige brings up, references are better than pointers because they express non-nullablity
2. Pointers are better than references because callers explicitly opt-in to mutation by using `&`

I'm pretty sure our codebase is not consistent about this either, but our style guide does prefer (2) over (1): https://docs.k9h.dev/0.1.101.dev+51.303.ac3e9a.4018d6/contributing/code-style.html#code-style

And we do have clang-tidy rules that warn on mutable references.

Since one cannot do both (1) and (2), conventions make up the difference.

If you are in the world of (1), you have the expectation that callers will pay attention to the signature of functions if they care about mutations of parameters ("people should read the docs!").

If you are in the world of (2), you have the expectation that either (a) a function accepting a null pointer is documented ("people should read the docs!") or objects of a certain type always exist ("don't be a monster").

I'd suggest the latter for us. That is, even if there is no transaction, we should have a dummy transaction context and all callees can assume that it is safe to dereference the transaction context.

witchel: Thanks very much.
When the rules and logic conflict, it is tough to know what to do.
I am a fan of C-style pointers for mutable arguments.
And `KATANA_LOG_DEBUG_ASSERT(txn != nullptr);` provides a zero-cost documentation to programmers that you expect a valid transaction context.

yige-hu:
@witchel @ddn0 I changed from passing by reference to passing by pointer. Please take a look. Also, I'm only added null checking to the functions dereferencing `TxnContext*`, but not those functions taking `TxnContext*` as a parameter without dereferencing it. I think it's enough, right?

I think I'll put changes on `AddNodeProperty` and `RemoveNodeProperty` in another PR, since changes should follow and overlap with this PR and are much easier.

libgalois/include/katana/PropertyGraph.h
```diff
@@ -170,7 +170,7 @@ class KATANA_EXPORT PropertyGraph {
     Result<void> (PropertyGraph::*add_properties_fn)(
         const std::shared_ptr<arrow::Table>& props);
     Result<void> (PropertyGraph::*upsert_properties_fn)(
-        const std::shared_ptr<arrow::Table>& props);
+        const std::shared_ptr<arrow::Table>& props, tsuba::TxnContext* ctx);
```
witchel: This is really a nit, but I'd prefer `txn ` (or `txc`) to `ctx `given that context is more generic.

python/katana/local/_graph.pyx
yige-hu: I added default value for `TxnContext` and put the `track the transaction in a thread local at the python level` as a Jira task KAT-2568 but don't know how to assign you as reporter. It says reporter cannot be changed.

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,92 @@
...
+
+private:
+  std::vector<std::string> node_properties_read;
+  std::vector<std::string> node_properties_write;
+  std::vector<std::string> edge_properties_read;
+  std::vector<std::string> edge_properties_write;
+  bool all_properties_read{false};
+  bool all_properties_write{false};
+  bool topology_read{false};
+  bool topology_write{false};
```
witchel: Since this is a class, all its members should end in `_`

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,92 @@
...
+
+  bool GetAllPropertiesWrite() const { return all_properties_write; }
+
+  bool GetTopologyRead() const { return topology_read; }
+
+  bool GetTopologyWrite() const { return all_properties_write; }
+
+private:
+  std::vector<std::string> node_properties_read;
+  std::vector<std::string> node_properties_write;
```
witchel: Should these be sets instead of vectors?  Otherwise, what are you going to do about duplicates?

python/katana/local/_graph.pxd
yige-hu: Set the "integrating transaction begin/commit/abort with `katana.remote`" a future task in Jira KAT-2570

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  void InsertEdgePropertyWrite(std::vector<std::string> names) {
+    edge_properties_write.insert(
+        edge_properties_write.end(), names.begin(), names.end());
+  }
+
+  void SetAllPropertiesRead() { all_properties_read = true; }
+
+  void SetAllPropertiesWrite() { all_properties_write = true; }
```
witchel: This style of setter is not very logical.  Who should call it and why?  It looks more like a friend interface.  (This is true of the SetTopologyRead/Write functions too.

libtsuba/src/RDGCore.cpp
witchel: This PR is dragging on, but this seems like a small amount of additional work, so why not throw it in?

python/katana/local/_graph.pyx
```diff
@@ -229,15 +229,16 @@ cdef class GraphBase:
         """
         handle_result_void(self.underlying_property_graph().AddNodeProperties(GraphBase._convert_table(table, kwargs)))
 
-    def upsert_node_property(self, table=None, **kwargs):
+    def upsert_node_property(self, TxnContext ctx=TxnContext(), table=None, **kwargs):
```
arthurp: The transaction argument needs to be after table. Otherwise it's not longer possible to call `upsert_node_property({"prop": data})` which was one of the expected use cases.

python/test/test_property_graph.py
```diff
@@ -126,7 +126,8 @@ def test_add_node_property_dataframe(graph):
 def test_upsert_node_property(graph):
     prop = graph.loaded_node_schema().names[0]
     t = pyarrow.table({prop: range(graph.num_nodes())})
-    graph.upsert_node_property(t)
+    ctx = TxnContext()
+    graph.upsert_node_property(ctx, t)
```
arthurp: NIT: Why explicitly use a ctx here? You don't use it for anything. So it's the same as the default.

python/katana/local/_graph.pyx
arthurp: This needs to be done for all the functions you changed in this file.

arthurp (dismissed):

python/katana/local/_graph.pyx
arthurp: Oh a better example of why this change is required is `upsert_node_property(user_dataframe)`.

ddn0: Isn't there some style in Python where the default argument values should be `None` or pure values as otherwise `ctx` in `ctx=TxnContext()` will be shared for all invocations and a mutation to `ctx` will leak to other invocations?

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,92 @@
+#ifndef KATANA_LIBTSUBA_TSUBA_TXNCONTEXT_H_
+#define KATANA_LIBTSUBA_TSUBA_TXNCONTEXT_H_
+
+namespace tsuba {
+
+class KATANA_EXPORT TxnContext {
+public:
+  TxnContext(){};
```
ddn0: If you aren't going to implement them differently than the compiler would, you should just omit the default constructor and destructor

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,92 @@
...
+  bool GetAllPropertiesRead() const { return all_properties_read; }
+
+  bool GetAllPropertiesWrite() const { return all_properties_write; }
+
+  bool GetTopologyRead() const { return topology_read; }
+
+  bool GetTopologyWrite() const { return all_properties_write; }
+
+private:
+  std::vector<std::string> node_properties_read;
```
ddn0: underscores after the name for member variables.

python/katana/local/_graph.pyx
arthurp: Oh yes!! Good catch. I forgot that. You must have None as the default and then you can use `ctx or TxnContext()` inside to create a new context iff the argument isn't already a real context.

libtsuba/src/RDGCore.cpp
```diff
@@ -164,15 +170,27 @@ RDGCore::AddEdgeProperties(const std::shared_ptr<arrow::Table>& props) {
...
-RDGCore::UpsertNodeProperties(const std::shared_ptr<arrow::Table>& props) {
-  return UpsertProperties(
-      props, &node_properties_, &part_header_.node_prop_info_list());
+RDGCore::UpsertNodeProperties(
+    const std::shared_ptr<arrow::Table>& props, tsuba::TxnContext* ctx) {
+  KATANA_LOG_DEBUG_ASSERT(ctx != nullptr);
+  auto written_prop_names = KATANA_CHECKED(UpsertProperties(
+      props, &node_properties_, &part_header_.node_prop_info_list()));
+  // store write properties into transaction context
+  ctx->InsertNodePropertyWrite(std::move(written_prop_names));
```
witchel: I'm not sure everyone agrees, but I feel like the parameter should be `std::vector<std::string>&&` if you expect clients to `std::move` into it.  But Tyler does this std::move as an optimization also.

witchel (commented):
It is great to see this.

libtsuba/include/tsuba/TxnContext.h
yige-hu: Again in the flatbuffer `RDGAccess`, there are types of access like `AllProperties` and `Topology`, meaning that some operation could possibly change all the properties (logically, not by inserting one by one. Right now I'm not sure where it can happen but it's defined in `RDGAccess` so correspondent things needs to be added here) and topology (I saw some topology changing code, but haven't investigated in details). With those codes, `SetAllProperties...()` and `SetAllTopology()` needs to be called, same as `SetNodeProperty...()`.

libtsuba/src/RDGCore.cpp
yige-hu: I think they're the same in terms of efficiency. The smart pointer passes the ownership, which is equivalent to passing callee the raw pointer.

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -0,0 +1,101 @@
...
+    edge_properties_write_.insert(names.begin(), names.end());
+  }
+
+  void SetAllPropertiesRead() { all_properties_read_ = true; }
+
+  void SetAllPropertiesWrite() { all_properties_write_ = true; }
+
+  void SetTopologyRead() { topology_read_ = true; }
+
+  void SetTopologyWrite() { topology_write_ = true; }
```
witchel, line 62: Why do you need these setters?

witchel (approved):
Thanks for wrapping this up.

libtsuba/include/tsuba/TxnContext.h
yige-hu: They are supposedly called when a function changes the topology of a graph (a global property). Also, the boolean can only be set from false to true but not reversely.

PR #664, opened 2021-11-03
Safer API for slices in ParquetReader

darabos:
From https://github.com/KatanaGraph/katana-enterprise/pull/2265#discussion_r741829752.

Taking the slice in `Make()` and then _maybe_ respecting it depending on if we call `ReadTable(uri)` or `ReadTable(uri, columns)` is error-prone. I don't really see the use case for reading rows 23 to 45 of multiple files anyway. Better move the slice parameter to the `ReadTable()` calls.

I'm also adding slice support in `ReadTable(uri, columns)` and fixing a typo. (`make_cannonical` &rarr; `make_canonical`)

I've checked by renaming it, and `ReadTable(uri, columns)` is only called from one place. In the enterprise repo. I'll update that caller in https://github.com/KatanaGraph/katana-enterprise/pull/2265. What's the right way to synchronize the change across the two repos?

Thank you!

### DIFF
--- a/libtsuba/include/tsuba/ParquetReader.h
+++ b/libtsuba/include/tsuba/ParquetReader.h
@@ -26,11 +26,7 @@ public:
   struct ReadOpts {
     /// if true (default) make sure canonical types are used and table columns
     /// are not chunked
-    bool make_cannonical{true};
-
-    /// if provided, slice the resulting table so that it only contains
-    /// Slice.length rows starting from Slice.offset
-    std::optional<Slice> slice{std::nullopt};
+    bool make_canonical{true};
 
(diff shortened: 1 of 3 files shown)


tylershunt (approved):
Thanks, I like this interface change (and the spell check :D).

We just need to make sure we update the master branch quickly after this gets merged. @wkyu2kg  can you help with that?

wkyu2kg:
@darabos Let us get this merged. Just rebase both PRs to the HEAD, and have them pass the tests. We can then merge the other PR soon after.

pyrito (approved):
@darabos this looks good to me! Please go ahead and merge this and rebase your other PR so that the submodules match! Thank you!

PR #667, opened 2021-11-03
Shared memory Louvain: Add int64 edge property to the property graph (if edge weight input is either of type int32 or uint32)

uditagar:
This PR adds the functionality to the shared memory louvain app to allow the edge weights to be converted to int64 type, if the input edge weights are of type int32 or uint32. This is required since the edge weights in the coarsened graph can fall into the range of int64.

### DIFF
--- a/libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
+++ b/libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
@@ -496,6 +496,30 @@ public:
 
 template <typename EdgeWeightType>
 static katana::Result<void>
+AddInt64EdgeWeight(
+    katana::PropertyGraph* pg, const std::string& input_edge_weight_property_name,
+        const std::string& edge_weight_property_name) {
+  using EdgeData = std::tuple<katana::PODProperty<EdgeWeightType>, katana::PODProperty<int64_t>>;
+
+  if (auto res = katana::analytics::ConstructEdgeProperties<EdgeData>(
+          pg, {edge_weight_property_name});
+      !res) {
+    return res.error();
+  }
+
+  auto typed_graph =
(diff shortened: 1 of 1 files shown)


libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -494,6 +494,36 @@ struct LouvainClusteringImplementation
...
+          pg, {},
+          {input_edge_weight_property_name, edge_weight_property_name})));
+  katana::do_all(
+      katana::iterate(typed_graph.all_edges()),
+      [&](auto e) {
+        typed_graph.template GetEdgeData<katana::PODProperty<int64_t>>(e) =
+            typed_graph
+                .template GetEdgeData<katana::PODProperty<EdgeWeightType>>(e);
+      },
+      katana::steal(), katana::loopname("CopyEdgeWeight"));
```
roshandathathri, line 523: Drop steal()

roshandathathri (approved):
LGTM

uditagar:
I wanted to ask if we need to push this for alpha or not, since it seems that there is still some problem with it with rmat10 test cases.

libgalois/src/analytics/louvain_clustering/louvain_clustering.cpp
```diff
@@ -593,15 +623,32 @@ katana::analytics::LouvainClustering(
         pg, temporary_edge_property.name(), output_property_name, plan);
   }
 
+  TemporaryPropertyGuard temporary_edge_property{pg->EdgeMutablePropertyView()};
```
gurbinder533, line 626: I don't think we should add another temp edge_property to the original graph. Please convert the edge-property to int64 when creating duplicate graph to work on.

gurbinder533 (changes requested):

PR #670, opened 2021-11-06
Add conda channels as command line arguments in testing

jdent23:


test/install_conda_packages_test.py
```diff
@@ -16,7 +16,13 @@ def main():
     parser.add_argument("docker_image", type=str, help="The docker image to use.")
 
     parser.add_argument("--python-package", default="katana-python", help="Name of python conda package to test")
+    parser.add_argument(
+        "--python-package-channels", default="-c katanagraph", help="Channels to use when installing the python package"
+    )
     parser.add_argument("--tools-package", default="katana-tools", help="Name of tools conda package to test")
+    parser.add_argument(
+        "--tools-package-channels", default="-c katanagraph", help="Channels to use when installing the tools package"
+    )
```
arthurp, line 25: I would really prefer these to be list arguments. Including `-c` in each argument is pretty weird.

arthurp (approved):
This is OK. Let's just make sure we only use this for debugging. Let's make sure we don't end up in a situation where people have to use this for anything in the normal flow of work.

jdent23:
Resolved by [this PR](https://github.com/KatanaGraph/katana/pull/671)
Closing.

PR #673, opened 2021-11-08
Disable -u while activating conda

arthurp:
The problem is that some Conda packages have activation scripts that doesn't work when in the "strict" mode we write our scripts in where an unbound variable reference is an error. And since the activation scripts are sourced they are affected by our settings.


arthurp:
I'm scatter shotting for reviewers since it's late on Sun and this needs to be done quickly. @srogatch, I'm thinking our time zone difference may help. Thanks y'all. Should be a pretty quick review.

denish44 (approved):
I work quick 😬 , also my phone was handy with the email alert.

PR #674, opened 2021-11-08
Handle malformed storage_format_version 2 RDGs

e-mcginnis:
Some stored storage_format_version 2 RDGs improperly
consider kUnknownEntityType an AtomicEntityType
Work around this when loading the entity_type_id_dictionary_
and entity_type_id_name_ into their in-memory formats

- add libgalois test that would have caught this
- Add checks to EntityTypeManager constructors
  to ensure kUnknownEntiytType is not considered
  an AtomicEntityType

- Ensure EntiyTypeIDsArrays are written out to disk
  when we go from uint8_t -> uint16_t EntityTypeIDs


roshandathathri (approved):
LGTM

PR #675, opened 2021-11-09
Use correct jupyter root directory.

arthurp:
I was not able to think how to easily test this. It does need to be tested, but it will be more work than I want to spend at this instant. I will post here when I have finished testing locally.

### DIFF
--- a/python/katana/bug/environment.py
+++ b/python/katana/bug/environment.py
@@ -5,6 +5,7 @@ import socket
 import subprocess
 import sys
 import tempfile
+import warnings
 import zipfile
 from pathlib import Path
 from typing import Any, Iterable, Optional, Union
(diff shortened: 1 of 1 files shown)


python/katana/bug/environment.py
```diff
@@ -242,6 +243,15 @@ def capture_environment(filename: Optional[Union[str, Path, Any]] = None):
 
             if get_ipython():
                 # If IPython is importable and actually loaded, display the link.
+                try:
+                    # Path.readlink doesn't exist before Python 3.9, so that cannot be used quite yet.
+                    ipython_root_directory = os.readlink(f"/proc/{os.environ['JPY_PARENT_PID']}/cwd")
+                except OSError:
+                    warnings.warn(
+                        "Could not determine Jupyter root path. Guessing that it is the current working "
+                        "directory. The download link may be incorrect."
```
origandrew, line 252: how about providing some manual way to determine the correct download link here? Like, "copy path of notebook" and use the following templated link:.... or the output of `!pwd` within the notebook

origandrew (approved):
small suggestion

denish44 (approved):

python/katana/bug/environment.py
arthurp: That's a good idea. Hmm. Though I actually think this will only happen if the code is running in ipython, but not Jupyter. So I wouldn't know what instructions to give.

arthurp: I think I'm gonna leave that for another time. I don't know what I would actually add.

PR #678, opened 2021-11-11
Fix and improve edge list graph loading.

arthurp:
* Add support for importing sorted edge lists.
* Fix incorrect import of unsorted edge lists.

Nothing was right. :-/

### DIFF
--- a/python/katana/local/import_data.py
+++ b/python/katana/local/import_data.py
@@ -50,18 +50,50 @@ def from_edge_list_matrix(edges: np.ndarray) -> Graph:
 @numba.njit()
 def _fill_indices(sort_order: np.ndarray, sources: np.ndarray, indices: np.ndarray):
     last_source = 0
-    for source_index, edge_index in enumerate(sort_order):
-        source = sources[source_index]
-        if source != last_source:
-            for j in range(last_source, source):
-                indices[j] = edge_index
-            last_source = source
+    # Code is duplicated because numba cannot perform type based cloning within a function and cannot pass ranges
+    # into a function.
+    if sort_order is None:
+        for edge_index, source in enumerate(sources):
+            if source != last_source:
+                for j in range(last_source, source):
+                    indices[j] = edge_index
+                last_source = source
+    else:
+        for edge_index, source_index in enumerate(sort_order):
+            source = sources[source_index]
+            if source != last_source:
+                for j in range(last_source, source):
+                    indices[j] = edge_index
+                last_source = source
     for j in range(last_source, len(indices)):
         indices[j] = len(sources)
 
 
 def from_edge_list_arrays(
     sources: np.ndarray, destinations: np.ndarray, property_dict: Dict[str, np.ndarray] = None, **properties: np.ndarray
+) -> Graph:
+    """
+    Convert an edge list represented as two parallel arrays into a :py:class:`~katana.local.Graph`.
+
+    This preserves node IDs, but **not** edge IDs.
+    """
+    return _from_edge_list_arrays_impl(sources, destinations, property_dict, edges_sorted=False, properties=properties)
+
+
+def from_sorted_edge_list_arrays(
+    sources: np.ndarray, destinations: np.ndarray, property_dict: Dict[str, np.ndarray] = None, **properties: np.ndarray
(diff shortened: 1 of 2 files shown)


python/katana/local/import_data.py
```diff
@@ -66,6 +75,29 @@ def from_edge_list_arrays(
...
+    """
+    return _from_edge_list_arrays_impl(sources, destinations, property_dict, edges_sorted=False, properties=properties)
+
+
+def from_sorted_edge_list_arrays(
+    sources: np.ndarray, destinations: np.ndarray, property_dict: Dict[str, np.ndarray] = None, **properties: np.ndarray
+) -> Graph:
+    """
+    Convert an **sorted** edge list represented as two parallel arrays into a :py:class:`~katana.local.Graph`. The
+    ``sources`` array must be sorted or the resulting graph will be incorrect.
```
gurbinder533, line 88: Why do we need a separate path for sorted?

python/katana/local/import_data.py
```diff
@@ -50,12 +50,21 @@ def from_edge_list_matrix(edges: np.ndarray) -> Graph:
...
-        source = sources[source_index]
-        if source != last_source:
-            for j in range(last_source, source):
-                indices[j] = edge_index
-            last_source = source
+    # Code is duplicated because numba cannot perform type based cloning within a function and cannot pass ranges
+    # into a function.
+    if sort_order is None:
+        for edge_index, source in enumerate(sources):
+            if source != last_source:
```
gurbinder533, line 57: Is this calculating prefix sum? I don't completely understand how will this give prefix-sum for unsorted sources?

python/test/test_convert_graph.py
```diff
@@ -23,8 +24,8 @@ def test_adjacency_matrix():
     assert list(g.get_edge_property("weight").to_numpy()) == [1, 2, 3]
 
 
-def test_trivial_arrays():
-    g = from_edge_list_arrays(np.array([0, 1, 10]), np.array([1, 2, 0]))
+def test_trivial_arrays_unsorted():
+    g = from_edge_list_arrays(np.array([0, 10, 1]), np.array([1, 0, 2]))
```
gurbinder533, line 28: Shouldn't this also output range for 11 nodes? There will be 11 nodes in the graph, isn't it?

python/katana/local/import_data.py
gurbinder533: For source[0, 10, 1] case, I am not sure how we will get indices[10] = 4? It is coming out to be 3 based on my rough calculations. I could be wrong.

python/katana/local/import_data.py
arthurp: I debated on this too. I decided two functions was best because they have different semantics. One call maintains edge IDs the other does not. So it's not just a performance different, it's a concrete correctness difference. And forcing the users to make a specific choice makes sure they know what they are choosing, at least in theory.

python/katana/local/import_data.py
arthurp: This code is only used in the presorted case. If sort_order is none the sources is already sorted.

python/test/test_convert_graph.py
arthurp: There will be 11. And that's what's (implicitly) asserted by the assert below, while it's also checking that the edge ranges are correct.

I don't see any place where this test implies anything other than 11 nodes.

python/katana/local/import_data.py
arthurp: And to answer your prefix sum question: It's not really prefix sum per se. It's basically prefix sum and counting the out edges for a given node fused into one loop, now that I think of it. Maybe I should split it apart. What do you think?

gurbinder533: You need to compute edges per sources for prefix sum. I think it is fine to keep it in a single function. I was thinking  we can name it "_prefix_sum" but current name is fine as well.

python/test/test_convert_graph.py
gurbinder533: Yes, I missed that. My bad.

gurbinder533 (approved):
LGTM.

PR #680, opened 2021-11-11
Improve support for dev setup on Ubuntu 18.04.

arthurp:
This does not actually make it work fully. There are still issues with installing numba (via pip) on 18.04. I think that never actually worker correctly. The issue is that the LLVM available on 18.04 is too old. We probably never noticed because the people using that platform were doing C++ development and not dealing with Python much. Or had manually upgraded their LLVM.

### DIFF
--- a/katana_requirements.yaml
+++ b/katana_requirements.yaml
@@ -265,7 +265,8 @@ pandoc:
   version: [ 2a, null ]
   labels:
     - conda/dev
-    - pip/dev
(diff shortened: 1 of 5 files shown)


arthurp:
Here is it. We should still phase out 18.04, but this will hopefully soften the blow a little since it at least improve our backward compatibility.

ddn0 (approved):

witchel (approved):
From what I can understand, looks good.
I guess 18.04 is moving into the rear view mirror soon for us.

PR #681, opened 2021-11-12
change line for each file mentioned in the backtrace

chakpongchung:
Feel free to reject this change if there are reasons I am not aware of.

Before the change the backtrace is too long to fit into one screen.
Not even within the screen of a ultrawide monitor with 4k resolution
using standard DPI. 

Before it was all within the same line, which means I need to scrolllllllll to the right to see what is going on:

```
ng/kg/forked/katana-enterprise/libquery/src/OpGraph.cpp:784): backtrace: (/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:224): backtrace: (/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:418): backtrace: (/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:418): backtrace: parameter name endDate is undefined (SymbolResolutionChecker.cpp:355)

```

After the change, they are all in different lines:

```
kg/forked/katana-enterprise/libquery/src/OpGraph.cpp:784): backtrace: 
(/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:224): backtrace: 
(/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:418): backtrace: 
(/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:418): backtrace: parameter name endDate is undefined (SymbolResolutionChecker.cpp:355)
TCK = Undefined Parameter Error
Katana = ParameterError:UndefinedParameter
```

I believe this change is beneficial
for debugging. At least now I dont need to scrolllllllll to the right and back most of the time, if not all.

Wouldn't it be nice to have?

### DIFF
--- a/libsupport/include/katana/Result.h
+++ b/libsupport/include/katana/Result.h
@@ -629,7 +629,7 @@ CheckedExpressionToValue(arrow::Status&&) {
     if (::katana::internal::CheckedExpressionFailed(result_name)) {            \
       return ::katana::internal::CheckedExpressionToError(result_name)         \
           .WithContext(__VA_ARGS__)                                            \
-          .WithContext(FMT_STRING("({}:{})"), __FILE__, __LINE__);             \
+          .WithContext(FMT_STRING("\n({}:{})"), __FILE__, __LINE__);           \
     }                                                                          \
     std::move(                                                                 \
         ::katana::internal::CheckedExpressionToValue(std::move(result_name))); \


libsupport/include/katana/Result.h
```diff
@@ -629,7 +629,7 @@ CheckedExpressionToValue(arrow::Status&&) {
     if (::katana::internal::CheckedExpressionFailed(result_name)) {            \
       return ::katana::internal::CheckedExpressionToError(result_name)         \
           .WithContext(__VA_ARGS__)                                            \
-          .WithContext(FMT_STRING("({}:{})"), __FILE__, __LINE__);             \
+          .WithContext(FMT_STRING("\n({}:{})"), __FILE__, __LINE__);           \
```
ddn0, line 632: If the line number information is about the current error why insert a new line before rather than after?

chakpongchung:
I have not tried 5k2k or 49 inch DQHD. I guess most of us are not using them.

ddn0:
> I have not tried 5k2k or 49 inch DQHD. I guess most of us are not using them.

I don't buy this line of argument.

Typically, we should not consider presentation issues when working with low-level libraries; it should really be the responsibility of parts of the code higher up the call-stack to deal with this (i.e., the part the code that actually prints the error message) rather than Result.h/WithContext which simply appends to a buffer.

If you want to add a newline to separate "frames" logically, that is possible; though, we should really make it clear what a frame is. The most sensible definition is the scope of one `WithContext` call.

In which case, the better solution would be to distinguish the line number context which is just extending the current frame and other `WithContext` calls which actually create new error frames. Ideally, we'd pick a frame separator that cannot be confused for anything a user could write (e.g., NUL) and then on printing, we'd render frames with newlines between them.

libsupport/include/katana/Result.h
chakpongchung: I tried that, does not seem better:
"({}:{})\n"

```
g/forked/katana-enterprise/libquery/src/OpGraph.cpp:784)
: backtrace: (/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:224)
: backtrace: (/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:418)
: backtrace: (/home/cpchung/kg/forked/katana-enterprise/libquery/src/error-checker-passes/SymbolResolutionChecker.cpp:418)
: backtrace: parameter name endDate is undefined (SymbolResolutionChecker.cpp:355)
TCK = Undefined Parameter Error
Katana = ParameterError:UndefinedParameter
```

ddn0: It seems more correct to me as now each line represents a frame and the line number is at the end of the corresponding frame rather than on the following line.

It looks weird because many of the intervening frames have no text associated with them and the initial `:` comes from the fact the message is intended to be rendered on one line.

arthurp: I kinda like the idea of splitting up the trace into multiple lines. But it will require more than changing this I think. Since we should format it more carefully. I think one line might be the best we can do if we handle line information as strings instead of actually having a backtrace data structure like most exception systems.

arthurp:
> In which case, the better solution would be to distinguish the line number context which is just extending the current frame and other `WithContext` calls which actually create new error frames. Ideally, we'd pick a frame separator that cannot be confused for anything a user could write (e.g., NUL) and then on printing, we'd render frames with newlines between them.

I would argue that if we want to go here we should just start storing the track trace as a data structure instead of text, and then format it at print time. That would allow formatting it differently depending on the context which could be useful. Not a huge gain for that change, but it definitely seems cleaner to me.

chakpongchung:
consensus reached. closing this PR

PR #682, opened 2021-11-13
Write python_env.sh debugging output to stderr

ddn0:
Keep python_env.sh additions out of stdout for cases where someone may
want to parse the output of a python command.


arthurp (approved):
All good changes. You could even remove the print in when there are arguments. I really only have it there so people can copy past it to get the environment into some other tool (I do it into my IDE). So only printing when nothing is being launched would be fine.

PR #683, opened 2021-11-13
[KAT-1987] Install cudatoolkit-dev instead of cudatoolkit package

srogatch:


srogatch:
Apparently this is not the actual issue. It seems that some other package has `cudatoolkit` as a dependency, while we need `cudatoolkit-dev` installed.

PR #684, opened 2021-11-14
Remove assert from Undirected View

uditagar:
This PR removes an assertion from UndirectedTopology class that checks if outgoing number of edges is greater than 0. This check is problematic for clustering on very small graphs since we can arise at a situation where the number of edges in a merged/coarsened graph is zero on one of the hosts.


srogatch (approved):

roshandathathri (commented):
LGTM

PR #686, opened 2021-11-15
Fix leiden application

gurbinder533:


libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -952,25 +958,25 @@ struct ClusteringImplementationBase {
...
-        katana::iterate(*graph),
-        [&](GNode n) { graph->template GetData<CurrentSubCommunityID>(n) = n; },
-        katana::steal());
+    katana::do_all(katana::iterate(*graph), [&](GNode n) {
+      graph->template GetData<CurrentSubCommunityID>(n) = n;
+    });
 
     // populate nodes into communities
-    std::vector<std::vector<GNode>> cluster_bags(2 * graph->size() + 1);
+    std::vector<std::vector<GNode>> cluster_bags(graph->size());
```
uditagar, line 963: I thought twice size was needed to allow the move to self clusters.

gurbinder533: This is just to separate out nodes to their respective communities. Since, community ids cannot be more than graph size, we don't need 2xsize.

uditagar (approved):
looks fine. I will work on it to match scanpy after it is merged.

PR #687, opened 2021-11-16
tsuba: modified most functions to use KATANA_CHECKED

pyrito:
Now that I have some time, I decided to do some code cleanup in our storage code to make use of `KATANA_CHECKED` wherever I saw it was most appropriate.

### DIFF
--- a/libtsuba/src/AddProperties.cpp
+++ b/libtsuba/src/AddProperties.cpp
@@ -20,11 +20,11 @@ katana::Result<std::shared_ptr<arrow::Table>>
 DoLoadProperties(
     const std::string& expected_name, const katana::Uri& file_path,
     std::optional<tsuba::ParquetReader::Slice> slice = std::nullopt) {
-  auto reader =
-      KATANA_CHECKED_CONTEXT(tsuba::ParquetReader::Make(), "loading property");
+  std::unique_ptr<tsuba::ParquetReader> reader =
+      KATANA_CHECKED(tsuba::ParquetReader::Make());
 
-  auto out = KATANA_CHECKED_CONTEXT(
-      reader->ReadTable(file_path, slice), "loading property");
+  std::shared_ptr<arrow::Table> out =
+      KATANA_CHECKED(reader->ReadTable(file_path, slice));
 
   std::shared_ptr<arrow::Schema> schema = out->schema();
   if (schema->num_fields() != 1) {
@@ -179,9 +179,10 @@ tsuba::AddPropertySlice(
             std::launch::async,
             [path, prop, begin,
              size]() -> katana::CopyableResult<std::shared_ptr<arrow::Table>> {
-              auto load_result = KATANA_CHECKED_CONTEXT(
-                  LoadPropertySlice(prop->name(), path, begin, size),
-                  "error loading {}", path);
+              std::shared_ptr<arrow::Table> load_result =
+                  KATANA_CHECKED_CONTEXT(
+                      LoadPropertySlice(prop->name(), path, begin, size),
+                      "error loading {}", path);
               return load_result;
             });
     auto on_complete = [add_fn,
@@ -202,8 +203,8 @@ tsuba::AddPropertySlice(
           std::move(future), path.string(), on_complete);
       continue;
     }
-    auto read_fut = KATANA_CHECKED(future.get());
-    KATANA_CHECKED(on_complete(read_fut));
+    std::shared_ptr<arrow::Table> props = KATANA_CHECKED(future.get());
+    KATANA_CHECKED(on_complete(props));
   }
 
   return katana::ResultSuccess();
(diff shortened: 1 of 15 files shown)


libtsuba/src/AddProperties.cpp
```diff
@@ -20,18 +20,11 @@ katana::Result<std::shared_ptr<arrow::Table>>
...
-    return reader_res.error().WithContext("loading property");
-  }
-  std::unique_ptr<tsuba::ParquetReader> reader = std::move(reader_res.value());
-
-  auto out_res = reader->ReadTable(file_path, slice);
-  if (!out_res) {
-    return out_res.error().WithContext("loading property");
-  }
+  auto reader =
+      KATANA_CHECKED_CONTEXT(tsuba::ParquetReader::Make(), "loading property");
```
tylershunt: We might take this opportunity to reduce stuttering in errors. I think a good rule of thumb is not to report context that the caller knows. In this case I don't think "loading property" is adding anything that the caller of this doesn't know.

libtsuba/src/AddProperties.cpp
```diff
@@ -186,22 +179,16 @@ tsuba::AddPropertySlice(
...
     auto on_complete = [add_fn,
                         prop](const std::shared_ptr<arrow::Table>& props)
         -> katana::CopyableResult<void> {
-      auto add_result = add_fn(props);
-      if (!add_result) {
-        return add_result.error().WithContext(
-            "adding {}", std::quoted(prop->name()));
-      }
+      KATANA_CHECKED_CONTEXT(
+          add_fn(props), "adding: {}", std::quoted(prop->name()));
```
tylershunt, line 192: This is an example of good context IMO the caller doesn't know which property triggered the error.

libtsuba/src/AddProperties.cpp
e-mcginnis: +1 
Though personally I do prefer we lean towards over-providing context rather than under-providing it.

libtsuba/src/RDG.cpp
```diff
@@ -613,29 +590,16 @@ tsuba::RDG::Store(
         rdg_dir(), handle.impl_->rdg_manifest().dir()));
   }
 
-  auto desc_res = WriteGroup::Make();
-  if (!desc_res) {
-    return desc_res.error();
-  }
   // All write buffers must outlive desc
-  std::unique_ptr<WriteGroup> desc = std::move(desc_res.value());
+  auto desc = KATANA_CHECKED(WriteGroup::Make());
```
e-mcginnis: nit:
I'm sure this is up for debate, but I prefer the explicit `std::unique_ptr<WriteGroup>` here vs the `auto`. I don't like having to look at the signature of `WriteGroup::Make()` or whatever we are calling to know what `desc` is. Especially since its not just a `WriteGroup` like one might assume, but a  `std::unique_ptr<WriteGroup>`.

libtsuba/src/AddProperties.cpp
pyrito: Agreed. There are certain places I removed it where it didn't provide much value, but in general I think adding some context can be helpful when looking at an error message. Unless @tylershunt you think it's confusing and should only be there if it provides needed context? E.g. if it's just a isolated function call, there may not be any context needed since the error gives the line number?

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -115,21 +112,14 @@ katana::Result<void>
 RDGPartHeader::Write(
     RDGHandle handle, WriteGroup* writes,
     RDG::RDGVersioningPolicy retain_version) const {
-  auto serialized_res = katana::JsonDump(*this);
-  if (!serialized_res) {
-    return serialized_res.error();
-  }
-
-  std::string serialized = std::move(serialized_res.value());
+  auto serialized = KATANA_CHECKED(katana::JsonDump(*this));
```
e-mcginnis: nit: same auto nit as above

libtsuba/src/AddProperties.cpp
witchel: I think `KATANA_CHECKED` is easier to read and should be preferred.  `KATANA_CHECKED_CONTEXT` is for cases where the context, beyond file and line number, is valuable and (always?) that includes data values.

scober (approved):
Thank you for doing this

libtsuba/src/RDG.cpp
pyrito: @witchel wanted to get your opinion on this. I'm a fan of using `auto` for most things since it's usually cleaner to write, but I can also see Evan's point here.

witchel: This is an ongoing debate, but I think the non-auto camp is winning at Katana.  I see the value of explicit types for people less familiar with the code.  We should prioritize the needs of this audience.

pyrito: Makes sense. I will make the changes accordingly

pyrito:
I made some updates and addressed some concerns from the reviewers. I left some context prints in areas where I thought it might be helpful but please feel free to disagree.

e-mcginnis (approved):
Looks great, thanks for cleaning this up!

PR #688, opened 2021-11-16
Unbind RDGTopology FileView when no longer used

e-mcginnis:
Unbind RDGTopology FileView when no longer used

Currently, we bind the RdgTopology FileView, copy its contents to arrays
and then move on. This wastes a ton of memory since we effectively have
two copies of the topology in memory at once.

Unbind the RDGTopology FileView after we copy out its contents to reduce
our memory usage.
Must then rebind the RDGTopology FileView when persisting in a new
location

modified libtsuba optional topology test to:
 - unbind the topologies before calling store
 - copy the csr topology data it was making optional topologies from since the csr topology is now unbound
 - cleanup topology validation code, and tmp rdg cleanup code

JIRA: KAT-2498

### DIFF
--- a/libgalois/src/GraphTopology.cpp
+++ b/libgalois/src/GraphTopology.cpp
@@ -7,6 +7,7 @@
 #include "katana/Logging.h"
 #include "katana/PropertyGraph.h"
 #include "katana/Random.h"
+#include "katana/Result.h"
 #include "tsuba/RDGTopology.h"
 
 void
@@ -163,6 +164,11 @@ katana::EdgeShuffleTopology::Make(tsuba::RDGTopology* rdg_topo) {
       &(rdg_topo->edge_index_to_property_index_map()[0]),
       &(rdg_topo->edge_index_to_property_index_map()[rdg_topo->num_edges()]),
       edge_prop_indices.begin());
+
+  // Since we copy the data we need out of the RDGTopology into our own arrays,
+  // unbind the RDGTopologys file store to save memory.
+  auto res = rdg_topo->unbind_file_storage();
+  KATANA_LOG_ASSERT(res);
 
   std::unique_ptr<EdgeShuffleTopology> shuffle =
       std::make_unique<EdgeShuffleTopology>(EdgeShuffleTopology{
@@ -409,6 +415,11 @@ katana::ShuffleTopology::Make(tsuba::RDGTopology* rdg_topo) {
       &(rdg_topo->node_index_to_property_index_map()[rdg_topo->num_nodes()]),
       node_prop_indices_copy.begin());
 
+  // Since we copy the data we need out of the RDGTopology into our own arrays,
+  // unbind the RDGTopologys file store to save memory.
+  auto res = rdg_topo->unbind_file_storage();
+  KATANA_LOG_ASSERT(res);
+
   std::unique_ptr<ShuffleTopology> shuffle =
       std::make_unique<ShuffleTopology>(ShuffleTopology{
           rdg_topo->transpose_state(), rdg_topo->node_sort_state(),
@@ -581,6 +592,11 @@ katana::EdgeTypeAwareTopology::Make(
       &(rdg_topo->adj_indices()
             [rdg_topo->num_nodes() * edge_type_index->num_unique_types()]),
       per_type_adj_indices.begin());
+
+  // Since we copy the data we need out of the RDGTopology into our own arrays,
+  // unbind the RDGTopologys file store to save memory.
+  auto res = rdg_topo->unbind_file_storage();
+  KATANA_LOG_ASSERT(res);
 
   return std::make_unique<EdgeTypeAwareTopology>(EdgeTypeAwareTopology{
       edge_type_index, e_topo, std::move(per_type_adj_indices)});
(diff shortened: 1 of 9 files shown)


libgalois/src/GraphTopology.cpp
```diff
@@ -164,6 +165,11 @@ katana::EdgeShuffleTopology::Make(tsuba::RDGTopology* rdg_topo) {
       &(rdg_topo->edge_index_to_property_index_map()[rdg_topo->num_edges()]),
       edge_prop_indices.begin());
 
+  // Since we copy the data we need out of the RDGTopology into our own arrays,
+  // unbind the RDGTopologys file store to save memory.
+  auto res = rdg_topo->unbind_file_storage();
+  KATANA_LOG_ASSERT(res);
```
pyrito, line 171: why not `KATANA_CHECKED`?

libgalois/src/GraphTopology.cpp
```diff
@@ -409,6 +415,11 @@ katana::ShuffleTopology::Make(tsuba::RDGTopology* rdg_topo) {
       &(rdg_topo->node_index_to_property_index_map()[rdg_topo->num_nodes()]),
       node_prop_indices_copy.begin());
 
+  // Since we copy the data we need out of the RDGTopology into our own arrays,
+  // unbind the RDGTopologys file store to save memory.
+  auto res = rdg_topo->unbind_file_storage();
+  KATANA_LOG_ASSERT(res);
```
pyrito, line 421: ditto

libgalois/src/GraphTopology.cpp
```diff
@@ -582,6 +593,11 @@ katana::EdgeTypeAwareTopology::Make(
             [rdg_topo->num_nodes() * edge_type_index->num_unique_types()]),
       per_type_adj_indices.begin());
 
+  // Since we copy the data we need out of the RDGTopology into our own arrays,
+  // unbind the RDGTopologys file store to save memory.
+  auto res = rdg_topo->unbind_file_storage();
+  KATANA_LOG_ASSERT(res);
```
pyrito, line 599: ditto

libgalois/src/GraphTopology.cpp
e-mcginnis: Currently, the whole stack of BlahTopology()::Make, PGViewCache::BuildOrGetBlahTopo, PGViewCache:;BuildView, don't return katana::Result<>. Making this `KATANA_CHECKED` would require changing the signatures of a dozen functions, and require changes to the enterprise repo as well.
 
Since my focus here is on memory usage, I feel doing that work is out of scope.
Also, if we can't unbind the `file_store` something huge has gone wrong and we should probably die anyway. 

- Resolving the dittos for now

pyrito: Ah right, I forgot that the parent function **also** needs to return a `Result` type. I agree, that task is out-of-scope for now.

libtsuba/src/RDGTopology.cpp
```diff
@@ -483,8 +501,34 @@ tsuba::RDGTopology::DoStore(
     //TODO: emcginnis need different naming schemes for the optional topologies
     katana::Uri path_uri = MakeTopologyFileName(handle);
 
+    // file store must be bound to make a copy of it in a new location
+    if (!file_store_bound_) {
+      // Can't just use RDGTopology::Bind() here since path() is empty
+      // Must bind to old_path since we are persisting in a new location
+      std::string old_path = metadata_entry_->old_path_;
```
pyrito, line 508: I don't understand this comment. Why do we have to use old path?

pyrito (approved):
LGTM

libtsuba/src/RDGTopology.cpp
e-mcginnis: When we change the RDG location, we clear the PartitionTopologyMetadata path (and many other paths) to indicate we are moving the RDG.
Now since we likely don't have the topology bound at this point, we have to somehow find the topology file at the path we just cleared.
To make locating the topology file possible, when we clear the path we first save it in `old_path` so that we can bind and then store it here.
This code branch is only taken when we haven't modified the topology so we really just want to copy the topology file to the new RDG location.

e-mcginnis:
added comment about memory savings to https://katanagraph.atlassian.net/browse/KAT-2498
but to sum it up, in the test case where the topology is roughly 6GB, we save roughly 6GB of memory by doing this.

libtsuba/src/RDGTopology.cpp
e-mcginnis: If that explanation makes sense, any suggestions on what to add to the comment to make things clearer?

tylershunt (approved):
Looks good! thanks.

amberhassaan:
Sorry, late to the party. 
@e-mcginnis : why don't we make `unbind` calls part of the destructor?

e-mcginnis:
> Sorry, late to the party. @e-mcginnis : why don't we make `unbind` calls part of the destructor?

We don't call the descructor for the RDGTopology instances until RDGCore gets torn down. The actual RDGTopology instances that have FileViews bound to files are part of the RDGTopologyManager, which is part of the RDGCore. When we are calling unbind in the PropertyGraph::Make() path we are working off of a RDGTopology pointer.

If we started destructing the RDGTopology instances we would lose the metadata we need to quickly locate and bind a topology.

PR #690, opened 2021-11-16
Add edge iterator to python graphs

astolman:
Previously, there was no easy way to step through all the edges of a graph in python, especially if you want the edge source, dest pairs. This PR adds a method to the python Graph objects which provides an iterator over all the edges in the graph. This iterator returns tuples of the form (src, dest).

This will be used in downstream data science applications. Being able to get an edge list like this from a graph is an important operation in conversion to DGL/PyG graph representation, as well as allow sampling over edges uniformly at random.

### DIFF
--- a/python/katana/local/_graph.pyx
+++ b/python/katana/local/_graph.pyx
@@ -22,6 +22,7 @@ from libcpp.vector cimport vector
 
 from ..native_interfacing.buffer_access cimport to_pyarrow
 from .entity_type cimport EntityType
+from .edge_iterator import EdgeIterator
 
 from abc import abstractmethod
 
@@ -143,6 +144,12 @@ cdef class GraphBase:
         Can be called from numba compiled code with the name `self.nodes()`.
         """
         return iter(range(self.num_nodes()))
+
+    def edge_iterator(self):
+        """
+        Returns an iterator over the edges of this graph.
+        """
+        return EdgeIterator(self)
 
     def edges(self, uint64_t n):
         """
--- a/python/katana/local/_graph.pyx
+++ b/python/katana/local/_graph.pyx
@@ -22,9 +22,10 @@ from libcpp.vector cimport vector
 
 from ..native_interfacing.buffer_access cimport to_pyarrow
 from .entity_type cimport EntityType
-from .edge_iterator import EdgeIterator
 
 from abc import abstractmethod
+
+from .edge_iterator import EdgeIterator
 
 __all__ = ["GraphBase", "Graph"]
 
--- a/python/katana/local/edge_iterator.py
+++ b/python/katana/local/edge_iterator.py
@@ -1,4 +1,4 @@
-class EdgeIterator():
+class EdgeIterator:
     """
     An iterator over the edges of a graph
     """
--- a/python/katana/local/edge_iterator.py
+++ b/python/katana/local/edge_iterator.py
@@ -1,0 +1,24 @@
+class EdgeIterator():
+    """
+    An iterator over the edges of a graph
+    """
+
+    def __init__(self, graph):
+        self._graph = graph
+        self._curr_node = 0
+        self._curr_edge = 0
+
+    def __next__(self):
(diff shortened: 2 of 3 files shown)


python/katana/local/_graph.pyx
```diff
@@ -144,6 +146,12 @@ cdef class GraphBase:
         """
         return iter(range(self.num_nodes()))
 
+    def edge_iterator(self):
+        """
+        Returns an iterator over the edges of this graph.
+        """
+        return EdgeIterator(self)
```
arthurp, line 153: Everywhere else in this API we use IDs instead of the pairs. I think you are right that we should move to a more OO approach to this where we provide more data about the edges. However, I don't think this is the way to do it. The problem is now there are two different ways to access the edges.

I would much prefer having an Edge class which is the pair you are using in this iterator, and also represents the ID of the edge. That way there is a single cohesive way to talk about edges. We have been talking about moving to a dataframe oriented API and I think this is a great chance. We should make the `edges` method return a dataframe like object holding the edges as triples of ID, src, dst and also providing transparent access to the properties on those edges.

python/katana/local/edge_iterator.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def __next__(self):
+        if self._curr_edge >= self._graph.num_edges():
+            raise StopIteration
+        if len(self._graph.edges(self._curr_node)) == 0:
+            self._curr_node += 1
+            return self.__next__()
+        if self._curr_edge > self._graph.edges(self._curr_node)[-1]:
+            self._curr_node += 1
+        next_edge = (self._curr_node, self._graph.get_edge_dest(self._curr_edge))
```
arthurp, line 19: Using an unadorned tuple encourages bad client code. The reason is there is no way for the client to be explicit about what they actually mean when accessing the values in the pair. It's much better to use a custom class or named tuple.

arthurp (changes requested):
I think we need a better design for this, instead of just adding a one off method to provide the access you want.

How quickly do you need this? So we can figure out how fast the API needs to be determined and implemented.

python/katana/local/_graph.pyx
astolman: The fact that we never use pairs and only use IDs is one of the biggest hurdles in converting to PyG. The PyG graph representation is just an mx2 tensor holding the edge list of the graph (basically COO). I don't hate the idea of an edge object instead, so long as its easy to get both endpoints of that edge.

I also like the idea of using pandas DataFrames to hold the edges since this works nicely with lots of python libraries.

astolman:
> I think we need a better design for this, instead of just adding a one off method to provide the access you want.
> 
> How quickly do you need this? So we can figure out how fast the API needs to be determined and implemented.

I would like this done ASAP. I was hoping it would be a small, uncontroversial change that would let me play with the data more and develop link prediction models, which is my top priority.  You raise good points in your comments, but I really don't want to get bogged down in this. If I add an Edge object and return those instead of tuples, would that be enough to earn your approval?

arthurp:
> If I add an Edge object and return those instead of tuples, would that be enough to earn your approval?

Let me try my hand at this tonight. If I cannot get it, then yes, your idea is fine as a stop gap. But let's mark it as unsupported and immediately deprecated.

astolman:
Fair enough. I was also thinking that I like the idea of just returning a pandas dataframe with the edge data. That also sounds very easy to implement and probably easy to extend it to have edge properties in the dataframe if you want them, too

arthurp:
I did it I think. Tell me what you think: https://github.com/KatanaGraph/katana/pull/692. The approach I took provides a path forward to use DataFrames more and also avoids copies for at least some use cases of the data frame. The problem with converting to a pandas dataframe directly is it requires copying the data set immediately. My approach is zero copy until you convert, so you can slice first and then convert to pandas or maybe just use our dataframe directly.

astolman:
Closing this since Arthur replaced it with a better PR.

PR #692, opened 2021-11-17
Expose edges as a dataframe

arthurp:
WIth: KatanaGraph/katana-enterprise#2541

This renames `edges` to `edge_ids` and creates new `edges` method which returns a dataframe object similar to those provided by pandas, but which is backed by our existing data structures so the dataframe is zero-copy. There is also support for converting the dataframe to pandas, but that must copy the underlying data.

This should supersede #690 and be a step on the way to implementing a dataframe oriented API. A future PR will add a similar dataframe for nodes.


python/katana/local/_graph.pyx
```diff
@@ -94,13 +126,13 @@ cdef class GraphBase:
...
 
-    def loaded_node_schema(self):
+    def loaded_node_schema(self) -> pyarrow.Schema:
         """
         Return the `pyarrow` schema for the node properties loaded for this graph.
         """
         return pyarrow_wrap_schema(self.underlying_property_graph().loaded_node_schema())
 
-    def loaded_edge_schema(self):
+    def loaded_edge_schema(self) -> pyarrow.Schema:
```
astolman: Not really relevant to this PR, but I think we might not want to be returning pyarrow objects at all in the API. Pyarrow is not a library data scientists are at all familiar with, and having to look up the arrow API to understand how to use what our API returns is a real speed bump when reading the docs for the first time.

python/katana/local/_graph.pyx
```diff
@@ -156,6 +188,63 @@ cdef class GraphBase:
...
+                return pivot
+            elif e >= edge_range.stop:
+                bottom = edge_range.stop
+            elif e < edge_range.start:
+                top = edge_range.start
+            else:
+                raise AssertionError("Unreachable")
+        raise ValueError(e)
+
+    def edges(self, n = None) -> DataFrame:
```
astolman: Would it be possible to make the second argument a collection of nodes, and not just limit it to a single node?

python/test/test_property_graph.py
```diff
@@ -52,17 +50,41 @@ def test_get_edge_dest(graph):
     assert graph.get_edge_dest(1) == 8014
 
 
+def test_edge_data_frame(graph):
+    edges = graph.edges()
+    assert len(edges.columns) == 18 + 3
+    assert len(edges) == 43072
+    assert edges.at[0, "dest"] == 8014
+    assert edges["dest"][1] == 8014
```
astolman, line 58: Can we also get a test asserting the value of edges["src"]? Would be useful as an example seeing how to access all the columns of the DataFrame.

astolman (approved):
A couple of nits, but nothing I feel strongly about. This looks like a huge quality of life improvement, thank you.

python/katana/local/_graph.pyx
arthurp: Hmm. Yes. Basically preslicing the data when it's generated. I'm not sure if there is much advantage in doing that over getting the data then slicing. The DataFrame is very cheap to make. Maybe we should remove the argument entirely. Make this a property. And then the user and project out what they want at the dataframe level.

python/katana/local/_graph.pyx
arthurp: Yea. I agree. We should phase it out. This PR is not the last step in that process at all.

python/test/test_property_graph.py
arthurp: Tests != documentation. I'll put an example in the doc string.

arthurp:
Hopefully I can do all this for nodes as well soonish.

python/katana/local/_graph.pyx
arthurp: My thinking about slicing the data was a bit wrong. I'll make this argument accept collections and add a TODO to polish the API later.

python/katana/local/_graph.pyx
```diff
@@ -367,8 +456,8 @@ cdef class Graph(GraphBase):
         return <uint64_t>self.underlying_property_graph()
 
     def global_out_degree(self, uint64_t node):
-        return len(self.edges(node))
+        return len(self.edge_ids(node))
 
-    def global_in_degree(self, uint64_t node):
-        # TODO(loc) needs shared-memory bi-directional view
-        raise NotImplementedError()
+    # def global_in_degree(self, uint64_t node):
```
gliga: why is this now under comments?

python/katana/local/_graph.pyx
```diff
@@ -156,6 +188,63 @@ cdef class GraphBase:
...
         return range(deref(edge_range.begin()), deref(edge_range.end()))
 
+    def get_edge_source(self, uint64_t e):
+        """
+        Get the source of an edge. This is O(log(N)) since it uses binary search to avoid having to store a source list.
+        :param e: an edge ID.
+        :return: The source node of that edge.
+        """
+        bottom = 0
+        top = self.num_nodes()
```
gliga: Isn't there some library that already has a binary search? I am surprised still need to write it in 2021 :)

gliga (approved):
I did not get the reason we are not relying on panda's dataframe, but the rest looks good.  Oh, I wish you separate renaming for edges from the DataFrame addition. And one test for the binary search would be good ;)

metagraph-plugin/tests/test_types.py
```diff
@@ -32,40 +32,34 @@ def test_edge_schema(katanagraph_rmat15_cleaned_di):
...
 
 
-def test_num_nodes(networkx_weighted_undirected_8_12, networkx_weighted_directed_8_12):
+def test_num_nodes_networkx(networkx_weighted_undirected_8_12, networkx_weighted_directed_8_12):
     assert len(list(networkx_weighted_undirected_8_12.value.nodes(data=True))) == 8
     assert len(list(networkx_weighted_directed_8_12.value.nodes(data=True))) == 8
 
 
-def test_num_nodes(networkx_weighted_undirected_8_12, networkx_weighted_directed_8_12):
+def test_num_nodes_networkx_2(networkx_weighted_undirected_8_12, networkx_weighted_directed_8_12):
```
denish44: test_num_edges_networkx?

denish44 (approved):

arthurp:
> I did not get the reason we are not relying on panda's dataframe, but the rest looks good.

Because pandas data frames only support holding `pandas.Series` as columns and those cannot be lazily computed and don't have any flexibility in their format, so all the data would need to be copied if we used pandas DataFrame. I actually looked at that first, but it wouldn't work without forcing a copy. With this approach, the data isn't copied in many many cases. Instead dataframe accesses are translated into normal Katana graph data accesses.

python/katana/local/_graph.pyx
arthurp: Ooops. I hate "this isn't implemented implementations", but that's not the point of this PR. I'll revert this.

python/katana/local/_graph.pyx
arthurp: There isn't one that I can use in this context. Cython has very poor generic programming support and doesn't really have a standard library anyway. So here we are.

I'll write a test, your concern if valid. I already shook some bugs out of it.

arthurp:
> Oh, I wish you separate renaming for edges from the DataFrame addition

I can refactor the commits if you like. You are right that rename, then add new operation with old name would be better to have in the git history.

metagraph-plugin/tests/test_types.py
arthurp: Ooops. Thanks. I couldn't figure out the difference between the tests and apparently I was just momentarily blind.

arthurp:
@astolman Sorry for the delay. This PR now has the node list support you want.

PR #694, opened 2021-11-17
Adding CDLP shared memory synchronous algorithm

amazloumi:


amazloumi:
Hi @gurbinder533, I have cc'ed @amberhassaan as well since we once discussed the algorithm, he might have some feedback and comments.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+using namespace katana::analytics;
+
+namespace {
+const unsigned int kInfinity = std::numeric_limits<unsigned int>::max();
+
+/// Limited number of iterations to limit the oscillation of the label
+/// in Synchronous algorithm. We dont need to limit it in Asynchronous algorithm.
+/// Set to 10 same as Graphalytics benchmark.
+const unsigned int kMaxIterations = 10;
```
uditagar, line 33: so is this the total number of iterations for synchronous algo? I think it should depend more on the diameter of the graph. we need to figure something out for the oscillations. Can you point me to some paper where they talk about this oscillation? thanks

amazloumi: This is the original paper (implemented Asynchronous but it is not deterministic): https://arxiv.org/abs/0709.2938
This is a benchmark implemented a Synchronous Version to make it deterministic: https://arxiv.org/pdf/2011.15028.pdf

The synchronous version has the oscillation problem.
The Asynchronous version has multiple correct outputs.

amazloumi: The Graphalytics Benchmark; implemented CDLP for different platforms; I am trying to install one of them for the third party test; for example in the PowerGraph platform, the max_iter is initialized to the default value of 10.
https://github.com/atlarge-research/graphalytics-platforms-powergraph/blob/master/src/main/c/main.cpp

uditagar: I get it now. So one way to stop this oscillation would be to not update your label (if the new label also have same frequency but lower id). This should guarantee determinism as well, since you are still reading values from previous iteration (when making any decisions).

uditagar: I will think more on it then. You can keep it as it is for now. We can decide on it later on. thanks

uditagar: ok, then it's fine since it matches the 3rd party default params.

lonestar/analytics/cpu/cdlp/cdlp_cli.cpp
```diff
@@ -0,0 +1,164 @@
...
+  totalTime.start();
+
+  //FIXME: I am not sure if it only requiers a symmetric graph
+  /*  if (!symmetricGraph) {
+    KATANA_LOG_FATAL(
+        "This application requires a symmetric graph input;"
+        " please use the -symmetricGraph flag "
+        " to indicate the input is a symmetric graph.");
+  }*/
+
```
uditagar: you can remove this block since this app does not require a symmetric graph

libgalois/src/analytics/cdlp/cdlp.cpp
amazloumi: "From our experiments, we found that irrespective of n (# of nodes in the graph), 95% of the nodes or more are classified correctly by the end of iteration 5."
however, this percentage is reported for asynchronous, I think we should add this percentage to our statistics; so based on this percentage of accurate node classification one can alter the number of iterations; however, from one point increasing the number of iteration does not change this percentage due to the label oscillation. But maybe computing this percentage is expensive maybe we can do it on user-demand. Also, this percentage should be considered in CdlpAssertValid; in a way that if it is more than a threshold it returns true.

lonestar/analytics/cpu/cdlp/cdlp_cli.cpp
amazloumi: It was originally taken from CC app and modified; Yes; it needs to be removed; I will.

libgalois/src/analytics/cdlp/cdlp.cpp
amazloumi: > I get it now. So one way to stop this oscillation would be to not update your label (if the new label also have same frequency but lower id). This should guarantee determinism as well, since you are still reading values from previous iteration (when making any decisions).

The oscillation might happen in a way that in each iteration, one of the two oscillated labels becomes more frequent in the immediate neighborhood.

a simple example is following directed triangle:
0->1
1->0
0->2
2->0
1->2

The label for "node 0" and the label for "node 1 and 2" will be oscillated between 0 and 1.

uditagar: yes, you are right

uditagar (commented):
Looks fine to me.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,402 @@
...
+    for (auto e : bidir_view.edges(node)) {
+      auto dest = bidir_view.edge_dest(e);
+      const auto ddata = graph.template GetData<NodeCommunity>(dest);
+      histogram[ddata]++;
+    }
+
+    /// Pick the most frequent communtiy for node
+    /// Pick the smallest one if more than one max frequent exist.
+    /// TODO: This needs to be fix for the oscillation of the label cases
+    /// It returns Failed for those cases.
```
amazloumi: "From our experiments, we found that irrespective of n (# of nodes in the graph), 95% of the nodes or more are classified correctly by the end of iteration 5."

however, this percentage is reported for asynchronous, I think this percentage should be considered in CdlpAssertValid; in a way that if it is more than a specific threshold it returns true. Right now, it fails when oscillations happen; since in that case, there are some nodes that have a label that is not the most frequent one in their immediate neighborhood.

I think we should add this percentage to our statistics as well; so based on this percentage of accurate node classification one can alter the number of iterations; however, from some point on, increasing the number of iteration does not change this percentage due to the label oscillation. Concerns: Computing this percentage can be expensive. Maybe, we can do it on user demand.

python/katana/local/analytics/_cdlp.pyx
```diff
@@ -0,0 +1,247 @@
...
+        in cases where communities take the form of a star graph
+        """
+        return CdlpPlan.make(_CdlpPlan.Synchronous())
+
+    #@staticmethod
+    #def asynchronous() -> CdlpPlan:
+        """
+        Unlike Synchronous algorithm, Asynchronous can use the current iteration
+        updated community IDs for some of the neighbors that have been already
+        updated in the current iteration and use the old values for the other neighbors
```
arthurp: If you are commenting code, comment the whole thing. ;-)

arthurp (commented):
Interfacing looks fine. But it needs a test.

python/katana/local/analytics/_cdlp.pyx
amazloumi: You are right; I have been more a c++ guy; I considered what is inside the triple quote as comments. will comment the whole thing; thanks for the comment. Did I get your comment correctly? :)

arthurp: Yep. :-) That's what I wanted

amazloumi: I think also about maybe removing these redundant comments here; these comments already are in the CPP implementation of each algorithm. maybe make them very brief.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+struct CdlpSynchronousAlgo {
+  using CommunityType = uint64_t;
+  struct NodeCommunity : public katana::PODProperty<CommunityType> {};
+
+  using NodeData = std::tuple<NodeCommunity>;
+  using EdgeData = std::tuple<>;
+  typedef katana::TypedPropertyGraph<NodeData, EdgeData> Graph;
+  typedef typename Graph::Node GNode;
+  using BiDirGraphView = katana::TypedPropertyGraphView<
```
amberhassaan: Please use Undirected view and then you'd write only one loop for both in and out edges.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,402 @@
...
+      return;
+
+    struct NodeDataPair {
+      GNode node;
+      CommunityType data;
+      NodeDataPair(GNode node, CommunityType data) : node(node), data(data) {}
+    };
+
+    size_t iterations = 0;
+    katana::InsertBag<NodeDataPair> applyBag;
```
amberhassaan: What is apply bag? Why is it needed?

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,402 @@
...
+    /// FIXME: in this implementation, in each iteration, all the nodes are active
+    /// for gather phase. If InsertBag does not accept duplicate items then this
+    /// can be improved to have only the affected nodes to be active in next iteration
+    while (iterations < max_iterations) {
+      // Gather Phase
+      katana::do_all(
+          katana::iterate(*graph),
+          [&](const GNode& node) {
+            const auto ndata_current_comm = graph->GetData<NodeCommunity>(node);
+            typedef boost::unordered_map<CommunityType, size_t> Histogram_type;
```
amberhassaan: You should create a `katana::PerThreadStorage<katana::gstl::Map<K,V>> per_thread_map` and then inside the loop, you can get a thread-local instance of the map by doing ` auto& histogram = *per_thread_map.getLocal();`

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,402 @@
...
+              auto dest = bidir_view.edge_dest(e);
+              const auto ddata = graph->GetData<NodeCommunity>(dest);
+              histogram[ddata]++;
+            }
+
+            // Pick the most frequent communtiy as the new community for node
+            // pick the smallest one if more than one max frequent exist.
+            auto ndata_new_comm = ndata_current_comm;
+            size_t best_freq = 0;
+            for (Histogram_type::const_iterator it = histogram.begin();
```
amberhassaan: Use range based for like:
```
for (const auto& [k,v]: histgogram) {...
```

libgalois/src/analytics/cdlp/cdlp.cpp
amberhassaan: Also, use `using X = Y;` instead of `typedef Y X;`. `using` is the new preferred way of doing typedefs.

amberhassaan (commented):
Initial set of comments. Mainly, use `Undirected` View instead of BiDirectional.

libgalois/src/analytics/cdlp/cdlp.cpp
amazloumi: It is merged! perfect, will do.

libgalois/src/analytics/cdlp/cdlp.cpp
amazloumi: since it is synchronous, applyBag is used to collect the next value for the nodes. when the gather phase finishes then the values will be assigned to nodes. hence, it guarantees that, in each iteration, nodes will use their neighbors' data from the previous iteration.

libgalois/src/analytics/cdlp/cdlp.cpp
amberhassaan: I still don't see it.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+using namespace katana::analytics;
+
+namespace {
+/// Limited number of iterations to limit the oscillation of the label
+/// in Synchronous algorithm. We dont need to limit it in Asynchronous algorithm.
+/// Set to 10 same as Graphalytics benchmark.
+const unsigned int kMaxIterations = 10;
+
+struct CdlpSynchronousAlgo {
+  using CommunityType = uint64_t;
```
gurbinder533, line 36: Please pull out the common code from algorithms to avoid duplication.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+  using Graph = katana::TypedPropertyGraphView<
+      katana::PropertyGraphViews::Undirected, NodeData, EdgeData>;
+  using GNode = typename Graph::Node;
+
+  CdlpPlan& plan_;
+  CdlpSynchronousAlgo(CdlpPlan& plan) : plan_(plan) {}
+
+  void Initialize(Graph* graph) {
+    katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
+      auto& ndata = graph->GetData<NodeCommunity>(node);
```
gurbinder533: `graph->GetData<NodeCommunity>(node) = node`

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+  using Graph = katana::TypedPropertyGraphView<
+      katana::PropertyGraphViews::Undirected, NodeData, EdgeData>;
+  using GNode = typename Graph::Node;
+
+  CdlpPlan& plan_;
+  CdlpAsynchronousAlgo(CdlpPlan& plan) : plan_(plan) {}
+
+  void Initialize(Graph* graph) {
+    katana::do_all(katana::iterate(*graph), [&](const GNode& node) {
+      auto& ndata = graph->GetData<NodeCommunity>(node);
```
gurbinder533: ditto.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  algo.Initialize(&graph);
+
+  katana::StatTimer execTime("CDLP");
+
+  execTime.start();
+  algo(&graph, max_iterations);
+  execTime.stop();
+
+  algo.Deallocate(&graph);
```
gurbinder533: I don't think deallocate is required in this case. It was introduced for some cc algorithms. Please remove this.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+    return CdlpWithWrap<CdlpAsynchronousAlgo>(
+        pg, output_property_name, plan, max_iterations);
+  */
+  default:
+    return ErrorCode::InvalidArgument;
+  }
+}
+
+katana::Result<CdlpStatistics>
+katana::analytics::CdlpStatistics::Compute(
```
gurbinder533, line 189: This function is now being used by louvain, cc, and cdlp, basically everything which is calculating communities. Is it possible to move it to some common .h file in libgalois/include/analytics to avoid code duplication.

lonestar/analytics/cpu/cdlp/cdlp_cli.cpp
```diff
@@ -0,0 +1,140 @@
...
+    cll::init(kMaxIterations));
+
+static cll::opt<CdlpPlan::Algorithm> algo(
+    "algo", cll::desc("Choose an algorithm (default value Synchronous):"),
+    cll::values(
+        clEnumValN(
+            CdlpPlan::kSynchronous, "Synchronous",
+            "Synchronous algorithm")/*,
+
+		/// TODO: Asynchronous Algorithm will be implemented later after Synchronous
```
gurbinder533: Please always leave ur user-id with TODO. Example: TODO (gill)

lonestar/analytics/cpu/cdlp/cdlp_cli.cpp
```diff
@@ -0,0 +1,140 @@
...
+  case CdlpPlan::kAsynchronous:
+    plan = CdlpPlan::Asynchronous();
+    break;
+  */
+  default:
+    std::cerr << "Invalid algorithm\n";
+    abort();
+  }
+
+  auto pg_result = Cdlp(pg.get(), "community", maxIterations, plan);
```
gurbinder533: Use variable for output_property_name instead of a string "community". See enterprise lonestar apps for example. We will update other apps too.

gurbinder533 (changes requested):

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+    struct NodeDataPair {
+      GNode node;
+      CommunityType data;
+      NodeDataPair(GNode node, CommunityType data) : node(node), data(data) {}
+    };
+
+    size_t iterations = 0;
+    katana::InsertBag<NodeDataPair> applyBag;
+
```
uditagar, line 66: should use `apply_bag` instead

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+    /// FIXME: in this implementation, in each iteration, all the nodes are active
```
uditagar: replace FIXME with TODO (yasin)

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -0,0 +1,302 @@
...
+          katana::loopname("CDLP_Gather"));
+
+      // No change! break!
+      if (applyBag.empty())
+        break;
+
+      // Apply Phase
+      katana::do_all(
+          katana::iterate(applyBag),
+          [&](const NodeDataPair nodeData) {
```
uditagar: replace `nodeData` with `node_data`

libgalois/src/analytics/cdlp/cdlp.cpp
amazloumi: A parent "CdlpAlgo" structure has been defined to keep the common code.

libgalois/src/analytics/cdlp/cdlp.cpp
amazloumi: I agree we should do this. This function for cc and cdlp is exactly the same while for louvain is more complex but we can come up with a common function for them and put it in a common .h file.  though I think it should be done in another task/pr.

lonestar/analytics/cpu/cdlp/cdlp_cli.cpp
amazloumi: a const std::string has been defined for property name.

gurbinder533:
@amazloumi Please make sure undirected view is added before we merge this.

amazloumi:
@gurbinder533 Yes; I have switched from Bidirectional view to Undirected view.

libgalois/src/analytics/cdlp/cdlp.cpp
```diff
@@ -32,7 +32,7 @@ namespace {
 /// Set to 10 same as Graphalytics benchmark.
 const unsigned int kMaxIterations = 10;
 
-struct CdlpSynchronousAlgo {
+struct CdlpAlgo {
```
uditagar, line 35: I am not sure if this is the right place for this common code. It was supposed to be at a place where all apps can use it. You can create a separate `ConnectedComponentsBase.h` file for this, I think.

gurbinder533 (approved):
Looks good.

PR #695, opened 2021-11-18
remove unnecessary assert

yshalabi:
this assert is wrong -- graphs can be cleared to end with zero-sized condensed type ID maps 

JIRA: KAT-2545

### DIFF
--- a/libtsuba/src/RDGTopology.cpp
+++ b/libtsuba/src/RDGTopology.cpp
@@ -159,11 +159,8 @@ RDGTopology::Map() {
   // EdgeTypeAwareTopologies have a larger adj_indices array than usual topologies
   if (topology_state_ ==
       tsuba::RDGTopology::TopologyKind::kEdgeTypeAwareTopology) {
-    KATANA_LOG_VASSERT(
-        edge_condensed_type_id_map_size_,
-        "mapping an EdgeTypeAwareTopology with a zero sized "
-        "condensed_type_id_map");
-    adj_indices_size = num_nodes_ * edge_condensed_type_id_map_size_;
+    adj_indices_size =
+        std::max(num_nodes_, num_nodes_ * edge_condensed_type_id_map_size_);
   }
 
   cursor += adj_indices_size;


e-mcginnis:
Can they though? We create empty `EdgeTypeAware` optional topologies?

yshalabi:
> Can they though? We create empty `EdgeTypeAware` optional topologies?

I’m not sure what you are you asking. If you are asking if it is legal for graphs to be cleared and new graphs constructed: the answer is yes.

e-mcginnis:
if we are constructing a new graph, and it is empty, what use do we have for an optional `EdgeTypeAwareTopology`? This assert only covers `TopologyKind::kEdgeTypeAwareTopology`. The default topo doesn't have `edge_condensed_type_id_map_size_`

yshalabi:
> if we are constructing a new graph, and it is empty, what use do we have for an optional `EdgeTypeAwareTopology`? This assert only covers `TopologyKind::kEdgeTypeAwareTopology`. The default topo doesn't have `edge_condensed_type_id_map_size_`

The failing scenario is not in constructing a new graph. The failure occurs when loading a graph, or constructing from a CSV, then clearing it, then adding nodes. The details are in the Jira ticket.

roshandathathri:
> Can they though? We create empty `EdgeTypeAware` optional topologies?

Yes, any topology, including optional ones, can be empty.

roshandathathri (approved):
LGTM

amberhassaan (approved):
I suppose the size should be zero if the graph is empty.

libtsuba/src/RDGTopology.cpp
```diff
@@ -146,10 +146,6 @@ RDGTopology::Map() {
   // EdgeTypeAwareTopologies have a larger adj_indices array than usual topologies
   if (topology_state_ ==
       tsuba::RDGTopology::TopologyKind::kEdgeTypeAwareTopology) {
-    KATANA_LOG_VASSERT(
-        edge_condensed_type_id_map_size_,
-        "mapping an EdgeTypeAwareTopology with a zero sized "
-        "condensed_type_id_map");
     adj_indices_size = num_nodes_ * edge_condensed_type_id_map_size_;
```
e-mcginnis: Even if the `edge_condensed_type_id_map_size_` == 0,  can we guarantee `num_nodes` is also `0`? . With just this change, there is a chance we aren't adjusting the cursor properly. 

It should read:
```  // EdgeTypeAwareTopologies have a larger adj_indices array than usual topologies
  if (topology_state_ ==
      tsuba::RDGTopology::TopologyKind::kEdgeTypeAwareTopology && edge_condensed_type_id_map_size_ != 0 ) {
    adj_indices_size = num_nodes_ * edge_condensed_type_id_map_size_;
  }
```

yshalabi: I don't think we can guarantee that num_nodes is 0. It is possible to clear edges exclusively.

libtsuba/src/RDGTopology.cpp
```diff
@@ -146,11 +146,7 @@ RDGTopology::Map() {
   // EdgeTypeAwareTopologies have a larger adj_indices array than usual topologies
   if (topology_state_ ==
       tsuba::RDGTopology::TopologyKind::kEdgeTypeAwareTopology) {
-    KATANA_LOG_VASSERT(
-        edge_condensed_type_id_map_size_,
-        "mapping an EdgeTypeAwareTopology with a zero sized "
-        "condensed_type_id_map");
-    adj_indices_size = num_nodes_ * edge_condensed_type_id_map_size_;
+    adj_indices_size = std::max(num_nodes_, num_nodes_ * edge_condensed_type_id_map_size_);
```
roshandathathri: nit: This hides the reason for using max. FWIW I prefer to keep it explicit even if it is more verbose:
    `adj_indices_size = num_nodes_ == 0 ? 0 : num_nodes_ * edge_condensed_type_id_map_size_`;

roshandathathri (approved):
LGTM

PR #700, opened 2021-11-20
GraphTopology: don't barf on corner case where partition is empty

tylershunt:
For some small graphs: if the user partitions them they have no nodes
or edges on some host. In this case some of these checks are too strong.

These adjustments let me load my tiny graph on 4 hosts.

I think this is causing the GraphML CI flake, the GraphML test graph is
small and random partitioning means that sometimes hosts are left
without any data.

### DIFF
--- a/libgalois/src/GraphTopology.cpp
+++ b/libgalois/src/GraphTopology.cpp
@@ -152,18 +152,22 @@ katana::EdgeShuffleTopology::Make(tsuba::RDGTopology* rdg_topo) {
   PropIndexVec edge_prop_indices;
   edge_prop_indices.allocateInterleaved(rdg_topo->num_edges());
 
-  katana::ParallelSTL::copy(
-      &(rdg_topo->adj_indices()[0]),
-      &(rdg_topo->adj_indices()[rdg_topo->num_nodes()]),
-      adj_indices_copy.begin());
-  katana::ParallelSTL::copy(
-      &(rdg_topo->dests()[0]), &(rdg_topo->dests()[rdg_topo->num_edges()]),
-      dests_copy.begin());
+  if (rdg_topo->num_nodes() > 0) {
+    katana::ParallelSTL::copy(
+        &(rdg_topo->adj_indices()[0]),
+        &(rdg_topo->adj_indices()[rdg_topo->num_nodes()]),
+        adj_indices_copy.begin());
+  }
+  if (rdg_topo->num_edges() > 0) {
+    katana::ParallelSTL::copy(
+        &(rdg_topo->dests()[0]), &(rdg_topo->dests()[rdg_topo->num_edges()]),
+        dests_copy.begin());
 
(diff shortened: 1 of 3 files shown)


libtsuba/include/tsuba/RDGTopology.h
```diff
@@ -143,20 +143,26 @@ class KATANA_EXPORT RDGTopology {
...
   /// Optional field, may not be present depending on the kind of topology this is
   /// Requires backing FileView to be mapped & bound, or the RDGTopology to be filled from memory
   const katana::EntityTypeID* node_condensed_type_id_map() const {
-    KATANA_LOG_VASSERT(
-        node_condensed_type_id_map_ != nullptr,
-        "Either this optional field is not present, or the RDGTopology must be "
-        "either bound & mapped, or filled from memory.");
+    if (node_condensed_type_id_map_size() > 0) {
+      KATANA_LOG_VASSERT(
+          node_condensed_type_id_map_ != nullptr,
```
roshandathathri, line 161: This feels weird to me. Why don't we create an empty map so that we avoid `nullptr` dereferences?

I'm fine with adding a TODO for this to get this merged.

roshandathathri (approved):
LGTM

libtsuba/include/tsuba/RDGTopology.h
tylershunt: Yeah, there are definitely better ways to do this; ideally we wouldn't reason about pointers like this at all. Part of the problem is that tsuba is managing its own memory, and doesn't know about things like numa allocation so all of these buffers that tsuba creates are copied into buffers that we think will preform better.

I think an interface where the caller passes tsuba some memory allocator callback would work well actually, then the tsuba code just fill buffers like it does now and we'd save a copy. Not going to land before Thanksgiving though.

roshandathathri: Makes sense. Thanks.

libgalois/src/GraphTopology.cpp
```diff
@@ -152,18 +152,22 @@ katana::EdgeShuffleTopology::Make(tsuba::RDGTopology* rdg_topo) {
...
-      &(rdg_topo->edge_index_to_property_index_map()[rdg_topo->num_edges()]),
-      edge_prop_indices.begin());
+  if (rdg_topo->num_nodes() > 0) {
+    katana::ParallelSTL::copy(
+        &(rdg_topo->adj_indices()[0]),
+        &(rdg_topo->adj_indices()[rdg_topo->num_nodes()]),
+        adj_indices_copy.begin());
+  }
+  if (rdg_topo->num_edges() > 0) {
+    katana::ParallelSTL::copy(
```
witchel, line 162: Kinda too bad that the parallel version isn't smart enough to deal with the zero-length case.

witchel (commented):
Post facto good.

PR #701, opened 2021-11-22
cache: fix LRU, expose resident properties

witchel:
Before #2540

### DIFF
--- a/scripts/setup_dev_ubuntu.sh
+++ b/scripts/setup_dev_ubuntu.sh
@@ -31,8 +31,8 @@ sudo bash -x "${REPO_ROOT}/scripts/setup_ubuntu.sh" --no-setup-toolchain-variant
 "${REPO_ROOT}/scripts/setup_conan.sh"
 
 # make clang-{tidy,format}-12 the default
-sudo update-alternatives --verbose --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-12 90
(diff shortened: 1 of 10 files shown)


scripts/setup_dev_ubuntu.sh
```diff
@@ -31,8 +31,8 @@ sudo bash -x "${REPO_ROOT}/scripts/setup_ubuntu.sh" --no-setup-toolchain-variant
 "${REPO_ROOT}/scripts/setup_conan.sh"
 
 # make clang-{tidy,format}-12 the default
-sudo update-alternatives --verbose --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-12 90
-sudo update-alternatives --verbose --install /usr/bin/clang-format clang-format /usr/bin/clang-format-12 90
+sudo update-alternatives --verbose --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-12 100
+sudo update-alternatives --verbose --install /usr/bin/clang-format clang-format /usr/bin/clang-format-12 100
```
roshandathathri, line 35: Why this change?

roshandathathri (approved):
LGTM

scripts/setup_dev_ubuntu.sh
witchel: Good question.

Because when I ran those lines on our servers with a weight of 90, they left clang-format as version 10.  Only by increasing the number did it actually update them to version 12.

PR #702, opened 2021-11-22
Uprev `build/inputs` to storage_format_version 3

e-mcginnis:
- Uprev `build/inputs/propertygraphs` to storage_format_version 3
  - required some restructuring of 
- Add tools to uprev inputs storage_format_version
- bump inputs to v29 
- Add input as a dependency for tsuba, galois tests
- Correct docs/testing to remove .github/workflows from list of katana-inputs-vN references


inputs/uprev_rdg_storage_format_version.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
witchel, line 1: I guess some people like bash, but I prefer python with `import sh`.  I want to minimize the amount of sh code in our codebase.

inputs/uprev_rdg_storage_format_version.sh
```diff
@@ -0,0 +1,73 @@
...
+set -eu -o pipefail
+
+if [[ ! -f CMakeCache.txt ]]; then
+    echo run from build directory >&2
+    exit 1
+fi
+
+NEW_VERSION=$1
+
+NEW_VERSION_DIR=storage_format_version_$NEW_VERSION/
```
witchel, line 26: That is a long directory name.  Should I just get over it?

libtsuba/test/CMakeLists.txt
```diff
@@ -1,22 +1,38 @@
...
-add_test(NAME clean-parquet COMMAND ${CMAKE_COMMAND} -E rm -rf "${CMAKE_CURRENT_BINARY_DIR}/parquet-test-wd")
-set_tests_properties(clean-parquet PROPERTIES FIXTURES_SETUP parquet-ready LABELS quick)
+set(name manifest)
+set(test_name ${name}-test)
+add_executable(${test_name} manifest.cpp)
+target_link_libraries(${test_name} tsuba)
+target_include_directories(${test_name} PRIVATE ../src)
+if(KATANA_USE_INPUTS)
+  add_dependencies(${test_name} input)
+endif()
```
witchel, line 8: I'm a little unclear on who sets KATANA_USE_INPUTS and why it needs this check needs to be propagated to so many places.  Doesn't this test specify an input from ${BASEINPUT}/propertygraphs?  Doesn't that mean it is using inputs?

tools/uprev-rdg-storage-format-version-worker/uprev-rdg-storage-format-version-worker.cpp
```diff
@@ -0,0 +1,79 @@
...
+        "in memory graph from load previous storage_format_version does not "
+        "match in memory graph loaded from new storage_format_version");
+  }
+
+  KATANA_LOG_WARN("uprev of {} stored at {}", input_rdg, g2_rdg_file);
+}
+
+int
+main(int argc, char** argv) {
+  katana::SharedMemSys sys;
```
witchel, line 73: If this utility can be run in distributed mode, and I think it shoud, this should be distributed memory.

tools/uprev-rdg-storage-format-version-worker/uprev-rdg-storage-format-version-worker.cpp
```diff
@@ -0,0 +1,79 @@
...
+static cll::opt<std::string> InputFile(
+    cll::Positional, cll::desc("<input rdg file>"), cll::Required);
+
+static cll::opt<std::string> OutputFile(
+    cll::Positional, cll::desc("<output rdg file>"), cll::Required);
+
+katana::PropertyGraph
+LoadGraph(const std::string& rdg_file) {
+  KATANA_LOG_ASSERT(!rdg_file.empty());
+  auto g_res = katana::PropertyGraph::Make(rdg_file, tsuba::RDGLoadOptions());
```
witchel, line 32: We can't do this, can we?  It probably doesn't matter.
`auto g_res = std::move(KATANA_CHECKED(katana::PropertyGraph::Make(rdg_file, tsuba::RDGLoadOptions())));`

tools/uprev-rdg-storage-format-version-worker/uprev-rdg-storage-format-version-worker.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void
+UprevGraph(std::string& input_rdg, std::string& output_path) {
+  katana::PropertyGraph g = LoadGraph(input_rdg);
+  std::string g2_rdg_file = StoreGraph(&g, output_path);
+  katana::PropertyGraph g2 = LoadGraph(g2_rdg_file);
+
+  if (!g.Equals(&g2)) {
+    KATANA_LOG_WARN("{}", g.ReportDiff(&g2));
```
witchel, line 60: Nice!

witchel (approved):
Thanks Evan, this is very useful tooling.

It looks like you are pretty conversant with the "create a temporary working directory" ctest idiom.  This is cleaner than I thought it would be.

inputs/uprev_rdg_storage_format_version.sh
e-mcginnis: I'm a fan of verbosity and clarity in organization. I could see an argument for just using "v#" but RDGs themselves already have versions so I think that would be confusing.

inputs/uprev_rdg_storage_format_version.sh
e-mcginnis: I'll add a TODO to consider reworking this to python

libtsuba/test/CMakeLists.txt
e-mcginnis: So previously if you built just `libtsuba` or just `libgalois`, you wouldn't build `input` as well and your tests would fail. So I wanted to add `input` as a dependency, but in some CI cases `KATANA_USE_INPUTS` is false and the inputs get provided in a different way. So we have to do this.

tools/uprev-rdg-storage-format-version-worker/uprev-rdg-storage-format-version-worker.cpp
e-mcginnis: yeah, we likely can.

PR #703, opened 2021-11-22
RDGTopology: allow storing a topology without edges [KAT-2579]

tylershunt:
A rare case since our partitioner assigns edges, but technically
possible if there are nodes without any edges in the graph.

I missed this point in my recent PR with point fixes for these cases.


roshandathathri (approved):
LGTM

PR #704, opened 2021-11-23
PropertyGraph: remove type properties after inference during V1 convert

tylershunt:
I noticed that RDGs we convert still keep references to these files.
They're not useful and can actually be confusing for cases when a caller
want's to iterate over properties.

They also cause problems for some code paths that aren't aware they
should ignore uint8 properties (like Ensure{Node,Edge}PropertyLoaded)
which is what got me looking at this.

Eventually, I think we should re-introduce support for the uint8 type,
but our codebase's dislike for the type is too far reaching to adress
in a release bugfix PR.

### DIFF
--- a/libgalois/src/PropertyGraph.cpp
+++ b/libgalois/src/PropertyGraph.cpp
@@ -347,9 +347,13 @@ katana::PropertyGraph::ConstructEntityTypeIDs() {
   node_entity_type_manager_ = EntityTypeManager{};
   node_entity_type_ids_ = EntityTypeIDArray{};
   node_entity_type_ids_.allocateInterleaved(num_nodes());
-  KATANA_CHECKED(EntityTypeManager::AssignEntityTypeIDsFromProperties(
-      num_nodes(), rdg_.node_properties(), &node_entity_type_manager_,
-      &node_entity_type_ids_));
+  auto node_props_to_remove =
+      KATANA_CHECKED(EntityTypeManager::AssignEntityTypeIDsFromProperties(
+          num_nodes(), rdg_.node_properties(), &node_entity_type_manager_,
+          &node_entity_type_ids_));
+  for (const auto& node_prop : node_props_to_remove) {
+    KATANA_CHECKED(RemoveNodeProperty(node_prop));
+  }
 
   int64_t total_num_edge_props = full_edge_schema()->num_fields();
   for (int64_t i = 0; i < total_num_edge_props; ++i) {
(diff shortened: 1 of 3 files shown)


witchel (approved):
Nice catch.  This will help.
Yeah, I guess we can give uint8 some love, but its a bit of a sticky wicket.

libgalois/src/PropertyGraph.cpp
```diff
@@ -363,9 +364,18 @@ katana::PropertyGraph::ConstructEntityTypeIDs() {
...
   edge_entity_type_ids_.allocateInterleaved(num_edges());
-  KATANA_CHECKED(EntityTypeManager::AssignEntityTypeIDsFromProperties(
-      num_edges(), rdg_.edge_properties(), &edge_entity_type_manager_,
-      &edge_entity_type_ids_));
+  auto edge_props_to_remove =
+      KATANA_CHECKED(EntityTypeManager::AssignEntityTypeIDsFromProperties(
+          num_edges(), rdg_.edge_properties(), &edge_entity_type_manager_,
+          &edge_entity_type_ids_));
+
+  for (const auto& node_prop : node_props_to_remove) {
```
roshandathathri: nit: isn't it better to eagerly drop this before we create edge entity types?

roshandathathri (approved):
LGTM

libgalois/src/PropertyGraph.cpp
tylershunt: Good point.

e-mcginnis (approved):
LGTM!

tylershunt:
Requires this fix in enterprise:
https://github.com/KatanaGraph/katana-enterprise/pull/2589

witchel:
Dear Dr. Hunt,

@e-mcginnis and I noticed a problem with v29 inputs, which have a stray City uint8 property in ldbc_003.  This PR fixes that problem and once it merges Evan will rev the inputs.  Is there anything holding it up? #2647?

Thank you ,
Emmett

tylershunt:
> Dear Dr. Hunt,
> 
> @e-mcginnis and I noticed a problem with v29 inputs, which have a stray City uint8 property in ldbc_003. This PR fixes that problem and once it merges Evan will rev the inputs. Is there anything holding it up? #2647?
> 
> Thank you , Emmett

Nothing at all holding it up. @e-mcginnis has been back channeling with me and and has taken ownership of this bit in #714

PR #705, opened 2021-11-23
View cache stores topologies behind shared pointers.

yan-zaretskiy:
This PR changes how topologies are stored in the PG view cache, to use shared pointers instead of unique pointers. This is the preliminary step in support of safe destruction of unused topologies.

### DIFF
--- a/libgalois/include/katana/GraphTopology.h
+++ b/libgalois/include/katana/GraphTopology.h
@@ -239,15 +239,15 @@ public:
     return edge_property_index(eid);
   }
 
-  static std::unique_ptr<EdgeShuffleTopology> MakeTransposeCopy(
+  static std::shared_ptr<EdgeShuffleTopology> MakeTransposeCopy(
       const PropertyGraph* pg);
-  static std::unique_ptr<EdgeShuffleTopology> MakeOriginalCopy(
+  static std::shared_ptr<EdgeShuffleTopology> MakeOriginalCopy(
       const PropertyGraph* pg);
 
-  static std::unique_ptr<EdgeShuffleTopology> Make(
+  static std::shared_ptr<EdgeShuffleTopology> Make(
       PropertyGraph* pg, const tsuba::RDGTopology::TransposeKind& tpose_todo,
       const tsuba::RDGTopology::EdgeSortKind& edge_sort_todo) noexcept {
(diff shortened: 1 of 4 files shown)


libgalois/include/katana/GraphTopology.h
```diff
@@ -709,7 +709,8 @@ class KATANA_EXPORT ProjectedTopology : public GraphTopologyTypes {
 template <typename Topo>
 class KATANA_EXPORT BasicTopologyWrapper : public GraphTopologyTypes {
 public:
-  explicit BasicTopologyWrapper(const Topo* t) noexcept : topo_ptr_(t) {
+  explicit BasicTopologyWrapper(const std::shared_ptr<const Topo>& t) noexcept
+      : topo_ptr_(t) {
```
rob-isaac: think convention is to pass by value and move, which allows users to move their shared pointer into the constructor if they are done with it.

libgalois/include/katana/GraphTopology.h
```diff
@@ -1187,7 +1191,8 @@ class KATANA_EXPORT EdgeTypeAwareTopology
   bool is_valid() const noexcept { return edge_shuff_topo_->is_valid(); }
 
   void invalidate() noexcept {
-    const_cast<EdgeShuffleTopology*>(edge_shuff_topo_)->invalidate();
+    std::const_pointer_cast<EdgeShuffleTopology>(edge_shuff_topo_)
```
rob-isaac, line 1195: we should figure out how to avoid const cast at some point, please add a todo

libgalois/include/katana/PropertyGraph.h
```diff
@@ -650,7 +650,7 @@ class KATANA_EXPORT PropertyGraph {
     return MakeResult(std::move(array));
   }
 
-  const GraphTopology& topology() const noexcept { return topology_; }
+  const GraphTopology& topology() const noexcept { return *topology_.get(); }
```
rob-isaac: can just do `*topology`

rob-isaac (approved):
LGTM, have some small style nits.

libgalois/include/katana/GraphTopology.h
yan-zaretskiy: Good point.

PR #706, opened 2021-11-23
RDG Slice work

scober:
This PR got a little out of hand, but it has a few main thrusts:

- All OpaqueIDs can now be passed directly to fmt functions (this was for convenience while debugging other issues)
- One-off interface improvement to `divideNodesBinarySearch()`
- Improve `RDGPrefix` and `RDGSlice` compatibility with partitioned graphs. This involved moving more logic out of `RDGSlice` and into `RDGCore` as well as providing the `GetPerPartitionCounts()` static function for helping load slices of partitioned graphs.
- On-demand loading and unloading of some metadata arrays in `RDGSlice`

### DIFF
--- a/libtsuba/include/tsuba/RDGPrefix.h
+++ b/libtsuba/include/tsuba/RDGPrefix.h
@@ -16,7 +16,8 @@ class RDGManifest;
 /// partitioning decisions
 class KATANA_EXPORT RDGPrefix {
 public:
-  static katana::Result<RDGPrefix> Make(RDGHandle handle);
+  static katana::Result<RDGPrefix> Make(
+      RDGHandle handle, uint32_t partition_id = 0);
 
   uint64_t num_nodes() const { return prefix_->header.num_nodes; }
   uint64_t num_edges() const { return prefix_->header.num_edges; }
@@ -49,7 +50,8 @@ private:
   uint64_t view_offset_;
   const CSRTopologyPrefix* prefix_{nullptr};
 
-  static katana::Result<RDGPrefix> DoMakePrefix(const RDGManifest& manifest);
+  static katana::Result<RDGPrefix> DoMakePrefix(
+      const RDGManifest& manifest, uint32_t partition_id);
 };
 
 /// EntityTypeIDArrayHeader describes the header in the on disk representation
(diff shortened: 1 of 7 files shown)


libtsuba/include/tsuba/RDGPrefix.h
```diff
@@ -16,7 +16,8 @@ class RDGManifest;
 /// partitioning decisions
 class KATANA_EXPORT RDGPrefix {
 public:
-  static katana::Result<RDGPrefix> Make(RDGHandle handle);
+  static katana::Result<RDGPrefix> Make(
```
e-mcginnis, line 19: nit: I would prefer to avoid defaulting `partition_id` if reasonably possible.
Are there a ton of use cases where the caller of `Make` doesn't have any way of knowing what `partition_id` it should use?

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -46,7 +46,7 @@ class KATANA_EXPORT RDGSlice {
   };
 
   static katana::Result<RDGSlice> Make(
-      RDGHandle handle, const SliceArg& slice,
+      RDGHandle handle, const SliceArg& slice, const uint32_t partition_id = 0,
```
e-mcginnis, line 49: ditto

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -119,16 +130,13 @@ class KATANA_EXPORT RDGSlice {
 
   std::unique_ptr<RDGCore> core_;
   // NB: we intentionally do not include a property cache here because that will
-  // complicate the property cache logic; one could be added in the future if it
-  // seems necessary
-
-  void InitArrowVectors();
```
e-mcginnis, line 125: Its not completely clear to me why we no longer would need to `InitArrowVectors()`? Does RDGCore now handle this?

scober: Yes it does

libtsuba/include/tsuba/RDGPrefix.h
scober: The largest benefit of this is that it makes the interface change non-breaking. I would be happy to come back and remove the default after all uses of this interface have been updated.

rossbach (approved):
LGTM

PR #707, opened 2021-11-24
libsupport: Remove redundant fmt specialization

ddn0:
support: Remove redundant fmt specialization

    fmt picks up on ostream& operator<<(ostream&, T) overloads and there is
    typically no need to specialize printing functionality for fmt if
    there is already an ostream overload.

    This also has the benefit of supporting both fmt::print and std::cout
    with the same customization point.

    There is an issue with fmt<8 when using fmt::join(items, ...) when Item
    only overloads operator<<, which is what prompted the change in
    153ab8a80 that this commit reverts. This issue is larger than printing
    OpaqueIDs though and specific to fmt::join.

This partially reverts https://github.com/KatanaGraph/katana/pull/706

### DIFF
--- a/libsupport/test/opaque-id.cpp
+++ b/libsupport/test/opaque-id.cpp
@@ -33,7 +33,31 @@ static_assert(
 static_assert(
     std::is_same_v<decltype(TestLongOrdered::sentinel()), TestLongOrdered>);
 
+struct IntID : public katana::OpaqueIDLinear<IntID, int> {
+  using OpaqueIDLinear::OpaqueIDLinear;
+};
+
+void
+TestPrint() {
+  int value = 1;
+  IntID id(value);
+
+  std::stringstream expected;
+  expected << value;
+
+  fmt::memory_buffer fmt_buf;
+  fmt::format_to(std::back_inserter(fmt_buf), "{}", id);
+
+  std::stringstream stl_buf;
+  stl_buf << id;
+
+  KATANA_LOG_ASSERT(to_string(fmt_buf) == expected.str());
+  KATANA_LOG_ASSERT(stl_buf.str() == expected.str());
+}
+
 int
 main() {
+  TestPrint();
+
   return 0;
 }
--- a/docs/reference/cxx/index.rst
+++ b/docs/reference/cxx/index.rst
@@ -5,5 +5,7 @@ C++ Library
 .. toctree::
    arrow
    error-handling
+   logging
    property-graph
+   strings
    tracing
--- a/docs/reference/cxx/logging.rst
+++ b/docs/reference/cxx/logging.rst
@@ -1,0 +1,28 @@
+=======
+Logging
+=======
+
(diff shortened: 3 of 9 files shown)


arthurp (approved):
I really thought this was needed. But maybe not. Do be careful to make sure this works on broken compilers like CUDA and old stuff that we still use.

libsupport/test/opaque-id.cpp
```diff
@@ -33,7 +33,20 @@ static_assert(
...
 
+struct IntID : public katana::OpaqueIDLinear<IntID, int> {
+  using OpaqueIDLinear::OpaqueIDLinear;
+};
+
+void
+TestPrint() {
+  IntID id(1);
+  fmt::print("{}\n", id);
+  std::cout << id << "\n";
```
pyrito: Is it possible to verify that these are being printed as expected?

pyrito (approved):

tylershunt (approved):
lgtm

ddn0:
It turns out the issue that @scober was facing is that `fmt::join` does not work with classes that only overload `ostream operator<<`. This was fixed in `libfmt` 8.

I still think the solution is not to add `libfmt` specific customization.

If necessary, we can add `katana::Join` and then remove it once we no longer support `libfmt` < 8.

scober:
My highest priority is to prevent the next person who faces this issue from re-implementing the `formatter` specialization. I think a comment in the `OpaqueID` base class + `katana::Join()` would be enough to prevent wasted work while we wait to bump `libfmt` versions.

arthurp:
> My highest priority is to prevent the next person who faces this issue from re-implementing the `formatter` specialization. I think a comment in the `OpaqueID` base class + `katana::Join()` would be enough to prevent wasted work while we wait to bump `libfmt` versions.

Serious question: Why not just bump the libfmt version right now? No reason to work around this in a way that will touch a lot of code if we can just upgrade.

ddn0:
> I think a comment in the OpaqueID base class

`fmt:join` not working with ostream is not specific to OpaqueID, so given the choice between writing a comment in OpaqueID (that no one will read) or just being more circumspect about `libfmt` specializations (that reviewers won't enforce), either seem to have about the same strength, though I have a preference for the latter because (a) this is a general issue and not specific to OpaqueID and (b) enforcing code norms is part of reviewing.

I can add something to the contributing / code style document about how to do `libfmt` specializations if you want.

> Why not just bump the libfmt version right now?

libfmt>8 is not widely available on all system platforms we support (e.g., ubuntu 20.04) and the workaround is very similar to what  you would write with libfmt >8 anyway (fmt::join => katana::Join), is easy to gradually refactor later, and the only con is that it makes extra copies. It seems better to continue supporting a wide variety of libfmt versions if we can.

e-mcginnis (approved):
lgtm

scober (approved):

PR #708, opened 2021-11-24
Add undirected view to Kcore

gurbinder533:


uditagar:
looks fine as long as the output is same. thanks

amazloumi (approved):
thank you. looks fine. we should do it for all apps that require symmetric graphs.

gurbinder533:
> thank you. looks fine. we should do it for all apps that require symmetric graphs.

Yes, exactly. We have already started. Leiden and Louvain are in PR. We need to do for CC and CDLP. Can you do it for CDLP?

lonestar/analytics/cpu/k-core/kcore_cli.cpp
```diff
@@ -107,7 +108,9 @@ main(int argc, char** argv) {
   }
 
   tsuba::TxnContext txn_ctx;
-  if (auto r = KCore(pg.get(), kCoreNumber, "node-in-core", &txn_ctx, plan);
+  if (auto r = KCore(
+          pg.get(), kCoreNumber, "node-in-core", &txn_ctx, symmetricGraph,
+          plan);
```
amazloumi, line 113: isn't it better to pass the extra arguments like symmetricGraph through "plan"?

amazloumi:
> > thank you. looks fine. we should do it for all apps that require symmetric graphs.
> 
> Yes, exactly. We have already started. Leiden and Louvain are in PR. We need to do for CC and CDLP. Can you do it for CDLP?

Yes, I will do it for CDLP. Actually already started.

lonestar/analytics/cpu/k-core/kcore_cli.cpp
gurbinder533: Not really. Plan should only have things which do not change the output. Symmetric flag will change the output if used incorrectly. Also, very soon we will push this functionality to the RDG itself so that we can auto check if graph is symmetric or not.

amazloumi: I mean, it seems that "plan" is controlling how the app behaves and be run. So we can keep all the configs in "plan". does that make sense?

amazloumi: > 

So in this case; it makes sense to me. thanks for the clarification.

amazloumi: > Not really. Plan should only have things which do not change the output. Symmetric flag will change the output if used incorrectly. Also, very soon we will push this functionality to the RDG itself so that we can auto check if graph is symmetric or not.

So in this case; it makes sense to me. thanks for the clarification.

lonestar/analytics/cpu/k-core/kcore_cli.cpp
```diff
@@ -79,10 +79,11 @@ main(int argc, char** argv) {
   total_timer.start();
 
   if (!symmetricGraph) {
```
amazloumi: I think it is better to show the following warning if the flag is set not vice versa.

uditagar (approved):
looks fine

lonestar/analytics/cpu/k-core/kcore_cli.cpp
gurbinder533: Yeah makes sense. Good Catch.

PR #709, opened 2021-11-24
Switching cc to use undirected view

uditagar:
This PR adds Undirected view support to CC app.


roshandathathri (changes requested):
Please evaluate the impact of the performance change. This is required to merge the PR and should be included in the commit/PR message.

uditagar:
So my conclusion is that the execution time increases by around 25% (even after using undirected view on an already symmetric graph). Most of the time is spent in creating the view/graph. thanks

PR #710, opened 2021-11-28
Separate libgalois from libgraph

amberhassaan:
This PR splits `libgalois` into `libgalois + libgraph`, where
`libgalois` is all the parallel runtime and API stuff, 
while `libgraph` is all the graph related code. 

As a result we get `libkatana_graph` in addition to `libkatana_galois`

Enterprise PR: https://github.com/KatanaGraph/katana-enterprise/pull/2627


arthurp:
Can you please `rebase -i` to make this into conceptually separate commits that can be reviewed separately? Most importantly split the find-and-replace changes from the non-mechanical changes.

libgalois/test/barriers.cpp
```diff
@@ -77,7 +77,7 @@ test(std::unique_ptr<katana::Barrier> b) {
 
 int
 main(int argc, char** argv) {
-  katana::SharedMemSys Katana_runtime;
+  katana::GaloisRT Katana_runtime;
```
ddn0: I don't think `GaloisRT` is any better than `GaloisSys` or even `GaloisRuntime`

amberhassaan: Agree. Will change to `GaloisRuntime`.

amberhassaan:
> Can you please `rebase -i` to make this into conceptually separate commits that can be reviewed separately? Most importantly split the find-and-replace changes from the non-mechanical changes.

Good suggestion. Please check now.

libgalois/include/katana/GaloisRT.h
```diff
@@ -17,35 +17,35 @@
...
+/// A GaloisRT represents global initialization required for the shared
 /// memory subsystem, i.e., thread pools and barriers. As a side-effect of
 /// construction, this class sets global internal state.
 ///
 /// Data structures that require per-thread allocation typically ask for the
 /// thread pool. If their construction is not guaranteed to happen after the
-/// construction of a SharedMem, initialization races can occur.
-class KATANA_EXPORT SharedMem {
+/// construction of a GaloisRT, initialization races can occur.
+class KATANA_EXPORT GaloisRT {
```
arthurp: If this is the `GaloisRuntime` what do we call the distributed runtime?

libgraph/test/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# Keep alphabetical order
```
arthurp, line 1: I respect that goal. But it's not realistic unless it's checked in CI. Maybe we could have a sort checker we apply to a few selected files.

libgraph/include/katana/SharedMemSys.h
```diff
@@ -3,6 +3,7 @@
```
arthurp, line 3: This needs documentation updates. We need to document what parts of the runtime are ending up in what library. Especially for those of us who got used to the old arrangement.

arthurp (approved):
This needs documentation changes!!! I think it's important to do in this PR, not later. The reason is that people need to know exactly what belongs in `graph` and what belongs in `galois` without it people will just put things in them at random. Please write up the split clearly in the `docs/` tree and probably in the readme too.

Once that's done, LGTM. No need to check back with me.

libgalois/test/barriers.cpp
roshandathathri: Why change in name? SharedMemSys and DistMemSys seemed reasonable to me

amberhassaan: So, we still have `SharedMemSys` in the new `libgraph` because, now, `SharedMemSys` initializes `libtsuba` and `libgalois`

roshandathathri (approved):
LGTM

GaloisRT -> GaloisRuntime

e-mcginnis (approved):
LGTM. Thanks for doing this Amber, looking forward to removing all of my TODOs that required this!

PR #711, opened 2021-11-29
[KAT-2552] Add inheritance in requirements.

arthurp:
For both labels and packaging systems.

This allows improved support for Ubuntu 18.04 as a dev platform and future changes that will appear between ubuntu version. It may be useful for other things. Ubuntu 18.04 non-conda dev machines are still going to be an issue because I'm pretty sure python API builds will not work and probably never did. Conda builds will still work fine.


scripts/katana_requirements/__main__.py
```diff
@@ -61,6 +62,12 @@ def setup_general_arguments(parser):
         default=[],
         action="append",
     )
+    parser.add_argument(
+        "--packaging-system",
+        "-p",
+        help="The packaging system to use. Only print packages which support this system.",
+        type=str,
```
ddn0, line 70: I don't suppose we can get a list of valid values here?

scripts/setup_ubuntu.sh
```diff
@@ -110,10 +110,15 @@ apt update
...
 apt install --yes --quiet python3-yaml python3-packaging
 
+if [ "$VERSION" == "18" ]; then
+  PKG_SYS_SUFFIX="-18.04"
+else
+  PKG_SYS_SUFFIX=""
+fi
 # Use the requirements tool to install apt packages with: apt-get satisfy --allow-downgrades --yes --quiet
-"$REPO_ROOT"/scripts/requirements install -a--allow-downgrades -a--yes -a--quiet -l apt -l apt/dev -f apt
+"$REPO_ROOT"/scripts/requirements install -a--allow-downgrades -a--yes -a--quiet -l apt -l apt/dev -p "apt$PKG_SYS_SUFFIX"
```
ddn0: I suppose `argparse` is okay with `-a--allow-downgrades` but would it be possible to be a little more explicit with `--arg=--allow-downgrades`?

scripts/setup_ubuntu.sh
```diff
@@ -122,7 +127,7 @@ fi
 
 # --no-binary is required to cause the pip package to use the debian package's native binaries.
 # https://lists.apache.org/thread.html/r4d2e768c330b6545649e066a1d9d1846ca7a3ea1d97e265205211166%40%3Cdev.arrow.apache.org%3E
-PYARROW_WITH_PARQUET=1 run_as_original_user "$REPO_ROOT"/scripts/requirements install -a--upgrade -a--no-binary -a:all:  -l deb/pip-no-binary -f pip
+PYARROW_WITH_PARQUET=1 run_as_original_user "$REPO_ROOT"/scripts/requirements install -a--upgrade -a--no-binary -apyarrow  -l deb/pip-no-binary -p pip
```
ddn0: ditto

scripts/requirements
```diff
@@ -4,5 +4,6 @@ REPO_ROOT="$(cd "$(dirname "$0")/.."; pwd)"
 
 export PYTHONPATH="$REPO_ROOT/external/katana/scripts:$REPO_ROOT/scripts:${PYTHONPATH:+:${PYTHONPATH}}"
 python3 -m katana_requirements "$@" || \
-  echo "HINT: If there are ImportErrors make sure you have the yaml and packaging python packages" \
-        "installed. (Apt packages: python3-yaml, python3-packaging)"
+  (echo "HINT:   If there are ImportErrors make sure you have the yaml and packaging python packages" \
```
ddn0: This hint is printed for all non-zero exit values, which is a little confusing.

It seems just as easy to make a dummy module that tries to `import yaml` and prints this hint when there is an actual `ImportError` than to ask users to scan lines of errors to know when to disregard this hint.

ddn0 (approved):

scripts/katana_requirements/__main__.py
arthurp: The list is from the config file. I will add a reference to the `requirements packaging-systems` which does list the options.

katana_requirements.yaml
```diff
@@ -183,12 +226,14 @@ libxml2:
     - conda
     - apt
 llvm:
+  name_overrides:
+    apt: llvm-10
```
witchel, line 230: `name_overrides` specify a specific version?

katana_requirements.yaml
```diff
@@ -225,10 +270,13 @@ nlohmann_json:
     - conda
     - conan
 numba:
+  name_overrides:
+    apt-18.04: null
+    pip-18.04: null
```
witchel, line 275: But on a package name a `name_override` specifies an OS version?

scripts/katana_requirements/__main__.py
```diff
@@ -71,38 +79,44 @@ def setup_general_arguments(parser):
...
 
-def labels_subcommand(args, inputs, data):
+def labels_subcommand(args, inputs, data: Requirements):
     # pylint: disable=unused-argument
 
     print("Loaded data from files: " + ", ".join(str(f) for f in inputs))
 
     print("Labels listed in data files:")
-    print_str_table(data.labels)
+    print_str_table({k: f"{v.description} (inherits from: {list(v.inherits)})" for k, v in data.labels.items()})
```
witchel, line 88: Whoa, lot going on there.  I actually like how it is formatted below.

scripts/setup_ubuntu.sh
```diff
@@ -110,10 +110,15 @@ apt update
...
 apt install --yes --quiet python3-yaml python3-packaging
 
+if [ "$VERSION" == "18" ]; then
+  PKG_SYS_SUFFIX="-18.04"
+else
+  PKG_SYS_SUFFIX=""
+fi
 # Use the requirements tool to install apt packages with: apt-get satisfy --allow-downgrades --yes --quiet
-"$REPO_ROOT"/scripts/requirements install -a--allow-downgrades -a--yes -a--quiet -l apt -l apt/dev -f apt
+"$REPO_ROOT"/scripts/requirements install --arg=--allow-downgrades --arg=--yes --arg=--quiet --label apt --label apt/dev --packaging-system "apt$PKG_SYS_SUFFIX"
```
witchel, line 119: I have to admit, I didn't know the `-a--allow-downgrades` syntax and I found it confusing.  I prefer the explicit `--arg`.

witchel (approved):
Pretty cool stuff, thank you for doing this.

katana_requirements.yaml
arthurp: It shouldn't, but many package names have major versions in there, so there isn't much to do about it.

katana_requirements.yaml
arthurp: A null name disables a packages for the specific packaging system (the keys in the YAML object here), which in this case is associated with an OS version. Does that clarify? I'm not really sure what you are asking.

scripts/katana_requirements/__main__.py
arthurp: Yea. I wasn't sure how to make it much clearer, so I just dumped all the data in an moved on.

scripts/setup_ubuntu.sh
arthurp: Yea. `-a` was  a bad idea in a script. This reminds me of my own advice, never use contractions in code without a really good reason and good documentation. I should listen to me. ;-)

arthurp:
I'm really tight on time. So I'm gonna merge this. @witchel, we should keep talking about this though if anything is unclear. Documentation improvements would be good.

katana_requirements.yaml
witchel: Yes, the packaging system is tied to the OS version.  I didn't even realize the null value was disabling that packaging system.

PR #712, opened 2021-11-29
Ensure there is always at least an empty topology

yan-zaretskiy:
A quick fix, to make `katana-enterprise` tests pass.


roshandathathri (approved):
LGTM

PR #713, opened 2021-11-30
add max-height style for notebook outputs

origandrew:
Some basic css to add scrolling to large notebook outputs. IMO, better to allow for verbose/long outputs than to alter code to try to truncate.

After:
Screen Shot 2021-12-01 at 1 05 11 PM

Before:
Screen Shot 2021-12-01 at 1 04 29 PM

Unfortunately, does not handle multiple outputs from a single input (running a loop of different graph operations - queries/imports) which needs alteration in how we output progress bars:
Screen Shot 2021-12-01 at 1 07 16 PM

### DIFF
--- a/docs/_static/style.css
+++ b/docs/_static/style.css
@@ -1,3 +1,3 @@
 div.nboutput.container div.output_area {
-	max-height: 60vh;
+	max-height: 50vh;
 }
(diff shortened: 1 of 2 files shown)


ddn0 (approved):
A screenshot before and after would be good to demonstrate the benefit and serve as a "test"

origandrew:
Thanks. Added screenshots and tweaked to 50vh after taking screenshots on a laptop.

PR #715, opened 2021-12-01
Graph helpers: replace some unsigned with size_t

scober:
This doesn't modify large vectors or anything where adding a few bytes
will blow up the memory use.

It is primarily to allow
divideNodesBinarySearch() to accept std::vector<size_t> instead of
std::vector<unsigned> for the scaleFactor parameter. This gives that
parameter some more expressiveness and makes it safer by making it
impossible to pass in a vector of too-large values.

### DIFF
--- a/libgalois/include/katana/GraphHelpers.h
+++ b/libgalois/include/katana/GraphHelpers.h
@@ -130,8 +130,8 @@ findIndexPrefixSum(
  *
  * @returns The total number of blocks to split among all divisions
  */
-KATANA_EXPORT uint32_t determine_block_division(
-    uint32_t numDivisions, std::vector<unsigned>& scaleFactor);
+KATANA_EXPORT size_t
+determine_block_division(size_t numDivisions, std::vector<size_t>& scaleFactor);
 
 }  // end namespace internal
 
@@ -169,10 +169,10 @@ KATANA_EXPORT uint32_t determine_block_division(
  */
 template <typename PrefixSumType, typename NodeType = uint64_t>
 auto
-divideNodesBinarySearch(
+DivideNodesBinarySearch(
     NodeType numNodes, uint64_t numEdges, size_t nodeWeight, size_t edgeWeight,
     size_t id, size_t total, PrefixSumType& edgePrefixSum,
-    std::vector<unsigned> scaleFactor = std::vector<unsigned>(),
+    std::vector<size_t> scaleFactor = std::vector<size_t>(),
     uint64_t edgeOffset = 0, uint64_t nodeOffset = 0) {
   typedef boost::counting_iterator<NodeType> iterator;
   typedef boost::counting_iterator<uint64_t> edge_iterator;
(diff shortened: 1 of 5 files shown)


libgalois/include/katana/GraphHelpers.h
```diff
@@ -172,7 +172,7 @@ auto
 divideNodesBinarySearch(
```
l-hoang: I'm assuming changing this to `DivideNodesBinarySearch` would be way too intrusive
for such a simple commit (to match function convention with the rest of katana)?

libgalois/include/katana/GraphHelpers.h
```diff
@@ -245,15 +245,29 @@ divideNodesBinarySearch(
                                   edgePrefixSum, nodesUpper - 1 + nodeOffset) -
                                   edgeOffset;
 
-  // katana::gDebug("Unit ", id, " nodes ", nodesLower, " to ",
-  //               nodesUpper, " edges ", edgesLower, " ",
-  //               edgesUpper);
+  katana::gDebug(
+      "Unit ", id, " nodes ", nodesLower, " to ", nodesUpper, " edges ",
+      edgesLower, " ", edgesUpper);
```
l-hoang: you should probably recomment this out or remove it entirely (this function is called quite a bit in distributed execution IIRC, and it will blow up debug output)

l-hoang (approved):
looks sane; only thing that bugs me is that these functions don't match the
rest of Katana naming convention at the moment, but that doesn't really
have anything to do with the content of the commit

libgalois/include/katana/GraphHelpers.h
scober: I would be okay with adding an overload that has the desired name. Then we could remove uses of the old version incrementally before removing the original.
EDIT: I actually already created an overload for this function to ease the transition of callers from `unsigned` to `size_t`. I will change the name of the new function to be `DivideNodesBinarySearch` and leave the compatibility overload as-is.

libgalois/include/katana/GraphHelpers.h
scober: I think I uncommented this to see if it would compile. I haven't seen `katana::gDebug` before. I guess it does! I will probably just remove it.

libgalois/include/katana/GraphHelpers.h
roshandathathri: Please don't add overloads. It can be confusing. Either remove it or continue using the old one.

scober: My issue with removing it is that that would be an interface modification. And an interface modification always requires a tightly coupled enterprise PR. If we instead do an interface addition followed by a migration to the new interface followed by an interface deletion then we can get the same effect without ever having to coordinate PRs across repos.

roshandathathri: Correction: Either replace with new one or continue using the old one.

I get your point. That's not ideal. If you're doing both additional of new one and removal of old one in quick succession, sure, that's fine.

scober: That is my intention. The overload is just a temporary measure to simplify the transition.

PR #719, opened 2021-12-01
RDGSlice: out of core properties

scober:
`RDGSlice` can now load and remove property columns on demand. The interface for this is closer to the `RDGSlice` interface for loading and removing metadata arrays than it is to the `RDG` interface for loading and unloading properties. That is because `RDGSlice` is read-only and provides only a small subset of the interface provided by `RDG`.

### DIFF
--- a/libtsuba/include/tsuba/RDGSlice.h
+++ b/libtsuba/include/tsuba/RDGSlice.h
@@ -94,6 +94,10 @@ public:
   /// The semantics of these two functions are about memory use. So they set
   /// their respective arrays to empty, removing them from memory without
   /// semantically removing them from the object.
+  katana::Result<void> unload_local_to_user_id();
+  katana::Result<void> unload_local_to_global_id();
+  /// deprecated duplicates of the unload_* functions for
+  /// backwards-compatibility
   katana::Result<void> remove_local_to_user_id();
   katana::Result<void> remove_local_to_global_id();
 
--- a/libtsuba/include/tsuba/RDGSlice.h
+++ b/libtsuba/include/tsuba/RDGSlice.h
@@ -46,7 +46,7 @@ public:
   };
 
   static katana::Result<RDGSlice> Make(
-      RDGHandle handle, const SliceArg& slice, const uint32_t partition_id = 0,
+      RDGHandle handle, const SliceArg& slice, uint32_t partition_id = 0,
       const std::optional<std::vector<std::string>>& node_props = std::nullopt,
       const std::optional<std::vector<std::string>>& edge_props = std::nullopt);
 
@@ -73,6 +73,30 @@ public:
   std::shared_ptr<arrow::Schema> full_edge_schema() const;
   const std::shared_ptr<arrow::Table>& node_properties() const;
   const std::shared_ptr<arrow::Table>& edge_properties() const;
+
+  // NB: The following interface differs from the one provided by RDG for
+  // property loading/unloading. You can't remove a property by its
+  // index in the loaded property table and you can't specify an index for a
+  // loaded property in that table. These differences are because RDGSlice has
+  // many fewer users than RDG and those users require a less rich interface.
+  // There is also no upsert call because RDGSlice is read-only.
+
+  /// load the specified node property and append it to the Table returned by
+  /// node_properties()
+  /// @param name the name of the property in full_node_schema()
+  katana::Result<void> load_node_property(const std::string& name);
+  /// remove the specified node property from the Table returned by
+  /// node_properties()
+  /// @param name the name of the property in full_node_schema()
+  katana::Result<void> unload_node_property(const std::string& name);
+  /// load the specified edge property and append it to the Table returned by
+  /// edge_properties()
+  /// @param name the name of the property in full_edge_schema()
+  katana::Result<void> load_edge_property(const std::string& name);
+  /// remove the specified edge property from the Table returned by
+  /// edge_properties()
+  /// @param name the name of the property in full_edge_schema()
+  katana::Result<void> unload_edge_property(const std::string& name);
 
   // topology and friends
   const FileView& topology_file_storage() const;
--- a/libtsuba/src/RDGSlice.cpp
+++ b/libtsuba/src/RDGSlice.cpp
@@ -11,6 +11,66 @@
 #include "tsuba/RDGPrefix.h"
 #include "tsuba/RDGTopology.h"
 
+namespace {
+// empty should be a function that sets the metadata array referred to by
+// array_name to empty when called - see load_local_to_global_id() for an
+// example.
+// This is necessary because depending on the version of the RDG we are loading,
+// the desired metadata array might not exist and the semantics of this function
+// are that it will "load" an empty array in that case. And unfortunately
+// array_name is not enough information on its own to for this function to empty
+// the array by itself.
+katana::Result<void>
+load_metadata_array(
+    const std::string& array_name, const std::function<void()>& empty,
+    tsuba::RDGCore* core) {
+  auto search_res = core->find_part_prop_info(array_name);
+  if (!search_res) {
+    if (search_res.error() == tsuba::ErrorCode::NotFound) {
+      empty();
+      return katana::ResultSuccess();
+    } else {
+      return search_res.error().WithContext(
+          "searching for array name in prop info list");
+    }
+  }
+
+  std::vector<tsuba::PropStorageInfo*> prop_infos{&search_res.value()};
+  KATANA_CHECKED(AddProperties(
+      core->rdg_dir(), tsuba::NodeEdge::kNeitherNodeNorEdge, nullptr, nullptr,
+      prop_infos, nullptr, [&core](const std::shared_ptr<arrow::Table>& props) {
+        return core->AddPartitionMetadataArray(props);
+      }));
+
+  return katana::ResultSuccess();
+}
+
+// empty should be a function that sets the metadata array referred to by
+// array_name to empty when called - see unload_local_to_global_id() for an
+// example.
+// The semantics of this function are that "unload" means "set to empty". The
+// point of this function is really to free memory and setting the array to
+// empty is the easiest way to do that for metadata arrays.
+katana::Result<void>
+unload_metadata_array(
+    const std::string& array_name, const std::function<void()>& empty,
+    tsuba::RDGCore* core) {
+  auto search_res = core->find_part_prop_info(array_name);
+  if (!search_res) {
+    if (search_res.error() != tsuba::ErrorCode::NotFound) {
+      return search_res.error().WithContext(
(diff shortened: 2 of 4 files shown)


libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -74,6 +74,32 @@ class KATANA_EXPORT RDGSlice {
   const std::shared_ptr<arrow::Table>& node_properties() const;
   const std::shared_ptr<arrow::Table>& edge_properties() const;
 
+  // NB: The following interface differs from the one provided by RDG for
+  // property loading/unloading.
+  // For example, this interface uses "remove" instead of "unload" to emphasize
+  // that property modifications will not be persisted (RDGSlice is read-only).
```
pyrito: This sounds confusing to me. "remove" vs. "unload" don't seem to make the distinction that the modifications will/won't be persisted.

libtsuba/src/RDGSlice.cpp
```diff
@@ -11,6 +11,142 @@
...
+  auto it = std::find_if(
+      core->part_header().part_prop_info_list().begin(),
+      core->part_header().part_prop_info_list().end(),
+      [&array_name](tsuba::PropStorageInfo& psi) {
+        return psi.name() == array_name;
+      });
+
+  if (it == core->part_header().part_prop_info_list().end()) {
+    // caller's function for emptying the array we care about
+    empty();
```
pyrito: I think this should be documented a bit more to explain why the empty function needs to be passed in. I see that this function is used internally mainly, but still worth a bit more explanation since the burden is placed on the caller.

libtsuba/src/RDGSlice.cpp
```diff
@@ -11,6 +11,142 @@
...
+  auto it = std::find_if(
+      core->part_header().part_prop_info_list().begin(),
+      core->part_header().part_prop_info_list().end(),
+      [&name](tsuba::PropStorageInfo& psi) { return psi.name() == name; });
+
+  if (it == core->part_header().part_prop_info_list().end()) {
+    return tsuba::ErrorCode::PropertyNotFound;
+  }
+
+  std::vector<tsuba::PropStorageInfo*> property{&(*it)};
```
pyrito: Interesing. So you're initializing a new vector with an iterator. We can do that?

pyrito: Why are you creating a vector here though? Is that the requirement for `AddPropertySlice`?

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -74,6 +74,32 @@ class KATANA_EXPORT RDGSlice {
...
+  // Other differences are more subtle - you can't remove a property by its
+  // index in the loaded property table and you can't specify an index for a
+  // loaded property in that table. These differences are because RDGSlice has
+  // many fewer users than RDG and those users require a less rich interface.
```
witchel, line 81: and because the index interface is bad.

libtsuba/include/tsuba/RDGSlice.h
witchel: I am also confused.  Does this delete the property?  It only discards the modifications?

libtsuba/src/RDGSlice.cpp
```diff
@@ -11,6 +11,142 @@
 #include "tsuba/RDGPrefix.h"
 #include "tsuba/RDGTopology.h"
 
+namespace {
+katana::Result<void>
+demand_load_metadata_array(
```
witchel: What does demand mean here?

libtsuba/src/RDGSlice.cpp
witchel: It looks to me like he is bracket initializing the vector with a single element that the iterator points to.

libtsuba/src/RDGSlice.cpp
```diff
@@ -11,6 +11,142 @@
...
+    const std::string& name, tsuba::NodeEdge node_edge, tsuba::RDGCore* core) {
+  if (node_edge == tsuba::NodeEdge::kNeitherNodeNorEdge) {
+    return KATANA_ERROR(
+        tsuba::ErrorCode::InvalidArgument,
+        "cannot remove property that is attached to neither nodes nor edges");
+  }
+  auto it = std::find_if(
+      core->part_header().part_prop_info_list().begin(),
+      core->part_header().part_prop_info_list().end(),
+      [&name](tsuba::PropStorageInfo& psi) { return psi.name() == name; });
```
witchel: I'm sort of musing here, but I've seen this pattern of iterating through `part_header().part_prop_info_list()` a bunch.  Should there be a `FilterProperties` function in RDGCore that takes in a function that returns a bool and does this traversal and returns matching `PropStorageInfo`s?

libtsuba/src/RDGSlice.cpp
```diff
@@ -11,6 +11,142 @@
...
+    return KATANA_ERROR(
+        tsuba::ErrorCode::InvalidArgument,
+        "cannot remove property that is not loaded");
+  }
+
+  std::shared_ptr<arrow::Table> table = node_edge == tsuba::NodeEdge::kNode
+                                            ? core->node_properties()
+                                            : core->edge_properties();
+  std::vector<int> indices = table->schema()->GetAllFieldIndices(name);
+  // only one property per name
```
witchel: This seems like a weird way to do it.  We guarantee that there are no duplicate property names (within nodes, within edges).  Why not just use `GetFieldByName`?

libtsuba/src/RDGSlice.cpp
```diff
@@ -289,80 +425,79 @@ tsuba::RDGSlice::local_to_global_id() const {
...
-          [slice = this](const std::shared_ptr<arrow::Table>& props) {
-            return slice->core_->AddPartitionMetadataArray(props);
-          }),
-      "populating local to global id");
+  KATANA_CHECKED(demand_load_metadata_array(
+      RDGCore::kLocalToUserIDPropName,
+      [&]() {
+        core_->set_local_to_user_id(
+            katana::NullChunkedArray(arrow::uint64(), 0));
+      },
```
witchel, line 448: Oh, this is the empty array?  Seems like we are playing fast and loose with types, would arrow::null() also work?  If so, do we need to pass in a lambda?

witchel (commented):
Looks good, but questions linger.

libtsuba/src/RDGSlice.cpp
pyrito: Yeah I see that now. At a first glance I thought he was initializing a vector based on an iterator that pointed to the beginning of some container and then it just used that iterator to create the rest of the new vector. I see that it's just a single element now.

libtsuba/include/tsuba/RDGSlice.h
scober: "remove" vs. "unload" is the distinction that `RDG` currently makes. Remove means "remove from memory + assert not modified", unload means "if necessary, write back update column and update in-memory data structures to point to the new file + remove from memory". I can add comments but I would like to maintain the same verbiage as `RDG`.

libtsuba/include/tsuba/RDGSlice.h
scober: lol. I did wonder why there was an index-based interface.

libtsuba/src/RDGSlice.cpp
scober: "on-demand" - as in "when the user requests it" as opposed to "when the object is constructed".

libtsuba/src/RDGSlice.cpp
scober: I think we could even offer a `SearchForPropInfo(std::string, std::vector<PartPropInfo>)`. I think searching for a single `PartPropInfo` by name is the most common case.

libtsuba/src/RDGSlice.cpp
scober: I could do that. I was looking at the `arrow::Table` api and saw an opportunity to be defensive/assert an invariant. I can drop it.

libtsuba/src/RDGSlice.cpp
scober: I think it is important that it be an empty array of `arrow::uint64()`.

libtsuba/include/tsuba/RDGSlice.h
scober: I added more comments to clarify the intention of the API. But while I was digging around it looked like maybe the `RDG` remove is actually stronger than I thought? It looks like it actually removes the property from the graph, not just from memory. And those are not the semantics of this remove. So maybe we should call these functions unload and the write-only-ness of the interface can just be the lack of an upsert?
UPDATE: I changed the interface to what I described above. Hopefully it is more clear.

libtsuba/src/RDGSlice.cpp
scober: I added some documentation, is it more clear now?

libtsuba/src/RDGSlice.cpp
scober: I changed this name because it seemed like "demand" was causing more confusion than clarity.

libtsuba/src/RDGSlice.cpp
pyrito: Yes it is much clearer now.

libtsuba/src/RDGSlice.cpp
e-mcginnis: Nit
While I think this is kind of elegant, it requires a bit more thinking than I'd prefer. 
Maybe to add a bit of clarity split this into:
```
tsuba::PropStorageInfo* prop = &(*it);
std::vector<tsuba::PropStorageInfo*> property{prop};
```
The compiler optimizes it all the same according to godbolt.

e-mcginnis (approved):
appreciate the clarification on the unload API. 
LGTM

libtsuba/src/RDGSlice.cpp
scober: This has been updated so that there is no longer an iterator at all (searching has been factored out into a `RDGCore` method).

pyrito (approved):
Merci beaucoup, Simon! Tout semble plus propre!

libtsuba/src/RDGSlice.cpp
witchel: Ok, but weird to me to require a lambda that always returns the same thing.  Maybe it is only the same in the code for this PR.

scober: There are other metadata arrays that could make use of this function to load and unload on demand. But the other issue is that these lambdas both use the `core_->set_XXX()` interface. And there is no real connection between the string name of the local to global id array and the function `set_local_to_global_id()`. So we have to pass something to the function letting it know not only what the empty version of the array is, but how to set it.

PR #720, opened 2021-12-01
PropertyGraphRetractor: add interface for manipulating cache

tylershunt:
There are cases where we need explicit control over when properties are
in memory. This exposes interfaces so that the partitioner and other
memory sensitive operations can take a heavier hand in managing property
memory.


pyrito (approved):
LGTM

PR #722, opened 2021-12-02
Avoid backwards dependency of get_inputs_directory on callers

ddn0:
A recent change to get_inputs_directory 7158d308776f3e makes its
behavior dependent on the directory structure of source trees that
include this source tree.

This seems unnecessary. If a caller knows something that
get_inputs_directory does not, the caller can handle that detail itself
rather than ask get_inputs_directory to search around in the caller's
context.

### DIFF
--- a/python/katana/example_data.py
+++ b/python/katana/example_data.py
@@ -11,14 +11,8 @@ from pathlib import Path
 __all__ = ["get_input", "get_input_as_url"]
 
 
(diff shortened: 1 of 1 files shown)


bhargavivadali (approved):

arthurp (approved):
I thought I vetoed this code. :-/ I guess I was not insistent enough, or I dreamed the whole thing.

PR #726, opened 2021-12-02
Document conda cudatoolkit-dev issue and workarounds

scober:
I am concerned that the `/dev/null` workaround is just crowding the docs and we would be better off dropping it. Thoughts?

### DIFF
--- a/docs/contributing/building.rst
+++ b/docs/contributing/building.rst
@@ -67,6 +67,18 @@ configured. Then, create and activate the development environment:
 The ``conda env update`` line can be run later to update your environment. Deactivate your environment
 ``conda deactivate``, then run the update commands, then reactivate ``conda activate katana-dev``.
 
+.. warning::
+
+   Not all conda installation scripts are well-behaved. In particular, the default behavior of the ``cudatoolkit-dev`` package's post-installation script is to write a log to the hard-coded path ``/tmp/cuda-installer.log`` and not delete it. Once one user on a system has written this file all other users will fail to install the package because they won't be able to write the log file. The recommended workaround for this is to set the ``TMP`` variable in your shell environment (if that variable is set, the script will write its log to ``$TMP/cuda-installer.log`` so you can avoid conflicts with other users).
+
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


arthurp (approved):
Let's not include the /dev/null "fix". That seems like a bit of a problem. Setting `TMP` is a reasonable hack. I would also recommend laying the blame right where it belongs: CUDA. this is not a conda problem. If CUDA has even vaguely sane packaging this would not a be an issue. All the CUDA installers suck. If nVidia weren't being actively anti-competative, they could never get away with building such a POS system.

scober:
@arthurp I think you will appreciate this absolute gem of a forum thread that helped us debug the issue: https://forums.developer.nvidia.com/t/change-the-path-of-install-log-when-installing-cuda-toolkit/183179

pyrito (approved):
Sad reacts only

docs/contributing/building.rst
```diff
@@ -67,6 +67,20 @@ configured. Then, create and activate the development environment:
...
 
+.. warning::
+
+   Not all conda installation scripts are well-behaved. In particular, the default behavior of the ``cudatoolkit-dev`` package's post-installation script is to write a log to the hard-coded path ``/tmp/cuda-installer.log`` and not delete it. Once one user on a system has written this file all other users will fail to install the package because they won't be able to write the log file. The recommended workaround for this is to set the ``TMP`` variable in your shell environment (if that variable is set, the script will write its log to ``$TMP/cuda-installer.log`` so you can avoid conflicts with other users).
+
+.. code-block:: bash
+
+   # If you are sharing a system with other users,
+   # you will need to use one of the following workarounds:
+   TMP=/tmp/$USER conda env update --name katana-dev --file $SRC_DIR/conda_recipe/environment.yml
```
srogatch, line 79: Don't you have to create the directory first? `mkdir /tmp/$USER`

srogatch (approved):

docs/contributing/building.rst
scober: That is a good point

PR #727, opened 2021-12-03
[KAT-1688] Detect read/write set for TxnContext from AddNode/EdgeProperties

yige-hu:
Jira issue: [KAT-1688]
PR in enterprise repo: [#2678]

This is the second part of changes related to read/write set detection. It handles read/write set starting from `RDGCore:: AddNodeProperties` and `RDGCore:: AddEdgeProperties` (the first part of porting was on upserts).

This PR mainly affects analytic operations.

Affected API changes allow functions to pass on the `TxnContext`, which contains the transaction's read/write set. The information will be later used for conflict detection during a transaction commit.


yige-hu:
@arthurp There are also some python API changes in this PR as well as the enterprise PR. Please take a look. Thanks!

libgalois/include/katana/PropertyGraph.h
```diff
@@ -198,8 +198,9 @@ class KATANA_EXPORT PropertyGraph {
     uint64_t ApproxMemUse() const { return ropv.ApproxMemUse(); }
 
     Result<void> AddProperties(
-        const std::shared_ptr<arrow::Table>& props) const {
-      return (g->*add_properties_fn)(props);
+        const std::shared_ptr<arrow::Table>& props,
+        tsuba::TxnContext* txn_ctx) const {
```
arthurp: You have the transaction context as the last argument here and the first for the analytics routines. This should be consistent. I think I prefer last, since if we were doing this in a modern language this argument wouldn't exist at all, so hiding it is good.

libgalois/include/katana/analytics/Utils.h
```diff
@@ -52,22 +52,22 @@ DefaultPropertyNames() {
 template <typename NodeProps>
 inline katana::Result<void>
 ConstructNodeProperties(
-    PropertyGraph* pg,
+    tsuba::TxnContext* txn_ctx, PropertyGraph* pg,
     const std::vector<std::string>& names = DefaultPropertyNames<NodeProps>()) {
```
arthurp: See discussion above. The context should be last I think.

libgalois/include/katana/analytics/independent_set/independent_set.h
```diff
@@ -57,8 +57,8 @@ class IndependentSetPlan : public Plan {
 /// The property named output_property_name is created by this function and may
 /// not exist before the call. The created property has type uint8_t.
 KATANA_EXPORT Result<void> IndependentSet(
-    PropertyGraph* pg, const std::string& output_property_name,
-    IndependentSetPlan plan = {});
+    tsuba::TxnContext* txn_ctx, PropertyGraph* pg,
+    const std::string& output_property_name, IndependentSetPlan plan = {});
```
arthurp: The context should be last as above.

python/katana/local/_graph.pyx
```diff
@@ -219,45 +219,49 @@ cdef class GraphBase:
...
         """
         Insert new node properties into this graph.
 
         :param table: A pyarrow Table or other dataframe-like object containing the properties. The table must have
             length ``self.num_nodes()``. (Optional)
+        :param txn_ctx: The tranaction context for passing read write sets.
         :param kwargs: Properties to add. The values must be arrays or sequences of length ``self.num_nodes()``. (Optional)
         """
-        handle_result_void(self.underlying_property_graph().AddNodeProperties(GraphBase._convert_table(table, kwargs)))
+        txn_ctx = TxnContext() if txn_ctx is None else txn_ctx
```
arthurp: I prefer `txn_ctx or TxnContext()`. I admit it's a bit magical, but it really does read clearly. It's either `txn_ctx` or a new context if that's not available, which is very clear once you've seen it. I think it's a good idiom. And it's common in Python.

arthurp: Please apply this throughout the file.

python/katana/local/analytics/_independent_set.pyx
```diff
@@ -151,8 +153,9 @@ def independent_set(Graph pg, str output_property_name,
     """
     output_property_name_bytes = bytes(output_property_name, "utf-8")
     output_property_name_cstr = <string>output_property_name_bytes
+    txn_ctx = TxnContext() if txn_ctx is None else txn_ctx
```
arthurp: Apply the idiom above throughout the whole PR.

python/katana/local/analytics/_k_core.pyx
```diff
@@ -106,7 +107,7 @@ cdef class KCorePlan(Plan):
         return KCorePlan.make(_KCorePlan.Asynchronous())
 
 
-def k_core(Graph pg, uint32_t k_core_number, str output_property_name, KCorePlan plan = KCorePlan()) -> int:
+def k_core(Graph pg, uint32_t k_core_number, str output_property_name, KCorePlan plan = KCorePlan(), TxnContext txn_ctx=None) -> int:
```
arthurp: NIT: put spaces around the `=` in default values.

python/katana/local/_graph.pyx
```diff
@@ -219,45 +219,49 @@ cdef class GraphBase:
 
         return pyarrow_unwrap_table(arrow_table)
 
-    def add_node_property(self, table=None, **kwargs):
+    def add_node_property(self, table=None, TxnContext txn_ctx=None, **kwargs):
```
arthurp: Let's make `txn_ctx` keyword only. I think it should be since the normal usage will never pass it. That will make this declaration:

```python
def add_node_property(self, table=None, *, TxnContext txn_ctx=None, **kwargs):
```

Note the bare star  between the normal arguments and the keyword-only arguments.

arthurp (changes requested):

libgalois/src/BuildGraph.cpp
```diff
@@ -1832,33 +1832,36 @@ ImportData::ValueFromArrowScalar(std::shared_ptr<arrow::Scalar> scalar) {
 }
 
 katana::Result<std::unique_ptr<katana::PropertyGraph>>
-katana::ConvertToPropertyGraph(katana::GraphComponents&& graph_comps) {
+katana::ConvertToPropertyGraph(
+    katana::GraphComponents&& graph_comps, tsuba::TxnContext* txn_ctx) {
```
witchel: Is logical that this function takes a transaction context?  I ask because it seems like we are creating a graph from non-graph components.  Are these non-graph components in need of the concurrency control/atomicity of transactions?  This is an honest question, I'm not sure.  I was thinking of all transaction data objects as associated with a graph, but maybe that is too limited.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -461,7 +461,7 @@ struct ClusteringImplementationBase {
         arrow::field((new_edge_property_name), std::make_shared<ArrowType>()));
     columns.emplace_back(edge_property);
     auto edge_data_table = arrow::Table::Make(arrow::schema(fields), columns);
-    if (auto r = pfg_to->AddEdgeProperties(edge_data_table); !r) {
+    if (auto r = pfg_to->AddEdgeProperties(edge_data_table, txn_ctx); !r) {
```
witchel: Does this function need to mark `new_edge_property_name` in `pfg_from` as read?

libgalois/src/BuildGraph.cpp
witchel: I have the same question about the `Construct*Properties` functions.

libtsuba/src/RDGCore.cpp
```diff
@@ -179,7 +188,7 @@ RDGCore::UpsertNodeProperties(
   auto written_prop_names = KATANA_CHECKED(UpsertProperties(
       props, &node_properties_, &part_header_.node_prop_info_list()));
   // store write properties into transaction context
-  txn_ctx->InsertNodePropertyWrite(std::move(written_prop_names));
+  txn_ctx->InsertNodePropertyWrite<std::unordered_set>(written_prop_names);
```
witchel: I'd like to see us agree on using std::set<std::string> for read/write sets.  I don't think the efficiency of `unordered_set `is significant and the ordering of `std::set` will be useful for determining conflicts and debugging.

witchel (commented):
This PR puts in some much needed plumbing to pass around TxnContext*.  That part is great.

I have two major questions.  First is whether transactions protect non-graph associated data.  I sort of thought not, but you are updating read/write sets when building graphs.

The second question is whether we use const as much as we can.  We are passing around TxnContexts, but if we are passing them to a function that would only ever need to read it, we should mark it const.  Maybe that never happens?

libtsuba/src/RDGCore.cpp
ddn0: Shouldn't we just take a range or pair of iterators? There is no reason for callers to know how a transaction context keeps track of access sets. They just need to do a multi-insert.

libgalois/include/katana/PropertyGraph.h
ddn0: The convention should be that TxnContext is the last argument; though, I think we will have some exceptions when the function already takes variable arguments like

https://github.com/KatanaGraph/katana/pull/727/files#diff-576d16ef31b9271d43238ea8dfe5c19a7cd07e1761e825554316b14fff165875R97-R98

libgalois/include/katana/TopologyGeneration.h
```diff
@@ -196,9 +197,11 @@ class KATANA_EXPORT PropertyGenerator {
 /// );
 template <typename... Args>
 KATANA_EXPORT Result<void>
-AddNodeProperties(katana::PropertyGraph* pg, Args&&... generators) {
+AddNodeProperties(
+    tsuba::TxnContext* txn_ctx, katana::PropertyGraph* pg,
```
ddn0: at the very least, can we put TxnContext towards the end? PropertyGraph*, TxnContext*, Args&&?

libgalois/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -438,8 +438,8 @@ struct ClusteringImplementationBase {
  * property graph, pg_from to pg_to.
  */
   static katana::Result<void> CopyEdgeProperty(
-      katana::PropertyGraph* pfg_from, katana::PropertyGraph* pfg_to,
-      const std::string& edge_property_name,
+      tsuba::TxnContext* txn_ctx, katana::PropertyGraph* pfg_from,
```
ddn0: txn_ctx last

libgalois/include/katana/analytics/betweenness_centrality/betweenness_centrality.h
```diff
@@ -80,7 +80,8 @@ KATANA_EXPORT extern const BetweennessCentralitySources
 ///          nodes; if this is an int process that number of source nodes.
 /// @param plan
 KATANA_EXPORT Result<void> BetweennessCentrality(
-    PropertyGraph* pg, const std::string& output_property_name,
+    tsuba::TxnContext* txn_ctx, PropertyGraph* pg,
```
ddn0: txn_context placement: ditto and throughout

libtsuba/include/tsuba/TxnContext.h
```diff
@@ -9,47 +9,35 @@ class KATANA_EXPORT TxnContext {
     node_properties_read_.insert(name);
   }
 
-  void InsertNodePropertyRead(std::unordered_set<std::string> names) {
-    node_properties_read_.insert(names.begin(), names.end());
-  }
-
-  void InsertNodePropertyRead(std::vector<std::string> names) {
+  template <template <class> class Container>
+  void InsertNodePropertyRead(Container<std::string>& names) {
```
ddn0: Let's just take a Range or Iterator:

```
template <typename It>
void InsertNodePropertyRead(It begin, It end) { .... }

template <typename Range>
void InsertNodePropertyRead(const Range& range) {
  InsertNodePropertyRead(range.begin(), range.end());
}
```

and avoid over-parameterizing on things we don't actually know the shape of or even want to control.

libgalois/include/katana/TopologyGeneration.h
yige-hu: @ddn0 I've been trying to put `TxnContext` at the end when it's possible. But a parameter pack needs to be put at the end afaik, isn't it? I remember I put `TxnContext` at the end here before, and got compilation errors.

libgalois/include/katana/PropertyGraph.h
yige-hu: @arthurp I prefer putting it last. But there are some cases where it cannot be put at the end. For example, some other function parameters having default values, or parameter packs as in the link posted by Donald above.

libgalois/include/katana/analytics/Utils.h
yige-hu: Same. In `ConstructNodeProperties`, `TxnContext` cannot be put at the end because `names` has default value.

libgalois/include/katana/analytics/independent_set/independent_set.h
yige-hu: Same. `plan` has default value.

python/katana/local/analytics/_k_core.pyx
yige-hu: Hmm.... I didn't put spaces around the `=` because at most of other places, the default values of other parameters were without spaces around their `=`-s (written by others, like the below comment). Do you want me to change them all to be with spaces, or just make `TxnContext` to be consistent with other parameters?

yige-hu: Right now I just make them consistent with their surrounding code.

arthurp: That's fine.

libgalois/include/katana/PropertyGraph.h
arthurp: I'm OK with having some exceptions like that. But even in that case I think it should be "as late as possible". Having it first in some cases and last in others seems really confusing.

libgalois/include/katana/analytics/Utils.h
arthurp: As above, I would prefer the context to come as late as possible in the list anyway.

libgalois/include/katana/TopologyGeneration.h
ddn0: At the very least, can you make it `(PropertyGraph*, TxnContext*, Args&&...)`?

libgalois/include/katana/analytics/ClusteringImplementationBase.h
yige-hu: Resolved because eventually `txn_ctx` gets updated at `RDGCore`.

libgalois/src/BuildGraph.cpp
yige-hu: Resolved because Monday's discussions agreed on leaving all the `TxnContext` to maintain completeness.

libtsuba/include/tsuba/TxnContext.h
yige-hu: Sure. Changed using range (but named it as `Container` to be consistent with Emmett's previous PR).

libgalois/src/BuildGraph.cpp
witchel: I do want you to keep an eye on what goes in your read and write sets.  For example, I don't think you should put the names of non-graph data into your read sets.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
witchel: How is that a resolution?

libtsuba/src/RDGCore.cpp
witchel: Oh, the Container template idea came from me because Tyler used it and I liked it.  I prefer a single Container argument over begin/end iterators, but that is a personal thing.  I find `begin()/end()` unwieldy.

libgalois/include/katana/PropertyGraph.h
yige-hu: My original thought was that something named `...Context` conventionally tends to be put as either first or last argument. But now I've changed all of the `TxnContext`-s to be last possible argument.

libgalois/include/katana/TopologyGeneration.h
yige-hu: My original thought was that something named `...Context` conventionally tends to be put as either first or last argument. But now I've changed all of the `TxnContext`-s to be last possible argument.

libgalois/include/katana/analytics/ClusteringImplementationBase.h
yige-hu:
```
fields.emplace_back(
        arrow::field((new_edge_property_name), std::make_shared<ArrowType>()));
    columns.emplace_back(edge_property);
    auto edge_data_table = arrow::Table::Make(arrow::schema(fields), columns);
    if (auto r = pfg_to->AddEdgeProperties(edge_data_table, txn_ctx); !r) {
      return r.error();
    }
```
`edge_data_table` is made with `new_edge_property_name` and passed into `AddEdgeProperties` with a modifiable `txn_ctx`. This `txn_ctx` eventually gets updated by `RDGCore::AddProperties` with exactly the property names in `new_edge_property_name`. `RDGCore::Upsert/Add/RemoveProperties` is the ultimate point where `TxnContext`-s get updated.

libgalois/src/BuildGraph.cpp
yige-hu: I think this needs discussion on whether we should ignore the `TxnContext` passed in some functions that, although those functions eventually calls `RDGCore::Upsert/AddProperties`, they won't actually cause any conflicts. (e.g. Adding new properties that won't have name conflicts with others' new property creation.) In these cases, maybe we can ignore the `TxnContext` updated by the lower layer of the call stack by not passing it on to these functions' callers.

arthurp (approved):

witchel (approved):
The plumbing is now in place, time to see what comes out of the pipes.

docs/reference/python/katana.local.rst
```diff
@@ -16,6 +16,5 @@ as part of a real data science pipeline that does not require scale out.
 
    katana.local.graph
    katana.local.analytics
-   katana.local.import_data
```
origandrew: do not remove

PR #728, opened 2021-12-03
Fix RDGCore::find_*_prop_info()

scober:
The original implementations returned a copy of the found PropStorageInfo
object, meaning that the state of the original objects could not be
modified. Change the semantics of the search functions to return a
pointer to the object stored by the RDGCore object.

Also fix an error-checking mismatch between ErrorCode::NotFound and
ErrorCode::PropertyNotFound.

### DIFF
--- a/libtsuba/src/RDGCore.cpp
+++ b/libtsuba/src/RDGCore.cpp
@@ -122,20 +122,17 @@ EnsureTypeLoaded(const katana::Uri& rdg_dir, tsuba::PropStorageInfo* psi) {
   return katana::ResultSuccess();
 }
 
-katana::Result<tsuba::PropStorageInfo>
+tsuba::PropStorageInfo*
 find_prop_info(
-    const std::string& name,
-    const std::vector<tsuba::PropStorageInfo>& prop_infos) {
+    const std::string& name, std::vector<tsuba::PropStorageInfo>* prop_infos) {
   auto it = std::find_if(
-      prop_infos.begin(), prop_infos.end(),
-      [&name](const tsuba::PropStorageInfo& psi) {
-        return psi.name() == name;
-      });
-  if (it == prop_infos.end()) {
-    return tsuba::ErrorCode::PropertyNotFound;
+      prop_infos->begin(), prop_infos->end(),
+      [&name](tsuba::PropStorageInfo& psi) { return psi.name() == name; });
+  if (it == prop_infos->end()) {
+    return nullptr;
   }
 
-  return *it;
+  return &(*it);
 }
 }  // namespace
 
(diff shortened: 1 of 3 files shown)


libtsuba/src/RDGCore.cpp
```diff
@@ -122,20 +122,17 @@ EnsureTypeLoaded(const katana::Uri& rdg_dir, tsuba::PropStorageInfo* psi) {
   return katana::ResultSuccess();
 }
 
-katana::Result<tsuba::PropStorageInfo>
+katana::Result<tsuba::PropStorageInfo*>
```
tylershunt: If we're going to return a pointer anyway, maybe take advantage of `nullptr` and make that the return value for no property found?

tylershunt (commented):
I'm curious why this wasn't necessary for `RDG`. Is it that `RDG` does a better job using `RDGCore`'s functionality?

libtsuba/src/RDGCore.cpp
```diff
@@ -122,20 +122,17 @@ EnsureTypeLoaded(const katana::Uri& rdg_dir, tsuba::PropStorageInfo* psi) {
...
-      });
-  if (it == prop_infos.end()) {
+      prop_infos->begin(), prop_infos->end(),
+      [&name](tsuba::PropStorageInfo& psi) { return psi.name() == name; });
+  if (it == prop_infos->end()) {
     return tsuba::ErrorCode::PropertyNotFound;
   }
 
-  return *it;
+  return &(*it);
```
witchel, line 135: I luv c++.

libtsuba/src/RDGSlice.cpp
```diff
@@ -57,12 +58,12 @@ unload_metadata_array(
     tsuba::RDGCore* core) {
   auto search_res = core->find_part_prop_info(array_name);
   if (!search_res) {
-    if (search_res.error() != tsuba::ErrorCode::NotFound) {
+    if (search_res.error() != tsuba::ErrorCode::PropertyNotFound) {
       return search_res.error().WithContext(
           "searching for array name in prop info list");
```
witchel: Is this context useful?

witchel (approved):
Right on, right on, right on.

libtsuba/src/RDGCore.cpp
scober: Yeah, that would be fine. Do we not usually return Results of pointers?

libtsuba/src/RDGSlice.cpp
scober: If that is the only place it can fail, maybe not.

libtsuba/src/RDGCore.cpp
tylershunt: For something like this where not found isn't necessarily an error I think it makes sense. I mean, we've already gotten the error code check wrong once 😅

libtsuba/src/RDGSlice.cpp
witchel: Without context, we get file and line number.

My rule of thumb is that if we are not printing a data item specific to that function, the context is not necessary and indeed is clutter.

scober:
> I'm curious why this wasn't necessary for `RDG`. Is it that `RDG` does a better job using `RDGCore`'s functionality?

I don't totally understand this question. The answer might be because `RDG` doesn't use this searching functionality because it was only recently added and I only used it in `RDGSlice` when I added it.

libtsuba/src/RDGSlice.cpp
scober: Seems like a sensible rule

libtsuba/src/RDGCore.cpp
witchel: I'd like to live in a world where NO error code is ever interpreted at a higher level as a non-error.  That only happens in a very few cases and I don't like it.

ddn0:
And no tests or anything ?

PR #729, opened 2021-12-03
include libdl error message when dynamically linking a katana plugin fails

yshalabi:
When dlopen fails we get an awkward error message:
```
WARNING: /source/external/katana/libsupport/src/Plugin.cpp:27: skipping plugin "/usr/lib/katana/plugins/some_plugin.so": failed to dynamically link plugin: Success
``` 

use dlerror to produce a more useful error message and retrieve the cause of the error (e.g. which specific library is missing which the plugin depends on).


arthurp (approved):
Ooops

PR #730, opened 2021-12-03
[KAT-2703] Enable uint64_t type for index

arekay:
Add `uint64_t` data type for louvian clustering. 
At some point we would like to have all the data types arrow supports be included here.


witchel (approved):
This change looks good.

Can't we leave out data types not present in Cypher like int16?

arekay:
> This change looks good.
> 
> Can't we leave out data types not present in Cypher like int16?

Yes, but we need a clean way. Currently we have to explicitly add data-types we intend to support. One option could be to use the mapping from cypher data-types to arrow data-types to generate the list of data-types we intend to support. Will create a JIRA for that task.

PR #731, opened 2021-12-03
Test PropStorageInfo lookup

scober:
@ddn0 correctly pointed that I should have added some tests for `find_node_prop_info()` and friends. And I especially should have added some tests after I found a bug in the first implementation. While writing those tests I realized that `RDGCore` wasn't a particularly good home for those functions anyway.

So.

Move `find_*_prop_info()` from `RDGCore` to `RDGPartHeader` so that they are now in the same class as the data structures they operate on. And add a test for those functions as well as some functions that modify the `PropStorageInfo` lists.


e-mcginnis (approved):
LGTM
Thanks for adding this test Simon!

libtsuba/test/rdg-part-header.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // upserts
+  tsuba::PropStorageInfo new_edge_prop(
+      "not value", arrow::fixed_size_binary(8));
+  under_test.UpsertEdgePropStorageInfo(std::move(new_edge_prop));
+
+  tsuba::PropStorageInfo updated_edge_prop("value", arrow::large_binary());
+  under_test.UpsertEdgePropStorageInfo(std::move(updated_edge_prop));
+
+  KATANA_LOG_ASSERT(under_test.node_prop_info_list().empty());
```
witchel, line 42: Shouldn't this be `!`?

libtsuba/test/rdg-part-header.cpp
```diff
@@ -0,0 +1,97 @@
...
+  KATANA_LOG_ASSERT(under_test.edge_prop_info_list().size() == 2);
+  KATANA_LOG_ASSERT(under_test.find_edge_prop_info("value") != nullptr);
+  KATANA_LOG_ASSERT(
+      under_test.find_edge_prop_info("value")->IsAbsent() == true);
+  KATANA_LOG_ASSERT(under_test.find_edge_prop_info("not value") != nullptr);
+  KATANA_LOG_ASSERT(
+      under_test.find_edge_prop_info("not value")->IsDirty() == true);
```
witchel: Why is "value" absent, while "not value" is dirty?

libtsuba/test/rdg-part-header.cpp
```diff
@@ -0,0 +1,97 @@
...
+  under_test.find_edge_prop_info("not value")
+      ->WasWritten("/tmp/did/not/actually/write/anything");
+  KATANA_LOG_ASSERT(
+      under_test.find_edge_prop_info("not value")->IsClean() == true);
+
+  // remove everything
+  KATANA_CHECKED(under_test.RemoveEdgeProperty("not value"));
+  under_test.RemoveEdgeProperty(0);
```
witchel, line 67: I'd think you'd have to remove both of them.

witchel (approved):
Nice tests.

I wonder about some of the asserts, but if they pass, they pass.

libtsuba/test/rdg-part-header.cpp
scober: This is tightly coupled to rmat15 as an input. I will add some comments/asserts indicating that. But with that in mind, rmat15 has one edge property and no node properties. So this is the correct assert.

witchel: Cool.  Comment is worthwhile.

libtsuba/test/rdg-part-header.cpp
scober: Nice catch! Turns out this was a bug, I am pushing a fix now.

libtsuba/test/rdg-part-header.cpp
scober: I tried to remove one by name and one by index.

PR #732, opened 2021-12-03
Use mamba if available. Override channels.

arthurp:
Overriding channels makes installation more consistent across
environments.


arthurp:
I think this will improve the failure we are seeing in CI, but I'm not sure. I have not been able to reproduce those failures (even in CI weirdly).

ddn0 (approved):

witchel (approved):
Worth a try.

PR #733, opened 2021-12-03
Expose multiple type-related functions

bowu:
This PR exposes multiple type-related functions in the python wrapper of `PropertyGraph`, such that they are available in both  local and distributed APIs. It also implements a simple wrapper for `EntityTypeManager`.


bowu:
> This PR exposes multiple type-related functions in the python wrapper of `PropertyGraph`, such that they are available in both local and distributed APIs. It also implements a simple wrapper for `EntityTypeManager`.

@arthurp I'm not sure what you mean by `integrate the edge type information into the edges dataframe` but could you see whether this PR is self-contained?

python/katana/local/_graph.pyx
```diff
@@ -457,22 +460,72 @@ cdef class GraphBase:
...
 
+    def get_node_type_manager(self):
+        """
+        :return the node type manager
+        """
+        return EntityTypeManager.make(&self.underlying_property_graph().GetNodeTypeManager())
+
+    def get_type_of_node(self, uint64_t n):
+        """
+        Return the type id of a node
```
arthurp: This isn't returning *the* type of a node. It's returning the most specific type. The node actually has many types as you can tell by the fact that `does_node_have_type` will return true for various different types.

Please change the documentation for this and the analogous edge function.

python/katana/local/_graph.pyx
```diff
@@ -457,22 +460,72 @@ cdef class GraphBase:
...
+        return self.underlying_property_graph().GetTypeOfEdge(e)
+
+    def does_edge_have_type(self, uint64_t e, EntityTypeID type_id):
+        """
+        Check whether a given edge has a certain type
+
+        :param e: edge id
+        :param type_id: type id
+        :return True if edge e has the given type and False otherwise
+        """
```
arthurp: This is OK for the moment. But I think we should move all this to the edge dataframe. That way people don't need to handle edge IDs as much.

Please add a TODO and file a JIRA issue so we don't forget to come back and improve this.

python/katana/local/entity_type_manager.pyx
```diff
@@ -0,0 +1,13 @@
...
+    cdef EntityTypeManager make(const CEntityTypeManager *manager):
+        m = <EntityTypeManager>EntityTypeManager.__new__(EntityTypeManager)
+        m.underlying_entity_type_manager = manager
+        return m
+
+    def get_atomic_entity_type_ids(self):
+        """
+        Return a list of all entity type IDs
+        """
+        return self.underlying_entity_type_manager.GetAtomicEntityTypeIDs()
```
arthurp: Since we are exposing this to the user, please provide a complete API. This doesn't mean you need to wrap every method. But this wrapper should provide access to all the features of the underlying class.

arthurp (commented):
I think this could use some expansion, but this is looking pretty good.

arthurp:
> @arthurp I'm not sure what you mean by `integrate the edge type information into the edges dataframe` but could you see whether this PR is self-contained?

Sorry I missed this in the first review. What I mean is that there is a dataframe for edges (returned by `GraphBase.edges`). This dataframe should have computed columns for the types. Probably booleans, one for each atomic type.

Now that I think about this, I think it can wait for another PR. Adding the basic API is a good enough start. It will have to change, but it can get us going without too much API design debate.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetNodeTypeManager()
+        return manager.GetEntityTypeIDToAtomicTypeNameMap()
+
+    def get_edge_entity_type_id_to_atomic_type_name_map(self):
+        """
+        Return the map from edge entity type IDs to type names
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetEdgeTypeManager()
+        return manager.GetEntityTypeIDToAtomicTypeNameMap()
```
arthurp: All these functions should be on `EntityTypeManager`. They are not operations on the graph.

One way to tell this is that you are having to access another object to do the main operation. This produces duplicated code which is a hint that something is wrong.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+    def get_atomic_edge_type_name(self, entity_type_id: int):
+        """
+        Return the type name if `entity_type_id` is a valid edge entity type ID. Otherwise,
+        return None
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetEdgeTypeManager()
+        r =  manager.GetAtomicTypeName(entity_type_id)
+        if r:
+            return r.value()
+        return None
```
arthurp: In Python, errors should show exceptions instead of returning a sentinel. In the case of these operations it should be `LookupError(id)`.

arthurp: This applies throughout the PR.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+        """
+        Return True iff the type `sub_type` is a subtype of the type `super_type` for
+        edges 
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetEdgeTypeManager()
+        return manager.IsSubtypeOf(sub_type, super_type)
+
+    def get_node_entity_type_id_to_atomic_type_name_map(self):
+        """
+        Return the map from node entity type IDs to type names
```
arthurp: The main map type in Python is called `dict`. I assume that's what you are getting here.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+    def get_atomic_edge_entity_type_ids(self):
+        """
+        Return a list of all edge entity type IDs
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetEdgeTypeManager()
+        return manager.GetAtomicEntityTypeIDs()
+
+    def node_is_subtype_of(self, sub_type: int, super_type: int):
+        """
+        Return True iff the type `sub_type` is a subtype of the type `super_type` for
```
arthurp: NIT: double back-ticks to get "code-font". Single back ticks are a link in RST and links to arguments don't work sadly.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+        if r:
+            return r.value()
+        return None
+
+    def get_atomic_node_entity_type_ids(self):
+        """
+        Return a list of all node entity type IDs
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetNodeTypeManager()
+        return manager.GetAtomicEntityTypeIDs()
```
arthurp: I think all the "get list" methods can be combined into a two properties one returning a dict of ID to `EntityType` objects, and the other returning a dict of `str` (name) to `EntityType`. If someone wants the EntityTypes only they can just use `.values()`. All the different accessors are needed in C++ because the C++ collections library is super impoverished.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+        Return the type name if `entity_type_id` is a valid node entity type ID. Otherwise,
+        return None
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetNodeTypeManager()
+        r =  manager.GetAtomicTypeName(entity_type_id)
+        if r:
+            return r.value()
+        return None
+
+    def get_atomic_edge_type_name(self, entity_type_id: int):
```
arthurp: These are not needed because we will have methods that return dicts so users can just to `entity_type_manager.by_id[id].name`. I think this is true of most of the single element accessors.

python/katana/local/_graph.pyx
```diff
@@ -481,6 +531,182 @@ cdef class GraphBase:
...
+
+    def node_is_subtype_of(self, sub_type: int, super_type: int):
+        """
+        Return True iff the type `sub_type` is a subtype of the type `super_type` for
+        nodes
+        """
+        cdef const CEntityTypeManager* manager = &self.underlying_property_graph().GetNodeTypeManager()
+        return manager.IsSubtypeOf(sub_type, super_type)
+
+    def edge_is_subtype_of(self, sub_type: int, super_type: int):
```
arthurp: This should support int *and* `EntityType` objects. You will need to do a dynamic check with `isinstance`. Python doesn't have overloading since it doesn't have type based dispatch as such.

arthurp: This change should be made to several functions in `EntityTypeManager`

arthurp (changes requested):
My comments assume that you will be extracting the type related methods into `EntityTypeManager`. So even though the code is correctly on the graph please imagine it's already moved when reading.

You should use the `EntityType` class. Currently it says it doesn't support intersection types (well it would if the author had spell check in their editor ;-) ), this will need to be fixed. My approach would be: 

Create two classes `EntityType` which only has an ID and a subclass `AtomicEntityType` which adds `name`. Atomic types will be instances of `AtomicEntityType` and intersections will be instances of only `EntityType`.

I think it should be straight forward and pretty pythonic.

python/katana/local/atomic_entity_type.pyx
```diff
@@ -0,0 +1,27 @@
...
+from katana.local.entity_type cimport EntityType
+
+
+cdef class AtomicEntityType(EntityType):
+    """
+    A class representing node/edge actomice types.
+    """
+    @property
+    def name(self):
+        return self._name
```
arthurp: This should be a string. Not `bytes`. Use decode as you did for `__str__`.

python/katana/local/entity_type_manager.pyx
```diff
@@ -0,0 +1,110 @@
...
+        """
+        if isinstance(entity_type, int):
+            entity_type_id = entity_type
+        elif isinstance(entity_type, EntityType):
+            entity_type_id = entity_type.type_id
+        else:
+            raise ValueError(f"{entity_type}'s type is not supported")
+        if not self.underlying_entity_type_manager.HasEntityType(entity_type_id):
+            raise LookupError(f"{entity_type_id} does not represent a type")
+        return handle_result_TypeNameSet(self.underlying_entity_type_manager.EntityTypeToTypeNameSet(entity_type_id))
```
arthurp: This method doesn't match the mappings we are using for the other parts of the interface. I think in the interface I proposed the the idea was to have a list of all entitytypes and then a function to get all atomic super types (which is the set of names you are returning here). Also this shouldn't return names, but should return `EntityType` objects.

python/test/test_property_graph.py
```diff
@@ -314,3 +314,60 @@ def func_operator(g, prop, out, nid):
...
+    assert len(node_type_set) == 11
+    assert len(edge_type_set) == 15
+
+    assert graph.does_node_have_type(0, 17)
+    assert graph.does_edge_have_type(0, 8)
+
+    node_atomic_types = graph.node_types.atomic_types
+    node_name_to_id_map = {name: node_atomic_types[name].type_id for name in node_atomic_types}
+    assert node_name_to_id_map == {
+        b"Tag": 12,
```
arthurp: Using the numeric IDs in the test will be fragile since we don't really guarantee stability of the numeric ID (I assume). This doesn't need to be fixed in this PR, but it is something to keep in mind.

l-hoang: Confirming this: I've been told many times by others that the stability of a GID is not guaranteed.
I've been doing my ID tests using numeric ID properties instead because of that (it's
a lot more painful to do it that way but it's better than having all these tests break if the
way GIDs work suddenly changes elsewhere).

python/katana/local/entity_type_manager.pyx
bowu: Makes sense. Since this PR doesn't touch `DynamicBitsetSlow`. I'll remove this function.

arthurp: Sounds good. Re can come back to it in the next PR.

python/test/test_property_graph.py
arthurp: @bowu please address this somehow. In this case I think checking everything by name would work fine, so that seems like the way to go.

arthurp (approved):
LG other than the testing issue.

PR #734, opened 2021-12-04
tsuba: Return errors reading RDGPartHeader rather than dying

ddn0:
Generally, one does not control what one's inputs are so return errors
rather than dying for invalid input. I left the asserts on the to_json
code paths under the assumption that we control the data so it is more
acceptable to fail violently.


ddn0:
My preference would be to separate validation from serialization as it is pretty normal to want to read possibly invalid files (so you can diagnose them, etc.). Conflating these two steps makes things harder to reuse.

tylershunt (approved):

e-mcginnis (approved):
lgtm

PR #736, opened 2021-12-06
Fix handling of commits with both tags and trivial merges.

arthurp:
The code simplifies trivial merges to make versions make more
sense when run in github CI (which generates trivial merges).
However, when tags are present it was not checking the original
commit and was only checking the "normalized" commit. This change
makes the system check both the original commit and the
"normalized" commit.

This commit also assumes tags on only enterprise are correct
instead of ignoreing them. It prints a warning, but will still
create the expected version. This should avoid "oops the artifacts
are broken" surprises.

### DIFF
--- a/scripts/katana_version/version.py
+++ b/scripts/katana_version/version.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import warnings
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


roshandathathri (approved):
LGTM

PR #737, opened 2021-12-06
Fix update-alternatives call in setup_dev_ubuntu.sh

srogatch:


arthurp (approved):
Please add them to the line above. Running without `-y` may mess things up. Also can we drop automatically installing clang-10? Is it still used anywhere?

srogatch:
> Please add them to the line above. Running without `-y` may mess things up. Also can we drop automatically installing clang-10? Is it still used anywhere?

Thanks for catching `-y`. I think `clang-10` is not used anywhere except `scripts/dev` that provides a possibility to build with `clang-10` still. But we can drop it here.

PR #738, opened 2021-12-06
Make list of cached properties specific to rdg_prefix

witchel:
Properties from different RDG prefixes can have the same name.

### DIFF
--- a/libsupport/include/katana/Cache.h
+++ b/libsupport/include/katana/Cache.h
@@ -19,6 +19,7 @@
 #include <boost/container_hash/hash.hpp>
 
 #include "katana/Logging.h"
+#include "katana/Strings.h"
 
 namespace katana {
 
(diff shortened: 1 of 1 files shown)


witchel:
Finishing tests, but it looks like this will fix 2540.  I will have a followup PR for Cache.h's handling of the RDG pointer.

tylershunt (approved):

libsupport/include/katana/Cache.h
```diff
@@ -116,30 +117,26 @@ class KATANA_EXPORT Cache {
   }
 
   // Return a vector of node properties present in the cache, in LRU order
-  std::vector<std::string> GetNodeProperties() const {
+  std::vector<std::string> GetNodeProperties(
+      const std::string& rdg_prefix) const {
     std::vector<std::string> node_props;
     for (const auto& elt : lru_list_) {
-      if (elt.is_node()) {
+      if (elt.is_node() && elt.rdg_prefix() == rdg_prefix) {
```
ddn0, line 124: The cache key and value type is generic, yet we assume certain methods exist like `elt.is_node` and `elt.rdg_prefix` ? 

Options:
- Document this restriction, add tests to at least check the concept requirements
- Move this class closer to its user as it is clearly not generic enough to support arbitrary key or value types
- Use a more generic interface (key::operator==, Get, Put).

ddn0 (approved):
Approving because it fixes a bug, but this is also a reason to pause to confirm if we really have the right interface

libsupport/include/katana/Cache.h
witchel: Yeah, I hear you.  I started out with Cache.h as generic and PropertyCache as the place for node/edge specialization.  But a lot of PropertyCache has leaked into Cache.h and it isn't clear that Cache gains any value from being generic.

In addition, the cache's support for multiple RDGs is shakey and I need to be able to evict a specific entry, so I'll revisit the interface.

PR #739, opened 2021-12-06
Reset default topology pointer to more constrained topology if available

yan-zaretskiy:
This commit has cosmetic and functional changes. The cosmetic change is that `EdgeTypeAwareTopology` now derives directly from `EdgeShuffleTopology`. The code in `GraphTopology.h` has been moved around so that all Topology class definitions precede TopologyWrapper definitions. The functional change is that the default topology has been moved to the topology cache. Moreover, now its pointer may be reseated to point to a more constrained topology, if one is created, to reduce the overall memory consumption of topologies in cache. Concretely, we update the default topology when a new `EdgeShuffleTopology` or an `EdgeTypeAwareTopology` is constructed.

Here are the memory consumption breakdown when running a simple query on the friendster graph:

| Item | Before  | After | Comment |
| ------|------- | ----------|--- |
| Original topology | 14 GB  | 41 GB  | Original topology now stores the identity edge index mapping. We agreed to address this in a subsequent PR. This bump does not affect the query use case, since we now drop the original topology when we create the AttributedGraph.  |
| Two AttributedGraph topologies  | 2 x 41 GB  | 2 x 41 GB | No change |
| AttributedGraph's property index -> out-edge ID mapping  | N/A  | 26 GB | Current AttributedGraph's implementation relies on property indexes and local edge IDs to be the same for the default topology. When we replace the default topology with a shuffled one, we have to introduce this additional mapping. This will be addressed in a subsequent PR. |
| Overall memory usage | 105 GB | 117 GB | The increase of 12 GB is the difference between the new mapping of 26 GB and dropping the original topology of 14 GB |

### DIFF
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -68,8 +68,8 @@ public:
       const Edge* adj_indices, size_t num_nodes, const Node* dests,
       size_t num_edges) noexcept;
 
-  GraphTopology(NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests) noexcept
-      : adj_indices_(std::move(adj_indices)), dests_(std::move(dests)) {}
+  GraphTopology(
+      NUMAArray<Edge>&& adj_indices, NUMAArray<Node>&& dests) noexcept;
 
   static GraphTopology Copy(const GraphTopology& that) noexcept;
 
@@ -159,11 +159,12 @@ public:
   ///@returns Degree of node N
   size_t degree(Node node) const noexcept { return edges(node).size(); }
 
-  virtual PropertyIndex edge_property_index(const Edge& eid) const noexcept {
-    return eid;
+  PropertyIndex edge_property_index(const Edge& eid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(eid < num_edges());
+    return edge_prop_indices_[eid];
   }
 
-  virtual PropertyIndex node_property_index(const Node& nid) const noexcept {
+  PropertyIndex node_property_index(const Node& nid) const noexcept {
     return nid;
   }
 
@@ -196,18 +197,6 @@ public:
     return edge_sort_state_;
   }
 
-  bool has_nodes_sorted_by(
-      const tsuba::RDGTopology::NodeSortKind& kind) const noexcept {
-    if (kind == tsuba::RDGTopology::NodeSortKind::kAny) {
-      return true;
-    }
-    return node_sort_state_ == kind;
-  }
-
-  tsuba::RDGTopology::NodeSortKind node_sort_state() const noexcept {
-    return node_sort_state_;
-  }
-
   void Print() const noexcept;
 
 protected:
@@ -221,9 +210,13 @@ protected:
     edge_sort_state_ = edge_sort_state;
   }
 
-  void SetNodeSortState(
-      tsuba::RDGTopology::NodeSortKind node_sort_state) noexcept {
-    node_sort_state_ = node_sort_state;
+  void SetEdgePropIndices(PropIndexVec&& edge_prop_indices) noexcept {
+    KATANA_LOG_DEBUG_ASSERT(edge_prop_indices_.size() == num_edges());
+    edge_prop_indices_ = std::move(edge_prop_indices);
+  }
+
+  const PropertyIndex* edge_property_index_data() const noexcept {
+    return edge_prop_indices_.data();
   }
 
 private:
@@ -242,8 +235,14 @@ private:
       tsuba::RDGTopology::TransposeKind::kNo};
   tsuba::RDGTopology::EdgeSortKind edge_sort_state_{
       tsuba::RDGTopology::EdgeSortKind::kAny};
-  tsuba::RDGTopology::NodeSortKind node_sort_state_{
-      tsuba::RDGTopology::NodeSortKind::kAny};
+
+  // TODO(amber): In the future, we may need to keep a copy of edge_type_ids in
+  // addition to edge_prop_indices_. Today, we assume that we can use
+  // PropertyGraph.edge_type_set_id(edge_prop_indices_[edge_id]) to obtain
+  // edge_type_id. This may not be true when we group properties
+  // when this is done, the Write path must also be updated to pass the edge_type_ids index
+  // to RDG. For now, we pass nullptr.
+  PropIndexVec edge_prop_indices_;
 };
 
 // TODO(amber): In the future, when we group properties e.g., by node or edge type,
@@ -269,11 +268,6 @@ public:
       const tsuba::RDGTopology::EdgeSortKind& kind) const noexcept {
     return (kind == tsuba::RDGTopology::EdgeSortKind::kAny) ||
            (kind == edge_sort_state());
-  }
-
-  PropertyIndex edge_property_index(const Edge& eid) const noexcept override {
-    KATANA_LOG_DEBUG_ASSERT(eid < num_edges());
-    return edge_prop_indices_[eid];
   }
 
   using Base::original_edge_id;
@@ -315,13 +309,6 @@ public:
     return find_edge(src, dst) != edges(src).end();
   }
 
-  // unfortunately, without access to this data EdgeTypeAware would be unable to
-  // create an RDGTopology from itself. We can't even make this protected
-  // as EdgeTypeAware inherits from BasicTopologyWrapper<EdgeShuffleTopology>
-  const PropertyIndex* edge_property_index_data() const noexcept {
-    return edge_prop_indices_.data();
-  }
-
 protected:
   void SortEdgesByDestID() noexcept;
 
@@ -356,23 +343,14 @@ protected:
       const tsuba::RDGTopology::EdgeSortKind& edge_sort_todo,
       AdjIndexVec&& adj_indices, EdgeDestVec&& dests,
       PropIndexVec&& edge_prop_indices) noexcept
-      : Base(std::move(adj_indices), std::move(dests)),
-        is_valid_(true),
-        edge_prop_indices_(std::move(edge_prop_indices)) {
-    KATANA_LOG_DEBUG_ASSERT(edge_prop_indices_.size() == num_edges());
+      : Base(std::move(adj_indices), std::move(dests)), is_valid_(true) {
     SetTransposeState(tpose_todo);
     SetEdgeSortState(edge_sort_todo);
+    SetEdgePropIndices(std::move(edge_prop_indices));
   }
 
 private:
   bool is_valid_ = true;
-  // TODO(amber): In the future, we may need to keep a copy of edge_type_ids in
-  // addition to edge_prop_indices_. Today, we assume that we can use
-  // PropertyGraph.edge_type_set_id(edge_prop_indices_[edge_id]) to obtain
-  // edge_type_id. This may not be true when we group properties
-  // when this is done, the Write path must also be updated to pass the edge_type_ids index
-  // to RDG. For now, we pass nullptr.
-  PropIndexVec edge_prop_indices_;
 };
 
 /// This is a fully shuffled topology where both the nodes and edges can be sorted
@@ -389,9 +367,21 @@ public:
 
   virtual ~ShuffleTopology();
 
-  PropertyIndex node_property_index(const Node& nid) const noexcept override {
+  PropertyIndex node_property_index(const Node& nid) const noexcept {
     KATANA_LOG_DEBUG_ASSERT(nid < num_nodes());
     return node_prop_indices_[nid];
+  }
+
+  bool has_nodes_sorted_by(
+      const tsuba::RDGTopology::NodeSortKind& kind) const noexcept {
+    if (kind == tsuba::RDGTopology::NodeSortKind::kAny) {
+      return true;
+    }
+    return node_sort_state_ == kind;
+  }
+
+  tsuba::RDGTopology::NodeSortKind node_sort_state() const noexcept {
+    return node_sort_state_;
   }
 
   static std::shared_ptr<ShuffleTopology> MakeFrom(
@@ -515,10 +505,13 @@ private:
       : Base(
             tpose_todo, edge_sort_todo, std::move(adj_indices),
             std::move(dests), std::move(edge_prop_indices)),
+        node_sort_state_(node_sort_todo),
         node_prop_indices_(std::move(node_prop_indices)) {
     KATANA_LOG_DEBUG_ASSERT(node_prop_indices_.size() == num_nodes());
-    SetNodeSortState(node_sort_todo);
   }
+
+  tsuba::RDGTopology::NodeSortKind node_sort_state_{
+      tsuba::RDGTopology::NodeSortKind::kAny};
 
   // TODO(amber): In the future, we may need to keep a copy of node_type_ids in
   // addition to node_prop_indices_. Today, we assume that we can use
@@ -1796,13 +1789,15 @@ struct PropertyGraphViews {
 };
 
 class KATANA_EXPORT PGViewCache {
-  std::shared_ptr<GraphTopology> original_topo_ =
-      std::make_shared<GraphTopology>();
+  std::shared_ptr<GraphTopology> original_topo_{
+      std::make_shared<GraphTopology>()};
+
   std::vector<std::shared_ptr<EdgeShuffleTopology>> edge_shuff_topos_;
   std::vector<std::shared_ptr<ShuffleTopology>> fully_shuff_topos_;
   std::vector<std::shared_ptr<EdgeTypeAwareTopology>> edge_type_aware_topos_;
   std::shared_ptr<CondensedTypeIDMap> edge_type_id_map_;
   // TODO(amber): define a node_type_id_map_;
+
   std::shared_ptr<ProjectedTopology> projected_topos_;
 
   template <typename>
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -48,6 +48,10 @@ class KATANA_EXPORT EdgeShuffleTopology;
 class KATANA_EXPORT EdgeTypeAwareTopology;
 class KATANA_EXPORT ProjectedTopology;
 
+/********************/
+/* Topology classes */
+/********************/
+
 /// A graph topology represents the adjacency information for a graph in CSR
 /// format.
 class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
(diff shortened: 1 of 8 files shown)


yan-zaretskiy:
I can't merge this PR until `katana` open is fixed, but I wanted folks to take a look at the actual changes.

mindv0rtex:
Before:

```
INFO: host=0 ms=33435 max_mem_gb=28.155 mem_gb=21.550 arrow_mem_gb=0.000|  starting construct attributedgraph
INFO: host=0 ms=33435 max_mem_gb=28.155 mem_gb=21.550 arrow_mem_gb=0.000 edge_mirrors=false | tags for construct attributedgraph
INFO: host=0 ms=107474 max_mem_gb=104.790 mem_gb=104.790 arrow_mem_gb=0.000 | finished construct attributedgraph
```

After:

```
INFO: host=0 ms=34806 max_mem_gb=28.155 mem_gb=21.550 arrow_mem_gb=0.000 | starting construct attributedgraph
INFO: host=0 ms=34806 max_mem_gb=28.155 mem_gb=21.550 arrow_mem_gb=0.000 edge_mirrors=false | tags for construct attributedgraph
INFO: host=0 ms=108429 max_mem_gb=90.845 mem_gb=90.844 arrow_mem_gb=0.000 | finished construct attributedgraph
```

Some explanation. The original CSR is 14 GB, which is the reduction we get back with this PR. Of the remaining 91 GB, 80 is occupied by the two topologies created for the bidirectional edge type aware view. I suspect the remaining 10 GB are properties (have not checked that last statement). Therefore, we got all the wins we could from dropping redundant topologies.

libgraph/include/katana/GraphTopology.h
```diff
@@ -154,11 +159,11 @@ class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
   ///@returns Degree of node N
   size_t degree(Node node) const noexcept { return edges(node).size(); }
 
-  PropertyIndex edge_property_index(const Edge& eid) const noexcept {
+  virtual PropertyIndex edge_property_index(const Edge& eid) const noexcept {
```
rob-isaac: I think this may be undesirable given that it can be in a critical path of our code

libgraph/include/katana/GraphTopology.h
```diff
@@ -172,8 +177,47 @@ class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
...
   void Print() const noexcept;
 
+protected:
+  tsuba::RDGTopology::TransposeKind tpose_state_ =
+      tsuba::RDGTopology::TransposeKind::kNo;
+  tsuba::RDGTopology::EdgeSortKind edge_sort_state_ =
+      tsuba::RDGTopology::EdgeSortKind::kAny;
+  tsuba::RDGTopology::NodeSortKind node_sort_state_ =
+      tsuba::RDGTopology::NodeSortKind::kAny;
+
```
rob-isaac: Initialize with braces rather than assignment

Nit: it is better to have these private and mutable accessors as protected

libgraph/include/katana/GraphTopology.h
```diff
@@ -319,19 +343,14 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
       PropIndexVec&& edge_prop_indices) noexcept
       : Base(std::move(adj_indices), std::move(dests)),
         is_valid_(true),
-        tpose_state_(tpose_todo),
-        edge_sort_state_(edge_sort_todo),
         edge_prop_indices_(std::move(edge_prop_indices)) {
     KATANA_LOG_DEBUG_ASSERT(edge_prop_indices_.size() == num_edges());
+    tpose_state_ = tpose_todo;
+    edge_sort_state_ = edge_sort_todo;
```
rob-isaac: why the move?

libgraph/include/katana/GraphTopology.h
```diff
@@ -1804,16 +1814,23 @@ class KATANA_EXPORT PGViewCache {
         pg, node_types, edge_types, *this);
   }
 
+  // Avoids a copy of the default topology.
+  const GraphTopology& GetDefaultTopologyRef() const noexcept;
+
+  void DropAllTopologies() noexcept;
```
rob-isaac, line 1836: We should never be dropping all topologies. Haven't looked at the impl yet but I would prefer a name like `DropUnusedTopologies` which reflects this. One topology will always be in use by the PG, others may also be in use by views

libgraph/include/katana/GraphTopology.h
```diff
@@ -1804,16 +1814,23 @@ class KATANA_EXPORT PGViewCache {
...
   std::shared_ptr<CondensedTypeIDMap> BuildOrGetEdgeTypeIndex(
       const PropertyGraph* pg) noexcept;
 
+  // The pop flag ensures that the returned topology is not
+  // in the edge_shuff_topos_ cache.
   std::shared_ptr<EdgeShuffleTopology> BuildOrGetEdgeShuffTopo(
       PropertyGraph* pg, const tsuba::RDGTopology::TransposeKind& tpose_kind,
-      const tsuba::RDGTopology::EdgeSortKind& sort_kind) noexcept;
+      const tsuba::RDGTopology::EdgeSortKind& sort_kind,
+      bool pop = false) noexcept;
```
rob-isaac: Nit: it is preferred a second function rather than use default arguments (second func can just call the other with the default argument)

rob-isaac (commented):
Still working through this but looks good so far

libgraph/include/katana/GraphTopology.h
yan-zaretskiy: This was added in support of `katana::PropertyGraphRetractor::DropTopologies()` which does drop all topologies. So I didn't change any semantics here.

libgraph/include/katana/GraphTopology.h
yan-zaretskiy: Base class variables cannot be initialized in the initializer list of a derived class's constructor. I could either add them to the `GraphTopology` constructor arguments or to put them into the constructor body. I chose the latter because I don't want to allow creating instances of `GraphTopology` with different values for these two variables.

libgraph/include/katana/GraphTopology.h
yan-zaretskiy: > Nit: it is better to have these private and mutable accessors as protected

I just realized we already have these methods implemented, I'll use them instead.

libgraph/include/katana/GraphTopology.h
yan-zaretskiy: I agree, but I don't see a way around it if we want to use the base topology pointer to point to a derived topology. Unfortunately, this is the only method that is overwritten in the derived classes, other methods are just added on top of what's already available in the base topology. I am open to suggestions.

libgraph/include/katana/GraphTopology.h
mindv0rtex: Thinking about it some more, pushing these to the base class constructor would be much cleaner.

libgraph/include/katana/GraphTopology.h
roshandathathri: Yes, this is NOT desirable unless we use `final` in the derived class and use the derived type explicitly.

libgraph/include/katana/GraphTopology.h
roshandathathri: Agreed

libgraph/include/katana/GraphTopology.h
roshandathathri: (1) Why do we need virtual functions? That's not clear to me.

(2) Why not have 1-level hierarchy? Everything is derived from GraphTopologyTypes but not derived further.

roshandathathri (commented):
Partial comments

yan-zaretskiy:
@roshandathathri The reason I put made that method `virtual` was to call the correct property table indexing method, otherwise you'd get wrong value if the underlying default topology is replaced in favor of the edge type aware topology. The issue is that, with this mechanism of dropping the default topology, we don't know what it actually will be until runtime. If we have a 1-level hierarchy, then we can't replace one type with another.

One way to drop `virtual` that I see is to add edge/node property index vectors to the base `GraphTopology` class, but that would hurt the memory consumption.

roshandathathri:
Thanks, Yan. I understand the problem now and the two alternatives. Let me try to summarize that and add a third alternative. 

Let's say we have 2 classes or types A and B that share a common interface C (both A and B could be derived from C or A could be derived from C and B could be derived from A - that's besides the point). The definition of some common function `foo` (in C) depends on the class/type. We have another class D that uses an object of C. 

Let's consider two topologies: `GraphTopology` and `EdgeShuffleTopology`. They share a common interface `GraphTopologyTypes` because `OutView` uses that common interface (i.e., objects of either class). In particular, we need both to have `node_prop_index()`. 

- In `GraphTopology`, `node_prop_index()` is an identify function, so does not require additional storage. 
- In `EdgeShuffleTopology`, `node_prop_index()` depends on the shuffle and is stored in a vector (abstractly, a map).

To support this:

1. `node_prop_index()` can be implemented as a virtual function. This will have high compute overhead because this function is used in performance-critical loops.
2. `GraphTopologyTypes` can include a `node_prop_array` member which will be used in its `node_prop_index`. Derived classes don't change the definition of the function but change the definition of the member. This has high memory overhead for default `GraphTopology` but soon, we will have shuffle topologies as our default topology, so this may not matter much. 
3. Same as (2) except that we will define a new array type that can be specialized to omit storing memory if it's an identity-mapping array. This will avoid memory overhead and would have lower compute overhead than virtual functions. It would be just an if-condition for the flag on each call.

For this PR, I'd suggest (2). We can add (3) as an optimization in a follow-up PR. 

@ddn0 @amberhassaan Thoughts/suggestions welcome.

libgraph/include/katana/GraphTopology.h
```diff
@@ -1771,6 +1796,8 @@ struct PropertyGraphViews {
 };
 
 class KATANA_EXPORT PGViewCache {
+  std::shared_ptr<GraphTopology> original_topo_ =
+      std::make_shared<GraphTopology>();
```
rob-isaac: Nit: Do braces init rather than assignment

libgraph/include/katana/GraphTopology.h
rob-isaac: Okay. I think this is the wrong behavior in that case though. We should not be constructing an empty default topology. We should either leave no topologies in the cache and force loading from storage or maintain a single valid topology.

libgraph/src/GraphTopology.cpp
```diff
@@ -906,10 +915,40 @@ katana::ProjectedTopology::MakeTypeProjectedTopology(
...
+    return false;
+  }
+
+  original_topo_ = other;
+  return true;
+}
+
+void
+katana::PGViewCache::DropAllTopologies() noexcept {
+  original_topo_ = std::make_shared<katana::GraphTopology>();
```
rob-isaac, line 959: As mentioned above, I don't like this behavior. Can be targeted in a future PR

rob-isaac (commented):
This is looking good except for the virtual functions in the topology. I think it could be okay to change this in a future PR if there is some benchmarking effort against master to show that there is not a significant slowdown. @roshandathathri let me know if you agree.

Also, this is introducing some  name changes which will break the enterprise repo, so there will need to be a corresponding PR there if there isn't already.

yan-zaretskiy:
@rob-isaac I am about to add the commit that removes virtual calls, need to run some tests locally first. The corresponding `katana-enterprise` PR is there already, it's in draft mode at the moment. I'll promote it soon.

yan-zaretskiy:
@rob-isaac If you look at how the topology dropping logic is called upstream, it is used in the methods that instruct complete topology deallocation from memory. For one, changing this behavior is outside the scope of this PR, but your suggestion for the future fix contradicts the intended use of this method.

rob-isaac:
> @rob-isaac If you look at how the topology dropping logic is called upstream, it is used in the methods that instruct complete topology deallocation from memory. For one, changing this behavior is outside the scope of this PR, but your suggestion for the future fix contradicts the intended use of this method.

I am fine with that behavior, but I am not fine with  constructing an empty topology afterwards. I think there should be no associated topologies instead. Agree that it is not in the scope of this PR though

libgraph/include/katana/PropertyGraph.h
```diff
@@ -662,7 +672,13 @@ class KATANA_EXPORT PropertyGraph {
     return MakeResult(std::move(array));
   }
 
-  const GraphTopology& topology() const noexcept { return *topology_; }
+  void DropAllTopologies() noexcept {
```
roshandathathri, line 676: @witchel This is probably what you were asking about in Monday's meeting.

@mindv0rtex Please add a TODO to provide more fine-grained control of dropping specific topologies.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -529,12 +523,20 @@ class KATANA_EXPORT PropertyGraph {
...
   EntityTypeID GetTypeOfNode(Node node) const {
-    return node_entity_type_ids_[node];
+    auto idx = node_property_index(node);
+    return node_entity_type_ids_[idx];
+  }
+
+  /// \return returns the most specific edge entity type for @param edge
+  EntityTypeID GetTypeOfEdgeFromTopoIndex(Edge edge) const {
+    auto idx = edge_property_index(edge);
+    return edge_entity_type_ids_[idx];
```
roshandathathri: nit: call the function below

libgraph/include/katana/GraphTopology.h
```diff
@@ -1737,11 +1748,18 @@ struct PGViewBuilder<PGViewEdgeTypeAwareBiDir> {
     auto in_topo = viewCache.BuildOrGetEdgeTypeAwareTopo(
         pg, tsuba::RDGTopology::TransposeKind::kYes);
 
+    // The new topology can replace the defaut one.
```
roshandathathri, line 1751: default

libgraph/include/katana/GraphTopology.h
roshandathathri: @witchel @tylershunt FYI

roshandathathri (approved):
LGTM

libgraph/include/katana/PropertyGraph.h
witchel: My plan for the memory management RFC is that we will have many Managers, like a view manager, that communicate with the overall memory manager.  The overall memory manager might say to the view manager, please reduce your memory use, so it would be helpful to know that there are topologies that are not in use and safe to drop.  It would also be useful to know (approximately) how much memory that frees up. (All of this not for this PR, obviously).

libgraph/src/PropertyGraphRetractor.cpp
```diff
@@ -15,7 +15,6 @@ katana::PropertyGraphRetractor::InformPath(const std::string& input_path) {
 
 Result<void>
 katana::PropertyGraphRetractor::DropTopologies() {
-  //TODO: emcginnis reset all topologies in PGViewCache
-  pg_->topology_ = std::make_shared<katana::GraphTopology>();
+  pg_->DropAllTopologies();
```
arthurp, line 18: Why do these methods have different names?

arthurp (commented):
LGTM. But I don't really have context, so I don't think I should be the final approval.

libgraph/include/katana/GraphTopology.h
```diff
@@ -58,13 +62,14 @@ class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
 
   GraphTopology(const GraphTopology&) = delete;
   GraphTopology& operator=(const GraphTopology&) = delete;
+  virtual ~GraphTopology();
```
amberhassaan, line 65: I saw a comment that said that you decided to remove `virtual` functions. So I think you should mark this destructor as `non virtual` too so that no Vtable is generated.

libgraph/include/katana/GraphTopology.h
```diff
@@ -172,8 +178,47 @@ class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
...
+  }
+
+  tsuba::RDGTopology::EdgeSortKind edge_sort_state() const noexcept {
+    return edge_sort_state_;
+  }
+
   void Print() const noexcept;
 
+protected:
+  void SetTransposeState(
```
amberhassaan, line 203: why have these set methods here? Why should the default topology support these?

libgraph/src/GraphTopology.cpp
```diff
@@ -936,7 +989,23 @@ katana::PGViewCache::BuildOrGetEdgeShuffTopo(
...
+    const tsuba::RDGTopology::TransposeKind& tpose_kind,
+    const tsuba::RDGTopology::EdgeSortKind& sort_kind) noexcept {
+  return BuildOrGetEdgeShuffTopoImpl(pg, tpose_kind, sort_kind, true);
+}
+
+std::shared_ptr<katana::EdgeShuffleTopology>
+katana::PGViewCache::BuildOrGetEdgeShuffTopoImpl(
+    katana::PropertyGraph* pg,
+    const tsuba::RDGTopology::TransposeKind& tpose_kind,
+    const tsuba::RDGTopology::EdgeSortKind& sort_kind, bool pop) noexcept {
```
amberhassaan, line 1007: How does `pop` work?

amberhassaan (commented):
OK, I need some more background to understand the changes. I'll meet with Yan separately.

libgraph/include/katana/GraphTopology.h
rob-isaac: It is to enable derived classes to set the private members without giving them protected status, which is generally considered bad practice. I think they are only set in the constructors though, so it prob makes more sense to just have a GraphTopology constructor that takes them as input.

yan-zaretskiy: @rob-isaac Even `protected`, you can't put them in the class initializer list... To your point, I specifically wanted to avoid these params being passed to the GraphTopology ctor, because for GraphTopology proper their values should be defaults. It's ugly, because we are trying to avoid virtual methods, otherwise I could have replaced member variables with virtual methods that return fixed values in GraphTopology proper.

arthurp: Why are we trying to avoid virtual members? Is there a reason to avoid them in all cases? Or is it reasonable to have them in specific cases where they are objectively cleaner than other approaches? 

Personally I would go with the virtual methods in this case and then benchmark it. Branch predictors are really good these days, so virtual call overhead can be quite low as long as each generated call instruction almost always calls the same target implementation. And I suspect that will be the case here since most parts of the system will always be using the same view.

rob-isaac: @mindv0rtex couldn't you maintain the current ctor, which sets the values to the defaults, but also maintain another ctor which explicitly sets these values? E.g:
```
public:
GraphTopology(<current_args>)
protected:
GraphTopology(<current_args>, shuffle_type, ect) // this will set the values
```
Then derived class constructors can call the second ctor to set the correct values instead of the defaults.
```
ShuffleTopology(<args>) : GraphTopology(<args>, /*shuffle type*/ true, ect), <other stuff>

PR #741, opened 2021-12-07
arrow 6.0, VisitArrow generalization, ArrayFromScalars simplification

danielmawhirter:
1. arrow 6.0
   - had arrow 4.0
   - adopt 6.0
2. VisitArrow Generalization
   - ArrowVisitor.h had a hardcoded list of types that it supported
   - Expand that to the universe of all Arrow types
   - NOTE: filtering the universe happens at the AcceptTypes level, so this doesn't require heavy changes
3. ArrayFromScalars simplification
   - Used a Visitor to implement ArrayFromScalars, required explicit handling per-type
   - Use arrow 6.0 API to append a vector<Scalar> to a Builder
   - NOTE: not changing AppendToBuilder(Scalar) because that's type-flexible (all numerics are inter-operable)


libsupport/src/ArrowVisitor.cpp
```diff
@@ -134,92 +134,13 @@ class AppendScalarToBuilderVisitor : public katana::ArrowVisitor {
...
-      if (scalar != nullptr && scalar->is_valid) {
-        const ScalarType* typed_scalar = static_cast<ScalarType*>(scalar.get());
-        builder->UnsafeAppend(typed_scalar->value);
-      } else {
-        builder->UnsafeAppendNull();
-      }
+    if (auto res = builder->AppendScalar(scalar); !res.ok()) {
+      return KATANA_ERROR(
+          katana::ErrorCode::ArrowError, "incompatible types: {}, {}",
+          scalar.type->ToString(), builder->type()->ToString());
```
rob-isaac: scalar.type(), right?

libsupport/include/katana/ArrowVisitor.h
```diff
@@ -204,40 +205,24 @@ struct ArrowDispatcher {
     auto type_id =
         GetArrowTypeID(std::get<index>(std::forward_as_tuple(args...)));
 
-    switch (type_id) {
-#define TYPE_CASE(EnumType)                                                    \
-  case arrow::Type::EnumType: {                                                \
+#define ENUM_TYPE_CASE(z, n, unused)                                           \
+  case static_cast<arrow::Type::type>(n): {                                    \
```
rob-isaac: looks like z is also unused?

libsupport/include/katana/ArrowVisitor.h
```diff
@@ -204,40 +205,24 @@ struct ArrowDispatcher {
...
-      TYPE_CASE(LIST)  // TODO(daniel) DEPRECATED
-      TYPE_CASE(LARGE_LIST)
-      TYPE_CASE(NA)
-#undef TYPE_CASE
+
+    switch (type_id) {
+      // 37 in 4.0
+      // 38 in 6.0
+      static_assert(arrow::Type::MAX_ID == 38);
+      BOOST_PP_REPEAT(38, ENUM_TYPE_CASE, ~);
```
rob-isaac: add comment that we are switching over all arrow types, as this is no longer clear

libsupport/src/ArrowVisitor.cpp
rob-isaac: Also, this could use KATANA_CHECKED_CONTEXT, right?

rob-isaac (approved):
Not familiar with the versioning stuff, but everything else looks good

libsupport/src/ArrowVisitor.cpp
danielmawhirter: arrow::Scalar is the exception, `type` is a member instead of a method

libsupport/include/katana/ArrowVisitor.h
danielmawhirter: That's accurate, it's an internal thing for the PP_REPEAT call (so that they can nest arbitrarily). I'll add a comment

ddn0 (approved):

katana_requirements.yaml
```diff
@@ -195,11 +195,11 @@ libarrow-python:
     conda: null
     pip: null
     apt: libarrow-python-dev
-  version: &arrow_version [ 4, 5 ]
+  version: &arrow_version [ 6 ]
```
arthurp: This isn't valid syntax. It needs to be either `[6, 7]` to mean at least 6 and strictly less than 7 or `[6, null]` meaning at least 6 and no upper bound.

katana_requirements.yaml
```diff
@@ -323,7 +323,7 @@ parquet-cpp:
     - conda
     - apt
   version_overrides:
-    apt: "=4.0.1-1"
+    apt: "=6.0.1-1"
```
arthurp, line 326: Please verify the current package number. It might not be `.0.1` if there has been a different number of patches.

libsupport/include/katana/ArrowVisitor.h
```diff
@@ -197,47 +198,49 @@ struct ArrowDispatcher {
...
-      TYPE_CASE(STRUCT)
-      TYPE_CASE(LIST)  // TODO(daniel) DEPRECATED
-      TYPE_CASE(LARGE_LIST)
-      TYPE_CASE(NA)
-#undef TYPE_CASE
+
+    switch (type_id) {
+      // repeat the ENUM_TYPE_CASE macro NUM_TYPES times
+      // ~ is an unparsed token as recommended by Boost docs
+      BOOST_PP_REPEAT(NUM_TYPES, ENUM_TYPE_CASE, ~);
```
arthurp: Is there a way to do this with template metaprogramming instead of preprocessor meta programming? @ddn0, thoughts? I think it should be possible to build the switch as a bunch of if statements in templates and that would probably be safer and easier to debug.

libtsuba/include/tsuba/ParquetWriter.h
```diff
@@ -20,7 +20,7 @@ class KATANA_EXPORT ParquetWriter {
     /// 2.0. In Arrow to Parquet version 1.0, nanosecond timestamps will get
     /// truncated to milliseconds.
     parquet::ParquetVersion::type parquet_version{
-        parquet::ParquetVersion::PARQUET_2_0};
+        parquet::ParquetVersion::PARQUET_2_6};
```
arthurp, line 23: Does this break any compatibility with data that has been created? Does this mean new data will not be readable by old versions? This was a problem in the past so we should flag it. @gliga

scripts/setup_ubuntu.sh
```diff
@@ -88,6 +88,7 @@ if [[ -n "${SETUP_TOOLCHAIN_VARIANTS}" ]]; then
 fi
 
 if [ "$VERSION" == "21" ]; then
+    # TODO(daniel) revise?
```
arthurp: Huh? This is not a very specific TODO next to some complex choices.

libsupport/include/katana/ArrowVisitor.h
arthurp: Another reason to not use the preprocessor is the need for actual numbers for counting instead of just using `arrow::Type::MAX_ID` directly.

arthurp (approved):
Please look into eliminating the preprocessor metaprogramming. Hopefully someone better at C++ can provide some input.

No need to check back after addressing my comments.

libtsuba/include/tsuba/ParquetWriter.h
ddn0: Arrow 4.0, PARQUET_2_0 (e.g., support for ns timestamps) effectively means arrow 6.0 PARQUET_2_6. There is no PARQUET_2_0 in arrow 6.0.

libsupport/include/katana/ArrowVisitor.h
ddn0: If you really wanted to do this with template metaprogramming, you could try:

```
template<int i> Dispatch(...) {
  if (v == i) {
    ...
  } else {
    return Dispatch<i-1>(...)
  }
}
```

I don't know if this would be any easier to debug though, and you should check the microbenchmark before and after. It might increase compile times, so you should check that too.

libtsuba/include/tsuba/ParquetWriter.h
danielmawhirter: The associated [task](https://issues.apache.org/jira/browse/ARROW-13794) for 6.0.0 read:
```
ParquetVersion::type currently has two members: PARQUET_1_0 and PARQUET_2_0.
The latter enables a hodgepodge of features which were actually added
in Parquet format 2.4 and 2.6 (and others?).

To allow more fine-grained selection of the format version, we should deprecate
this value and add PARQUET_2_4 and PARQUET_2_6 values.
```

I lack any further info to understand how this will affect our backward compatibility. I would suspect that, since this only appears in the Writer, we should be able to read old data, but old versions may not be able to read newly-written data

gliga: Thanks for flagging this case.  I cannot comment on compatibility but I will be on the lookout for anything related to this change.  @e-mcginnis can comment more if this would  impact anything about stored graphs.

e-mcginnis: Thanks for the alert
Based on
>I would suspect that, since this only appears in the Writer, we should be able to read old data, but old versions may not be able to read newly-written data

that really sounds like we should be upreving the `storage_format_version`, since software without this PR will be unable to read rdgs output by software with this PR

The exact process to do this is still under active refinement, but the most important thing to do is bump https://github.com/KatanaGraph/katana/blob/1fbad29acbf94dca53cf04e8c59b76d14556cb6b/libtsuba/src/RDGPartHeader.h#L599 in this PR

e-mcginnis (changes requested):
blocking approval based on storage_format_version compatibility 
see: https://github.com/KatanaGraph/katana/pull/741#discussion_r767018548

libtsuba/include/tsuba/ParquetWriter.h
ddn0: Just to be clear, PARQUET_2_0 (in libarrow 4.0) is PARQUET_2_6 (in libarrow 6.0):

https://github.com/apache/arrow/pull/11031/files#diff-fc3371fc0696bd56bf179db429eabc666d90eb854c934a09cbe58d511bd13546R43

If you thought PARQUET_2_0 was okay before, you should think the same about PARQUET_2_6.

e-mcginnis: ah, I understand what you mean now @ddn0
though your use of "effectively" here 
> Arrow 4.0, PARQUET_2_0 (e.g., support for ns timestamps) effectively means arrow 6.0 PARQUET_2_6

worries me a bit

looking at that diff, it looks like we shouldn't see any difference between our usage of Arrow 4.0, PARQUET_2_0 and  Arrow 6.0 PARQUET_2_6

I think the proper test to ensure this is the case would be:
1) take an existing interesting rdg (ldbc_003 should be sufficient)
2) write out a copy of that rdg using the katana build by this PR 
3) load that rdg using katana master, load the original rdg, ensure they match

@danielmawhirter I can help you out with the specifics of how to go about this
Ensuring the original_rdg_propertygraph.Equals(new_rdg_property_graph) should be sufficient for (3) 

I'll get a test put together to do this more automatically in the future, but for now we will have to do it manually.

ddn0: "effectively" as in for "all intents and purposes".

Parquet only has two file format versions: 1 and 2.

Writers normalize their data types based on the minor version.

In arrow 6.0,

2.4: Writers use the new logical types supported by the parquet file format 2 but always truncate timestamps to microseconds
2.6: Writers do not truncate timestamps to microseconds.

In arrow 4.0, when the writer version is "2.0", writers use the new logical types supported by the parquet file format 2 and do not truncate timestamps to microseconds.

Compliant parquet file format 2.0 readers should have always supported both 2.4 and 2.6 style writers because they are just using the file format features supported by the underlying file. Though this is somewhat academic because I think the only implementation of parquet +2.0 readers or writers is libarrow.

You are welcome to make an explicit test for this though.

e-mcginnis: thats convincing enough for me

e-mcginnis (approved):

tylershunt (approved):

libsupport/include/katana/ArrowVisitor.h
```diff
@@ -212,32 +220,52 @@ struct ArrowDispatcher {
     return Call<ArrowTypes..., ArrowType>(                                     \
         std::forward<Visitor>(visitor), std::forward<Args>(args)...);          \
   }
-      TYPE_CASE(INT8)
+      TYPE_CASE(NA)
```
danielmawhirter, line 223: @amansavani

PR #742, opened 2021-12-08
Add views to pagerank and fix pull algorithms

gurbinder533:


libgalois/src/analytics/pagerank/pagerank-push.cpp
```diff
@@ -31,16 +31,21 @@ struct NodeResidual : katana::AtomicPODProperty<PRTy> {};
 
 using NodeData = std::tuple<NodeValue, NodeResidual>;
 using EdgeData = std::tuple<>;
-typedef katana::TypedPropertyGraph<NodeData, EdgeData> Graph;
-typedef typename Graph::Node GNode;
+// typedef katana::TypedPropertyGraph<NodeData, EdgeData> Graph;
+// typedef typename Graph::Node GNode;
+
+using Graph = katana::TypedPropertyGraphView<
```
uditagar: I am thinking (even for other apps) that we should not create a default view, instead just use the full property graph. Default views somehow have very high overhead at the moment.

uditagar (approved):
everything looks fine, except Default view. This Default view is plaguing all apps (in which it is implemented).

libgalois/src/analytics/pagerank/pagerank-push.cpp
gurbinder533: We should definitely use default view but it should not have overhead. I did not measure it but why should it have any overhead. Did you measure it?

amazloumi (commented):
Everything looks fine to me!

libgalois/src/analytics/pagerank/pagerank-push.cpp
uditagar: for louvain it takes extra 70 seconds with 32 threads on friendster. What I meant is to use a template `Graph`, which can be used to run our apps with views as well as with the full property graph. just `using Graph = ...` thing will change.

gurbinder533: No, we should move everything to use views I feel. Mix and matching different types is not good.

PR #743, opened 2021-12-08
Fix RDGSlice schemas

scober:
RDGSlice declares `full_node_schema()` and `full_edge_schema()` but doesn't actually define them. Define them.

Also change the semantics of the `PropStorageInfo` state machine for property slices. In particular, don't mark sliced properties dirty when they are first loaded. This simplifies the logic loading and unloading properties in `RDGSlice`.

Also add a `RDGSlice` test.


libtsuba/src/AddProperties.cpp
```diff
@@ -227,12 +227,13 @@ tsuba::AddPropertySlice(
         -> katana::CopyableResult<void> {
       KATANA_CHECKED_CONTEXT(
           add_fn(props), "adding: {}", std::quoted(prop->name()));
+      // NB: Sliced properties don't fit super cleanly into the PropStorageInfo
+      // model. This property is dirty in the sense that there is no file on
+      // storage that exactly matches it but it is clean in the sense that it
+      // has not been modified. Leave it as clean to simplify loading/unloading
+      // logic in RDGSlice.
       prop->WasLoaded(props->field(0)->type());
```
pyrito, line 235: Just to confirm, it's considered "dirty" in some sense because we have a slice of a partition which technically doesn't exist on disk on its own?

libtsuba/test/rdg-slice.cpp
```diff
@@ -0,0 +1,120 @@
...
+  KATANA_LOG_ASSERT(rdg_slice.node_properties()->num_columns() == 2);
+  KATANA_LOG_ASSERT(
+      rdg_slice.full_edge_schema()->num_fields() == expected_num_edge_props);
+  KATANA_LOG_ASSERT(rdg_slice.edge_properties()->num_columns() == 2);
+
+  // load and unload some non-existent properties
+  auto res = rdg_slice.load_node_property("does not exist");
+  KATANA_LOG_ASSERT(!res && res.error() == tsuba::ErrorCode::PropertyNotFound);
+  res = rdg_slice.unload_node_property("does not exist");
+  KATANA_LOG_ASSERT(!res && res.error() == tsuba::ErrorCode::PropertyNotFound);
```
pyrito, line 85: nit: repeat the paradigm for edge property as well, just for code coverage

pyrito (approved):
I appreciate the test!

libtsuba/src/AddProperties.cpp
scober: Yes. A truly clean property would be able to be recreated in memory from files on storage. This property can be recreated from a file on storage + some in-memory metadata (the slice arg).

scober:
> I appreciate the test!

I'm trying to be better about testing.

libtsuba/src/RDGCore.cpp
```diff
@@ -121,10 +121,33 @@ EnsureTypeLoaded(const katana::Uri& rdg_dir, tsuba::PropStorageInfo* psi) {
   }
   return katana::ResultSuccess();
 }
+
+std::shared_ptr<arrow::Schema>
+schemify(const std::vector<tsuba::PropStorageInfo>& prop_info_list) {
```
witchel, line 126: `Schemify`.
As in Schemafication complete!

libtsuba/test/rdg-slice.cpp
```diff
@@ -0,0 +1,124 @@
...
+      rdg_slice.full_node_schema()->num_fields() == expected_num_node_props);
+  KATANA_LOG_ASSERT(rdg_slice.node_properties()->num_columns() == 0);
+  KATANA_LOG_ASSERT(
+      rdg_slice.full_edge_schema()->num_fields() == expected_num_edge_props);
+  KATANA_LOG_ASSERT(rdg_slice.edge_properties()->num_columns() == 0);
+
+  // load all properties
+  for (int64_t i = 0; i < expected_num_node_props; ++i) {
+    KATANA_CHECKED(rdg_slice.load_node_property(
+        rdg_slice.full_node_schema()->field(i)->name()));
```
witchel, line 48: While this code is natural to write, it causes the repeated creation of the full schema.  I wonder if it would be pedagogically superior to `auto full_schema = rdg_slice.full_node_schema();` and then use `full_schema `in the loop.

witchel (approved):
Looks good with those tests.

libtsuba/test/rdg-slice.cpp
scober: I would be open to that. But I kind of like this style. One of the things that this version tests is that loading properties doesn't affect the full schema. If we changed to your proposal, we would have to explicitly test that somewhere else. I suppose explicit tests are probably better than implicit tests.

witchel: Well, at least you thought about it.  If you are checking that loading properties doesn't change the full schema, it probably merits a comment.

PR #745, opened 2021-12-09
Add views to BC

amazloumi:
Not sure where to use the Transpose view. I have used the Default view for both Level and Outer algorithms.
Please leave a comment about it when reviewing so I can change it to Transpose view if needed somewhere.


uditagar:
Yasin, can you please check the performance of this commit with the one on master? both total and execution time. Let me know if you face any issues.

amazloumi:
> Yasin, can you please check the performance of this commit with the one on master? both total and execution time. Let me know if you face any issues.

On what machine can I run the performance test? And also if any specific graphs and/or configs usually are used for this purpose?

uditagar:
> > Yasin, can you please check the performance of this commit with the one on master? both total and execution time. Let me know if you face any issues.
> 
> On what machine can I run the performance test? And also if any specific graphs and/or configs usually are used for this purpose?

just do an aws. let me know if you run into any problem with that one. just use friendster for testing for now.

libgraph/src/analytics/betweenness_centrality/level.cpp
```diff
@@ -121,8 +122,8 @@ LevelSSSP(
           KATANA_LOG_ASSERT(src_data.current_dist == current_level);
 
           for (auto e : graph->edges(n)) {
-            auto dest = graph->GetEdgeDest(e);
-            auto& dst_data = (*graph_data)[*dest];
+            auto dest = graph->edge_dest(e);
```
gurbinder533, line 125: Avoid `auto` if possible.

gurbinder533 (approved):
Please add tests for all the algorithms in CMake files for ctests.

uditagar (approved):
This is good to go. seems there are no performance issues with Default views, it seems.

PR #746, opened 2021-12-09
adding views to jaccard

amazloumi:


uditagar:
same performance benchmarking required for this as well. Thanks

gurbinder533 (approved):
LGTM.

uditagar:
Do not commit this to master until we are sure that there is no change in total/execution time with respect to `master`.

uditagar (approved):
This is good to go. seems there are no performance issues with Default views, it seems.

PR #748, opened 2021-12-10
tsuba: Parallelize CopyRDG

pyrito:
This PR adds a `do_all` loop to parallelize the CopyRDG operation in tsuba. The code previously made all the copies serially, so this is a minor optimization in issuing the copy requests in parallel. 

JIRA: https://katanagraph.atlassian.net/browse/KAT-1558

### DIFF
--- a/libtsuba/src/tsuba.cpp
+++ b/libtsuba/src/tsuba.cpp
@@ -273,26 +273,26 @@ katana::Result<void>
 tsuba::CopyRDG(std::vector<std::pair<katana::Uri, katana::Uri>> src_dst_files) {
   katana::DynamicBitset manifest_uri_idxs;
   katana::do_all(
-    katana::iterate(size_t{0}, src_dst_files.size()),
-        [&](size_t i) {
-      auto [src_file_uri, dst_file_uri] = src_dst_files[i];
-      // We save the names of all the manifest files and we write them out at the end.
-      if (tsuba::RDGManifest::IsManifestUri(src_file_uri)) {
-        manifest_uri_idxs.set(i);
-        return;
-      }
-      // NB: can't use KATANA_CHECKED here because of the do_all, so we keep track of the 
-      // error for now and then return it later
-      tsuba::FileView fv;
-      auto fv_res = fv.Bind(src_file_uri.string(), true);
-      if (!fv_res) {
-        KATANA_LOG_FATAL("Failed to bind {}", src_file_uri.string());
-      }
-      auto fs_res = tsuba::FileStore(dst_file_uri.string(), fv.ptr<char>(), fv.size());
-      if (!fs_res) {
-        KATANA_LOG_FATAL("Failed to store to {}", dst_file_uri.string());
-      }
-    });
+      katana::iterate(size_t{0}, src_dst_files.size()), [&](size_t i) {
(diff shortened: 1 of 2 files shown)


libtsuba/src/tsuba.cpp
```diff
@@ -270,24 +271,32 @@ tsuba::CreateSrcDestFromViewsForCopy(
...
+      // error for now and then return it later
+      tsuba::FileView fv;
+      auto fv_res = fv.Bind(src_file_uri.string(), true);
+      if (!fv_res) {
+        KATANA_LOG_FATAL("Failed to bind {}", src_file_uri.string());
+      }
+      auto fs_res = tsuba::FileStore(dst_file_uri.string(), fv.ptr<char>(), fv.size());
+      if (!fs_res) {
+        KATANA_LOG_FATAL("Failed to store to {}", dst_file_uri.string());
+      }
```
arthurp: You don't seem to be collecting the errors to use later. You are just crashing. Please actually do what the comment says. We have a copyable version of ErrorInfo which should work for this.

pyrito: Oh right, so I should have updated the comment, but I was trying to store the error and return the error later but some of the collective calls are in `Network.h` which wasn't available in the open repo?

libtsuba/src/tsuba.cpp
```diff
@@ -270,24 +271,32 @@ tsuba::CreateSrcDestFromViewsForCopy(
...
+        auto fv_res = fv.Bind(src_file_uri.string(), true);
+        if (!fv_res) {
+          KATANA_LOG_FATAL("Failed to bind {}", src_file_uri.string());
+        }
+        auto fs_res =
+            tsuba::FileStore(dst_file_uri.string(), fv.ptr<char>(), fv.size());
+        if (!fs_res) {
+          KATANA_LOG_FATAL("Failed to store to {}", dst_file_uri.string());
+        }
+      });
```
roshandathathri, line 295: (1) Please use `katana::chunk_size<1>()` if each iteration of the loop is expected to be a heavy-weight operation. In this case, I think it is heavyweight.

(2) Please use `katana::steal()` if each iteration is expected to take significantly different amounts of time. I'm not sure about that in this case.

pyrito: 1.) what exactly is the chunk size?
2.) I would expect each `FileStore` call to take varying amounts of time based on the size of what I'm copying.

roshandathathri: (1) It's the unit of work distribution among threads. Even without stealing, it can impact performance. The default chunk size is 32. Consider 2 threads and 285 iterations, that is 9 chunks. One thread with get 5 and the other 4. That could lead to sub-optimal performance if each iteration takes significant amount of time. 

(2) Then use `steal()`. Chunk size would matter even more in this case because that's the unit of stealing/work-distribution.

roshandathathri: Smaller chunk size => larger number of chunks => more overhead in distribution/stealing

There's a trade-off. I'm assuming in this case, we do not millions of small files, so smaller chunk sizes are better.

pyrito:
Ok I'm closing this because I realize now that the memory usage in the parallel setting is going to be unpredictable.

PR #749, opened 2021-12-10
DynamicBitset: adds a serial count function for use in parallel regions

l-hoang:
DynamicBitset's count function cannot be called in a serial region as it
uses do_all. This commit adds a serial version of the count function so
that users can count the number of set bits in the different bitsets
in parallel.

### DIFF
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -309,11 +309,20 @@ public:
   void bitwise_xor(const DynamicBitset& other1, const DynamicBitset& other2);
 
   /**
-   * Count how many bits are set in the bitset
+   * Count how many bits are set in the bitset. Do not call in a parallel
+   * region as it uses a parallel loop.
    *
    * @returns number of set bits in the bitset
    */
   size_t count() const;
+
+  /***
+   * Count number of set bits in the bitset serially. Useful if
+   * you need to count different bitsets on different threads.
+   *
+   * @returns number of set bits in the bitset
+   */
+  size_t SerialCount() const;
 
   /**
    * Returns a vector containing the set bits in this bitset in order
--- a/libgalois/src/DynamicBitset.cpp
+++ b/libgalois/src/DynamicBitset.cpp
@@ -108,6 +108,21 @@ katana::DynamicBitset::count() const {
   return ret.reduce();
 }
 
+size_t
+katana::DynamicBitset::SerialCount() const {
+  size_t ret = 0;
+  for (uint64_t n : bitvec_) {
+#ifdef __GNUC__
+    ret += __builtin_popcountll(n);
+#else
+    n = n - ((n >> 1) & 0x5555555555555555UL);
+    n = (n & 0x3333333333333333UL) + ((n >> 2) & 0x3333333333333333UL);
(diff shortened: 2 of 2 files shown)


libgalois/src/DynamicBitset.cpp
```diff
@@ -108,6 +108,21 @@ katana::DynamicBitset::count() const {
   return ret.reduce();
 }
 
+size_t
+katana::DynamicBitset::SerialCount() const {
+  size_t ret = 0;
+  for (uint64_t n : bitvec_) {
+#ifdef __GNUC__
```
srogatch, line 115: Will it take this path in `clang` too?

libgalois/src/DynamicBitset.cpp
```diff
@@ -108,6 +108,21 @@ katana::DynamicBitset::count() const {
...
+    ret += __builtin_popcountll(n);
+#else
+    n = n - ((n >> 1) & 0x5555555555555555UL);
```
srogatch, line 118: I would prefer to resort to `int _popcnt64 (__int64 a)` intrinsic from `<immintrin.h>`. Perhaps it can be just used always, as it is unlikely that someone will use Katana on a CPU without `POPCNT`. See https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=pop&ig_expand=5445

srogatch (commented):
Shall the same be added to `libgluon/include/katana/CUDA/DynamicBitset.h` ?

libgalois/src/DynamicBitset.cpp
l-hoang: I'm not actually sure; I just copied what was there in the other version.

l-hoang:
> Shall the same be added to libgluon/include/katana/CUDA/DynamicBitset.h ?

I would probably say not until it is necessary unless there is a pressing need to do it now (i.e. things will break if I
don't do it). It's unlikely you'll want to be doing things serially on the CPU if you have a GPU, for instance.

libgalois/src/DynamicBitset.cpp
```diff
@@ -108,6 +108,21 @@ katana::DynamicBitset::count() const {
...
+    n = (n & 0x3333333333333333UL) + ((n >> 2) & 0x3333333333333333UL);
+    ret += (((n + (n >> 4)) & 0xF0F0F0F0F0F0F0FUL) * 0x101010101010101UL) >> 56;
+#endif
```
roshandathathri, line 121: nit: have a common function for doing this in both serial and parallel

libgalois/include/katana/DynamicBitset.h
```diff
@@ -309,12 +309,21 @@ class KATANA_EXPORT DynamicBitset {
...
    */
   size_t count() const;
 
+  /***
+   * Count number of set bits in the bitset serially. Useful if
+   * you need to count different bitsets on different threads.
+   *
+   * @returns number of set bits in the bitset
+   */
+  size_t SerialCount() const;
```
roshandathathri, line 325: nit: I'd prefer a better API but I don't have concrete suggestions now, so this is fine for now.

roshandathathri (approved):
LGTM

PR #750, opened 2021-12-12
Property cache now uses paths, is exclusive with RDG

witchel:
The cache takes a simple katana::Uri as its key, which
allows a straightforward and generic cache implementation
without concepts specific to property caches.

Also, the property cache is now exclusive with
the graph, meaning a property column either resides
in the graph or in the cache, never both.  This makes
the cache more like fast storage, it can always be cleared,
and we never evict an in-use property.

Or how Dr. Hunt showed me how to stop worrying and love the file system namespace.  Tyler suggested that the cache be exclusive with the graph and with that perspective, there is no longer a need to know things like the name of the property or whether it is a node or edge property.  Cool!

Though, @tylershunt, I wonder if it is worthwhile to embed whether a property is a node or edge property in the property's file name.  This would just be for redundancy.  We might also want to make it more clear what the legal format of a file name is. 
NAME-randomstring is the current format, but it could also be NAME-n|e-randomstring.

@ddn0, I hope you see this as an improvement to the cache code.


libsupport/test/cache.cpp
```diff
@@ -7,45 +7,11 @@
 #include "katana/Logging.h"
 #include "katana/Random.h"
 
-enum class NodeEdge { kNode, kEdge };
-
-// This code is replicated from libtsuba::PropertyCache.h to allow testing of
-// libsupport::Cache.h (libtsuba depends on libsupport, not the other way around)
```
pyrito, line 13: Do you also have to remove `PropertyCacheKey` in `libtsuba::PropertyCache.h`?

libsupport/include/katana/Cache.h
```diff
@@ -53,41 +53,33 @@ struct CacheStats {
...
 
-template <typename Key, typename Value, typename CallerPointer = void*>
+template <typename Value>
 class KATANA_EXPORT Cache {
+  using Key = katana::Uri;
   using ListType = std::list<Key>;
   struct MapValue {
     Value value;
     // This allows us to delete the old position in the LRU list without a scan
     typename ListType::iterator lru_it;
```
pyrito, line 62: So an iterator is used as an explicit pointer to a position in the doubly linked list correct? I was wondering why we didn't use something like an index or pointer, but this looks a lot cleaner. Very interesting

libsupport/include/katana/Cache.h
```diff
@@ -251,10 +216,9 @@ class KATANA_EXPORT Cache {
   CacheStats cache_stats_;
 
   std::function<size_t(const Value& value)> value_to_bytes_;
-  std::function<void(const Key& key, uint64_t approx_bytes, CallerPointer rdg)>
```
pyrito, line 254: Why was there a need for a specific `evict_cb` (obviously not needed anymore, but just trying to understand the context)?

libsupport/include/katana/URI.h
```diff
@@ -12,12 +12,6 @@
 namespace katana {
 
 class KATANA_EXPORT Uri {
-  std::string scheme_;
-  std::string path_;
-  std::string string_;
-
-  Uri(std::string scheme, std::string path);
```
pyrito, line 19: Appreciate moving this down to an explicit `private` block, but out of curiosity, would these have been public by default?

libsupport/test/cache.cpp
pyrito: Unless this is still being used. I'm not too sure.

witchel: There is no more` libtsuba::PropertyCache.h`.

libsupport/include/katana/Cache.h
witchel: I like it.  I coped the design from boost.
`https://www.boost.org/doc/libs/1_67_0/boost/compute/detail/lru_cache.hpp`

libsupport/test/cache.cpp
pyrito: Oops! Skipped over the file that was completely deleted :)

pyrito (approved):

libsupport/include/katana/Cache.h
witchel: My first cache design held all the properties that are present in the graph, plus (hopefully) more.  If a property got evicted, the cache needed to notify the graph to unload the property.  That isn't a great model because currently graph programmers tell us they need a property via the `EnsurePropertyLoaded`/`UnloadProperty `interface.  The code they write will not work if a property gets evicted while it is ensure loaded.

libsupport/include/katana/URI.h
witchel: They are private by default.  But our coding standard is to put private data at the bottom of the class.

tylershunt:
> Though, @tylershunt, I wonder if it is worthwhile to embed whether a property is a node or edge property in the property's file name.

I was actually hoping to reduce information in file names; even removing the name of the property from them.

witchel:
> > Though, @tylershunt, I wonder if it is worthwhile to embed whether a property is a node or edge property in the property's file name.
> 
> I was actually hoping to reduce information in file names; even removing the name of the property from them.

.......why?

libtsuba/include/tsuba/RDGSlice.h
```diff
@@ -20,6 +20,8 @@ namespace tsuba {
 class RDGManifest;
 class RDGCore;
 
+enum class NodeEdge { kNode = 10, kEdge, kNeitherNodeNorEdge };
```
tylershunt: Is this still needed? I saw that you saved it from the PropertyCache header but I'm not sure why.

libsupport/include/katana/Cache.h
```diff
@@ -53,41 +53,33 @@ struct CacheStats {
   uint64_t insert_hit_count{0ULL};
 };
 
-template <typename Key, typename Value, typename CallerPointer = void*>
+template <typename Value>
 class KATANA_EXPORT Cache {
```
tylershunt, line 56: Just to make sure: this makes a `Cache` an explicit holder of things that are backed by storage? That makes sense to me a good thing to be true, but wanted to make sure.

libsupport/include/katana/Cache.h
```diff
@@ -16,10 +16,10 @@
 #include <optional>
```
tylershunt: nit: this header no longer uses `std::optional`

tylershunt (approved):
LGTM

tylershunt:
> > > Though, @tylershunt, I wonder if it is worthwhile to embed whether a property is a node or edge property in the property's file name.
> > 
> > 
> > I was actually hoping to reduce information in file names; even removing the name of the property from them.
> 
> .......why?

When things have common prefixes it makes them more likely that GS/S3 will restrict our bandwidth to them (not on purpose but they tend to end up on the same machine). I can't find the exact discussion now but this came up a while ago as something we might want to change. CC @ddn0

libtsuba/include/tsuba/RDGSlice.h
witchel: There is code in RDGSlice.cpp that use it.  The code looks useful.

witchel: But I moved the definition from the .h file into the .cpp file.  That seems worthwhile.

libsupport/include/katana/Cache.h
witchel: Yes, and I added documentation to `PropertyCache `to make this clear.

witchel:
> > > > Though, @tylershunt, I wonder if it is worthwhile to embed whether a property is a node or edge property in the property's file name.
> > > 
> > > 
> > > I was actually hoping to reduce information in file names; even removing the name of the property from them.
> > 
> > 
> > .......why?
> 
> When things have common prefixes it makes them more likely that GS/S3 will restrict our bandwidth to them (not on purpose but they tend to end up on the same machine). I can't find the exact discussion now but this came up a while ago as something we might want to change. CC @ddn0

You have made me so sad.  So very, very sad.

witchel:
> > > > > Though, @tylershunt, I wonder if it is worthwhile to embed whether a property is a node or edge property in the property's file name.
> > > > 
> > > > 
> > > > I was actually hoping to reduce information in file names; even removing the name of the property from them.
> > > 
> > > 
> > > .......why?
> > 
> > 
> > When things have common prefixes it makes them more likely that GS/S3 will restrict our bandwidth to them (not on purpose but they tend to end up on the same machine). I can't find the exact discussion now but this came up a while ago as something we might want to change. CC @ddn0
> 
> You have made me so sad. So very, very sad.

How about we just start with the random string and then append the property name and n or e?

PR #751, opened 2021-12-13
Deprecate "build/inputs" use "test-datasets" git submodule instead

e-mcginnis:
This PR is part of a larger effort to reduce the number of repositories we keep testing inputs in, to increase the documentation associated with our test inputs, as well as improve our ability to quickly and easily create new versions of our rdg datasets. 

From the open katana perspective, the largest effect of these changes is that "build/inputs" is no longer used, and instead https://github.com/KatanaGraph/test-datasets is a submodule, located at `<katana-repo>/external/test-datasets`

This PR and the enterprise PR can be used to provide feedback on https://github.com/KatanaGraph/test-datasets

The commits in this PR are broken up by what part of the code base they touch (python or cmake), separating out the tooling changes from widespread find&replace changes

"build/inputs" was divided into
`test-datasets/rdg_datasets` for all rdg test inputs
`test-datasets/csv_datasets` for all csv test inputs (none actually came from "build/inputs, but some were brought in from other repos)
`test-datasets/misc_datasets` for all datasets that don't nicely fall into one of the above categories

Additional information can be found here https://github.com/KatanaGraph/test-datasets/blob/master/README.md

Any RDGs, or other test inputs, that are too large to live in the git repository should still be stored on aws/gcp as usual. None of the existing files in "build/inputs" fall into this category. 

JIRA: KAT-2776, KAT-2777


cmake/Modules/TestDatasets.cmake
```diff
@@ -0,0 +1,39 @@
+## csv, rdg test inputs
+set(TEST_DATASETS "${PROJECT_SOURCE_DIR}/external/test-datasets")
+set(RDG_TEST_DATASETS "${PROJECT_SOURCE_DIR}/external/test-datasets/rdg_datasets")
+set(CSV_TEST_DATASETS "${PROJECT_SOURCE_DIR}/external/test-datasets/csv_datasets")
+set(MISC_TEST_DATASETS "${PROJECT_SOURCE_DIR}/external/test-datasets/misc_datasets")
+
+## latest supported rdg storage_format_version
+#TODO(emcginnis) get this envar in RDGPartHeader.h instead of having to hard code it here and there
```
arthurp, line 8: If this value isn't needed until test time (which I assume is the case), you could just have a tiny C++ program that prints the value you need that is built, but not installed.

python/katana/example_data.py
```diff
@@ -3,67 +3,127 @@
...
-        if not invalidate:
-            return inputs_dir
-        try:
-            shutil.rmtree(inputs_dir)
-        except OSError:
-            inputs_dir.unlink()
-    inputs_dir.mkdir(parents=True, exist_ok=True)
-    fn, _headers = urllib.request.urlretrieve(
-        "https://katana-ci-public.s3.us-east-1.amazonaws.com/inputs/katana-inputs-v30.tar.gz"
-    )
```
arthurp, line 42: I think you missed a critical feature of this code: It will download the inputs if the user doesn't have a build directory. This allowed users to get these dataset easily from within a notebook since the code would automatically download the data for them when they required it.

It should be possible do download the data if it's not available. Github provides zip files of repo content. In this case: https://github.com/KatanaGraph/test-datasets/archive/refs/heads/master.zip.

See main comment for a bit more discussion of this.

arthurp (changes requested):
The inputs blob was a problem and I like this change.

There is a complication though. The inputs were serving dual purpose as both:

1. Our testing inputs.
2. Example inputs for users.

I think these two use cases are not really compatible in the long term, so I think they should be separated. However, I think having easy to get example inputs for users is import to allow people to play with the system and to use in demo notebooks.

Below I propose automatically downloading the repo content if it's not available. I think that's the right path at the moment. However, I would love to split out the testing and examples uses into two separate parts of the system in the future.

python/katana/example_data.py
e-mcginnis: Is this actually an issue since the git repo is now part of the source, rather than the build?

I'm not sure I understand why we would want to support someone cloning the katana repo without the test-datasets submodule?

(That said, I'm now sure I have to update the docs/README for downloading katana)

yshalabi (changes requested):
1. It might be that we want katana-enterprise only RDGs. That was partly why I created that repo.
2. Why is it not a better solution is to require that all graphs that are not intended to test RDG storage engine to materialize their graphs at test time? I am moving all the graphs I created to take precisely this approach to reduce what needs to be maintained.  If we adopt this standard, the question of RDG format is no longer a concern for anybody other than storage-engine developers.  This is especially easy now that we support creating graphs from scratch (i.e. from empty graph). 

I think we should aim to ensure that the problem of maintaining RDG input formats is only a concern for devs working on the storage format. But I don't understand why we need anything more than a convenience script for storage-engine developers to generate new property graphs for inputs.tgz as the storage formats are updated. 

Maybe this is more appropriate for an RFC discussion, but I think it would useful to answer these questions before committing to this approach.

e-mcginnis:
@yshalabi please see the RFC https://docs.google.com/document/d/1YeSHwAQ2yplmRdDBfqli4spShxSBdwqeQMb_bb_eLEY/edit#
and the RFC that spawned that RFC: https://docs.google.com/document/d/1OHYP-In0yrdrkTjvxQS_tEvV1R-uYD0fO5MljnfuLfM/edit#

> It might be that we want katana-enterprise only RDGs. That was partly why I created that repo.

I agree there needs to be discussion around how to best manage enterprise only rdgs. I can see two easy options:
1) we store them on gcp/aws along with our other enterprise only RDGS
2) we create a "test-datasets-enterprise" repo which wraps around the "test-datasets" repo

> Why is it not a better solution is to require that all graphs that are not intended to test RDG storage engine to materialize their graphs at test time? 

That can definitely be done, however the test RDGs are supposed to be a source of truth. If we use the changed code to generate the test RDGs, its very possible that we break our source of truth in a way that the test can't easily catch. 
In general though, I am for folks removing their RDGs in favor of generating them on the fly. 

All that said, we still need to track a number of RDGs and other datasets. So the issue of how best to manage them doesn't go away. 

> I think we should aim to ensure that the problem of maintaining RDG input formats is only a concern for devs working on the storage format

This is discussed in the second RFC I linked above. Anyone that handles RDGs has to be aware that they have a storage_format_version.
This PR doesn't make maintaining the RDG input formats any more of a concern for other devs than it already is. If anything, this PR makes understanding the storage_format_version of RDGs more transparent.

> inputs.tgz

the downsides of our inputs blob could be discussed at length but mainly
1) there is no documentation on what the files in the input blobs are
2) there is no process to regenerating the rdgs in the input blobs
3) there isn't even git history to see the who/what/why for changes to the input blobs. Best you can do is find the commit that reved the inputs blob version

yshalabi:
> @yshalabi please see the RFC https://docs.google.com/document/d/1YeSHwAQ2yplmRdDBfqli4spShxSBdwqeQMb_bb_eLEY/edit# and the RFC that spawned that RFC: https://docs.google.com/document/d/1OHYP-In0yrdrkTjvxQS_tEvV1R-uYD0fO5MljnfuLfM/edit#
> 
Thanks. Would have been nice to broadcast the RFC on storage-core or on the rfcs channel. For a PR like this, I would expect that we would want to achieve buy-in and consensus before merging the PR.

> > It might be that we want katana-enterprise only RDGs. That was partly why I created that repo.
> 
> I agree there needs to be discussion around how to best manage enterprise only rdgs. I can see two easy options:
> 
> 1. we store them on gcp/aws along with our other enterprise only RDGS
> 2. we create a "test-datasets-enterprise" repo which wraps around the "test-datasets" repo
> 

Option one is a non-starter when it comes for CI and testing. If we have issues with non-version controlled blobs in inputs.tgz we should have the even more issues with RDGs stored in cloud storage.

> > Why is it not a better solution is to require that all graphs that are not intended to test RDG storage engine to materialize their graphs at test time?
> 
> That can definitely be done, however the test RDGs are supposed to be a source of truth. If we use the changed code to generate the test RDGs, its very possible that we break our source of truth in a way that the test can't easily catch. In general though, I am for folks removing their RDGs in favor of generating them on the fly.
> 

I am not sure I am following. Imagine a test to verify we can count graphs with 10 nodes. Options:
1. the test author can create an RDG with 10 nodes and store it in a repo along with some code to regenerate that RDG in the event of a new RDG format
2. the test author can have a fixture which creates a graph with 10 nodes and a test which asserts that the count matches 10

Aside from the fact that the fixture is version-controlled and subjected to code review, if someone changes it to generate more/less nodes the test fails. 

> All that said, we still need to track a number of RDGs and other datasets. So the issue of how best to manage them doesn't go away.
> 

Yes, this is a pain point and it might not go away but it would be dramatically reduced. I would expect that the only RDGs left in CI at that point would be ones used by the teams working code related to in-memory/on-disk graphs.

> > I think we should aim to ensure that the problem of maintaining RDG input formats is only a concern for devs working on the storage format
> 
> This is discussed in the second RFC I linked above. Anyone that handles RDGs has to be aware that they have a storage_format_version. This PR doesn't make maintaining the RDG input formats any more of a concern for other devs than it already is. If anything, this PR makes understanding the storage_format_version of RDGs more transparent.
> 

Well, maybe I see it differently but I disagree that "anyone that handles RDGs has to be aware that they have a storage_format_version". This is an implementation detail that end-users ideally do not need to understand. We should also aspire that the same hold for non-storage katana developers.

Just so my opinion is clear: I think it is unnecessary for non-storage devs to have to write a migration script for test RDGs when we support importing CSVs into RDGs and we support creating RDGs from scratch. It seems cleaner to materialize graphs for non-storage tests at test time.

> > inputs.tgz
> 
> the downsides of our inputs blob could be discussed at length but mainly
> 
> 1. there is no documentation on what the files in the input blobs are

Yeah, but that is tangential. A smaller PR can solve this.

> 2. there is no process to regenerating the rdgs in the input blobs

Yeah, right now it is ad-hoc. Is the re-org and process you are proposing the only way to achieve this?

> 3. there isn't even git history to see the who/what/why for changes to the input blobs. Best you can do is find the commit that reved the inputs blob version

Again, tangential. Why is a "history.txt" in the inputs directory an unacceptable solution?

e-mcginnis:
> Thanks. Would have been nice to broadcast the RFC on storage-core or on the rfcs channel. For a PR like this, I would expect that we would want to achieve buy-in and consensus before merging the PR.

@yshalabi both of the RFCs were posted in #kgfs which is the main storage channel. 



> Just so my opinion is clear: I think it is unnecessary for non-storage devs to have to write a migration script for test RDGs when we support importing CSVs into RDGs and we support creating RDGs from scratch. It seems cleaner to materialize graphs for non-storage tests at test time.

I don't believe anyone is requiring that, I certainly am not. The RFC lays out that importing from csv is much much preferred. 
In addition the `test-datasets` repo provides tooling so writing an import script is as easy as specifying what csv to import from.
here is one: https://github.com/KatanaGraph/test-datasets/blob/master/rdg_datasets/bad_smiles_graph/import.py

Migration scripts are actually even simpler:
https://github.com/KatanaGraph/test-datasets/blob/master/rdg_datasets/two_triangles/migrate.py


Much of the rest of this discussion we can take to the RFCs, as it covers some of your points in much more detail than I will get into here.

witchel (commented):
This looks good to me.  What a mess!  So much ctest, notebooks, and python.

witchel:
> > 1. we store them on gcp/aws along with our other enterprise only RDGS
> > 2. we create a "test-datasets-enterprise" repo which wraps around the "test-datasets" repo
> 
> Option one is a non-starter when it comes for CI and testing. If we have issues with non-version controlled blobs in inputs.tgz we should have the even more issues with RDGs stored in cloud storage

Just to comment, I maintain many non-CI tests (ctests without the quick label) that read from s3 (and now gs).  I have had to update them a bit over the months, but it is quite manageable.

e-mcginnis:
@witchel 
> What a mess! So much ctest, notebooks, and python.

I know :/ I spent a while trying to figure out how to split this into multiple PRs but that ended up being especially difficult with so many moving parts. I hope my commit division is helpful at the very least.

python/katana/example_data.py
pyrito: I don't think it should be a problem since it should be a part of the source module now. 

On a tangent, I think the advantage of having the downloadable inputs was that you could always download the latest version of the dataset (convenient for users) whereas now you have to update a submodule, which can be a bit of a pain to some. My hope is that as our RDG format changes stabilize this will become less of an issue.

pyrito (commented):
Thanks for doing all this, @e-mcginnis ! From what I see in the RFC and implementation, I see a few main benefits of your approach:

1. Test inputs now have some sort of history, so we can at least follow along with what changed, what was added, etc. Git history is powerful and will definitely help me when I'm trying to test my code.
2. We now have some structure/formality in generating new datasets with the import, generate, and migrate scripts. 

However, I think @arthurp does bring up a good point that downloading the inputs is convenient for users to work with, but I think having the repo should serve that purpose just as well. The main difference I see here is that if you always download the testing/example inputs, then all you have to do is update the ground truth files on S3/GCS and everyone will automatically download the latest dataset when needed. With a submodule, you'd have to update where the `test-datasets` repo was pointing to. I'm hoping we reach a point of stability to where this won't be as big of an issue.

@yshalabi also brought up a good point of private vs public datasets. I think for the most part we will be working with smaller datasets that shouldn't expose any sensitive information. However, it's possible that we will show our hand on things that we may be potentially working on, but this could play to our benefit as opposed to being a disadvantage. I think there might be some confusion - I think we very much do not want users to have to deal with different RDG versions and those nitty gritty details. I think a lot of the migration, version changing, etc is for our internal use. I think @e-mcginnis basically said this.

python/katana/example_data.py
e-mcginnis: Talked to @arthurp out of band who said:

> The issue is that the datasets are not just for testing they are also for end user experimentation.
> The user will not have the source tree in hand.

so I will be addressing this

e-mcginnis: alright, I have addressed this issue. It makes me sad that we must hardcode a sha to download, but the other option is to just have the code always pull test-datasets master, which could cause breakages in notebooks/users code. 
Theoretically, we could be smart and store the sha of test-datasets in use at build time, but I will leave that as a TODO.

tylershunt (approved):
I'm happy with this; I think if/when we want/need enterprise graphs that don't come from S3 we could add another enterprise only submodule that follows this pattern.

python/katana/example_data.py
```diff
@@ -4,66 +4,181 @@
...
+                tmp_path = path / "external" / "test-datasets"
+                if validate_path(tmp_path):
+                    found_path = tmp_path
+
+        return found_path
+
+    def download_datasets(invalidate_cache=False) -> Path:
+        # TODO(emcginnis) it would be really really nice if this got updated automatically
+        # when the submodule ref held by open katana is updated
+        datasets_sha = "14d993df871ddda26964958e271a8bc04d012d40"
```
arthurp: This should be a top level constant, and it should probably check the hash when using the data from the source tree, that would help catch failures to update this, or typos.

We don't yet have a generated Python file to hold build time configuration, but I think we should. This isn't the only place it would be helpful. With such a file, this hardcoded hash would no longer be needed.

e-mcginnis: Good idea, implemented a git sha check on the local test-datasets repo.

arthurp (approved):

pyrito (approved):

yshalabi (commented):
removing requested changes

yshalabi (approved):

e-mcginnis:
repeatedly failing CI with 
```
 Err:24 https://apt.kitware.com/ubuntu focal/main amd64 cmake-data all 3.22.1-0kitware1ubuntu20.04.1
  Could not connect to apt.kitware.com:443 (66.194.253.25), connection timed out
E: Failed to fetch https://apt.kitware.com/ubuntu/pool/main/c/cmake/cmake-data_3.22.1-0kitware1ubuntu20.04.1_all.deb  Could not connect to apt.kitware.com:443 (66.194.253.25), connection timed out
Err:25 https://apt.kitware.com/ubuntu focal/main amd64 cmake amd64 3.22.1-0kitware1ubuntu20.04.1
E: Failed to fetch https://apt.kitware.com/ubuntu/pool/main/c/cmake/cmake_3.22.1-0kitware1ubuntu20.04.1_amd64.deb  Unable to connect to apt.kitware.com:https:
  Unable to connect to apt.kitware.com:https:
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```
will try again in the morning

PR #753, opened 2021-12-14
Include all requisite packages

witchel:
I'm sorry about that!

### DIFF
--- a/libsupport/include/katana/Cache.h
+++ b/libsupport/include/katana/Cache.h
@@ -12,8 +12,12 @@
 // execute insert code with the parallel-hashmap write lock held, it seemed like there
 // would be some form of race condition.
 
+#include <cstdint>
+#include <functional>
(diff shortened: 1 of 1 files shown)


scober (approved):
LGTM. Although I don't want to manually scan this whole file, so I am not sure that we have reached full "include what you use" status. It does seem like there are tools that will do this so maybe we should add this to CI?

witchel:
> LGTM. Although I don't want to manually scan this whole file, so I am not sure that we have reached full "include what you use" status. It does seem like there are tools that will do this so maybe we should add this to CI?

It seems hard for individual programmers to be fastidious about this.  I didn't realize `uint32_t `required `<cstdint>`.

ddn0 (approved):
technically `<memory>` (`shared_ptr`) and `<iterator>` (`distance`) too

PR #756, opened 2021-12-15
support: add environment controllable feature flags

tylershunt:
RFC via PR! This is a mechanism that we have explicitly built into some
of the work for the next release, so I decided to come with a fully
concrete proposal.

Here we provide the macro "KATANA_EXPERIMENTAL_FEATURE(FeatureName)"
where FeatureName is some feature-specific string. It defines a function
(Feature_FeatureName_IsEnabled()) that allows other code to check when
the feature flag is set.

Those wishing to use the experimental feature simply put "FeatureName"
into a comma delimited list provided via the environment variable
"KATANA_ENABLE_EXPERIMENTAL". If "FeatureName" is part of that string
the above function will return true.

The mechanism is very simple on purpose; rather than allow flags that
are "True" by default I thought it was a good idea to encourage
developers to eliminate feature flags once the features they guard are
stable or turn feature flags into actual user facing configuration
options if the ability to toggle a flag is actually useful.

The main thing I like about this design:
 * flags can be defined close to to features they guard
 * related to the above: adding/changing a flag does not require a
   cross-repo commit
 * flags can be as private or as public as they need to be

This comes with the drawback that technically flags can conflict (though they have
to be in different libraries to compile),
meaning a developer could accidentally enable some far flung feature
when they only meant to test their own. I don't think that's a huge
concern as the macro is easy to grep for.
I also added an explicit WARN message that prints when this is detected.

### DIFF
--- a/libgraph/src/SharedMemSys.cpp
+++ b/libgraph/src/SharedMemSys.cpp
@@ -20,10 +20,12 @@
 #include "katana/SharedMemSys.h"
 
 #include "katana/CommBackend.h"
+#include "katana/Experimental.h"
 #include "katana/Galois.h"
 #include "katana/GaloisRuntime.h"
 #include "katana/Logging.h"
 #include "katana/Plugin.h"
+#include "katana/Strings.h"
 #include "katana/TextTracer.h"
 #include "tsuba/FileStorage.h"
 #include "tsuba/tsuba.h"
@@ -44,6 +46,23 @@ katana::SharedMemSys::SharedMemSys(std::unique_ptr<ProgressTracer> tracer)
   if (auto init_good = tsuba::Init(&comm_backend); !init_good) {
     KATANA_LOG_FATAL("tsuba::Init: {}", init_good.error());
   }
+
+  auto features_on = katana::internal::ExperimentalFeature::ReportEnabled();
+  if (!features_on.empty()) {
+    auto feature_string = katana::Join(features_on, ",");
+    tracer->GetActiveSpan().SetTags(
+        {{"experimental_features_enabled", feature_string}});
+  }
+
+  auto unrecognized =
+      katana::internal::ExperimentalFeature::ReportUnrecognized();
+  if (!unrecognized.empty()) {
+    KATANA_LOG_WARN(
+        "these values from KATANA_ENABLE_EXPERIMENTAL did not match any "
+        "features:\n\t{}",
+        katana::Join(unrecognized, " "));
+  }
+
   katana::ProgressTracer::Set(std::move(tracer));
 }
 
--- a/libsupport/include/katana/Experimental.h
+++ b/libsupport/include/katana/Experimental.h
@@ -1,0 +1,122 @@
+#ifndef KATANA_LIBSUPPORT_KATANA_EXPERIMENTAL_H_
+#define KATANA_LIBSUPPORT_KATANA_EXPERIMENTAL_H_
+
+#include <memory>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "katana/config.h"
+
+namespace katana {
+namespace internal {
+
+/// ExperimentalFeature tracks the state of feature flags set in the environment;
+/// It is not intended to be used directly; please see the macro KATANA_EXPERIMENTAL_FEATURE
+/// below
+class KATANA_EXPORT ExperimentalFeature {
+public:
+  static ExperimentalFeature* Register(
+      const std::string& feature_name, const std::string& filename,
+      int line_number);
+
+  /// report the feature flags that were checked on codepaths that were executed and
+  /// the flag was set to true
+  static std::vector<std::string> ReportEnabled();
+
+  /// report the feature flags that were used but stayed false
+  static std::vector<std::string> ReportDisabled();
+
+  /// report the feature flags that were provided but did not match any registered flag
+  static std::vector<std::string> ReportUnrecognized();
+
+  bool IsEnabled() { return is_enabled_; }
+
+  ExperimentalFeature(const ExperimentalFeature& no_copy) = delete;
+  ExperimentalFeature& operator=(const ExperimentalFeature& no_copy) = delete;
+  ExperimentalFeature(ExperimentalFeature&& no_move) = delete;
+  ExperimentalFeature& operator=(ExperimentalFeature&& no_move) = delete;
+  ~ExperimentalFeature() = default;
+
+  const std::string& name() const { return name_; }
+  const std::string& filename() const { return filename_; }
+  int line_number() const { return line_number_; }
+
+private:
+  ExperimentalFeature(std::string name, std::string filename, int line_number)
+      : name_(std::move(name)),
+        filename_(std::move(filename)),
+        line_number_(line_number) {
+    CheckEnv();
+  }
+
+  void CheckEnv();
+
+  std::string name_;
+  std::string filename_;
+  int line_number_{};
+  bool is_enabled_{};
+
+  static std::mutex registered_features_mutex_;
+  static std::unordered_map<std::string, std::unique_ptr<ExperimentalFeature>>
+      registered_features_;
+};
+
+}  // namespace internal
+}  // namespace katana
+
+/// KATANA_EXPERIMENTAL_FEATURE creates a flag that can be set from the environment.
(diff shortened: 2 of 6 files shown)


tylershunt:
Edit: I've updated the implementation to use the macro style checks (latter option below).

One thing I wasn't sure about is the hidden function declaration. I just picked this as a starting point so I'm not super attached to it, if folks would rather have another macro to check the state of the flag I could do that too. If we did that instead of:
```
KATANA_EXPERIMENTAL_FEATURE(UnstableButUseful);
if (Feature_UnstableButUseful_IsEnabled()) {
   ...
}
```
we could have:
```
KATANA_EXPERIMENTAL_FEATURE(UnstableButUseful);
if (KATANA_EXPERIMENTAL_ENABLED(UnstableButUseful)) {
   ...
}
```

roshandathathri:
This is a great idea. Of the two options, I'm in favor of:
```
KATANA_EXPERIMENTAL_FEATURE(UnstableButUseful);
if (KATANA_EXPERIMENTAL_ENABLED(UnstableButUseful)) {
   ...
}
```

libgraph/src/SharedMemSys.cpp
```diff
@@ -45,6 +47,20 @@ katana::SharedMemSys::SharedMemSys(std::unique_ptr<ProgressTracer> tracer)
     KATANA_LOG_FATAL("tsuba::Init: {}", init_good.error());
   }
   katana::ProgressTracer::Set(std::move(tracer));
+
+  auto features_on = ExperimentalFeature::ReportEnabled();
+  if (!features_on.empty()) {
+    KATANA_LOG_VERBOSE(
+        "running with experimental features:\n\t{}",
```
roshandathathri: If in the future, this could be enabled in a deployment, then we should have a trace output for this.

libsupport/include/katana/Experimental.h
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Example:
+///
+/// Program env:
+///   KATANA_ENABLE_EXPERIMENTAL="UnstableButUseful"
+///
+/// active_development.cpp:
+///    KATANA_EXPERIMENTAL_FEATURE(UnstableButUseful);
+///
+///    void important_function() {
```
roshandathathri, line 91: We're going with the assumption that classes/types, function declarations, or other compile-time changes are not going to depend on the experimental features. That's fine with me. It's good to state that up-front.

libsupport/include/katana/Experimental.h
```diff
@@ -0,0 +1,96 @@
...
+///      if (Feature_UnstableButUseful_IsEnabled()) {
+///        // do something useful
+///      } else {
+///        // be conservactive
+///      }
+///    }
+///
+/// Flags declared in different parts of the codebase can conflict. A good
+/// practice is to choose good name for your feature and grep for this marco
+/// to be sure it does not collide with another.
```
roshandathathri, line 101: Can we throw a warning or an error in this case?

roshandathathri (approved):
LGTM

libsupport/include/katana/Experimental.h
tylershunt: hmmmmmm; yeah now that you mention it I think I can get this.

libsupport/include/katana/Experimental.h
tylershunt: Will add that. If we want compile time experimental features like that I think it should be a separate mechanism.

tylershunt:
> This is a great idea. Of the two options, I'm in favor of:
> 
> ```
> KATANA_EXPERIMENTAL_FEATURE(UnstableButUseful);
> if (KATANA_EXPERIMENTAL_ENABLED(UnstableButUseful)) {
>    ...
> }
> ```

Yeah the more I think about it the more I find myself in that camp as well; I'll switch it to the macro style.

libsupport/include/katana/Experimental.h
roshandathathri: Agreed. We should try to stick to environmental variable approach as much as possible because that allows us to try these features on customer deployments.

e-mcginnis (approved):
Looks great to me! Looking forward to using this. Thanks Tyler!

libsupport/include/katana/Experimental.h
```diff
@@ -0,0 +1,122 @@
...
+  static std::unordered_map<std::string, std::unique_ptr<ExperimentalFeature>>
+      registered_features_;
+};
+
+}  // namespace internal
+}  // namespace katana
+
+/// KATANA_EXPERIMENTAL_FEATURE creates a flag that can be set from the environment.
+/// The macro takes a feature_name which should be an unquoted, unique string that
+/// looks like a function name. Developers can then use the macro
```
witchel, line 71: I guess "looks like a function name" also means no embedded commas in the name.

libsupport/include/katana/Experimental.h
```diff
@@ -0,0 +1,122 @@
...
+  class ExperimentalFeature;                                                   \
+  static auto* katana_experimental_feature_ptr_##feature_name =                \
+      ::katana::internal::ExperimentalFeature::Register(                       \
+          #feature_name, __FILE__, __LINE__);                                  \
+  }                                                                            \
+  static_assert(                                                               \
+      std::is_same<                                                            \
+          ::katana::internal::ExperimentalFeature,                             \
+          katana::internal::ExperimentalFeature>::value,                       \
+      "KATANA_EXPERIMENTAL_FEATURE must not be inside a namespace block")
```
witchel, line 116: Static asserts always blow my mind.

Not even within the katana namespace?

libsupport/src/Experimental.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+namespace {
+
+class ExperimentalFeatureEnvState {
+public:
+  static ExperimentalFeatureEnvState* Get() {
+    std::call_once(init_flag_, [&]() {
+      state_ = std::unique_ptr<ExperimentalFeatureEnvState>(
+          new ExperimentalFeatureEnvState());
+    });
```
witchel, line 18: Whoa, is this the new double-checked locking idiom?  Oh heck, Wikipedia says it is.  Cool.

libsupport/src/Experimental.cpp
```diff
@@ -0,0 +1,132 @@
...
+    return state_.get();
+  }
+
+  bool WasInEnv(const std::string& feature) {
```
witchel, line 22: The past tense here is throwing me off.  Isn't this `WasUsed`?

witchel (approved):
Nice feature!

libsupport/include/katana/Experimental.h
tylershunt: Thats true; I'm also embedding the string provided into a variable so the compiler will yell at you if you try to do things like that.

libsupport/include/katana/Experimental.h
tylershunt: The check macro needs to know what the namespace is for the variable the declare macro creates. It's always put in `katana::internal`. If you used this inside a katana namespace block the variables full qualified name would be `katana::katana::internal::...`

libsupport/src/Experimental.cpp
tylershunt: Big fan of `std::call_once`

libsupport/src/Experimental.cpp
tylershunt: Here I'm talking about the snapshot that I took of the environment. I thought `IsInEnv` would imply that I am checking the environment every time.

Elsewhere I'm using `Used` to talk about features listed in the environment that map to features that were declared. `WasInEnv` was meant to be the union of `Used` and `Unrecognized`.

libsupport/include/katana/Experimental.h
```diff
@@ -0,0 +1,122 @@
...
+///    void important_function() {
+///      if (KATANA_EXPERIMENTAL_ENABLED(UnstableButUseful)) {
+///        // do something useful
+///      } else {
+///        // be conservative
+///      }
+///    }
+///
+/// Flags declared in different parts of the codebase can conflict. A good
+/// practice is to choose good name for your feature and grep for this marco
```
ddn0: marco => macro

marco polo! marco polo!

ddn0 (approved):

PR #760, opened 2021-12-16
python: Added URL class

pyrito:
This PR adds a URL class that maintains `pathlib`-like semantics to manage paths and such.

### DIFF
--- a/python/katana/url.py
+++ b/python/katana/url.py
@@ -8,15 +8,11 @@ class URL:
     the class in the following ways:
 
     - URL("file:///home/") / "a" / "b" => URL("file:///home/a/b")
-    - URL("file:///home/", "a", "b") => URL("file:///home/a/b")
     - "file:///home/" / URL("a") => URL("file:///home/a")
     """
 
-    def __init__(self, *args):
-        if len(args) > 1:
-            self.url = self._join_path(args[0], *args[1:])
-        else:
-            self.url = args[0]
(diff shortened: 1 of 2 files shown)


python/katana/url.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class URL:
+    """
+    URL class that maintains semantics similar to pathlib. This class does not
+    intend to act as a universal class for all path related things! Users can use
+    the class in the following ways:
+
+    - URL("file:///home/") / "a" / "b" => URL("file:///home/a/b")
+    - URL("file:///home/", "a", "b") => URL("file:///home/a/b")
```
arthurp: I'm still not a fan of this second constructor. Why do you want to have it?

arthurp (approved):
Looks good over all. Just make sure you get the tests in here.

python/katana/url.py
pyrito: I was mimicking `pathlib` since they have something similar with how `Path` is constructed.

arthurp: Oh do they? WTF. The python standard library has some really odd stuff in it. My instinct is that good API is more important than exactly paralleling lesser known features of another library. But it's up to you.

pyrito: That's a good point. Less is more

ddn0 (approved):

PR #761, opened 2021-12-17
[KAT-1688] Detect read/write set for TxnContext (3/3) from RemoveNode/EdgeProperty

yige-hu:
Jira issue: [KAT-1688]
PR in enterprise repo: [#2791](https://github.com/KatanaGraph/katana-enterprise/pull/2791)

This is the 3/3 part of changes related to read/write set detection. It handles read/write set starting from `RDGCore:: RemoveNodeProperty` and `RDGCore:: AddEdgeProperty` (the previous two parts of porting was on upsert and add).

Affected API changes allow functions to pass on the TxnContext, which contains the transaction's read/write set. The information will be later used for conflict detection during a transaction commit.

I didn't pass the read/write set information for `TemporaryPropertyGuard`, since my understanding is that the properties it creates are temporary, and named in a per thread, non-repeating way. Correct me if wrong, @arthurp .

The changes in `PropertyGraph::Make` is mainly caused by old to new graph format conversion. I don't think it's necessary if there's no such a conversion. But I still left the read/write set passing for the `PropertyGraph` constructions. Or shouldn't I?


libgraph/include/katana/analytics/Utils.h
```diff
@@ -154,7 +160,9 @@ class KATANA_EXPORT TemporaryPropertyGuard {
 
   TemporaryPropertyGuard(
       PropertyGraph::MutablePropertyView pv, std::string name)
-      : property_view_(pv), name_(std::move(name)) {}
+      : property_view_(pv), name_(std::move(name)) {
+    txn_ctx_ = std::make_unique<tsuba::TxnContext>();
```
arthurp, line 159: Is this safe? You seem to be just making up a new context here. But if we are already in some user transaction won't this mess stuff up?

arthurp:
> I didn't pass the read/write set information for `TemporaryPropertyGuard`, since my understanding is that the properties it creates are temporary, and named in a per thread, non-repeating way. Correct me if wrong, @arthurp .

I comment asking about this exactly. I didn't read apparently. 

The names of temporary properties are unique and are deleted when when the object goes out of scope. I don't know if that makes a "fake" context safe. I don't really understand the planned semantics for TxnContext.

libgraph/include/katana/analytics/Utils.h
yige-hu: As long as they don't conflict with other transactions, it's fine to ignore the information. Since the names of temporary properties are unique, I don't think `TemporaryPropertyGuard`-s in different transactions will conflict with each other.

ddn0: Is there a difference between RemoveProperty on a view versus a property graph itself?

Does removing a property from a view just remove it from the view or does it mean also remove it from the underlying property graph?

And will TemporaryPropertyGuard go away when we start returning ephemeral graphs?

yige-hu: Good question. I don't have enough background knowledge to answer this. I think the same issue also applies to `AddProperties` and `UpsertProperties`?

The only thing I can tell is that `MutablePropertyView`'s `AddProperties`, `UpsertProperties` and `RemoveProperty` call into `RDGCore::AddProperties`, `RDGCore:: UpsertProperties` and `RDGCore:: UpsertProperties`.

yige-hu: I also have a question. Do we differentiate property names for different graphs/views (can property names for different graphs/views identical)? i.e. Do we also need to track RDG names while tracking property names in the `TxnContext`?

arthurp: I have assumed that adding a property to a view will add it to the underlying graph (though possibly in some optimized way that doesn't fully reify it immediately). I certainly hope this is the case, otherwise our views are not really views they are partial copies.

I don't know if ephemeral graphs would replace TemporaryPropertyGuard, but it does seem possible. We would need to be pretty careful about the "scope" of the ephemeral graphs though and we might need multiple "nested" ephemeral graphs: One for the users transaction and one for the algorithm to work in for example. The algorithm would then need to "commit" it's ephemeral graph back into the one for the transaction (creating a new version I guess), but not all they way back to the original graph. This seems complex enough that it might be better to use ARM explicitly to reduce the need for this. But I'm not at all wedded to that.

ddn0: Given that one of the constructors of TemporaryPropertyGuard takes an explicit property name, is it really safe to ignore all of its RemoveProperty calls? It is certainly possible that someone passes non-fresh property name and uses TemporaryPropertyGuard to remove it.

Anyway, I think this logic should be maintained in the transaction context level rather than picking and choosing among callers of transaction context.

arthurp: 1) The TemporaryPropertyGuard adds a unique number based on the thread ID and a counter to make the property unique.

2) I agree. Best to always track everything and allow the transaction machinery figure out what's safe. That way we have fewer assumptions scattered over the code base to trap us if we need to make changes to how transactions work.

ddn0: > The TemporaryPropertyGuard adds a unique number based on the thread ID and a counter to make the property unique.

This may be the intent, but we still have this constructor:
```
TemporaryPropertyGuard(PropertyGraph* pv, std::string name)
```

which is different than
```
explicit TemporaryPropertyGuard(PropertyGraph::MutablePropertyView pv)
        : TemporaryPropertyGuard(pv, GetPropertyName()) {}
```

witchel: What is the purpose of TemporaryPropertyGuard?  To add a temporary property that is then removed?  That does sound like a job better suited to transactions.  This issue needs to be resolved before the PR is approved.

As far as Yige's questions, my understanding is that different graphs can absolutely have the same property names and you should be tracking RDG prefix in addition to property names.  We do want transactions spanning multiple graphs.

witchel (commented):
Issues with TemporaryPropertyGuard linger.

ddn0 (approved):
Making TemporaryPropertyGuard more "transaction native" seems separate from the basic act of threading a transaction context.

Now that TemporaryPropertyGuard doesn't take an explicit property name, it is more clear that introduced properties by the guard are supposed to be unique, though I think we all agree this should really be checked at the tsuba-level as just because user-code thinks their names are unique doesn't mean they actually are.

My preference would be for TemporaryPropertyGuard to also take a transaction context and we can avoid any special cases in user-code/application-code (c.f. tsuba code).

If this looks ungainly, all the more reason for someone to go back and improve the interface of TemporaryPropertyGuard to be more transaction-idiomatic, but I think this is separate from the current work.

libgraph/include/katana/analytics/Utils.h
yige-hu: I've removed the redundant constructors of `TemporaryPropertyGuard`, so that now `name_` cannot be specified by users. I think this is the easiest way to fix this.

PR #762, opened 2021-12-17
entity manager: relax equality check

witchel:
Two entity type managers are "equal" if they manage the
same number of entities and the set of string representations of
these entities match.  This is a bit weaker than one might
expect.

This code is only used in testing.  I added to the documentation of the EntityTypeManager that the truth is the string representation.

@scober raises the following excellent question: Do we guarantee that two entity type ids never map to the same set of strings?  I think we need this property, but I don't grok the entity type manager and I don't know how we can provide this property or if it will be efficient.  Can anyone help?

### DIFF
--- a/libsupport/include/katana/EntityTypeManager.h
+++ b/libsupport/include/katana/EntityTypeManager.h
@@ -17,10 +17,27 @@
 
 namespace katana {
 
-/// EntityTypeID uniquely identifies an entity (node or edge) type
-/// EntityTypeID for nodes is distinct from EntityTypeID for edges
-/// This type may either be an atomic type or an intersection of atomic types
-/// EntityTypeID is represented using 8 bits
+/// EntityTypeID uniquely identifies an entity (node or edge) type, which is
+/// a set of user-defined strings.
+/// Types may either be an atomic type or an intersection of atomic types,
+/// where an atomic type is a unique user-defined string and
+/// an intersection of atomic types is a set of corresponding user-defined strings.
+/// The EntityTypeID space for nodes is distinct from EntityTypeID for edges.
+///
+/// Types are sets of strings; all other representations are optimizations.  Clients
+/// should never compare EntityTypeIDs across different managers.  Only comparison of
+/// the string representations are valid.
+///
+/// The type manager should guarantee that two entity type ids never map to the same
+/// set of strings, but the current implementation does not guarantee it for
+/// performance reasons.
+/// The GetOrAdd* functions only add a type if it isn't already present.
+/// There are two Add* functions: AddAtomicEntityType and AddNonAtomicEntityType.
+/// AddAtomicEntityType returns an error if the type is already present.
+/// AddNonAtomicEntityType has a debug assert but does not return an error if the type
+/// is already present because that check is too costly.
+///
+/// EntityTypeID is represented using 16 bits
 using EntityTypeID = uint16_t;
 static constexpr EntityTypeID kUnknownEntityType = EntityTypeID{0};
 static constexpr EntityTypeID kInvalidEntityType =
@@ -239,6 +256,8 @@ public:
   }
 
   /// Get the intersection of the types named in \p names
+  /// Note that AtomicEntityTypeIds map to their "NonAtomic" identity set
+  /// so this function will return EntityTypeIDs for atomic types.
   ///
   /// \returns the EntityTypeID of the intersection type.
   template <typename Container>
@@ -459,7 +478,7 @@ public:
   static size_t CalculateSetOfEntityTypeIDsSize(EntityTypeID max_id);
 
   /// bool Equals() IS A TESTING ONLY FUNCTION, DO NOT EXPOSE THIS TO THE USER
-  bool Equals(const EntityTypeManager& other) const;
+  bool IsIsomorphicTo(const EntityTypeManager& other) const;
 
   /// std::string ReportDiff() IS A TESTING ONLY FUNCTION, DO NOT EXPOSE THIS TO THE USER
   std::string ReportDiff(const EntityTypeManager& other) const;
--- a/libsupport/src/EntityTypeManager.cpp
+++ b/libsupport/src/EntityTypeManager.cpp
@@ -473,57 +473,36 @@ katana::EntityTypeManager::PrintEntityTypes() const {
   return std::string(buf.begin(), buf.end());
 }
 
+// This tests if two different type managers have the same entity types with the same
+// subtype relationship.  Such managers are isomorphic, but not necessarily equal
+// because they can assign different EntityTypeIDs to the same type.
 bool
-katana::EntityTypeManager::Equals(
+katana::EntityTypeManager::IsIsomorphicTo(
     const katana::EntityTypeManager& other) const {
-  if (entity_type_id_to_atomic_entity_type_ids_ !=
-      other.entity_type_id_to_atomic_entity_type_ids_) {
+  auto end = GetNumEntityTypes();
+  auto other_end = other.GetNumEntityTypes();
+  // If we have a different number of entity types, managers are not equivalent.
+  if (end != other_end) {
     KATANA_LOG_DEBUG(
-        "this.entity_type_id_to_atomic_entity_type_ids_.size() = {}, "
-        "other.size() = {}. SetOfEntityTypeIDsSize = {}, "
-        "other.SetOfEntityTypeIDsSize = {}",
-        entity_type_id_to_atomic_entity_type_ids_.size(),
-        other.entity_type_id_to_atomic_entity_type_ids_.size(),
-        SetOfEntityTypeIDsSize_, other.SetOfEntityTypeIDsSize_);
-
-    KATANA_LOG_DEBUG(
-        "this.entity_type_id_to_atomic_entity_type_ids_.at(0).size = {}, "
-        "other.size = "
-        "{}",
-        entity_type_id_to_atomic_entity_type_ids_.at(0).size(),
-        other.entity_type_id_to_atomic_entity_type_ids_.at(0).size());
+        "this.GetNumEntityTypes() = {}, "
+        "other.GetNumEntityTypes() == {}",
+        end, other_end);
     return false;
   }
-  if (atomic_entity_type_id_to_type_name_ !=
-      other.atomic_entity_type_id_to_type_name_) {
-    KATANA_LOG_DEBUG(
-        "this.atomic_entity_type_id_to_type_name_.size() = {}, other.size() = "
-        "{}",
-        atomic_entity_type_id_to_type_name_.size(),
-        other.atomic_entity_type_id_to_type_name_.size());
-    return false;
-  }
-
-  if (atomic_type_name_to_entity_type_id_ !=
-      other.atomic_type_name_to_entity_type_id_) {
-    KATANA_LOG_DEBUG(
-        "this.atomic_type_name_to_entity_type_id_.size() = {}, other.size() = "
-        "{}",
-        atomic_type_name_to_entity_type_id_.size(),
-        other.atomic_type_name_to_entity_type_id_.size());
-    return false;
-  }
-
-  if (atomic_entity_type_id_to_entity_type_ids_ !=
-      other.atomic_entity_type_id_to_entity_type_ids_) {
-    KATANA_LOG_DEBUG(
-        "this.atomic_entity_type_id_to_entity_type_ids_.size() = {}, "
-        "other.size() = {}. SetOfEntityTypeIDsSize = {}, "
-        "other.SetOfEntityTypeIDsSize = {}",
-        atomic_entity_type_id_to_entity_type_ids_.size(),
-        other.atomic_entity_type_id_to_entity_type_ids_.size(),
-        SetOfEntityTypeIDsSize_, other.SetOfEntityTypeIDsSize_);
-    return false;
+  // Every type name set in one manager must be in both
+  for (size_t i = 0; i < end; ++i) {
+    auto res = EntityTypeToTypeNameSet(i);
+    if (!res) {
+      KATANA_LOG_DEBUG(
+          "equality fails, no type name set for entity number {}\n", i);
+      return false;
+    }
+    auto tns = res.value();
+    auto res_type_id = other.GetNonAtomicEntityTypeFromStrings(tns);
+    if (!res_type_id) {
+      KATANA_LOG_DEBUG("this type name set ({}) not found in other", tns);
+      return false;
+    }
   }
   return true;
 }
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -494,11 +494,13 @@ katana::PropertyGraph::Equals(const PropertyGraph* other) const {
     return false;
   }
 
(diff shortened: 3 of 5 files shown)


libsupport/include/katana/EntityTypeManager.h
```diff
@@ -17,10 +17,16 @@
 
 namespace katana {
 
-/// EntityTypeID uniquely identifies an entity (node or edge) type
-/// EntityTypeID for nodes is distinct from EntityTypeID for edges
-/// This type may either be an atomic type or an intersection of atomic types
-/// EntityTypeID is represented using 8 bits
+/// EntityTypeID uniquely identifies an entity (node or edge) type, which is
+/// a Boolean property.
```
tylershunt: A type is a boolean property? This statement is confusing to me; why isn't `EntityTypeID` a bool then?

witchel: The term property is way overloaded, I should change that.  Any suggestions?
A type is a Boolean though, right?

tylershunt: I thought a type was a set of labels?

witchel: What is a label?

tylershunt: One of these strings in the sets we keep talking about.

witchel: Ok, see if you like what I have now.

tylershunt: Molto bene! tyty.

libsupport/src/EntityTypeManager.cpp
```diff
@@ -476,54 +476,24 @@ katana::EntityTypeManager::PrintEntityTypes() const {
 bool
 katana::EntityTypeManager::Equals(
     const katana::EntityTypeManager& other) const {
```
arthurp: If we are weakening the semantics of this to not include subtyping relations (which is the only part you are missing I think, since differences in numbering are not semantically meaningful), then I think we should rename this function. Maybe: `TestEquivalent` or something.

I think we should reserve `Equals` for real equality. Since people don't like using `operator==` for that.

arthurp: One thing that reinforces my thought is that I'm introducing tools to automatically wrap "conventional" methods on classes. I think `Equals` should be one of them.

e-mcginnis: +1 to renaming to TestEquivalent

witchel: Can you explain the subtyping point?
I think `City:Place` is a subtype of `City`, but is that not necessarily so?

arthurp: My point was just that the subtyping relationships need to be checked for anything that could be considered true equality. You are only checking the set of names if I read correctly.

libsupport/include/katana/EntityTypeManager.h
```diff
@@ -17,10 +17,16 @@
 
 namespace katana {
 
-/// EntityTypeID uniquely identifies an entity (node or edge) type
-/// EntityTypeID for nodes is distinct from EntityTypeID for edges
-/// This type may either be an atomic type or an intersection of atomic types
-/// EntityTypeID is represented using 8 bits
+/// EntityTypeID uniquely identifies an entity (node or edge) type, which is
+/// a set of strings (labels).
```
roshandathathri: Label and type can conflate with each other. Cypher uses "node labels" and "edge types". We call both of them as "atomic types". Instead of this, I'd prefer something like:

```cpp
/// EntityTypeID uniquely identifies an entity (node or edge) type, which is
/// a set of user-defined strings.
/// Types may either be an atomic type or an intersection of atomic types,
/// where an atomic type is a unique user-defined string and 
/// intersection of atomic types is a set of corresponding user-defined strings.
/// The EntityTypeID space for nodes is distinct from EntityTypeID for edges.
///
/// Types are sets of strings; all other representations are optimizations.  Clients
/// should never compare EntityTypeIDs across different managers.  Only comparison of
/// the string representations are valid.
///
/// EntityTypeID is represented using 16 bits
```

libsupport/src/EntityTypeManager.cpp
```diff
@@ -476,54 +476,24 @@ katana::EntityTypeManager::PrintEntityTypes() const {
...
-      other.atomic_entity_type_id_to_entity_type_ids_) {
-    KATANA_LOG_DEBUG(
-        "this.atomic_entity_type_id_to_entity_type_ids_.size() = {}, "
-        "other.size() = {}. SetOfEntityTypeIDsSize = {}, "
-        "other.SetOfEntityTypeIDsSize = {}",
-        atomic_entity_type_id_to_entity_type_ids_.size(),
-        other.atomic_entity_type_id_to_entity_type_ids_.size(),
-        SetOfEntityTypeIDsSize_, other.SetOfEntityTypeIDsSize_);
-    return false;
+  for (size_t i = 0; i < end; ++i) {
```
e-mcginnis, line 493: On the surface this doesn't look quite right. 

Shouldn't we also have to ensure that all names in `this.atomic_type_name_to_entity_type_id_.keys()` are present in `other.atomic_type_name_to_entity_type_id_.keys()`? 

From the name of `GetNonAtomicEntityTypeFromStrings` it really looks like we are only checking the managers NonAtomicEntityTypes here. 

Looking closer at `GetNonAtomicEntityTypeFromStrings` When `i` is an AtomicEntityTypeID, aka `tns` only has one name in it. 

`GetEntityTypeIDs` will function properly and give us the identity `SetOfEntityTypeIDs` for the AtomicEntityType name. 
which we pass to `GetNonAtomicEntityType`

the important part being
```
  auto existing_id = std::find(
      entity_type_id_to_atomic_entity_type_ids_.cbegin(),
      entity_type_id_to_atomic_entity_type_ids_.cend(), type_id_set_resized);
```

Since in `entity_type_id_to_atomic_entity_type_ids_` an AtomicEntityTypeID maps to its identity `SetOfEntityTypeIDs` this works. 


I think it would be helpful to add a comment here that we rely on AtomicEntityTypeIds mapping to its "NonAtomic" identity set so it is clear that all EntityTypes (Atomic & NonAtomic) are getting compared?

libsupport/src/EntityTypeManager.cpp
roshandathathri: +1 to `*Equivalent`

I didn't understand the subtyping point as well. This checks entity-type-IDs are the same (including the atomic types). We could relax it further to say that entity-types are the same but the IDs for those could be different but I don't think that's required.

libsupport/src/EntityTypeManager.cpp
```diff
@@ -476,54 +476,24 @@ katana::EntityTypeManager::PrintEntityTypes() const {
...
+    auto res = EntityTypeToTypeNameSet(i);
+    KATANA_LOG_ASSERT(res);
```
roshandathathri: Return an error?

roshandathathri (approved):
LGTM

roshandathathri:
> Do we guarantee that two entity type ids never map to the same set of strings? 

We should. AFAIK we do.

witchel:
> > Do we guarantee that two entity type ids never map to the same set of strings?
> 
> We should. AFAIK we do.

Do you know where?

libsupport/src/EntityTypeManager.cpp
witchel: > I think it would be helpful to add a comment here that we rely on AtomicEntityTypeIds mapping to its "NonAtomic" identity set so it is clear that all EntityTypes (Atomic & NonAtomic) are getting compared?

The code in GraphUpdate that uses EntityTypes also relies on this property.  I suspect the partitioner relies on it as well.  I added something along these lines to `GetNonAtomicEntityTypeFromStrings`, but I don't understand the Atomic/NonAtomic design well enough to go any further.

libsupport/src/EntityTypeManager.cpp
witchel: I renamed this function to TypeEquivalent and I added this comment.
Note, I don't understand what this comment means.
`// This relation is TypeEquivalent, not Equals because it does not test subtype relationships.`

roshandathathri:
> > > Do we guarantee that two entity type ids never map to the same set of strings?
> > 
> > 
> > We should. AFAIK we do.
> 
> Do you know where?

Was just looking over the code. There are many functions that are `GetOrAdd*`. They only add if it isn't already present. There are two `Add*` functions: `AddAtomicEntityType` and `AddNonAtomicEntityType`. The former returns an error if it is already present. The latter has a debug assert but does not return an error if it is already present and has an interesting comment from @arthurp:

```cpp
  // Ideally this would return an error instead of failing. But checking is
  // probably too slow. Remember kids, fast is more important than correct.
  // Exclude this check if our EntityTypeID == 0, as we expect it to have the empty set
```

So I was wrong. We should but we do NOT for performance reasons.

witchel:
> So I was wrong. We should but we do NOT for performance reasons.

Ok.  I will add a comment.

roshandathathri:
> > > > Do we guarantee that two entity type ids never map to the same set of strings?
> > > 
> > > 
> > > We should. AFAIK we do.
> > 
> > 
> > Do you know where?
> 
> Was just looking over the code. There are many functions that are `GetOrAdd*`. They only add if it isn't already present. There are two `Add*` functions: `AddAtomicEntityType` and `AddNonAtomicEntityType`. The former returns an error if it is already present. The latter has a debug assert but does not return an error if it is already present and has an interesting comment from @arthurp:
> 
> ```c++
>   // Ideally this would return an error instead of failing. But checking is
>   // probably too slow. Remember kids, fast is more important than correct.
>   // Exclude this check if our EntityTypeID == 0, as we expect it to have the empty set
> ```
> 
> So I was wrong. We should but we do NOT for performance reasons.

I searched our uses of `AddNonAtomicEntityType`. Other than unit tests, it is ONLY used in 3 places and all 3 places are member functions: 
1. `Init()`: This is to add to the unknown entity type, so it will be unique.
2. `GetOrAddNonAtomicEntityType()`: This will only add if it doesn't exist.
3. `DoAssignEntityTypeIDsFromProperties()`: This maintains a set of set of strings to ensure that it does not add the same set of strings again. 

Thus, there should be no way currently for two entity type IDs to map to the same set of strings. We should make `AddNonAtomicEntityType` private so that we can continue to enforce this (there's already a warning in the doxygen that explains that this should NOT be used unless the user guarantees that the set of strings do not already exist). I do NOT know enough about the tests but I think we can use `GetOrAddNonAtomicEntityType()` in those tests instead. @e-mcginnis and @scober, please check and confirm.

libsupport/src/EntityTypeManager.cpp
witchel: This had some unanticipated downstream, e.g., Equals being used in SanityCheckPartition.cpp and g-serialize.cpp, and v3-uint16-entity-type-ids.cpp.  I'm changing them to TypeEquivalent, but I'm not sure if the previous version of Equals was better suited for their use.  Of course, I can have both functions, but I was seeing spurious problems with the previous definition when I recreate an existing graph from the empty graph.

arthurp: > I renamed this function to TypeEquivalent and I added this comment.
> Note, I don't understand what this comment means.
> `// This relation is TypeEquivalent, not Equals because it does not test subtype relationships.`

OK. Let me try again:

An `EntityTypeManager` conceptually stores two kinds of information:

1. A set of types (concretely, names and IDs), and
2. The "inheritance heirarchy", or more formally the "subtype relation". This is a set of super types for each non-atomic (intersection) type. This is equivalent to having lists of subtypes, though that is less obvious from the implementation.

The function seem to check that the set of type names are the same, but does not check that the super type lists are the same. Though maybe I totally misunderstood.

roshandathathri: We do NOT store anything in the class for (2). It is derived from (1); that is (1) implies (2). So I don't see what check this function is missing. Can you please see if there is a (counter-)example?

arthurp: You definitely store The supertyping data: `entity_type_id_to_atomic_entity_type_ids_`.

As for your second point, you are right. You do actually check subtyping, but via strings. Sorry. My bad. My comment about subtyping was incorrect.

Sorry.

roshandathathri: Thanks for the clarification!

witchel: Thanks for the conversation. I assume then that y'all are ok with me leaving this method titled Equals.

arthurp: I'm on the fence. The problem is that the API does expose the IDs and this method explicitly *doesn't* check if IDs are the same. So if the client code only uses type names, then yes this is equality, but if the client code uses IDs, it's not. If there is any real way to get IDs from outside this class (PropertyGraph for instance) then the user could easily be using IDs directly making this equality invalid. For context: in OO the normal concept of equality is sustainability. So things are equal if you could substitute one for the other.

I think I would prefer this be called "Equivalent" (or even better "Isomorphic" ;-) ). That would also avoid my python wrappers automatically exposing this as `==` in Python which is probably not a great idea for the reasons above.

witchel: Fair enough. I can get behind that sentiment.

e-mcginnis:
> > > > > Do we guarantee that two entity type ids never map to the same set of strings?
> > > > 
> > > > 
> > > > We should. AFAIK we do.
> > > 
> > > 
> > > Do you know where?
> > 
> > 
> > Was just looking over the code. There are many functions that are `GetOrAdd*`. They only add if it isn't already present. There are two `Add*` functions: `AddAtomicEntityType` and `AddNonAtomicEntityType`. The former returns an error if it is already present. The latter has a debug assert but does not return an error if it is already present and has an interesting comment from @arthurp:
> > ```c++
> >   // Ideally this would return an error instead of failing. But checking is
> >   // probably too slow. Remember kids, fast is more important than correct.
> >   // Exclude this check if our EntityTypeID == 0, as we expect it to have the empty set
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I was wrong. We should but we do NOT for performance reasons.
> 
> I searched our uses of `AddNonAtomicEntityType`. Other than unit tests, it is ONLY used in 3 places and all 3 places are member functions:
> 
>     1. `Init()`: This is to add to the unknown entity type, so it will be unique.
> 
>     2. `GetOrAddNonAtomicEntityType()`: This will only add if it doesn't exist.
> 
>     3. `DoAssignEntityTypeIDsFromProperties()`: This maintains a set of set of strings to ensure that it does not add the same set of strings again.
> 
> 
> Thus, there should be no way currently for two entity type IDs to map to the same set of strings. We should make `AddNonAtomicEntityType` private so that we can continue to enforce this (there's already a warning in the doxygen that explains that this should NOT be used unless the user guarantees that the set of strings do not already exist). I do NOT know enough about the tests but I think we can use `GetOrAddNonAtomicEntityType()` in those tests instead. @e-mcginnis and @scober, please check and confirm.

I'm on board with this, PR incoming

roshandathathri:
Thanks Evan!

e-mcginnis:
PR: https://github.com/KatanaGraph/katana/pull/767

PR #767, opened 2021-12-20
make AddNonAtomicEntityType private use only

e-mcginnis:
AddNonAtomicEntityType does not ensure that the NonAtomicEntityType
is not already present, which could result into two different
EntityTypeIDs mapping to the same set of strings.
GetOrAddNonAtomicEntityType does ensure duplicates EntityTypeIDs are not
created, so enforce it is used over AddNonAtomicEntityType by making the
latter private.

see discussion on https://github.com/KatanaGraph/katana/pull/762 for more context

Enterprise PR https://github.com/KatanaGraph/katana-enterprise/pull/2809


roshandathathri (approved):
LGTM

witchel (approved):
Thank you Evan!

PR #771, opened 2021-12-21
libgraph: Adds fixed-binary POD array property and a view over the array

l-hoang:
Adds a new property type to Properties.h: FixedSizeBinaryPODArray.
This property is an array on every single node of some fixed size and is
represented with a FixedSizeBinary arrow array. Unlike the existing
ArrayProperty, it does not use arrow Lists (which add overhead to the
existing ArrayProperty since offsets are maintained even though the
ArrayProperty currently only uses fixed sized lists) and supports
arbitrary POD types. This commit also adds a view over
the arrow FixedSizeBinaryArray type assuming it contains some fixed
number of POD types.

The test this commit adds makes sure that this new view is sane.

-----

Meta-level comment:

The reason this has been added is to use it to create and persist FixedSizeBinary
arrays onto disc to use as the de-facto storage for arrays.

### DIFF
--- a/libgraph/include/katana/Properties.h
+++ b/libgraph/include/katana/Properties.h
@@ -622,8 +622,8 @@ private:
 /// of elements with trivial constructors (std::is_trivial) and standard layout
 /// (std::is_standard_layout). (i.e., POD)
 ///
+/// TODO(gurbinder) Currently, only double is supported. Support for other types need to be added.
 /// \tparam T the C type of array elements.
-/// Currently, only double is supported. Support for other types need to be added.
 /// \tparam N the size of the arrays in arrow::LargeLIstArray
 template <typename T, size_t N>
 class ArrayPropertyView {
@@ -665,8 +665,8 @@ private:
 /// with trivial constructors (std::is_trivial) and standard layout
 /// (std::is_standard_layout). (i.e., POD)
 ///
+/// TODO(gurbinder) Currently, only double is supported. Support for other types need to be added.
 /// \tparam T the C type of array elements.
-/// Currently, only double is supported. Support for other types need to be added.
 /// \tparam N the size of the array
 template <typename T, size_t N>
 struct ArrayProperty {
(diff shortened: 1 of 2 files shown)


libgraph/test/property-view-fixed-binary.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "tsuba/TxnContext.h"
+
+namespace {
+
+constexpr int kNumRows = 10;
+constexpr int kNumArrayEntries = 5;
+
+/// Generates a fixed sized binary array for checking with a view.
+katana::Result<std::shared_ptr<arrow::FixedSizeBinaryArray>>
+GenerateTestArray() {
```
scober: I don't know this part of the code super well, but isn't there a file for testing property views already? Does this need to be in a separate file?

l-hoang: I can place it in the older test file

tylershunt (approved):
I don't know this part of the code very well, but the code makes sense to me and looks low risk since it's purely additive. So let's go!

libgraph/include/katana/Properties.h
```diff
@@ -699,5 +699,126 @@ struct ArrayProperty {
...
+
+    arrow::FixedSizeBinaryBuilder fixed_sized_binary_builder(fixed_size_type);
+    KATANA_CHECKED(fixed_sized_binary_builder.AppendEmptyValues(num_rows));
+
+    std::shared_ptr<arrow::Array> array_of_fixed_size_binaries =
+        KATANA_CHECKED_CONTEXT(
+            fixed_sized_binary_builder.Finish(),
+            "failed to finish fixed size binary builder");
+
+    return katana::Result<std::shared_ptr<arrow::Table>>(arrow::Table::Make(
```
danielmawhirter: `return arrow::Table::Make(`
should work fine, there's no arrow::Result involved

libgraph/include/katana/Properties.h
```diff
@@ -699,5 +699,126 @@ struct ArrayProperty {
...
+        arrow::schema({arrow::field(name, fixed_size_type)}),
+        {array_of_fixed_size_binaries}));
```
danielmawhirter: std::move(array...

danielmawhirter (approved):
LGTM

PR #772, opened 2021-12-28
scripts: make ifdefs consistent

tylershunt:
Maybe a pedantic point but it's better to be consistent:

This change extends our ifndef checking script to check and fix
`#pragma once` style headers

### DIFF
--- a/libgalois/include/katana/AtomicHelpers.h
+++ b/libgalois/include/katana/AtomicHelpers.h
@@ -17,7 +17,8 @@
  * Documentation, or loss or inaccuracy of data of any kind.
  */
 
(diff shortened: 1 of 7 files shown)


srogatch:
What's wrong with `#pragma once`?

tylershunt:
> What's wrong with `#pragma once`?

Nothing, but we had decided to keep the convention of using ifndef guards (used in most places); and I think the worst outcome is to have a code base where they're mixed.

tylershunt:
Never mind; more trouble than I want to chew on right now. I'll go back to living with it 😄

PR #774, opened 2021-12-31
Pybind11 transition, step 1: Initial integration and test conversion [KAT-587, KAT-837]

arthurp:
This sets up the framework and converts the reduction objects to be wrapped by pybind11 instead of Cython. Future PRs will convert other parts of the API to Cython and expand the framework.

*Notably missing feature:* This PR does NOT introduce any ability to have compatibility between Cython and pybind11 wrappers. That is coming next as it will be required to provide a smooth transition.

[KAT-587, KAT-837]


arthurp:
I am not going to simplify the history of this PR (I will squash it of course) unless someone wants me to because changing the history will make the future PRs harder since I'm going to use git tricks to resurrect changes I excluded from this PR.

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,145 @@
...
+/// \tparam Args The argument types of the function.
+template <typename... Args>
+constexpr DefWithNumbaImpl<Args...> DefWithNumba{};
+
+/// Register a Python class for use from Numba compiled code.
+///
+/// This calls `katana.native_interfacing.numba_support.register_class`
+template <typename T>
+void
+RegisterNumbaClass(pybind11::class_<T>& cls) {
```
pyrito, line 220: What exactly does this function do?

python/katana/native_interfacing/wrappers.py
```diff
@@ -86,7 +87,7 @@ def typeof(val, c):
 
         return Type
 
-    def register_method(self, func_name, typ, cython_func_name=None, addr=None, dtype_arguments=None):
+    def register_method(self, func_name, typ, cython_func_name=None, addr=None, dtype_arguments=None, data=None):
```
pyrito: who would call this function?

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+struct ForReduceMax {
+  template <typename T>
+  using type = katana::GReduceMax<T>;
+};
+
+struct ForReduceLogicalOr {
+  template <typename T>
+  using type = katana::GReduceLogicalOr;
+};
```
pyrito, line 37: I understand the need for templating for the others, but no one is using `typename T` here, so why is it needed?

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,90 @@
...
+void
+katana::python::InitReductions(py::module_& m) {
+  katana::InstantiateForStandardTypes(
+      m, "ReduceSum", ReducibleFunctor<ForGAccumulator>());
+  katana::InstantiateForStandardTypes(
+      m, "ReduceMax", ReducibleFunctor<ForReduceMax>());
+  katana::InstantiateForStandardTypes(
+      m, "ReduceMin", ReducibleFunctor<ForReduceMin>());
+  ReducibleFunctor<ForReduceLogicalOr>().operator()<bool>(m, "ReduceOr");
+  ReducibleFunctor<ForReduceLogicalAnd>().operator()<bool>(m, "ReduceAnd");
```
pyrito: Is there any purpose to doing all this template metaprogramming? Don't we just need to call `katana::GReduceLogicalAnd` etc. ?

pyrito (commented):
First round of questions based on the current state

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: It sets up a class so that `DefWithNumba` can be used on it's methods. This should be documented explicitly.

libkatana_python_native/src/Reductions.cpp
arthurp: We need to call `ReducibleFunctor` with multiple layers of template arguments so it looks complex here, but it's really just a template function call. The reason for `ForReduceLogicalOr` (and similar) is that C++14 (or whatever exactly we support) doesn't allow templates to be passed as arguments to other templates, so passing `katana::GReduceMin` directly  isn't allowed. So we need to have these wrapper structs that carry the template. Since it's required for the parametric reducers like `GReduceMin` we need to use it for `ReduceLogicalOr` as well to avoid having to duplicate `ReducibleFunctor`.

python/katana/native_interfacing/wrappers.py
arthurp: This is called from register_method in numba_support.py and that is called from the C++ code.

libkatana_python_native/src/Reductions.cpp
arthurp: It's needed to match the type level interface provided by the structs above since they are all used in the same code below.

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -0,0 +1,95 @@
...
+template <typename T, typename Enable = void>
+struct DefEqualsEquals {
+  void operator()(pybind11::class_<T>& cls [[maybe_unused]]) {}
+};
+
+template <typename T>
+struct DefEqualsEquals<
+    T, std::void_t<decltype(std::declval<T>() == std::declval<T>())>> {
+  void operator()(pybind11::class_<T>& cls) {
+    cls.def("__eq__", &T::operator==);
```
denish44, line 40: This is mapping the python equal override to c++ == operator?

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -0,0 +1,95 @@
...
+template <typename T>
+struct DefEquals<
+    T, std::void_t<decltype(std::declval<T>().Equals(std::declval<T>()))>> {
+  void operator()(pybind11::class_<T>& cls) { cls.def("__eq__", &T::Equals); }
+};
+
+/// DefCopy defs `__copy__` and `copy` based on the copy constructor if it is
+/// available.
+template <typename T, typename Enable = void>
+struct DefCopy {
```
denish44, line 95: I understand the python to c++ mapping but for each, what are these maybe_unused portions?

denish44 (approved):
Just a few question and some of what kirthan asked also. Not too familiar with this code but will be picking it up as I go along.

Is there a file I can look at to see how some python class is making use of this binding that your adding?

libkatana_python_native/include/katana/python/Conventions.h
arthurp: Yes. I am using the C++ `==` to implement the python `==`.

libkatana_python_native/include/katana/python/Conventions.h
arthurp: Those are the "default" cases which get used if the type involved *does not* have the methods or operators I'm checking for. This is used to allow `DefConventions` to be called on any type and it will bind whatever conventions are actually present on the type and ignore any that are not.

libkatana_python_native/src/Reductions.cpp
pyrito: Ok I think I kind of understand what you are trying here.

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,146 @@
...
+    numba_support.attr("register_function")(
+        func, (uintptr_t)f, 0,
+        PythonTypeTraits<remove_cvref_t<Return>>::ctypes_type(),
+        PythonTypeTraits<remove_cvref_t<Args>>::ctypes_type()...);
+  }
+
+public:
+  // TODO(amp): The generated wrappers used from numba code are created per
+  //  *signature*, not per function. So two functions will the same overall
+  //  signature (including containing class) will use the same wrapper. This
```
pyrito: nit: it should be "So two functions with the same overall signature..."

libkatana_python_native/src/Reductions.cpp
arthurp: I added a few in code comments to try to help future readers as well. I really wish templates were less stupidly designed.

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,146 @@
...
+        pybind11::module_::import("katana.native_interfacing.numba_support");
+    numba_support.attr("register_function")(
+        func, (uintptr_t)f, 0,
+        PythonTypeTraits<remove_cvref_t<Return>>::ctypes_type(),
+        PythonTypeTraits<remove_cvref_t<Args>>::ctypes_type()...);
+  }
+
+public:
+  // TODO(amp): The generated wrappers used from numba code are created per
+  //  *signature*, not per function. So two functions will the same overall
```
danielmawhirter: functions with

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,95 @@
...
+}  // namespace
+
+void
+katana::python::InitReductions(py::module_& m) {
+  katana::InstantiateForStandardTypes(
+      m, "ReduceSum", ReducibleFunctor<ForGAccumulator>());
+  katana::InstantiateForStandardTypes(
+      m, "ReduceMax", ReducibleFunctor<ForReduceMax>());
+  katana::InstantiateForStandardTypes(
+      m, "ReduceMin", ReducibleFunctor<ForReduceMin>());
```
pyrito, line 95: I think I'm finally understanding what's going on here. If I understand correctly, this function is basically instantiating those types that are accessible within the class, correct? And we are doing this for all the different Reduction functions.

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,95 @@
...
+  // These calls are complex because they would usually be made from the
+  // template above. However, it's better than duplicating the implementation of
+  // ReducibleFunctor.
+  ReducibleFunctor<ForReduceLogicalOr>().operator()<bool>(m, "ReduceOr");
+  ReducibleFunctor<ForReduceLogicalAnd>().operator()<bool>(m, "ReduceAnd");
```
pyrito: And here, we are defining the class name for ReduceOr with specific metadata class metadata, as well as some of the conventional methods. 

Why exactly are we using an operator here? I don't see the benefit of using this compared to another templated function.

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+template <typename For>
+struct ReducibleFunctor {
+  template <typename T>
+  py::object operator()(py::module_& m, const char* name) {
+    py::class_<typename For::template type<T>> cls(
+        m, name,
+        "A reducer object that can updated with new values and combines the "
+        "results efficiently using the appropriate operator.\n"
+        "\n"
```
pyrito: nit: "A reducer object that can updated"? Do you mean "A reducer object that can be updated"?

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,95 @@
...
+    cls.def(
+        py::init([](T v) {
+          auto self = std::make_unique<typename For::template type<T>>();
+          self->update(v);
+          return self;
+        }),
+        "Create a new instance with an initial value.");
+    katana::RegisterNumbaClass(cls);
+    katana::DefWithNumba<const T&>(
+        cls, "update", &For::template type<T>::update,
```
pyrito: So this will call the update method on `GReduceLogicalAnd` or whatever else it's templated with?

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -0,0 +1,101 @@
...
+template <typename T>
+struct DefEqualsEquals<
+    T, std::void_t<decltype(std::declval<T>() == std::declval<T>())>> {
+  void operator()(pybind11::class_<T>& cls) {
+    cls.def("__eq__", &T::operator==);
+  }
+};
+
+/// DefEqualsEquals will def ``__eq__`` (Python ``==``) based on Equals() or
+/// operator==() if one is available (`Equals` is preferred).
```
danielmawhirter:
```
/// DefEquals will def ``__eq__`` (Python ``==``) based on Equals() if it is available
/// or defer to DefEqualsEquals
```

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,95 @@
...
+  py::object operator()(py::module_& m, const char* name) {
+    py::class_<typename For::template type<T>> cls(
+        m, name,
+        "A reducer object that can updated with new values and combines the "
+        "results efficiently using the appropriate operator.\n"
+        "\n"
+        "This class can be passed into numba compiled code and it's methods "
+        "can be used from there.\n");
+    cls.def(py::init<>());
+    cls.def(
```
pyrito, line 56: Is this defining a constructor?

libkatana_python_native/src/Reductions.cpp
```diff
@@ -0,0 +1,95 @@
...
+  template <typename T>
+  py::object operator()(py::module_& m, const char* name) {
+    py::class_<typename For::template type<T>> cls(
+        m, name,
+        "A reducer object that can updated with new values and combines the "
+        "results efficiently using the appropriate operator.\n"
+        "\n"
+        "This class can be passed into numba compiled code and it's methods "
+        "can be used from there.\n");
+    cls.def(py::init<>());
```
pyrito, line 55: Is this a default constructor?

libkatana_python_native/include/katana/python/TypeTraits.h
```diff
@@ -0,0 +1,79 @@
...
+PYTHON_TYPE_TRAITS_BY_PREFIX(int64);
+
+PYTHON_TYPE_TRAITS(float, "float32", "c_float");
+PYTHON_TYPE_TRAITS(double, "float64", "c_double");
+
+#undef PYTHON_TYPE_TRAITS_BY_PREFIX
+#undef PYTHON_TYPE_TRAITS
+
+/// There is no numpy dtype for bool. Requesting it will give a C++ compiler
+/// error.
```
pyrito, line 45: Seriously? I thought numpy had a bool that was 1 byte?

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,146 @@
...
+  template <typename Cls, typename Return, typename ImplCls, typename... Extra>
+  void operator()(
+      pybind11::class_<Cls>& cls, const char* name,
+      Return (ImplCls::*pmf)(Args...), const Extra&... extra) const {
+    // This leaks a single pointer sized struct for each defined numba function.
+    // Repeated import could theoretically cause this to matter, but it's very
+    // unlikely.
+    def_class_method<Cls, Return>(
+        cls, name, pmf, new MemberFunction<ImplCls, Return, Args...>{pmf},
+        extra...);
```
danielmawhirter: This should work just fine, I have no problem with it
Without understanding the full decision space, I would think a more common way to do this would be an effective Singleton, with a static declaration in the MemberFunction struct and a definition in a cpp file

danielmawhirter (approved):
I did not find any major issues on the c++ side

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,146 @@
...
+  //  predictor. The fix for this is to somehow distinguish different
+  //  functions/methods at the template level. This could be done with a counter
+  //  or a name argument of some kind.
+
+  template <typename Return, typename... Extra>
+  void operator()(
+      pybind11::module_& m, const char* name, Return (*pf)(Args...),
+      const Extra&... extra) const {
+    def_func(m, name, pf, extra...);
+  }
```
pyrito: When is this function called?

libkatana_python_native/include/katana/python/TemplateSupport.h
```diff
@@ -0,0 +1,64 @@
...
+/// \param types A dict to fill with instantiations.
+/// \return The resulting class.
+template <typename T, typename F>
+pybind11::object
+InstantiateForType(
+    pybind11::module_ m, const std::string& basename, F f,
+    pybind11::dict types) {
+  pybind11::object dtype = katana::PythonTypeTraits<T>::default_dtype();
+  pybind11::object cls = f.template operator()<T>(
+      m, (basename + "[" + katana::PythonTypeTraits<T>::name + "]").c_str());
```
pyrito: I see. We are creating a different class for each template instantiation, is that correct? Kind of a "duh" moment since that's the idea behind templates but I see that we have to do this to bind things in Python.

libkatana_python_native/include/katana/python/TypeTraits.h
```diff
@@ -0,0 +1,79 @@
...
+    return ctypes.attr("c_void_p");
+  }
+
+  /// Get a Python object which represents the type in Python for purposes of
+  /// numba or type checking.
+  static pybind11::object representation() { return pybind11::type::of<T>(); }
+};
+
+#define PYTHON_TYPE_TRAITS(T, numpy_name, ctypes_name)                         \
+  template <>                                                                  \
```
pyrito, line 12: Why the empty template?

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,146 @@
...
+
+  static remove_cvref_t<Return> call(
+      ConstMemberFunction<Cls, Return, Args...>* func, const Cls* self,
+      remove_cvref_t<Args>... args) {
+    return (self->*(func->member_func))(args...);
+  }
+};
+
+template <typename... Args>
+class DefWithNumbaImpl {
```
pyrito: Where and how is Numba being used? IIRC, it's used for compiling python code, correct?

pyrito (commented):
A lot of stuff going on here, but I think I'm starting to understand more of what's going on now.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: I'm not totally sure what you mean. I think the static on `MemberFunction` must be in the header because we want and need a new instance of static function for every instantiation of the `MemberFunction` template. Also a singleton will not work because the function *must* be static so we can take a pointer to it and cast that pointer to `void*`. Method pointers are not castable because they have different calling conventions than normal function (on at least some compilers).

libkatana_python_native/src/Reductions.cpp
arthurp: Yep. `katana::InstantiateForStandardTypes` is a template that instantiates another template, `ReducibleFunctor<T>::operator()`, at a number of different types. This allow us to create a cross product of element types and reductions while only listing the element types once and the reductions once, and then doing the cross product using nested template instantiations.

libkatana_python_native/src/Reductions.cpp
arthurp: `operator()` *is* a template function. :-) The reason I have to explicitly reference `operator()` instead of just using call syntax is that C++ doesn't allow `ReducibleFunctor<ForReduceLogicalOr>()<bool>(m, "ReduceOr")`. I don't totally understand the reasons why, but it boils down to the fact that call operator expansion (going from `f()` to `f.operator()()`) doesn't allow for template arguments. If fact checking again it appears it doesn't allow it EVER. So I will rename `operator()` to something else since not everyone knows that operators are just normal methods with special syntax associated with them and a few rules.

libkatana_python_native/src/Reductions.cpp
arthurp: This line will not call it. But the python method `cls.update` will call `update` on whatever the template argument specifies as you say.

libkatana_python_native/src/Reductions.cpp
arthurp: Yes.

libkatana_python_native/src/Reductions.cpp
arthurp: Yes. Though in python they are call "zero-arg" or "nullary" since they don't get used magically like they do in C++.

libkatana_python_native/include/katana/python/TypeTraits.h
arthurp: It does. But C++ `bool` is **not** always one byte. :-)

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: This particular overload may not be used yet. In fact I'm pretty sure it isn't. What is allows for is:

```
DefWithNumba(cls, "method", [](Cls* self) { return self->thing() + 1; });
```

This allows lambdas or static methods that make small changes to the C++ method semantics.

libkatana_python_native/include/katana/python/TemplateSupport.h
arthurp: Yep. There are no generics in Python (there aren't generics in C++ either, but that's a different rant), so we need to create a new class for each instantiation.

libkatana_python_native/include/katana/python/TypeTraits.h
arthurp: It's how you write fully specialized template specializations in C++. Otherwise it would be a new template instead of a specialization.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: Numba is what we use to compile user code that will run inside galois/katana parallel loops. So it's used in the `@do_all_operator` and `@for_each_operator` function decorators (indirectly, but that's where it's triggered).

libkatana_python_native/include/katana/python/Conventions.h
arthurp: I prefer to document behavior instead of implementation. But I will add a reference to DefEqualsEquals to help people tracing the implementation.

libkatana_python_native/src/Reductions.cpp
pyrito: That's interesting. I did not know that C++ disallowed that - seems strange that this work around is even permitted if the other is not.

libkatana_python_native/include/katana/python/TypeTraits.h
pyrito: Forgot about good ol' template specialization.

libkatana_python_native/include/katana/python/NumbaSupport.h
danielmawhirter: My understanding:
In order to wrap a method for python, python wants a function pointer that accepts `(self, args...)`. In order to give it that, we're currently creating a function `MemberFunction::call` which accepts `(self, MemberFunction*, args)`, which uses the instance to remember what particular member function it's referring to. This requires a `new` instance of MemberFunction per method to be wrapped. This method generates one `MemberFunction::call` function pointer per _signature_.

Proposal:
Let's instead generate one function pointer per _method_.
```c++
#include <functional>
#include <iostream>

class Thing {
public:
  float my_method(int x, float y) { return x * y; }
};

// MemberFunction is a template on the *Signature*
template <typename Cls, typename Return, typename... Args>
struct MemberFunction {
  // call is a template on the *particular member function*
  template <Return (Cls::*member_func)(Args...)>
  static Return call(Cls *self, Args... args) {
    return (self->*member_func)(args...);
  }
};

int main() {
  std::function<float(Thing *, int, float)> function =
      MemberFunction<Thing, float, int, float>::call<&Thing::my_method>;
  Thing t;
  float z = function(&t, 6, 7.0);
  std::cout << z << "\n";
  return 0;
}
```

Only using std::function for readability, and I omitted some cvref things, but hopefully my idea is clear. I believe this can replace
`(uintptr_t)&Caller::call, (uintptr_t)caller`
in the registration line, so we don't have to make a `new` MemberFunction for each one. These will exist statically, and without duplication

arthurp: Ah! I understand what you mean now. I think I tried that and it did not work. I will try again though.

The issue I think I ran into before is that member pointers are not allowed as template arguments. As far as I know no pointers are. I don't think C++ treats them as `constexpr` because they are strictly speaking not known until dynamic link time.

arthurp: And I tried it (as you probably did) and it works. So I'm gonna need to rethink that and figure out what I failed to do correctly before. Good catch. This is exactly why I asked for someone to look at this. I clearly misunderstood what is allowed as a template parameter.

danielmawhirter: I can't promise that the code I wrote is compliant c++, works with all compilers, or works across compilation units, as I don't really know. clang-12 seems fine with it. Happy to help!

libkatana_python_native/src/Reductions.cpp
arthurp: Yep. There are a lot of cases where C++ allows a work around, but does not allow doing something directly. I find it maddening.

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,161 @@
...
+        PythonTypeTraits<remove_cvref_t<Return>>::ctypes_type(),
+        PythonTypeTraits<remove_cvref_t<Args>>::ctypes_type()...);
+  }
+
+public:
+  // Overload to handle non-capturing lambdas and static functions. These
+  // should all take self as the first argument.
+  template <typename Return, typename... Extra>
+  void operator()(
+      pybind11::module_& m, const char* name, Return (*pf)(Args...),
```
amberhassaan: Are you assuming that `Extra...` and `Args...` match in types?

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,161 @@
...
+
+  // Overload to handle const methods on classes.
+  template <typename Cls, typename Return, typename ImplCls, typename... Extra>
+  void operator()(
+      pybind11::class_<Cls>& cls, const char* name,
+      Return (ImplCls::*pmf)(Args...) const, std::true_type,
+      const Extra&... extra) const {
+    // As above.
+    def_class_method<Cls, Return>(
+        cls, name, pmf, new ConstMemberFunction<ImplCls, Return, Args...>{pmf},
```
amberhassaan: How is `new ConstMemberFunc...` deleted?

amberhassaan (commented):
minor comments . In hindsight, I should have read up on `pybind11` a bit more but I was able to understand some parts.

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -0,0 +1,102 @@
...
+  void operator()(pybind11::class_<T>& cls) {
+    // If Equals doesn't exist then check for ==.
+    DefEqualsEquals<T>{}(cls);
+  }
+};
+
+template <typename T>
+struct DefEquals<
+    T, std::void_t<decltype(std::declval<T>().Equals(std::declval<T>()))>> {
+  void operator()(pybind11::class_<T>& cls) { cls.def("__eq__", &T::Equals); }
```
denish44, line 62: Should we also add the other operators like >=, <=, etc?

denish44 (approved):
LG, lot of questions from others so following that and will also need to look at pybind11 a bit to understand some parts but overall the idea makes sense.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: It is not. But there is a fix coming for this. Give me a couple of hours to setup the new design due to @danielmawhirter.

libkatana_python_native/include/katana/python/Conventions.h
arthurp: Hmm. Yes. Yes, we should. I'll do that.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: No. `Extra...` is a bunch of "option" arguments that are used by pybind11. This code never interprets them. I just pass them through. The are only used to configure the generated python wrapper with things like doc strings and argument names and default argument values and stuff.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: I pushed a rework of the code to use the approach you recommend above.

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -14,138 +14,207 @@ namespace detail {
...
+          decltype(func)>);
+  detail::StaticFunctionBinder<func>::template def_method<func>(
+      m, name, extra...);
+}
+
+/// Define a Python and numba method.
+/// This is used as follows:
+/// @code
+/// DefWithNumba&lt;&Cls::method>(cls, "method");
+/// @endcode
```
pyrito, line 183: nit: Do you have to say `&lt` to represent `<` ?

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -14,138 +14,207 @@ namespace detail {
...
+
+/// A utility class encapsulating the type manipulation required for binding
+/// const member functions.
+template <typename _Cls, typename _Return, typename... _Args>
+struct ConstMemberFunctionBinderImpl {
+  using Cls = _Cls;
+  using Return = _Return;
+  template <typename Impl>
+  using BaseMemberFunctionPointer =
+      Return (Impl::*)(remove_cvref_t<_Args>...) const;
```
pyrito, line 103: What's going on here?

libkatana_python_native/src/Reductions.cpp
```diff
@@ -61,19 +61,20 @@ struct ReducibleFunctor {
...
     katana::RegisterNumbaClass(cls);
-    katana::DefWithNumba<const T&>(
-        cls, "update", &For::template type<T>::update,
+    katana::DefWithNumba<py::overload_cast<const T&>(
+        &For::template type<T>::update)>(
+        cls, "update",
         "Update this reducer with ``v`` performing the operation.");
-    katana::DefWithNumba<>(
-        cls, "reduce", &For::template type<T>::reduce,
+    katana::DefWithNumba<&For::template type<T>::reduce>(
```
pyrito, line 68: Ah so this is the change from the earlier conversation where you were passing in a reference to the function instead of just passing in the function as a template parameter?

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: Yes. Yes. I do. Doxygen is a PITA.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: I'm defining a type member which is a member function pointer with a configurable class and has the same return type, but has arguments which are modified by removing `const` `volatile` and `&` (reference). The configurable class is needed to handle methods that are defined on a base class, but are being used on subclass instances.

libkatana_python_native/src/Reductions.cpp
arthurp: Yes. Though we are not templating on the function. We are templating on the function pointer. And it's suprisingly complex to get this working, much more template magic this way than with a normal argument.

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -0,0 +1,230 @@
...
+  static_assert(std::is_same_v<
+                typename detail::MemberFunctionBinder<
+                    member_func>::MemberFunctionPointer,
+                decltype(member_func)>);
+  detail::MemberFunctionBinder<member_func>::template def_class_method<
+      member_func>(cls, name, extra...);
+}
+
+// TODO(amp): Implement missing overloads/variants for: lambdas (as methods and
+//  functions), static methods.
```
danielmawhirter, line 210: Can't you just use the Function api for static methods?

danielmawhirter (approved):
Current state LGTM

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: Not really, because they would not be placed correctly in Python since static methods in C++ should become methods on the class in Python. But you could if you don't care about that.

pyrito (approved):
Some of the templating magic you've done are still a bit of a mystery to me, but everything else seems reasonably sane to me.

cmake/Modules/PythonSetupSubdirectory.cmake
```diff
@@ -134,25 +134,31 @@ function(_generate_build_configuration_txt)
...
+      list(APPEND LINK_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},LINKER:-rpath=$<TARGET_LINKER_FILE_DIR:${TARGET_NAME}>>")
+      list(APPEND LINK_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},LINKER:-rpath=$<JOIN:$<TARGET_PROPERTY:${TARGET_NAME},BUILD_RPATH>,;LINKER:-rpath=>>")
+      list(APPEND LINK_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},$<TARGET_LINKER_FILE:${TARGET_NAME}>>")
+    endif()
+    list(APPEND LINK_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},LINKER:-rpath=$<JOIN:$<TARGET_PROPERTY:${TARGET_NAME},INSTALL_RPATH>,;LINKER:-rpath=>>")
+    list(APPEND LINK_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},$<TARGET_PROPERTY:${TARGET_NAME},INTERFACE_LINK_DEPENDS>>")
+    list(APPEND LINK_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},$<TARGET_PROPERTY:${TARGET_NAME},LINK_DEPENDS>>")
+    # TODO(amp): Including INTERFACE_COMPILE_OPTIONS and using pybind11 causes LTO to be enabled, but only the compiler
+    #  option is actually passed to the setuptools compilers. So just disable this for the moment.
+    # list(APPEND COMPILE_OPTIONS "$<TARGET_GENEX_EVAL:${X_TARGET_NAME},$<TARGET_PROPERTY:${TARGET_NAME},INTERFACE_COMPILE_OPTIONS>>")
```
arthurp, line 156: @ddn0 I do not like this hack, but I cannot for the life of me determine why the pybind11 LTO options are being picked up here. And turning it off doesn't seem to break things. Yet. What do you think?

ddn0: I'm not following. What is the issue with just propagating LTO here?

arthurp: Sorry. I didn't explain well.

LTO must be enabled at both compile and link time. If it's only enabled at compile time the linker will choke on the object files because LTO objects are totally different. We cannot pass through LINK_OPTIONS because some of the GPU related generator expressions (HOST_ONLY or something) cause an unavoidable CMake failure when used in custom files. So the linker does not get the LTO options set. Passing through the compiler options puts us in a state where the linker chokes, so I just didn't. I don't like it, but it works for now. Honestly the only fix I can see for all this is to stop using CMake or write a "dump flags" backend for CMake (in the codebase, there is no external backend support).

PR #776, opened 2022-01-01
Replace repeated explicit autodoc options with defaults in the config file

arthurp:
An enterprise PR will come later. They don't need to be in sync or anything.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,8 +5,9 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here.
 
-import pathlib
 import os
+import pathlib
+
 import katana
 
 try:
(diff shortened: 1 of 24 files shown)


docs/conf.py
```diff
@@ -92,7 +101,7 @@
 # -- Options for extensions --------------------------------------------------
 
 # nbsphinx configuration values
-html_sourcelink_suffix = '' # Don't add .txt suffix to source files
+html_sourcelink_suffix = ""  # Don"t add .txt suffix to source files
```
JamesFCoffey:
```suggestion
html_sourcelink_suffix = ""  # Don't add .txt suffix to source files
```

arthurp: oooops. ;-) I over did the find and replace.

docs/reference/python/katana.local.graph.rst
```diff
@@ -7,7 +7,3 @@ A ``Graph`` is created by calling one of the import routines from
 :py:mod:`~katana.local.import_data`.
 
 .. autoclass:: katana.local.Graph
-   :members:
-   :inherited-members:
-   :undoc-members:
-   :special-members: __init__, __iter__, __getitem__, __setitem__, __len__
```
JamesFCoffey, line 13: Should `__iter__` and `__len__` be removed from the current documentation? There exists `Graph.num_nodes` for getting the number of nodes. How should users iterate through the nodes?

docs/reference/python/katana.local.datastructures.rst
```diff
@@ -3,6 +3,4 @@ Data Structures
 ===============
 
 .. automodule:: katana.local.datastructures
-   :members:
-   :undoc-members:
    :special-members: __init__, __iter__, __getitem__, __setitem__, __len__
```
JamesFCoffey, line 6: I noticed these special members were deleted elsewhere. Why are they included here?

JamesFCoffey (commented):
Thanks for setting the defaults in the `conf.py` file. I have some questions regarding the inclusion of special members.  In the places where you kept them listed, why are they needed? Are they all to be removed eventually (except `__init__`)?

docs/reference/python/katana.local.graph.rst
arthurp: I shouldn't have removed them here. We may be changing this API a bit, but for the moment, `__len__` and `__iter__` should be documented on `Graph` as you say. I will add this back.

docs/reference/python/katana.local.datastructures.rst
arthurp: They should have been left in the graph case. See above.

arthurp: The additional special members are left here because some of the data structures have these special methods and they should be documented. I wanted to include all of these in the default, but that causes a huge number of warnings because these methods are then effectively listed on all classes even those that do not have these methods which is (reasonably) a warning in sphinx.

JamesFCoffey (approved):

PR #777, opened 2022-01-02
Construct new arrow table by picking rows from an existing one

witchel:
Also, `MakeEmptyArrowTable`, and allow `NullChunkedArray `to return a `Result`.

### DIFF
--- a/libsupport/src/ArrowInterchange.cpp
+++ b/libsupport/src/ArrowInterchange.cpp
@@ -7,6 +7,7 @@
 #include <arrow/array/concatenate.h>
 
 #include "katana/Random.h"
+#include "katana/Result.h"
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


libsupport/src/ArrowInterchange.cpp
```diff
@@ -48,19 +49,51 @@ ApproxArrayDataMemUse(const std::shared_ptr<arrow::ArrayData>& data) {
 
 }  // anonymous namespace
 
-std::shared_ptr<arrow::ChunkedArray>
+katana::Result<std::shared_ptr<arrow::Table>>
+katana::TakeRows(
+    const std::shared_ptr<arrow::Table>& original,
+    const std::vector<bool>& picker) {
```
danielmawhirter: If you make an arrow::BooleanArray instead of vector<bool> and perform the operation column-wise, this would match the [Filter](https://arrow.apache.org/docs/cpp/compute.html#selections) API. That should be meaningfully more efficient than ConcatenateTables

danielmawhirter (approved):
I'd prefer to have this comment addressed, but I won't block for it

libsupport/src/ArrowInterchange.cpp
witchel: Yeah, I like this idea.  I'll look into it.

roshandathathri (approved):
LGTM

PR #778, opened 2022-01-03
Pin fmt to <8.1.

arthurp:
8.1 fixes a bug/misfeature which implicitly converted enum class values into ints for formatting. We rely on this bad behavior all over the place.

See: KAT-3182


micah-white (approved):
LGTM.

astolman (approved):
thanks for the quick fix

PR #780, opened 2022-01-04
Docs: fix root privileges comment

scober:
The docs state that people can contact sudoers@katanagraph.com for
issues requiring root privileges. That is not true for all people
building this code. Reword as a more generic "users with root
privileges".

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,8 +82,11 @@ html_css_files = ["style.css"]
 autodoc_preserve_defaults = True
 autodoc_member_order = "groupwise"
 
-suppress_warning = []
+# the open documentation links to the enterprise documentation in some places
+# let those links be broken in the open-only documentation and don't complain
+suppress_warnings = ["ref.ref"]
 nitpick_ignore = []
+
 
 # Exclude drafts from external documentation
 if tags.has("external"):
--- a/docs/contributing/building.rst
+++ b/docs/contributing/building.rst
@@ -69,7 +69,7 @@ The ``conda env update`` line can be run later to update your environment. Deact
 
 .. warning::
 
-   Not all conda installation scripts are well-behaved. In particular, the default behavior of the ``cudatoolkit-dev`` package's post-installation script is to write a log to the hard-coded path ``/tmp/cuda-installer.log`` and not delete it. Once one user on a system has written this file all other users will fail to install the package because they won't be able to write the log file. The only known workaround is to manually delete the log file. It is best practice to delete your own log file immediately after you have updated your environment. If you are on a system where someone else has written the log file and you can't overwrite it please contact that user to delete it. You may also write to sudoers@katanagraph.com.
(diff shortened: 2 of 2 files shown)


pyrito (approved):

srogatch (approved):

tylershunt (approved):

scober:
I re-requested everyone's review because I am curious about people's thoughts on my docs-building hack

docs/conf.py
```diff
@@ -82,9 +82,12 @@
 autodoc_preserve_defaults = True
 autodoc_member_order = "groupwise"
 
-suppress_warning = []
+# the open documentation links to the enterprise documentation in some places
+# let those links be broken in the open-only documentation and don't complain
+suppress_warnings = ["ref.ref"]
```
tylershunt, line 87: does this mean we don't get this check in enterprise?

scober: I _believe_ that the enterprise build has its own `conf.py` and symlinks to all of the open docs files. So the enterprise build should not be suppressing this warning.

tylershunt: Then this seems fine to me; in general I'm not sure how to feel about broken links on our public docs site.

scober: The link that I added this for is hidden behind an `only:: internal`. So for this specific use case we only have a broken link in internal documentation that doesn't include the enterprise docs, which I'm not super worried about. I'm worried about ignoring warnings at this granularity - I would have preferred to suppress the warning for a single broken link.

tylershunt: OK I guess ci on enterprise would detect broken links in the open repo anyway? (Trying to figure out how much of this cake we have that we get to eat)

scober: Yeah, I think a truly broken link would get caught in enterprise CI, which is better than nothing. I think our other options are 

1. Don't link from open repo docs to enterprise repo docs - which makes addenda-style enterprise docs less useful.
2. Keep trying to find a more granular way to ignore broken links - which I think could be expensive in developer time given how much time I have already dedicated to this.

tylershunt: thanks; this is fine with me but I don't feel like I have enough context/authority over docs to be final approval here.

tylershunt:
Woops, and I fat fingered the merge button; crap sorry! I'll fix it if others have problems

docs/conf.py
origandrew: Sorry, this is not okay. Suppressing this warning would result in broken links not being caught by testing and potentially making its way into public-facing documentation

docs/contributing/building.rst
```diff
@@ -6,7 +6,7 @@ Building
 
 .. only:: internal
 
-   If you are building katana-enterprise make sure to also read the :doc:`enterprise build addenda <index>`.
+   If you are building katana-enterprise make sure to also read the :ref:`enterprise-building`.
```
origandrew, line 9: Please create a building-internal.rst file on the enterprise side that has this using include directive. There are a number of ways to accomplish this effect. If you'd like some advice on how to do this, reach out to me.

origandrew:
Can you revert this PR? That warning suppression is not okay as it could potentially result in a broken public documentation site

PR #781, opened 2022-01-04
Tsuba reorg: Make libtsuba dependent on libgalois and thus parallelizable

amberhassaan:
NOTE: Please don't be intimidated by the large changeset. Each commit can be viewed as a self-contained change. I've tried to (and plan on) making the commits self contained. 

libtsuba and libsupport have some code that needs to be parallelized but could not be parallelized because of the dependency structure. 

Now that parallel loops and runtime have been separated into libgalois, we can make libtsuba dependent on libgalois and move code that needs to be parallelized out of libsupport. 

[KAT-3205]

[KAT-3205]: https://katanagraph.atlassian.net/browse/KAT-3205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libgraph/include/katana/GraphTopology.h
```diff
@@ -202,19 +202,19 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
   EdgeShuffleTopology& operator=(const EdgeShuffleTopology&) = delete;
 
   bool is_transposed() const noexcept {
-    return has_transpose_state(tsuba::RDGTopology::TransposeKind::kYes);
+    return has_transpose_state(katana::RDGTopology::TransposeKind::kYes);
```
tylershunt: Now that we're all one big happy namespace family we could just remove these rather than replacing them. Not the end of the world to be more verbose though.

libgraph/src/PropertyGraph.cpp
```diff
@@ -113,15 +113,15 @@ WriteEntityTypeIDsArray(
       ff->Write(&data, sizeof(katana::EntityTypeIDArrayHeader));
 
   if (!aro_sts.ok()) {
-    return katana::ArrowToTsuba(aro_sts.code());
+    return katana::ArrowToKatana(aro_sts.code());
```
tylershunt, line 116: This might not be useful anymore since `KATANA_CHECKED` can just handle Arrow status objects. We can leave that bit as future work though I totally get wanting this PR to be pure refactor and not change error messages or anything like that.

tylershunt (approved):
Looks good overall; I trusted your grep replace and didn't look too closely at those mechanical changes. The clean up w.r.t. errors makes sense to me.

One bummer was that we didn't get to benefit from the shared namespace by reducing verbosity, but we can grow into that over time :D.

libgraph/src/PropertyGraph.cpp
amberhassaan: Thank you and I agree. I don't want to make any "real changes" in this PR given that the changeset is so big. I will make note of the points you raised and address them in the near future. I'm planning a subsequent PR that moves `EntityManager` and parallelizes it.

arthurp (approved):

e-mcginnis (approved):
looks great to me, very pleased with the direction our dependency graph is going. Thanks for doing this!

PR #787, opened 2022-01-08
support: add a new id type for indexing (supports prefix sum)

tylershunt:
This is a coercion resistant wrapper for an index. They support index-y
things like being added together an subtracted, but their type keeps
you from using them in the wrong place.

The protections against misuse aren't as strong as the other Opaque
types. IndexIDs are mainly useful for cases when the keys are literally
indexes to something dense like an array, in these cases it's convenient
to do prefix sums over indexes (for instance) when assigning things in
blocks.

### DIFF
--- a/libsupport/include/katana/IndexID.h
+++ b/libsupport/include/katana/IndexID.h
@@ -1,0 +1,78 @@
+#ifndef KATANA_LIBSUPPORT_KATANA_INDEXID_H_
+#define KATANA_LIBSUPPORT_KATANA_INDEXID_H_
+
+#include "katana/OpaqueID.h"
+
+namespace katana {
+
+/// A numeric ID type suitable for indexing. IndexIDs support:
+///
+/// - Addition and subtraction of other IDs (the result of which is an ID)
+/// - Increment and decrement
+/// - Dereference
+///
+/// NB: IndexID is less safe in general than `OpaqueID` and `OpaqueLinearID`.
+///     Use those stricter variants whenever possible.
+template <typename _IDType, typename _Value>
+struct IndexID : public OpaqueIDOrderedWithValue<_IDType, _Value> {
+public:
+  using OpaqueIDOrderedWithValue<_IDType, _Value>::OpaqueIDOrderedWithValue;
+
+  // iterator traits
+  using difference_type = _IDType;
+  using value_type = _IDType;
+  using pointer = _IDType*;
+  using reference = _IDType&;
+  using iterator_category = std::random_access_iterator_tag;
+
+  // must provide a dereference to be an iterator
+  reference operator*() { return *static_cast<_IDType*>(this); }
+
+  _IDType& operator++() {
+    ++this->value_;
+    return *static_cast<_IDType*>(this);
+  }
+
+  _IDType& operator--() {
+    --this->value_;
+    return *static_cast<_IDType*>(this);
+  }
+
+  _IDType operator++(int) {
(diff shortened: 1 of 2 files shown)


roshandathathri (approved):
LGTM

There was already a substraction/difference operator, so I don't see any issue with an addition operator.

libsupport/include/katana/OpaqueID.h
```diff
@@ -241,6 +241,11 @@ struct OpaqueIDLinear : public OpaqueIDOrderedWithValue<_IDType, _Value> {
     return *static_cast<_IDType*>(this);
   }
 
+  _IDType& operator+=(const _IDType& rhs) {
+    this->value_ += rhs.value();
+    return *static_cast<_IDType*>(this);
+  }
```
arthurp: This isn't correct theoretically. The issue is that Linear Opaque IDs are *not* numbers. They are ordered and have differences, but they have no "value" inherently (`value()` is a backdoor). If you want to do a prefix sum, you should do it on the `DifferenceType` not on the IDs themselves. Pick a starting ID and compute differences out from that. This makes it explicit when you are working on numerical differences instead of IDs themselves (which again are not numbers and have no numeric value).

All that said, we can add this, but I don't think it should be on `Linear` since this changes that semantics of the OpaqueID it's used on to combine the idea of IDs and differences. I think that should be a separate subclass. Maybe we could call it `NumericID`. It would not have a difference type and would return IDs from subtraction as well as supporting addition of IDs.

scober: I agree with this. After spending some time with `OpaqueID`s I have been convinced that differentiating between `DifferenceType` and `ValueType` is a valuable feature.

tylershunt: > This isn't correct theoretically. The issue is that Linear Opaque IDs are _not_ numbers. They are ordered and have differences, but they have no "value" inherently (`value()` is a backdoor). If you want to do a prefix sum, you should do it on the `DifferenceType` not on the IDs themselves. Pick a starting ID and compute differences out from that. This makes it explicit when you are working on numerical differences instead of IDs themselves (which again are not numbers and have no numeric value).
 
Makes sense; I'll add another type. I'm using these as typed indexes so I'm implicitly ascribing numeric qualities to them already.

arthurp: That makes sense. Maybe `IndexID` would be a good name for it. Since that use case is common and makes sense.

tylershunt:
I declared a distinct type; I re-used some of the existing `Opaque...` machinery though; I hope that's ok!

libsupport/include/katana/IndexID.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "katana/OpaqueID.h"
+
+namespace katana {
+
+/// A numeric ID type suitable for indexing. IndexIDs support:
+///
+/// - Addition and subtraction of other IDs (the result of which is an ID)
+/// - Increment and decrement
+/// - Dereference
```
arthurp, line 12: Please add a note here that we should prefer `OpaqueID` and `OpaqueIDLinear` to this because those provide more safety. This type will allow accidental math with int which was part of what opaque IDs were designed to prevent. So while there will be cases where this is appropriate people should use the stricter variants if possible.

arthurp (commented):
Reusing machinery is great. They are definitely part of the same family with different levels of safety.

PR #789, opened 2022-01-09
Louvain Clustering: Add both default and undirected views

gurbinder533:


lonestar/analytics/cpu/louvain_clustering/louvain_clustering_cli.cpp
```diff
@@ -91,10 +91,11 @@ main(int argc, char** argv) {
   totalTime.start();
 
   if (!symmetricGraph) {
```
amazloumi: same here, I think it is better to show the following warning if the flag is set.

uditagar (approved):
looks fine

PR #791, opened 2022-01-10
Add parallel reading / writing for Sphinx [KAT-2621]

JamesFCoffey:
I reduced the `sphinx-build` time from it's upper bound at 157s (on my machine). The lower bound of 52s without Breathe / C++ API docs. In this commit, I got the time reduced to 137s using parallel reading / writing.

**Alternatives approaches tried:**

- I looked into the [slow building](https://github.com/michaeljones/breathe/issues/439) associated with the `Breathe` extension. The solution seems to be editing the `Breathe` package itself so that it doesn't rely on `xml.dom` and uses `lxml`. The project has not implemented this change since 2019 when discovered. This does sound like a good avenue for Developer Relations to contribute to an  open source project and make improvements that benefit Katana Graph. **Does this stay as a bug or be transformed into planned work?** 
- I looked at the performance of removing DOT from Doxygen, but that increased the time by 2 minutes.
- Using parallel processing reduced the time modestly but was also dependent on the number of workers, with 4 workers appearing optimal. (Using `auto` did not provide a performance increase.)

  - `-j 2` yielded 138s
  - `-j 4` yielded 137s
  - `-j 6` yielded 142s
  - `-j 8` yielded 157s

JIRA: KAT-2621

### DIFF
--- a/cmake/Modules/BuildCommon.cmake
+++ b/cmake/Modules/BuildCommon.cmake
@@ -463,6 +463,8 @@ function(add_katana_sphinx_target target_name)
   if (BUILD_DOCS STREQUAL "internal")
     set(sphinx_options "-W;${sphinx_options}")
   endif ()
+  # Enable parallel reading / writing, 4 workers gave optimal time reduction
+  set(sphinx_options "-j 4;${sphinx_options}")
 
   add_custom_target(
       ${target_name}_sphinx_docs
--- a/cmake/Modules/BuildCommon.cmake
+++ b/cmake/Modules/BuildCommon.cmake
@@ -63,6 +63,7 @@ set(KATANA_NUM_TEST_GPUS "" CACHE STRING "Number of test GPUs to use (on a singl
 set(KATANA_USE_LCI OFF CACHE BOOL "Use LCI network runtime instead of MPI")
 set(KATANA_NUM_TEST_THREADS "" CACHE STRING "Maximum number of threads to use when running tests (default: min(number of physical cores, 8))")
 set(KATANA_AUTO_CONAN OFF CACHE BOOL "Automatically call conan from cmake rather than manually (experimental)")
+set(KATANA_NUM_DOC_THREADS "" CACHE STRING "Maximum number of threads to use when reading / writing with Sphinx (default: min(number of physical cores, 4))")
 
 ###### Configure (users don't need to go beyond here) ######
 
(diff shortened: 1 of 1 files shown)


cmake/Modules/BuildCommon.cmake
```diff
@@ -463,6 +463,8 @@ function(add_katana_sphinx_target target_name)
   if (BUILD_DOCS STREQUAL "internal")
     set(sphinx_options "-W;${sphinx_options}")
   endif ()
+  # Enable parallel reading / writing, 4 workers gave optimal time reduction
+  set(sphinx_options "-j 4;${sphinx_options}")
```
arthurp: Is `-j 4` correct? I feel like it should be `-j;4` so the two arguments are split at the CMake list level.

BTW, where did you measure the time that showed 4 workers was best? And how different was, say, 2 and 6 workers? I ask because if it's really sensitive to this parameter we might need to make it configurable so CI can use an optimized value that's different from the local build defaults.

arthurp (approved):
I'm a little worried about hard coding 4 workers. But I cannot think of a better way to do it. I think this is OK.

JamesFCoffey:
> I'm a little worried about hard coding 4 workers. But I cannot think of a better way to do it. I think this is OK.

Yeah, me too. I'm opposed to "magic numbers", but most have the ability for 4 workers these days and it did give the best performance. It traded fast reading for waiting for workers.

arthurp:
You could add an option similar to this:

https://github.com/KatanaGraph/katana/blob/88e78bd6dc31dba14ce9b54a05a546e4cc2d74e0/cmake/Modules/BuildCommon.cmake#L64

That would make it tweakable if we needed to and eliminate the "undocumented magic number" element of it. But the default can still be provided, so if no one has issues we never need to touch it.

cmake/Modules/BuildCommon.cmake
JamesFCoffey: - `-j 2` yielded 138s
  - `-j 4` yielded 137s
  - `-j 6` yielded 142s
  - `-j 8` yielded 157s

I measured this in my shell.

JamesFCoffey:
@arthurp , made it tweakable.

arthurp:
Love it. Thanks! This will also make testing more easy if we need to.

PR #793, opened 2022-01-11
add unstable_storage_format_version support KAT-2778

e-mcginnis:
Adds UnstableRDGStorageFormat flag to enable storing/loading unstable
storage format RDGs

Developers should use the unstable storage format while developing features for Katana software which alter, add, or remove, RDG data structures to avoid disturbing everyone else that relies on the RDG format to be stable.

Katana software executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable supports loading/storing RDGs in the unstable RDG storage format.
Katana software executed *without* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable *does not* support loading/storing RDGs in the unstable RDG storage format.

Katana software uses the boolean flag `unstable_storage_format` in the RDGs part header to identify unstable format RDGs

If Katana software is executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable:
1. katana software stores RDGs:
   - with `storage_format_version = latest_storage_format_version`
   - with `unstable_storage_format = true`
2. RDGs with the "unstable_storage_format" are loaded

If Katana software is executed *without* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable:
1. katana software stores RDGs:
   - with `storage_format_version = latest_storage_format_version`
   - with `unstable_storage_format = false`
2. RDGs with the "unstable_storage_format" are not loaded, and an error is thrown

In Katana software the feature flag can be checked by using
```
KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat);
```

### DIFF
--- a/docs/contributing/rdg-storage-format.rst
+++ b/docs/contributing/rdg-storage-format.rst
@@ -1,0 +1,70 @@
+=========================
+Katana RDG Storage Format
+=========================
+
+Katana persists its graphs on disk in a format called `RDG: Resilient Distributed Graph`. This `RDG` is a directory filled with files that define the graph.
+The `storage_format_version` of an RDG determines the layout of the RDG
+An example of an RDG can be found here
+https://github.com/KatanaGraph/test-datasets/tree/master/rdg_datasets/ldbc_003/storage_format_version_3
+
+Katana software uses the `storage_format_version` of a loaded RDG to determine some of the features it supports
+
+Katana software always stores the RDG as the latest `storage_format_version`, sometimes with the `unstable_storage_format` modifier (covered below)
+RDGs stored in a newer `storage_format_versions` are incompatible with older software.
+All versions of Katana after a `storage_format_version` is introduced should support loading/reading RDGs stores as that `storage_format_version`.
+
+Early RDGs from before `storage_format_version` was introduced are assumed to be `storage_format_version=1`
+
+Rarely, `storage_format_versions` can be deprecated. When this occurs tooling will be provided to convert all RDGs stored in that `storage_format_version` to the next supported `storage_format_version`.
+
+Unstable RDG Storage Format
+===========================
+
+Developers should use the unstable storage format while developing features for Katana software which alter, add, or remove, RDG data structures to avoid disturbing everyone else that relies on the RDG format to be stable. 
+
+Katana software executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable supports loading/storing RDGs in the unstable RDG storage format.
+Katana software executed *without* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable *does not* support loading/storing RDGs in the unstable RDG storage format.
+
+Katana software uses the boolean flag `unstable_storage_format` in the RDGs part header to identify unstable format RDGs
+
+If Katana software is executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable:
+#. katana software stores RDGs:
+ 
+   #. with `storage_format_version = latest_storage_format_version`
+   #. with `unstable_storage_format = true`
+
+#. RDGs with the "unstable_storage_format" are loaded
+
+
+If Katana software is executed *without* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable:
+#. katana software stores RDGs:
+   #. with `storage_format_version = latest_storage_format_version`
+   #. with `unstable_storage_format = false`
+
+#. RDGs with the "unstable_storage_format" are not loaded, and an error is thrown
+
+
+In Katana software the feature flag can be checked by using
+```
+KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat);
+```
+
+Developing a new revision of the RDG Storage Format
+===================================================
+
+This process will be largely determined by what new data structures are being persisted in the RDG, but the following are some guidelines.
+
+#. Figure out the in-memory representation of the feature. Figure out how the in-memory representation will be generated from when an RDG without the new data structures is loaded.
+#. Add support for persisting the in-memory representation on disk in the RDG. Gate persisting these new data structures behind `KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat);`. When `UnstableRDGStorageFormat` is not set, the persisted RDG should look exactly like an RDG without support for the new data structures.
+#. Add support for loading the new data structures. Gate loading these new data structures behind `KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat);`. When `UnstableRDGStorageFormat` is not set, the code should behave like it loaded an RDG without the new data structures.
+#. Write tests for storing/load the new feature.
+#. Stabilize the features in-memory representation
+#. Stabilize the features on-disk representation
+#. When the representations are sufficiently stable
+
+   #. increase the `kLatestPartitionStorageFormatVersion` in `RDGStorageFormatVersion.h`.
+   #. Mirror this change for `KATANA_RDG_STORAGE_FORMAT_VERSION` in `TestDatasets.cmake`
+   #. uprev the rdgs in the `test-datasets` repo https://github.com/KatanaGraph/test-datasets
+   #. update the version of `test-datasets` used by Katana
+   #. update the `DATASETS_SHA` with the newest master sha from `test-datasets` in `example_data.py`
+
--- a/docs/contributing/rdg-storage-format.rst
+++ b/docs/contributing/rdg-storage-format.rst
@@ -38,6 +38,7 @@ If Katana software is executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableR
 
 If Katana software is executed *without* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable:
 #. katana software stores RDGs:
+
    #. with `storage_format_version = latest_storage_format_version`
    #. with `unstable_storage_format = false`
 
--- a/libtsuba/src/RDGPartHeader.cpp
+++ b/libtsuba/src/RDGPartHeader.cpp
@@ -7,6 +7,7 @@
 #include "PartitionTopologyMetadata.h"
 #include "RDGHandleImpl.h"
 #include "katana/ErrorCode.h"
+#include "katana/Experimental.h"
 #include "katana/FaultTest.h"
 #include "katana/FileView.h"
 #include "katana/Logging.h"
@@ -24,6 +25,7 @@ const char* kEdgePropertyKey = "kg.v1.edge_property";
 const char* kPartPropertyFilesKey = "kg.v1.part_property_files";
 const char* kPartPropertyMetaKey = "kg.v1.part_property_meta";
 const char* kStorageFormatVersionKey = "kg.v1.storage_format_version";
+const char* kUnstableStorageFormatFlagKey = "kg.v1.unstable_storage_format";
 // Array file at path maps from Node ID to EntityTypeID of that Node
 const char* kNodeEntityTypeIDArrayPathKey = "kg.v1.node_entity_type_id_array";
 // Array file at path maps from Edge ID to EntityTypeID of that Edge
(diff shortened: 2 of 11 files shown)


e-mcginnis:
blocked by https://github.com/KatanaGraph/katana/pull/795

libtsuba/src/RDGPartHeader.h
```diff
@@ -316,6 +316,19 @@ class KATANA_EXPORT RDGPartHeader {
...
+  }
+
+  /// Any feature which results in changes to the storage format version
+  /// but has not yet stabilized should ensure this is called
+  /// The feature can then be developed by setting the
+  /// KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"
+  /// env var. See RDGStorageFormatVersion.h for more details
+  void set_unstable_storage_format() {
+    unstable_storage_format_ = true;
+  }
```
pyrito: Would it ever make sense to give the users the option to set this back to false? I can't think of one, but figured I'd pose the question.

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,61 @@
+=========================
+Katana RDG Storage Format
```
pyrito, line 2: I'm surprised that we didn't explain what an RDG was in any other portion of the contributing docs.

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,61 @@
...
+=========================
+
+Katana persists its graphs on disk in a format called `RDG`. This `RDG` is a directory filled with files that define the graph.
```
pyrito: nit: define what the acronym for RDG stands for

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -316,12 +338,54 @@ katana::from_json(const json& j, katana::RDGPartHeader& header) {
...
+        "will be permitted.");
+    if (header.unstable_storage_format_) {
+      KATANA_LOG_WARN(
+          "Loading an RDG stored in an unstable storage format. If any issues "
+          "are encountered, please regenerate this RDG before filing any bugs, "
+          "as the unstable storage format can change without notice.");
+    }
+  } else {
+    if (header.unstable_storage_format_) {
+      throw std::runtime_error(
```
pyrito, line 374: Is there any reason why we are throwing runtime errors here instead of following the other pattern of returning a `Result<T>`?

pyrito (approved):
LGTM, just had a couple of questions/comments

libtsuba/src/RDGPartHeader.h
e-mcginnis: I don't think so. The tradition to unstable should be a one way latch. 
That said, thinking about it now, I should really gate usage of this API on the feature flag being enabled to catch when functions call this without checking the flag.

docs/contributing/rdg-storage-format.rst
e-mcginnis: someone would have to explain it to me first hehehe :)

pyrito: RDG: Resilient Distributed Graph, a tribute to Spark's Resilient Distributed Dataset (RDD).

libtsuba/src/RDGPartHeader.cpp
e-mcginnis: Great question. From what I understand the way the `nlohmann` `to_json` and `from_json` functions have to be void? I can't see this specified in any of the docs specifically, but every example uses void. 
I would love to be proven wrong here though.

docs/contributing/rdg-storage-format.rst
e-mcginnis: Thanks! Added

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,61 @@
+=========================
+Katana RDG Storage Format
+=========================
+
+Katana persists its graphs on disk in a format called `RDG: Resilient Distributed Graph`. This `RDG` is a directory filled with files that define the graph.
+The layout of an RDG is defined by its `storage_format_version`
+An example of an RDG can be found here
+https://github.com/KatanaGraph/test-datasets/tree/master/rdg_datasets/ldbc_003/storage_format_version_3
+
+Katana software uses the `storage_format_version` of a loaded RDG to determine what features are already persisted in the RDG.
```
witchel: Would this read better as the following?

Katana software uses the `storage_format_version` of a loaded RDG to determine some of the features it supports.

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,61 @@
...
+Unstable RDG Storage Format
+===========================
+
+The unstable RDG storage format is used while new features which add data structures to the RDG are developed. Support for reading/storing the unstable RDG storage format is controlled with the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable.
+
+The unstable RDG storage format lets developers work on and test persisting new feature data structures, without disturbing everyone else that relies on the RDG format to be stable.
+
+RDGs persisted by Katana software ran with `UnstableRDGStorageFormat` set have the `unstable_storage_format` flag set
+The `UnstableRDGStorageFormat` feature flag determines the capabilities of the software,
+while the `unstable_storage_format` flag determines the state of the RDG
```
witchel: Instead of "ran" I'd say "executed."

Watch out for passive voice.  I think you can make Katana or katana software as the subject for most of these cases.

I find this paragraph a little confusing.  I think you mean UnstableRDGStorageFormat is the feature and unstable_storage_format is the concrete version number, but it is murky.

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,61 @@
...
+
+If "UnstableRDGStorageFormat" is found in the envar "KATANA_ENABLE_EXPERIMENTAL":
+1) katana software will store RDGs
+   - as "latest_storage_format_version"
+   - with the "unstable_storage_format" flag set
```
witchel: I've read this 3 or 4 times now and I'm still not 100% what you mean.  Can we say with version=latest_storage_format_version or version=unstable_storage_format?

witchel: Oh, it is a Boolean flag in the part header?

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -316,12 +338,54 @@ katana::from_json(const json& j, katana::RDGPartHeader& header) {
...
   } else {
     header.storage_format_version_ = kPartitionStorageFormatVersion1;
   }
 
+  // load the unstable_storage_format flag if it is present in the RDG
+  // RDGs created before support for unstable_storage_format was added do not have this flag.
+  // If it is not present, the RDG is assumed to *NOT* be stored in an unstable_storage_format since the rdg was created before unstable formats were introduced
+  if (auto it = j.find(kUnstableStorageFormatFlagKey); it != j.end()) {
+    it->get_to(header.unstable_storage_format_);
+    // Ensure unstable_storage_format RDGs are not loaded when the feature flag is not set
```
witchel: So this should return an error?

libtsuba/test/storage-format-version/unstable-storage-format-version-flag-on.cpp
```diff
@@ -0,0 +1,89 @@
...
+/// the unstable RDG can vary wildly from the stable RDG at test time since features may
+/// be under development
+/// Because of this, it is possible that this test catches bugs in storage which are related
+/// to unstable features, in addition to bugs related to the unstable storage format
+/// feature flag itself
+///
+/// Tests the following while the feature flag is enabled:
+/// 1) loading a stable RDG
+/// 2) loading a stable RDG and storing it as unstable
+/// 3) loading an unstable RDG and storing it as unstable
```
witchel, line 26: Nice.

witchel (approved):
Looks good, thank you.

libtsuba/src/RDGPartHeader.cpp
tylershunt: This is right; our wrappers for `nlohmann`'s functions catch these exceptions and turn them into `Result<T>`'s

tylershunt (approved):
LGTM

libtsuba/src/RDGPartHeader.cpp
e-mcginnis: Apologies, that comment should have been on the if-else block below. Fixed.

docs/contributing/rdg-storage-format.rst
e-mcginnis: I've attempted to clarify the documentation. Let me know what you think.

docs/contributing/rdg-storage-format.rst
e-mcginnis: Thanks for calling out the passive voice, I'm trying to get better about writing clear concise documentation. Hopefully this is better now.

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,71 @@
...
+Katana software executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable supports loading/storing RDGs in the unstable RDG storage format.
+Katana software executed *without* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable *does not* support loading/storing RDGs in the unstable RDG storage format.
+
+Katana software uses the boolean flag `unstable_storage_format` in the RDGs part header to identify unstable format RDGs
+
+If Katana software is executed *with* the `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"` environment variable:
+#. katana software stores RDGs:
+ 
+   #. with `storage_format_version = latest_storage_format_version`
+   #. with `unstable_storage_format = true`
```
witchel, line 34: One more comment and then I'll let this go.

I figured stable version was 3 and if we can do unstable versions, it would be 4.  But it sounds like you want the unstable version to be 3 with an additional Boolean flag.  Why is that better?

docs/contributing/rdg-storage-format.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Early RDGs from before `storage_format_version` was introduced are assumed to be `storage_format_version=1`
+
+Rarely, `storage_format_versions` can be deprecated. When this occurs tooling will be provided to convert all RDGs stored in that `storage_format_version` to the next supported `storage_format_version`.
+
+Unstable RDG Storage Format
+===========================
+
+The unstable RDG storage format lets developers work on and test persisting new data structures without disturbing everyone else that relies on the RDG format to be stable.
+The unstable RDG storage format should be used while developing features for Katana software which alter, add, or remove, data structures in the RDG format.
```
witchel: Developers should use the unstable...

docs/contributing/rdg-storage-format.rst
e-mcginnis: By making it a flag, it is much more obvious to developers, the software, and tools, that this RDG is in an unstable format. Usually making things more obvious results in clearer code.

So the idea is that we have formats marked by version 3, 3 + unstable, 4, 4 + unstable, etc

>I figured stable version was 3 and if we can do unstable versions, it would be 4

If we went with that suggestion, any RDGs written while 4 was an unstable storage format would be indistinguishable from RDGs written while 4 was stable at the json level, which would cause problems if the unstable format RDG doesn't match the stabilized format. Until something that expected the stable format fails dramatically, we would be unable to identify an RDG as unstable.
To get around this, we could say 4 was the unstable format and 5 was the stable format. But now we have to know that even versions are unstable and odd versions are stable or something like that. 

Hope that clarifies why adding a flag is much clearer than appending unstability to the existing storage_format_version.

docs/contributing/rdg-storage-format.rst
e-mcginnis: gave this another shot

docs/contributing/rdg-storage-format.rst
witchel: You make a good point.  The counterpoint is that any additional on-storage state is tough to manage, so even/odd versions are simpler.  But you've convinced me.

e-mcginnis: Completely agree, we don't want to be throwing things in willy-nilly. My first go at implementing this was actually using even/odd versions but it felt obtuse :shrug: 
fingers crossed I don't regret this decision in the future

libtsuba/src/RDGPartHeader.cpp
pyrito: Gotcha, thanks!

PR #794, opened 2022-01-11
Add utility class to collect errors on threads

roshandathathri:
Currently, if more than one thread has an error,
only one of those errors are reported. That is,
the function that combines/merges/collects
errors from different threads just picks one of them.

Add unit tests that also serve as usage examples.

@reviewers, `KATANA_THREAD_CHECKED*` macros are in libsupport 
even though libsupport has no parallelism/threads. They work because 
they are macros that are expected to be used only with libgalois 
(specifically, Reduction.h in libgalois). I'm not sure if there's a better 
place to put these macros.


arthurp (changes requested):
1. I think the macros should go in libgalois. There is no reason they have to be in libsupport and having them declared but not working is a recipe for confusion.
2. I feel strongly that `KATANA_THREAD_CHECKED` and friends needs to be called something else. The reason is that they function fundamentally differently from the `KATANA_CHECKED` in that they don't return the error. This means that if someone used them instead simply thinking that they are thread safe versions you would get more confusion.

Specific recommendation:

* Move everything including macros into `libgalois` in a new header file `ResultReduction.h` (or something like that). 
* Rename the macros to `KATANA_REDUCTION_CHECKED`.

I think this will make the changed semantics a lot clearer.

BTW, why are you throwing away all but one error? Wouldn't it be more useful to collect the errors and combine them? (Java has built in support for this in it's exception "causes" information. It's very useful for cases like this since all the exceptions can still be handled separately if you need to.)

roshandathathri:
> Specific recommendation:
> 
> * Move everything including macros into `libgalois` in a new header file `ResultReduction.h` (or something like that).
> * Rename the macros to `KATANA_REDUCTION_CHECKED`.
> 
> I think this will make the changed semantics a lot clearer.

Yes, thanks!

> BTW, why are you throwing away all but one error? Wouldn't it be more useful to collect the errors and combine them?

Hehe, left for future work. Seriously, I don't know how to combine errors, especially if the error codes are different.

arthurp:
> Seriously, I don't know how to combine errors, especially if the error codes are different.

It's almost as if there has been 30 years of research on this kind of thing in many different exception handling systems. ;-) I can think of 2-3 ways to do it when we get to that point.

libgalois/include/katana/ResultReduction.h
```diff
@@ -0,0 +1,86 @@
...
+              .WithContext(FMT_STRING("({}:{})"), __FILE__, __LINE__));        \
+      return;                                                                  \
+    }                                                                          \
+    std::move(                                                                 \
+        ::katana::internal::CheckedExpressionToValue(std::move(result_name))); \
+  })
+
+/// KATANA_REDUCTION_CHECKED_CONTEXT is similar to KATANA_CHECKED_CONTEXT
+/// except that instead of returning the error, it copies the error
+/// into error_info by calling error_info.update(<CopyableErrorInfo>)
```
arthurp: If you write `GReduceCopyableErrorInfo::update(CopyableErrorInfo)` here instead it will turn into a link in the documentation. Might be nicer that way.

arthurp (approved):
:-) Thanks! This look good to me. Thanks for the clear documentation on the macros.

libgalois/include/katana/ResultReduction.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include "katana/Result.h"
+
+namespace katana {
+
+namespace internal {
+
+// TODO(roshan) combine/merge two error infos
+struct AnyCopyableErrorInfo {
+  std::optional<CopyableErrorInfo> operator()(
```
amberhassaan, line 13: Hey, this isn't combining anything, unlike the comment's suggestion?

libgalois/test/reduce-error-info.cpp
```diff
@@ -0,0 +1,103 @@
...
+ErrorOnTrue(bool error, bool return_value) {
+  if (error) {
+    return KATANA_ERROR(katana::ErrorCode::NotFound, "error");
+  }
+  return return_value;
+}
+
+void
+TestVoidFunc() {
+  katana::GReduceCopyableErrorInfo error_info;
```
amberhassaan: I think we need better names. How about `katana::CombinedErrorInfo`? The fact that it's implemented via `GReducible` or even Reduction is an implementation detail and doesn't serve the API user. 

Similarly, `KATANA_REDUCTION_CHECKED` should be renamed to `KATANA_COMBINE_ERROR_INFO`. Feels clearer to me as to what to expect from the macro.

libgalois/test/reduce-error-info.cpp
```diff
@@ -0,0 +1,103 @@
...
+  katana::do_all(katana::iterate(size_t{0}, size_t{100}), [&](size_t i) {
+    KATANA_REDUCTION_CHECKED(error_info, ErrorOnTrue((i % 10) == 0));
+    KATANA_LOG_ASSERT((i % 10) != 0);
+  });
+  std::optional<katana::CopyableErrorInfo> reduced_error_info =
```
amberhassaan: I would hide the `reduce()` API behind something easier, e.g. a conversion operator `operator CopyableErrorInfo() {...}`.

amberhassaan (commented):
Suggested some changes and wrote down my complaints :).

libgalois/include/katana/ResultReduction.h
roshandathathri: "Any" (or "Or") is a reduction but it is NOT really combining two or more errors.

libgalois/test/reduce-error-info.cpp
roshandathathri: Sounds good to me.

libgalois/test/reduce-error-info.cpp
roshandathathri: It's not there in the test but the typical usage would be:
```cpp
  if (auto error = error_info.reduce()) {
    return *error;
  }
```
It would be great if there's a more concise way to write that code. I can change it to:
```cpp
  if (auto error = error_info()) {
    return *error;
  }
```
Is that what you're suggesting?

amberhassaan: Yes.

libgalois/test/reduce-error-info.cpp
amberhassaan: Actually, after having thought about it (and seeing your comment about ANY or OR), I've changed my thinking:

1) We should implement both ANY and ALL semantics. Currently, it seems you've implemented ANY, or first thread to register an error (in the thread ID order). In the future, we can support ALL, where we combine errors into a `bit vector` or  some such structure. 
2) We should rename the object to convey the intent of ANY or ALL. In this case, for ANY semantics, it should be `katana::GatherAnyError` or `GatherAtMostOneError`. For ALL case, it should be `katana::GatherAllErrors`. The macro can be similarly named to `KATANA_GATHER_ERROR_INFO`.

roshandathathri: Done. I was also thinking of renaming ResultReduction.h to CombinedErrorInfo.h but didn't do it to be more consistent with Result.h and Reduction.h

roshandathathri: I don't think there's a need for both ANY and ALL. We should switch from ANY to ALL in the future. Ideally, it would be ALL-DISTINCT. Having more options can be confusing. IMO the current `CombinedErrorInfo` should be updated later to combine all distinct error infos into one error info. Currently, ANY is sufficient so that we can at least detect and return some error.

roshandathathri:
Unless there are more comments/discussion, I'll be merging this once CI tests pass.

PR #795, opened 2022-01-11
Link result-bench with thread library

ddn0:
commit 74b7400d23b123f8c1639201387e42104e3dac40 (HEAD -> fix/threads, ddn0/fix/threads)
Author: Donald Nguyen <ddn0@users.noreply.github.com>
Date:   Mon Jan 10 23:34:43 2022 -0800

    Prefer strict channel priority

    A recent release of the readline library 8.1.2 in the conda channel
    pkgs/main highlighted an issue with how conda selects libraries. Under
    the default ("flexible") policy, newer versions are preferred regardless
    of which channel they come from.

    This can cause issues because the pkgs/main channel (the typically
    slower moving channel) uses a different compiler toolchain from
    the conda-forge channel (the faster moving channel that we want to take
    our libraries from).

    We were saved from this behavior because most conda-forge packages are
    newer than pkgs/main counterparts, but this is not guaranteed in any
    way.

    Instead, use the "strict" policy, which searches channels in priority
    order and does not move to a lower priority channel unless no satisfying
    version exists in a higher priority channel. This prefers conda-forge
    packages over pkgs/main packages (if a satisfying version exists in
    conda-forge).

    The strict policy will be the default in Conda 5.0.

    Unfortunately, strict priority cannot be applied throughout as some
    callers require the flexible policy to be satisfiable. Call out those
    exceptions explicitly and defer fixing callers to future work.

commit a22074828b1e307cb41e08890b02f1377e42a96a
Author: Donald Nguyen <ddn0@users.noreply.github.com>
Date:   Mon Jan 10 20:45:48 2022 -0800

    Make threads a transitive dependency of galois

    While pthread symbols do not get exposed by libgalois, whatever thread
    library is used by libgalois (via the STL) can be observed by callers.
    For example, katana::on_each executes a task on each thread, and it is
    reasonable to assume that the thread primitives callers use in on_each
    use the same thread library as libgalois itself rather than having
    different versions of threads.

    Change the CMake target_link_library to PUBLIC to propagate the thread
    library rather than keeping it PRIVATE to libgalois.

    Along the way, add missing dependencies to threads in result-bench.
    Previously, this was resolved via its transitive dependencies, through
    common packages of benchmark (?) but that is no longer the case.
    Instead, link directly.


arthurp (commented):
Please include the build docs changes in the commit message. Also make sure to update the conda uses in CI and other scripts to use `channel_priority`.

Actually should the conda changes be a separate PR?

amberhassaan (approved):
Looks good.

ddn0:
> Actually should the conda changes be a separate PR?

I can if you want but that makes it hard to actually test these changes in the enterprise repo.

arthurp:
> > Actually should the conda changes be a separate PR?
> 
> I can if you want but that makes it hard to actually test these changes in the enterprise repo.

I really was asking "should you". The answer clearly is no. :-) Carry on.

arthurp (approved):

ddn0:
@arthurp re-requesting a review as I had to make some more changes to get everything to pass CI. See `--no-channel-priority` here and related changes in https://github.com/KatanaGraph/katana-enterprise/pull/2913

PR #796, opened 2022-01-11
enable FixedSizeBinary in ScalarsToArray [KAT-2881]

danielmawhirter:
Enable properties of FixedSizeBinary type
Currently updating a FixedSizeBinary property fails with type not implemented
Allow building and updating FixedSizeBinary

### DIFF
--- a/libsupport/src/ArrowVisitor.cpp
+++ b/libsupport/src/ArrowVisitor.cpp
@@ -148,7 +148,8 @@ struct ToArrayVisitor : public katana::ArrowVisitor {
 
   using ResultType = katana::Result<std::shared_ptr<arrow::Array>>;
 
-  using AcceptTypes = std::tuple<katana::AcceptAllArrowTypes>;
+  using AcceptTypes = std::tuple<katana::tuple_cat_t<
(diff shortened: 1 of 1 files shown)


libsupport/src/ArrowVisitor.cpp
```diff
@@ -148,7 +148,8 @@ struct ToArrayVisitor : public katana::ArrowVisitor {
 
   using ResultType = katana::Result<std::shared_ptr<arrow::Array>>;
 
-  using AcceptTypes = std::tuple<katana::AcceptAllArrowTypes>;
+  using AcceptTypes = std::tuple<katana::tuple_cat_t<
+      katana::AcceptAllArrowTypes, std::tuple<arrow::FixedSizeBinaryType>>>;
```
danielmawhirter, line 152: NOTE: arrow::FixedSizeBinary should be rolled into AcceptAllArrowTypes eventually, but that touches most visitors, so I'm dragging my feet on it

rob-isaac (approved):
Looks good. Just FYI I dropped the task of generating an RDG with FixedSizeBinary property via GraphUpdate because @l-hoang aparently got his hands on one already.

PR #797, opened 2022-01-11
Add simple version parsing to version script.

arthurp:
I don't know why I did this. But I did, so we might as well put it in the repo. It may be useful for scripts which operate on bug reports.


roshandathathri (approved):
LGTM although I don't have sufficient context

PR #803, opened 2022-01-12
Fix conda build: link pthread

srogatch:


ddn0:
See https://github.com/KatanaGraph/katana/pull/795 instead

srogatch:
> See #795 instead

Wow, you're not sleeping in the middle of your night because of the issue :)
Thanks for pointing out, I'm closing the PR then.

PR #804, opened 2022-01-13
Leiden: add steal option

uditagar:
This PR improves the leiden clustering time by 3X using `katana::steal` option in `Refine` function.


gurbinder533 (approved):
Good catch.

PR #805, opened 2022-01-13
[KAT-3406] Set CMAKE_CUDA_ARCHITECTURES if gpu component is requested

srogatch:
Don't require `CMAKE_CUDA_ARCHITECTURES` if `gpu` component is being built.


arthurp (approved):
The change set seems fine, but the PR title and content seem to be quite different. Please fix that to reduce confusion.

I'm guessing the component is already checked other places in the code too.

PR #806, opened 2022-01-13
Add GitHub action to publish documentation to a new branch [KAT-286]

JamesFCoffey:
The GitHub action takes the docs built during CI and places them in a separate branch called `documentation` that can be used for publishing the open source docs to GitHub pages.

The action is triggered on `push` and so the token should work this time despite originating from a fork.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -389,12 +389,20 @@ jobs:
 
       # Upload path is relative to work directory and "~" and "${{env.HOME}}" do
       # not seem to be expanded, so hard code HOME for now.
-      - name: Upload documentation (C++ and Python)
+      - name: Upload documentation artifact (C++ and Python)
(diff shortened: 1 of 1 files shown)


arthurp (approved):
Have you tested this? Can you test this?

LGTM otherwise. I love how simple it is.

JamesFCoffey:
> Have you tested this? Can you test this?
> 
> LGTM otherwise. I love how simple it is.

@arthurp Testing is what worried me. I tried it in nektos/act but it looks like only 18.04 is available in the large docker image.

Before I merge, what's the worst that could happen? I just send another PR to reverse it?

arthurp:
Yea. I might cause CI failures on master commits. But that's not actually a super big deal. Go ahead. Just keep an eye on it after it merges for a bit to make sure it works.
 
Thanks for doing this BTW.

PR #810, opened 2022-01-17
Enable move constructor for ProgressScope

roshandathathri:
... so that it can be returned in Result\<ProgressScope\>.

### DIFF
--- a/libsupport/include/katana/ProgressTracer.h
+++ b/libsupport/include/katana/ProgressTracer.h
@@ -184,9 +184,9 @@ class KATANA_EXPORT [[nodiscard]] ProgressScope {
 public:
   ~ProgressScope();
   ProgressScope(const ProgressScope&) = delete;
-  ProgressScope(ProgressScope&&) = delete;
+  ProgressScope(ProgressScope&&) = default;
   ProgressScope& operator=(const ProgressScope&) = delete;
-  ProgressScope& operator=(ProgressScope&&) = delete;
+  ProgressScope& operator=(ProgressScope&&) = default;
 
   /// Get the scope’s underlying span to add tagging/logging,
   /// span relationships, and handle multiple active spans at a time


patrickkenney9801:
Could you please give an example of how you plan to use/see this being used Roshan?

roshandathathri:
I need this: `Result<ProgressScope> StartDistributedTrace()`  because any distributed communication can return an error. That'll be a sprawling enterprise PR. I'll include you as reviewer.

witchel (commented):
I am not the man to sign off on this change.

patrickkenney9801:
If possible I would like to avoid adding a move constructor to scope since it allows for a lot of antipatterns and it's unmovable for that reason
Maybe we could change the behavior of StartDistributedTrace instead if that's the sole motivator?

roshandathathri:
> it allows for a lot of antipatterns

What are examples of these? Can users of tracing do this unintentionally?

> Maybe we could change the behavior of StartDistributedTrace instead if that's the sole motivator?

That's not feasible. It involves communication and any communication can return errors, so we need to handle it cleanly.

patrickkenney9801:
Antipatterns would include:
Creating a new scope in a child function, returning the scope to the parent caller, and then RAII exits at the end of the parent function instead of the child function making results misleading and hard for devs to diagnose what happened

Creating a new scope in a parent function, passing it to a child function, and then RAII exits at the end of the child function instead of the parent function as expected making results misleading and hard for devs to detect what actually happened

patrickkenney9801:
I am suggesting that we could handle communication errors cleanly inside the function call
Worst comes to worse, if communication fails we could retry or just create placeholder scopes to return not using communication

roshandathathri:
> I am suggesting that we could handle communication errors cleanly inside the function call
> Worst comes to worse, if communication fails we could retry or just create placeholder scopes to return not using communication

Please create an enterprise PR for this. Assume any call to `NetworkInterface` can return an error. Then we can evaluate options. 

We should NOT be asserting errors. We can default to some sane behavior but if we do not return an error to the user, I don't see a way to ensure that the execution is sane. 

Another option is to change the signature to something like this:
`Result<void> StartDistributedTrace(ProgressScope*)`

patrickkenney9801:
I'll do that, will try to get it out tomorrow
My conda environment has been updating since noon...

roshandathathri:
Thanks @patrickkenney9801

PR #811, opened 2022-01-18
meta: slightly improve several interfaces

tylershunt:
Odds and ends to support some improvements in the pipeline:
 * PropertyGraph: add interface to load from a file handle. This is
   useful when we're loading distributed graphs because we already do
   a bunch of this work and it's awkward to get back to a URI

 * RDGLoadOptions: add a static `Defaults` constructor. While I think
   just defining structs that behave sanely when default is good. I
   like the way the verbose "Defaults()" call looks, especially when
   part of a default argument. I also thinks it gives callers greater
   confidence that the default struct will do something reasonable.

 * Experimental: My linter complains that this pointer is not constant
   when it's declared; and there is no reason it shouldn't be


scober (approved):
LGTM!

PR #814, opened 2022-01-18
Feature/fix ci

ddn0:


test/install_conda_packages_test.Dockerfile
```diff
@@ -31,6 +31,8 @@ RUN set -eu; \
     . /activate_miniconda.sh; \
     mamba update --quiet --yes --all; \
     ${CONDA_CLEAN}
+# The above forces python to be the version we need.
+# TODO(amp): Remove the explicit version.
```
arthurp: I think you removed the change but kept the comment. or something. The `python=3.8` line above is gone.

arthurp: Wait. Never mind. *I* did that. This comment should be in ` .github/workflows/activate_miniconda.sh`.

ddn0: I manually patched things because I force pushed your changes away in the other PR, so there might be some transcription errors

PR #815, opened 2022-01-18
tracer: clearly mark tracer text output [KAT-1849]

patrickkenney9801:
The hint for disabling the tracer will print once at the start of the first operation

### DIFF
--- a/libsupport/src/TextTracer.cpp
+++ b/libsupport/src/TextTracer.cpp
@@ -57,7 +57,8 @@ BuildText(
 
   fmt::format_to(
       std::back_inserter(buf),
-      "INFO: host={} ms={} max_mem_gb={:.3f} mem_gb={:.3f} arrow_mem_gb={:.3f}",
+      "TRACE: host={} ms={} max_mem_gb={:.3f} mem_gb={:.3f} "
+      "arrow_mem_gb={:.3f}",
       host_id, msec_since_begin,
       katana::ProgressTracer::GetMaxMem() / 1024.0 / 1024.0,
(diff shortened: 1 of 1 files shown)


libsupport/src/TextTracer.cpp
```diff
@@ -55,9 +55,12 @@ BuildText(
   static auto begin = katana::Now();
   auto msec_since_begin = katana::UsSince(begin) / 1000;
 
+  if (!host_data.empty()) {
+    fmt::format_to(std::back_inserter(buf), "To disable TRACE output use flag --disable-trace\n");
```
roshandathathri: Use `KATANA_WARN_ONCE`?

roshandathathri (approved):
LGTM

libsupport/src/TextTracer.cpp
patrickkenney9801: That works a lot nicer thank you

libsupport/src/TextTracer.cpp
```diff
@@ -55,9 +55,13 @@ BuildText(
   static auto begin = katana::Now();
   auto msec_since_begin = katana::UsSince(begin) / 1000;
 
+  if (!host_data.empty()) {
+    KATANA_WARN_ONCE("To disable TRACE output use flag --disable-trace\n");
```
ddn0: Would it make more sense to put this warning closer to the code that defines `--disable-trace`?

patrickkenney9801: that makes sense, I'll move it next to where the flag is handled in the enterprise repo

ddn0 (approved):

PR #816, opened 2022-01-19
Better Cython integration

arthurp:


libkatana_python_native/include/katana/python/CythonIntegration.h
```diff
@@ -8,12 +8,27 @@
...
+/// - a static method _make_from_address which takes T* as
+///   an int and an owner object, and returns a new Python object wrapping it.
+///   Does not take ownership, but will attempt to keep the owner alive as long
+///   as the wrapper is alive.
+/// - a static method _make_from_address_own which takes T*
+///   as an int, and returns a new Python object wrapping it. Takes
+///   ownership.
+/// - a static method _make_from_address_shared which takes shared_ptr<T>*
+///   as an int, and returns a new Python object wrapping it. Creates a new
+///   shared_ptr internally.
```
pyrito: What happens to the ref count on the original shared_ptr that is passed in? Does that counter get incremented? What is the internal shared_ptr needed for?

libkatana_python_native/include/katana/python/CythonIntegration.h
```diff
@@ -8,12 +8,27 @@
...
+/// - a static method _make_from_address_unique which takes unique_ptr<T>*
+///   as an int, and returns a new Python object wrapping it. Moves out of
+///   the unique_ptr.
```
pyrito: Wdym by "moves out of the unique_ptr"? Does this mean we don't take ownership, rather make a copy?

libkatana_python_native/include/katana/python/CythonIntegration.h
arthurp: `shared_ptr` represents the concept of a reference counted pointer. If Cython holds a reference it will (if written even vaguely correctly) hold a reference count as well.

libkatana_python_native/include/katana/python/CythonIntegration.h
arthurp: I mean it takes ownership of the referenced object via `std::move`. Bad wording sorry.

libkatana_python_native/include/katana/python/CythonIntegration.h
arthurp: I'll remove the comment about the internal shared pointer. That's an implementation detail.

libkatana_python_native/include/katana/python/CythonIntegration.h
```diff
@@ -119,6 +134,7 @@ namespace katana {
 // Add Cython classes here as needed. Remove them when they are moved to
 // pybind11.
 CYTHON_REFERENCE_SUPPORT(katana::PropertyGraph, "katana.local", "Graph");
+CYTHON_REFERENCE_SUPPORT(katana::TxnContext, "katana.local", "TxnContext");
```
pyrito, line 137: Can I use this for the distributed case or would that need to be wrapped again?

python/katana/local/_graph.pyx
```diff
@@ -592,3 +602,12 @@ cdef class Graph(GraphBase):
...
         raise NotImplementedError()
+
+
+cdef class TxnContext:
+    @property
+    def __katana_address__(self):
+        """
+        Internal.
+        """
+        return <uint64_t>&self._txn_ctx
```
pyrito, line 613: Why was this needed?

pyrito (approved):
LGTM - everything looks sane to me

libkatana_python_native/include/katana/python/CythonIntegration.h
arthurp: You can use this.

python/katana/local/_graph.pyx
arthurp: To enable the cython integration above.

pyrito: Is this all that's needed?

arthurp: For `CythonReference` yes. If you need `MakeCythonWrapper` you also need to implement at least one of the `_make_from_address` methods.

PR #820, opened 2022-01-19
Promote graph projection to a subclass of PropertyGraph

yan-zaretskiy:
This PR removes graph projections from the topology cache of a `PropertyGraph` and makes it a subclass of `PropertyGraph` instead. This way one can compose projections with other graph views (transposed, undirected, etc).

Note that projections share the RDG instance with their parent property graph. At this point, it is NOT safe to call RDG related methods on the projection, because the projected topologies will conflict with unprojected ones. This will be addressed in the subsequent PR.


libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+class KATANA_EXPORT TransformationView final : public PropertyGraph {
+public:
+  TransformationView() = default;
+  TransformationView(TransformationView&& other) = default;
+
+  virtual ~TransformationView();
+
+  /// Make a projected graph from a property graph. Shares state with
+  /// the original graph.
+  static std::unique_ptr<TransformationView> MakeProjectedGraph(
```
uditagar: rename this to `MakeTransformationView`

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+      : PropertyGraph(pg, std::move(projected_topo)),
+        transformation_(std::move(transformation)) {}
+
+  /// this function creates an empty projection with num_new_nodes nodes
+  static std::unique_ptr<TransformationView> CreateEmptyEdgeProjectedTopology(
+      const PropertyGraph& pg, uint32_t num_new_nodes,
+      const DynamicBitset& bitset);
+
+  /// this function creates an empty projection
+  static std::unique_ptr<TransformationView> CreateEmptyProjectedTopology(
```
uditagar, line 103: `CreateEmptyTransformationView`

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+  TransformationView(
+      const PropertyGraph& pg, GraphTopology&& projected_topo,
+      Transformation&& transformation) noexcept
+      : PropertyGraph(pg, std::move(projected_topo)),
+        transformation_(std::move(transformation)) {}
+
+  /// this function creates an empty projection with num_new_nodes nodes
+  static std::unique_ptr<TransformationView> CreateEmptyEdgeProjectedTopology(
```
uditagar, line 98: `CreateEmptyEdgeTransformationView`

uditagar:
Why not change the name of the class from `TransformedView` to `TransformedPropertyGraph`. Right now, it's not even clear from the name that it's a subclass of `PropertyGraph`.

libgraph/test/projection.cpp
```diff
@@ -23,15 +23,12 @@ static cll::opt<std::string> nodeTypes(
...
-using ProjectedPropertyGraphView = katana::PropertyGraphViews::ProjectedGraph;
-
 struct TempNodeProp : public katana::PODProperty<uint64_t> {};
 using NodeData = std::tuple<TempNodeProp>;
 using EdgeData = std::tuple<>;
 
-using ProjectedGraphView = katana::TypedPropertyGraphView<
-    ProjectedPropertyGraphView, NodeData, EdgeData>;
-using GNode = typename ProjectedGraphView::Node;
+using ProjectedGraph = katana::TypedPropertyGraph<NodeData, EdgeData>;
```
uditagar: this doesn't make sense

libgraph/include/katana/PropertyGraph.h
```diff
@@ -116,7 +126,28 @@ class KATANA_EXPORT PropertyGraph {
 
   friend class PropertyGraphRetractor;
 
+protected:
+  RDG& rdg() { return *rdg_; }
+  const RDG& rdg() const { return *rdg_; }
+
+  // This constructor is meant to be used by transformation views to share
+  // state with the original graph.
+  PropertyGraph(const PropertyGraph& parent, GraphTopology&& topo) noexcept
```
arthurp: I think this method actually make sense in general, not just for projections. It's just a way to share property data while using a new topology. Please document it as such. I definitely agree it should be protected. If there are additional restrictions (not allowing reordering nodes or whatever), please document those too.

libgraph/include/katana/TransformationView.h
arthurp: I disagree. This method creates a TransformationView, but it's not the only kind that could be created. A transformation view isn't exclusively for projections.  There could be others, so this should be named to match the kind of graph we are creating.

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+
+  /// Return the number of nodes of the original property graph.
+  uint64_t NumOriginalNodes() const final {
+    return transformation_.original_to_transformed_nodes_.size();
+  }
+
+  /// Return the number of edges of the original property graph.
+  uint64_t NumOriginalEdges() const final {
+    return transformation_.original_to_transformed_edges_.size();
+  }
```
arthurp: I think it would be better to remove this and instead provide a method to get the original graph and then the user can query it directly. That way there isn't an explosion of "original" methods which exactly mirror the "current graph" methods.

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+        arrow::BitUtil::BytesForBits(length));
+  }
+
+  /// Bitmask of edges included in the transformation view. Should be used to construct arrow tables.
+  std::shared_ptr<arrow::Buffer> EdgeBitmask() const noexcept {
+    auto length = transformation_.original_to_transformed_edges_.size();
+    return std::make_shared<arrow::Buffer>(
+        transformation_.edge_bitmask_data_.data(),
+        arrow::BitUtil::BytesForBits(length));
+  }
```
arthurp: Is there a way to return this that is typed? Is there a bit vector type in arrow? I don't like the idea of returning a raw untyped buffer here. If there is no way to make it safer, I understand.

libgraph/include/katana/TransformationView.h
arthurp: As above, this is not the only kind of view that could be created.

libgraph/include/katana/TransformationView.h
arthurp: As above, this is not the only kind of view that could be created.

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+
+private:
+  struct Transformation {
+    NUMAArray<Node> original_to_transformed_nodes_;
+    NUMAArray<Edge> original_to_transformed_edges_;
+    NUMAArray<uint8_t> node_bitmask_data_;
+    NUMAArray<uint8_t> edge_bitmask_data_;
+  };
```
arthurp: Why bundle this up?

libgraph/include/katana/TransformationView.h
arthurp: IT looks like you mostly have these methods as hooks so that the super class can get the right number for specific operations. If that's the only use, then just making these protected so no one outside the class uses them and expands on them is probably fine.

arthurp (commented):
This looks pretty good. I do have some comments and there are definitely things I don't understand.

Request another review from me if you want me to look at it again, or just merge it with someone else's approval once you have addressed my comments in one way or another.

libgraph/include/katana/TransformationView.h
yan-zaretskiy: No reason besides personal taste. I don't feel strongly about it, though.

libgraph/test/projection.cpp
yan-zaretskiy: I am sorry, @uditagar, could you elaborate? The reason for this change is that, since a projection is now a property graph, I can build an instance of `TypedPropertyGraph` on it directly, I don't need `TypedPropertyGraphView` anymore. Was this what prompted your comment?

uditagar: then I guess you can just use `Graph` instead of `ProjectedGraph`

libgraph/include/katana/TransformationView.h
uditagar: sure

libgraph/include/katana/TransformationView.h
arthurp: I feel like it creates a level of "I gotta look somewhere else to understand this" to a number of constructors. But I also don't have a strong opinion. C++ code is hard to follow no matter what you do, so things like this are not that important.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -861,6 +877,12 @@ class KATANA_EXPORT PropertyGraph {
   /// Return the number of local edges
   uint64_t NumEdges() const { return topology().NumEdges(); }
 
+  /// Return the number of nodes of the original property graph.
+  virtual uint64_t NumOriginalNodes() const { return NumNodes(); }
+
+  /// Return the number of edges of the original property graph.
+  virtual uint64_t NumOriginalEdges() const { return NumEdges(); }
+
```
rob-isaac: would be good to group all virtual methods together so it is easier not to forget one when writing a derived class

libgraph/include/katana/TransformationView.h
rob-isaac: I would slightly prefer making these private members of the `TransformationView`

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,105 @@
...
+        transformation_.edge_bitmask_data_.data(),
+        arrow::BitUtil::BytesForBits(length));
+  }
+
+private:
+  struct Transformation {
+    NUMAArray<Node> original_to_transformed_nodes_;
+    NUMAArray<Edge> original_to_transformed_edges_;
+    NUMAArray<uint8_t> node_bitmask_data_;
+    NUMAArray<uint8_t> edge_bitmask_data_;
```
rob-isaac: Bitmask data needs to be in PropertyGraph, otherwise we won't be able to do Transformations of Projections properly.

E.g. 
```
PropertyGraph proj_pg = TranformationView::MakeProjection(orig_pg);
/// will be a transpose on the full graph rather than the projection
PropertyGraph tansposed_pg = TransformationView::MakeTransposed(proj_pg);
```

rob-isaac (commented):
Partial Review, will continue on Monday

libgraph/include/katana/TransformationView.h
rob-isaac: To follow up on this, I think a naïve implementation of a bitmask of 1s for the un-transformed PG is probably fine for now, and we can optimize it out with either a compressed bitmap (preferred) or a std::optional<bitmap> in the future

yan-zaretskiy: I don't follow, I'm afraid. I'll ask you on Monday in person :-)

rob-isaac (approved):
This looks good to me. A unit test for Graph Projections would be nice if there isn't one already

yan-zaretskiy:
> This looks good to me. A unit test for Graph Projections would be nice if there isn't one already

There is one in `libgraph/test/projection.cpp`, which I tweaked for this PR. The real test is in `katana-enterprise`, for Louvain clustering.

libgraph/include/katana/TransformationView.h
yan-zaretskiy: My quick reading of the Arrow docs didn't reveal a way to do it. The Arrow API expects the null bitmap to be supplied as a raw bytes buffer anyhow, which is what we return here.

arthurp: Yuck. :-( I wish more C++ programmers knew how to do statically typed OO programming. It's amazingly rare for a language that is nominally both.

arthurp:
@bowu is waiting for Python level support for this, and I am waiting on this C++ PR before doing the Python support. So I'm tagging this high-priority to avoid a priority inversion. Please focus on this if possible.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -35,6 +36,18 @@ ProjectAsArrowArray(const T* buf, const size_t len) noexcept {
   return std::make_shared<ArrowArrayType>(len, arrow::Buffer::Wrap(buf, len));
 }
 
+template <typename Props>
+std::vector<std::string>
+DefaultPropertyNames() {
```
amberhassaan, line 41: minor: `MakeDefaultPropertyNames`

amberhassaan (approved):
Looks good.

libgraph/include/katana/GraphTopology.h
```diff
@@ -161,15 +169,9 @@ class KATANA_EXPORT GraphTopology : public GraphTopologyTypes {
...
-      const Edge& eid) const noexcept {
-    KATANA_LOG_DEBUG_ASSERT(eid < NumEdges());
-    return edge_prop_indices_[eid];
-  }
+  PropertyIndex GetEdgePropertyIndexFromOutEdge(const Edge& eid) const noexcept;
 
-  PropertyIndex GetNodePropertyIndex(const Node& nid) const noexcept {
-    return nid;
-  }
+  PropertyIndex GetNodePropertyIndex(const Node& nid) const noexcept;
```
roshandathathri, line 174: Why were these definitions made out-of-line?

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,117 @@
...
+
+#include "katana/PropertyGraph.h"
+
+namespace katana {
+
+/// A TransformationView is a topological transformation of the property graph.
+/// It is derived from the PropertyGraph so that it can be treated by query and
+/// analytics routines as a regular property graph.
+class KATANA_EXPORT TransformationView final : public PropertyGraph {
+public:
```
roshandathathri, line 15: redundant

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /// @param eid the input eid (must be original edge id)
+  Edge OriginalToTransformedEdgeId(const Edge& eid) const noexcept {
+    KATANA_LOG_DEBUG_ASSERT(eid < NumEdges());
+    return original_to_transformed_edges_[eid];
+  }
+
+  /// @param nid the input node id (must be original node id)
+  Node OriginalToTransformedNodeId(const Node& nid) const noexcept {
```
roshandathathri, line 77: nit: `NodeID` throughout - to be consistent with the rest of our codebase

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,117 @@
...
+        edge_bitmask_data_.data(), arrow::BitUtil::BytesForBits(length));
+  }
+
+  /// Return the number of nodes of the original property graph.
+  uint64_t NumOriginalNodes() const override {
+    return original_to_transformed_nodes_.size();
+  }
+
+  /// Return the number of edges of the original property graph.
+  uint64_t NumOriginalEdges() const override {
```
roshandathathri, line 41: Aren't all of these `final` instead of `override`?

libgraph/include/katana/TransformationView.h
```diff
@@ -0,0 +1,117 @@
...
+        edge_bitmask_data_(std::move(edge_bitmask_data)) {}
+
+  /// this function creates an empty projection with num_new_nodes nodes
+  static std::unique_ptr<TransformationView> CreateEmptyEdgeProjectedTopology(
+      const PropertyGraph& pg, uint32_t num_new_nodes,
+      const DynamicBitset& bitset);
+
+  /// this function creates an empty projection
+  static std::unique_ptr<TransformationView> CreateEmptyProjectedTopology(
+      const PropertyGraph& pg, const DynamicBitset& bitset);
```
roshandathathri, line 104: nit: Why aren't these `Make...` instead of `Create...`?

libgraph/src/PropertyGraph.cpp
```diff
@@ -746,19 +761,32 @@ katana::PropertyGraph::Write(
...
 }
 
+// We do this to avoid a virtual call, since this method is often on a hot path.
+katana::GraphTopology::PropertyIndex
+katana::PropertyGraph::GetEdgePropertyIndexFromOutEdge(
+    const Edge& eid) const noexcept {
+  return topology().GetEdgePropertyIndexFromOutEdge(eid);
+}
+
+// We do this to avoid a virtual call, since this method is often on a hot path.
```
roshandathathri, line 771: Do what? 

Also, why aren't these defined inline in the class?

roshandathathri (commented):
Too late to the party. LGTM. Some minor comments.

PR #822, opened 2022-01-21
Test to catch storage format changes KAT-2859

e-mcginnis:
Currently, changes to the on disk storage format do not actively require the developer to change the storage_format_version. The only thing keeping the definition of some defined storage_format_version from drifting over time is constant, perfect, developer vigilance. Obviously relying on this is not realistic, and changes have likely made it in without notice.
It would be nice to better define what on disk storage format goes with a specific storage_format_version, which would make mistakes less likely and easier to catch. But that is a long way off. For now a lame test will have to do. 

This is intended to pair nicely with the unstable_storage_format flag, and as an added bonus, prevents new features that developers intended to hide behind the unstable_storage_format flag from accidentally getting persisted in a stable storage format RDG. 

The test attempts to compare a known good RDG of the latest storage_format_version to the same RDG stored by the built katana software.
Our format is almost entirely deterministic, so this comparison goes smoothly enough. The randomness is how files in an RDG are named. The files have random strings appended to them, so the comparison works around this. 

Unfortunately, It is entirely possible for a change to make it past this test, as the test is only as good as the known good RDG that is uses. If modifications are made to some optional feature of the storage format which isn't present in the known good RDG that the test uses, the modification will not trigger this test to fail. The only way around this would be to generate a "maximal" RDG which contains every possible feature, but there is not currently an easy way to do that. And it would have to be regenerated with every new feature.

PS: I'd love to drop the random strings that are added to files in an RDG, and instead maybe append a hash of the data stored in the file or something else deterministic but unique at the same time? We are sooo close to being deterministic.


python/test/test_storage_format.py
```diff
@@ -0,0 +1,273 @@
...
+    1) the storage_format_version was changed but the rdg test datasets have not been updated
+    2) the on disk storage format was changed, but the storage_format_version was not
+    3) unstable storage format changes that are in use without the unstable storage format flag
+
+    TODO(emcginnis): this test would be best if we had some way to create and use a 'maximal' RDG, aka one with as many
+        optional features present as possible
+    """
+    orig_rdg = get_rdg_dataset("ldbc_003")
+    orig_graph = Graph(orig_rdg)
+    new_rdg = tempfile.mkdtemp()
```
pyrito, line 33: you should use a context manager when using a tempdir to be as explicit as possible

python/test/test_storage_format.py
```diff
@@ -0,0 +1,273 @@
...
+                    f"{orig_file} json key number does not match {other_file},"
+                    f"diff = {list_diff(orig.keys(), other.keys())}"
+                )
+                return False
+
+            # flatten the json files.
+            # this also remove the random strings appended to some of the json value strings
+            # to make comparing the two possible
+            orig_flat = flatten(orig, verbose)
+            other_flat = flatten(other, verbose)
```
pyrito: I was wondering about this but do you have to flatten them? Can you not check for equivalence using `==`, which I thought was recursive...but perhaps not?

python/test/test_storage_format.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    load up a known good rdg, store a copy
+    compare the storage format of the known good rdg to our stored copy
+    Attempts to catch the following situations:
+    1) the storage_format_version was changed but the rdg test datasets have not been updated
+    2) the on disk storage format was changed, but the storage_format_version was not
+    3) unstable storage format changes that are in use without the unstable storage format flag
+
+    TODO(emcginnis): this test would be best if we had some way to create and use a 'maximal' RDG, aka one with as many
+        optional features present as possible
```
pyrito: if only, if only

python/test/test_storage_format.py
```diff
@@ -0,0 +1,273 @@
...
+                if val not in other_flat:
+                    print(f"val = {val} not found in other json file")
+                    return False
+
+            if verbose:
+                print(f"json files {orig_file} and {other_file} match")
+            return True
+
+
+def validate_rdg_storage_format_subset(orig_path, other_path, verbose=False):
```
pyrito: `other_path` refers to newly created RDG correct? This name is a little confusing

python/test/test_storage_format.py
```diff
@@ -0,0 +1,273 @@
...
+        found = False
+        for other_file in other_path.glob(f"{orig_no_rand}*"):
+            if filecmp.cmp(orig_file, other_file, shallow=False):
+                if verbose:
+                    print(f"{orig_file} in {orig_path} is equal to {other_file} in {other_path}")
+                found = True
+
+        if not found:
+            print(f"Failed to find matching file for {orig_no_rand}-* in {other_path}")
+            return False
```
pyrito: So currently it looks like your test checks to see whether something in the original RDG exists in the new one, correct? It might be worth checking for files that are in the new RDG but not in the original RDG as a safe guard too. If the test can notify the user in some way I feel like that could still be useful feedback - esp for the same version, but maybe this is overkill.

pyrito (approved):
LGTM - just a few comments

python/test/test_storage_format.py
e-mcginnis: hm, you do have a point. 
I was originally flattening them as a way to do comparison before I was also sanitizing out the random file names. 
I could probably get away without flattening them as well now, but would still have to traverse the entire dict to sanitize out the random strings. Let me think on this.

python/test/test_storage_format.py
e-mcginnis: using the context manager mean the tmp dir goes away completely after this is done, which really makes debugging a pain. 
I will add a call to cleanup the temp dir at the end on a successful run though.

python/test/test_storage_format.py
e-mcginnis: The test checks the files in both directions :) 
```
def validate_rdg_storage_format_match(first_path, second_path, verbose=False):
    """
    prove the storage format of the two RDGs match by proving the first RDG is at
    least a subset of the second, then the second is at least a subset of the first
    """

```

pyrito: Oops! Missed this one

python/test/test_storage_format.py
e-mcginnis: I'll better document/name this

python/test/test_storage_format.py
e-mcginnis: honestly though I understand the confusion. I could make this clearer.

python/test/test_storage_format.py
e-mcginnis: so I've decided I like flattening them because it allows for explicitly calling out what is different:
```
            subset_flat = flatten(subset, verbose)
            superset_flat = flatten(superset, verbose)

            for val in subset_flat:
                if val not in superset_flat:
                    print(f"val = {val} not found in superset json file")
```
When the test fails, it will tell you exactly what value at what keys is different. 

We could say only flatten them when `==` fails so that we can print this out, but we still have to traverse the entire dict to sanitize all of the file names random strings so that wouldn't actually save us any time. 

I'll add a TODO to clean this up if/when we stop putting random strings in our file names.

pyrito: Sounds good - verbosity would help in situations like this. Unclear error messages are the worst.

python/test/test_storage_format.py
e-mcginnis: alright, overhauled the variable/function names and docstrings to hopefully better communicate what is actually going on here.

pyrito:
> Unfortunately, It is entirely possible for a change to make it past this test, as the test is only as good as the known good RDG that is uses. If modifications are made to some optional feature of the storage format which isn't present in the known good RDG that the test uses, the modification will not trigger this test to fail. The only way around this would be to generate a "maximal" RDG which contains every possible feature, but there is not currently an easy way to do that. And it would have to be regenerated with every new feature.

Doesn't the current test have some strict expectations on what is required in the storage format? You are comparing everything file by file, so I'd be surprised to how a change could make it past this test. Did you have a specific example?

e-mcginnis:
> > Unfortunately, It is entirely possible for a change to make it past this test, as the test is only as good as the known good RDG that is uses. If modifications are made to some optional feature of the storage format which isn't present in the known good RDG that the test uses, the modification will not trigger this test to fail. The only way around this would be to generate a "maximal" RDG which contains every possible feature, but there is not currently an easy way to do that. And it would have to be regenerated with every new feature.
> 
> Doesn't the current test have some strict expectations on what is required in the storage format? You are comparing everything file by file, so I'd be surprised to how a change could make it past this test. Did you have a specific example?

For example someone could change how an EdgeTypeAware optional topology is stored. 
Since:
1) the EdgeTypeAware topology is an optionally present structure
2) our base "known good" RDG does not have an EdgeTypeAware topology

a change to the way an EdgeTypeAware optional topology is stored would slip by this test. 
The short term solution would be to generate an RDG with all possible optional structures which are available now. That would work up until someone added a new optional structure. That new optional structure now provides a new home for un detectable changes to hide, unless the person that added the new optional structure also remembers to update the known good rdg. 

The more I talk about it, the more I think I'll do the short term solution. I was torn on it, since it create another test input to manage, but I think the extra test coverage is worth it.

pyrito:
Let's say that the new RDG produces the optional structure and writes it to the RDG. When you do the reverse check of seeing whether files in the new RDG are in the original, wouldn't an error get thrown since there is a mismatch in files?

e-mcginnis:
> Let's say that the new RDG produces the optional structure and writes it to the RDG. When you do the reverse check of seeing whether files in the new RDG are in the original, wouldn't an error get thrown since there is a mismatch in files?

If the new RDG for some reason produced the optional structure, then yes it would be caught. 
But what would cause the RDG to be created with the optional structure when the test created it? As of now, nothing would.

witchel (commented):
This looks good and I like the idea of a maximal RDG.

Because parquet files are compressed, we'd have to compress them and then hash them to name the file by content hash.  That strikes me as too inefficient to be worth it.  We could name them by the hash of their uncompressed value.

Or, for testing only, we could have a mode where we insert a counter value instead of the random string.

PR #823, opened 2022-01-21
IndexID: define other arithmetic operations

tylershunt:
I needed the `>>` operator for binary search. It occurred to me that
there might be other arithmetic operations needed in the future, and no
good reason not to provide them on this less safe OpaqueID cousin.

### DIFF
--- a/libsupport/include/katana/IndexID.h
+++ b/libsupport/include/katana/IndexID.h
@@ -7,7 +7,8 @@ namespace katana {
 
 /// A numeric ID type suitable for indexing. IndexIDs support:
 ///
-/// - Addition and subtraction of other IDs (the result of which is an ID)
+/// - Arithmetic operations with other IDs (the result of which is an ID)
+/// - Bitwise operations with other IDs (the result of which is an ID)
 /// - Increment and decrement
 /// - Dereference
 ///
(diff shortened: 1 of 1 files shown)


libsupport/include/katana/IndexID.h
```diff
@@ -51,26 +51,100 @@ struct IndexID : public OpaqueIDOrderedWithValue<_IDType, _Value> {
...
+  _IDType operator<<(_Value val) const { return _IDType(this->value() << val); }
+
+  _IDType operator&(const _IDType& rhs) const { return *this & rhs.value(); }
+  _IDType operator&(_Value val) const { return _IDType(this->value() & val); }
+
+  _IDType operator|(const _IDType& rhs) const { return *this | rhs.value(); }
+  _IDType operator|(_Value val) const { return _IDType(this->value() | val); }
+
+  _IDType operator^(const _IDType& rhs) const { return *this ^ rhs.value(); }
+  _IDType operator^(_Value val) const { return _IDType(this->value() ^ val); }
```
arthurp, line 148: I understand the need for the numeric operations, however I think the bit-level operations should be avoided. They are not operations on numbers. They are operations on bit vectors. C numeric types are both, but I don't see why we should follow that, it's a clear breach of encapsulation.

That said, I will not block the PR over this.

libsupport/include/katana/IndexID.h
```diff
@@ -7,7 +7,7 @@ namespace katana {
 
 /// A numeric ID type suitable for indexing. IndexIDs support:
 ///
-/// - Addition and subtraction of other IDs (the result of which is an ID)
+/// - Arithmetic operations with other IDs (the result of which is an ID)
```
arthurp, line 10: If you include them, you should say "bitwise operations". They are not arithmetic.

arthurp (approved):

libsupport/include/katana/IndexID.h
tylershunt: I explicitly need `>>` and `<<`  because that's how the stl implements `{lower,upper}_bound`. I don't feel strongly about the other operations; I just added them because it seemed weird to define some bit operations and not others.

arthurp: STL has manual strength reduction!?!?!?! OMG. That's really bad. Oh well. Include them. It's C++, nothing is logical. ;-)

libsupport/include/katana/IndexID.h
tylershunt: Sure thing; I only did this because [this c++ doc site](https://en.cppreference.com/w/cpp/language/operator_arithmetic) groups them together as "arithmetic". Happy to supercede that with whatever folks like though.

arthurp: Include both. Just for people like me who don't think about things the same as the C++ docs. It will not confuse anyone I don't think.

libsupport/include/katana/IndexID.h
tylershunt: > It's C++, nothing is logical. ;-)

But apparently everything is arithmetic!

arthurp: And yet they have no exponentiation operation. Sigh.

PR #824, opened 2022-01-21
python: Changed CythonReference type cast

pyrito:


arthurp:
Does this fix your problem?

pyrito:
> Does this fix your problem?

Running it in CI now so we will see.

pyrito:
@arthurp it worked! At least the builds are not failing

arthurp (approved):

PR #825, opened 2022-01-21
Memory managers and DistPropertyGraphs must have a storage location

witchel:
This PR consists of four changes across open and enterprise.

 o Introduce the memory manager and resource managers a la RFC-132, but only the property manager for now.
 o Add an explicit replacement policy to the cache, which only evicts on a Reclaim or GetAndEvict call.  With tests.
 o Introduce GraphLifetime which takes the lifetime logic from QueryClient.
 o Require all DistPropertyGraphs to have a storage location.

I tried to split these into 3 commits to facilitate review, but I think I forgot to add the GraphLifetime file until commit 3.

For the memory manager and managers, I am most interested in feedback on the overall design.  Is the singleton pattern appropriate?  Does the API look sensible?  Let's put off for now issues of policy.

The query folks are probably most interested (and relieved) about GraphLifetime.  I think this change is mostly uncontroversial, but suggestions for structure and API also welcome.  Note that while I use GraphLifetime in QueryClient, I think it could be used even more (and will in a future PR after QueryClient has seen more work).

Finally, I wanted to make it easier for whoever modifies QueryClient by only allowing graphs that have a valid storage location--no more in-memory graphs.  I thought this task might be a bit hairy, but it ended up being somewhat hirsute.  My approach might be described as "backyard dentistry."  I did some ugly things to make this work, like adding "./" as the location for many test graphs, and I added InformPath to PropertyGraphRetractor (while adding a TODO to eliminate this call from the public interface).  I'd appreciate it if people would more or less confine their comments to simple mods here, rather than suggest detailed changes to different tests.  My hope is that individual test maintainers might take on the responsibility of fine-tuning storage location information.

Jira: KAT-2846, KAT-2847, KAT-3454


libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+
+class KATANA_EXPORT MemoryManager {
+public:
+  MemoryManager(const MemoryManager&) = delete;
+  MemoryManager(MemoryManager&&) = delete;
+  MemoryManager& operator=(const MemoryManager&) = delete;
+  MemoryManager& operator=(MemoryManager&&) = delete;
+
+  // https://codereview.stackexchange.com/questions/173929/modern-c-singleton-template
+  static auto& MM() {
```
roshandathathri: Why not just `Get()`? I'll probably make `getSystemNetworkInterface()` a static member `Get()` of `NetworkInterface`

libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+    static MemoryManager mm_;
+    return mm_;
+  }
+
+  // Let the MM know about this manager.
```
roshandathathri: Doxygen `///` throughout

libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+  void Register(Manager* manager);
+
+  // This manager is defunct.
+  // \p manager must have zero active and standby memory.
+  void Unregister(Manager* manager);
+
+  // Request \p bytes from the MM for active memory
+  // Application cannot continue if it does not get memory
```
roshandathathri: Then shouldn't the return value be a bool if it was successful or not? Am I missing something? 

Looking at its use, it's assumed that the allocation was successful before calling this. If so, please document that. Request makes it seem like it is requested before allocation.

libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+
+namespace katana {
+// The memory manager singleton (MM).  Not thread safe.
+// The MM controls policy and does bookkeeping.  All memory allocation
+// is done by the system, mostly the C++ standard library.
+// The MM does not manage per-allocation tokens, it only manages sizes
+// Clients must call the proper functions or the MM will make
+// bad decisions.
+
+class KATANA_EXPORT MemoryManager {
```
roshandathathri: nit: names are a bit confusing. This is a manager of managers, right? Maybe this is a MemorySupervisor and the other class is a MemoryManager?

libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+  MemoryManager();
+  // Make sure our state is sane, log if not
+  void Sanity();
+  void LogState(const std::string& str);
+
+  // Get managers to return goal bytes of standby memory
+  void ReclaimMemory(count_t goal);
+  bool CheckRegistered(Manager* manager);
+
+  struct ManagerInfo {
```
roshandathathri: Why isn't this in the Manager class?

libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+  std::unordered_map<Manager*, ManagerInfo> managers_;
+
+  count_t active_{};
+  void ActiveMinus(ManagerInfo& info, count_t bytes);
+  void ActivePlus(ManagerInfo& info, count_t bytes);
+  count_t standby_{};
+  void StandbyMinus(ManagerInfo& info, count_t bytes);
+  void StandbyPlus(ManagerInfo& info, count_t bytes);
+  count_t Used() { return active_ + standby_; }
+  count_t physical_{};
```
roshandathathri: nit: move variables to the end of the class

libgalois/include/katana/MemoryManager.h
```diff
@@ -0,0 +1,86 @@
...
+    ManagerInfo() {}
+    count_t active{};
+    count_t standby{};
+  };
+  std::unordered_map<Manager*, ManagerInfo> managers_;
+
+  count_t active_{};
+  void ActiveMinus(ManagerInfo& info, count_t bytes);
+  void ActivePlus(ManagerInfo& info, count_t bytes);
+  count_t standby_{};
```
roshandathathri: Are these active and standby a sum of the managers' active and standby? If so, please doxygen comment.

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <cstdint>
+#include <string>
+
+#include "katana/Manager.h"
+#include "katana/MemoryManager.h"
+
+/// Managers tend to the memory consumption for specific, large resources, e.g.,
+/// properties or views.  They interact with the central MemoryManager (MM singleton)
+/// to coordinate memory use.
```
roshandathathri: nit: make this specific to PropertyManager

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,64 @@
...
+  ///   e.g., property for the property manager
+  virtual const std::string& MemoryCategory() { return "property"; }
+
+  /// Free standby memory, attempting to free \p goal bytes.
+  /// Returns the number of bytes freed, which can only be less than goal if the
+  /// manager's standby total is less than goal.
+  virtual count_t FreeStandbyMemory(count_t goal);
+
+  // void Insert(
+  //     const katana::Uri& cache_key, const std::shared_ptr<arrow::table>& props);
```
roshandathathri: Drop?

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,64 @@
...
+
+  void AddProperty(const katana::Uri& property_path) {
+    auto property = cache_->Get(property_path);
+    if (property.has_value()) {
+      auto bytes = katana::ApproxTableMemUse(property.value());
+      MM().StandbyToActive(this, bytes);
```
roshandathathri: Didn't understand how this is compiling. Shouldn't this be `MemoryManager::MM()...`?

libgalois/include/katana/Manager.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <cstdint>
+#include <string>
+
+/// Managers tend to the memory consumption for specific, large resources, e.g.,
+/// properties or views.  They interact with the central MemoryManager (MM singleton)
+/// to coordinate memory use.
+
+// Using a signed type to make underflow more apparent
+using count_t = int64_t;
```
roshandathathri, line 9: namespace katana?

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,64 @@
...
+#include <string>
+
+#include "katana/Manager.h"
+#include "katana/MemoryManager.h"
+
+/// Managers tend to the memory consumption for specific, large resources, e.g.,
+/// properties or views.  They interact with the central MemoryManager (MM singleton)
+/// to coordinate memory use.
+
+class PropertyManager : public Manager {
```
roshandathathri: namespace katana?

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,64 @@
...
+    auto property = cache_->Get(property_path);
+    if (property.has_value()) {
+      auto bytes = katana::ApproxTableMemUse(property.value());
+      MM().StandbyToActive(this, bytes);
+      property_cache_->Evict(property_path);
+    }
+  }
+
+  void PropertyLoadedCallback(const std::shared_ptr<arrow::table>& property) {
+    MM.BorrowActive(*this, property.size());
```
roshandathathri: Ditto

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,64 @@
...
+  //     const katana::Uri& cache_key, const std::shared_ptr<arrow::table>& props);
+
+  void AddProperty(const katana::Uri& property_path) {
+    auto property = cache_->Get(property_path);
+    if (property.has_value()) {
+      auto bytes = katana::ApproxTableMemUse(property.value());
+      MM().StandbyToActive(this, bytes);
+      property_cache_->Evict(property_path);
+    }
+  }
```
roshandathathri: Shouldn't this return a bool to let users know whether the property was loaded from the cache or not?

IIUC if AddProperty() doesn't succeed, then users load the property explicitly which calls PropertyLoadedCallback(). Is that so? Or is the property cache expected to be shared with the loader?

libgalois/test/chembl-queries.txt
```diff
@@ -0,0 +1,13 @@
+match (f:compound_properties)-[:has_molecule]-(b:molecule_dictionary{oral:1})<-[:has_molecule]-(d:drug_indication{mesh_id:'D006333'}), (b)<-[:has_molecule]-(l:activities)-[:has_assay]->(r:assays)-[:has_target]->(t:target_dictionary)<-[:has_target]-(c:target_components)-[:has_comp_seq]->(o:component_sequences) where l.pchembl_value > 5.0 and f.full_mwt >= 199.0 return count(b);
```
roshandathathri: Not the right place for queries. This should be in the enterprise repo.

roshandathathri (commented):
Comments only on the \*Memory\* headers. Some of the comments might be naive. Need another pass to internalize the design.

libgalois/include/katana/MemoryManager.h
witchel: I was being a bit cutesy.  I like `MM `as the name of the memory manager singleton, but `Get `is probably more uniform.

libgalois/include/katana/MemoryManager.h
witchel: All of these functions are bookkeeping, not allocation.

So yes, allocation already succeeded, as will `BorrowActive`, so no bool return.

witchel: I changed the documentation in this file to make this more clear.

libgalois/include/katana/MemoryManager.h
witchel: Great idea!  Done.

libgalois/include/katana/MemoryManager.h
witchel: This is what the memory supervisor thinks about all the individual managers.  Mostly it thinks about how many bytes they have standby and how many active.

libgalois/include/katana/MemoryManager.h
witchel: I interspersed variables with the operators (Plus/Minus) for them.

libgalois/include/katana/PropertyManager.h
witchel: Yes it should.  This code isn't active yet.

libgalois/include/katana/PropertyManager.h
witchel: You are right, but I'm not sure if it should return a bool or the `shared_ptr `to the property table.

It turns out that giving all DistPropertyGraphs a storage location took so long I never hooked up the new property cache manager.  But I'd like to do that on the next PR.

libgalois/include/katana/Manager.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+/// Managers track the memory consumption for specific, large resources, e.g.,
+/// properties or views.  They interact with the central MemorySupervisor (MS singleton)
+/// to coordinate memory use.  They do not allocate memory, they only track it.
```
roshandathathri: nit: move to before `class Manager`.

libgalois/include/katana/PropertyManager.h
```diff
@@ -0,0 +1,62 @@
...
+
+  count_t FreeStandbyMemory(count_t goal) {
+    count_t total = 0;
+    if (goal >= cache_.size()) {
+      total = cache_.size();
+      MM.ReturnStandby(total);
+      cache_.clear();
+    } else {
+      total = cache_->Reclaim(goal);
+      MM().ReturnStandby(total);
```
roshandathathri: If my point on callers of this function hold, then this call should be moved to the supervisor's calling code.

roshandathathri (approved):
LGTM. I'm excited for this.

Not necessarily in this PR but it would be good to clarify a usage workflow. IIUC users of both classes will only directly interact with the manager and the manager internally coordinates with its supervisor. If so, who is responsible for the registering the manager with the supervisor? That wasn't clear. Maybe the manager's constructor registers itself?

libgalois/include/katana/PropertyManager.h
witchel: I'm not sure what you mean.  The supervisor calls FreeStandbyMemory in a manager, and how much it asks for from each manager is one of its most important policy decisions.

roshandathathri: Meant to delete that comment. Please ignore it.

PR #826, opened 2022-01-22
remove redundant header from entity_type_id_array files KAT-2562

e-mcginnis:
also use this opportunity to:
- remove some left over uint8_t EntityTypeID support
- test out the unstable_storage_format tooling

KAT-2562

### DIFF
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -63,7 +63,7 @@ CheckTopology(
 /// favor of this method.
 katana::Result<katana::PropertyGraph::EntityTypeIDArray>
 MapEntityTypeIDsArray(
-    const katana::FileView& file_view, bool is_uint16_t_entity_type_ids) {
+    const katana::FileView& file_view) {
   const auto* data = file_view.ptr<katana::EntityTypeIDArrayHeader>();
   const auto header = data[0];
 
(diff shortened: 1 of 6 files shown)


witchel (approved):
It is pretty cool to see the unstable feature flags in action.  And this is a nice, simple test case.

There can only be one outstanding format change at a time, but maybe that is good discipline.

If we approve this, then the next PR changes the format to the next rev and converts old formats?  Is there someplace to centralize the logic for older formats?

e-mcginnis:
> There can only be one outstanding format change at a time, but maybe that is good discipline.

There can be multiple outstanding changes to the format at once, there isn't really anything preventing it. More changes could be tacked on to this and everything would work fine. As for handing the transition from unstable features to supported stable features, we probably don't want to get in the habit of only transitioning specific unstable features. Instead we should probably say all unstable features in the code at time X are now fully supported, stable features.

> If we approve this, then the next PR changes the format to the next rev and converts old formats?

yep, thats the idea. 

> Is there someplace to centralize the logic for older formats?

unfortunately, no. Mostly this is because we could do a better job of defining what it means to be some storage_format_version_X. Right now multiple functions across libtsuba and libgraph define what a storage_format_version_X looks like for the piece they care about, which will not scale well as we go forward. We need some central definition in libtsuba to rely on. I've been rolling around solutions in my head for this for a while, so watch out for an RFC.

PR #828, opened 2022-01-23
Fix tests for undirected and transposed views

gurbinder533:


arthurp (approved):

amberhassaan (commented):
I think the existing names are better because they're more descriptive and don't add too much to the length of the name. 
So I propose dropping this PR. Why do you want to make this change?

rob-isaac (approved):

gurbinder533:
> I think the existing names are better because they're more descriptive and don't add too much to the length of the name. So I propose dropping this PR. Why do you want to make this change?

Yes, we don't need to change. I think I did search and replace and these files got edited as well. I will close this. Thanks

PR #829, opened 2022-01-23
allow DeleteNodeIndex to succeed in case index is absent

chakpongchung:
For this PR

https://github.com/KatanaGraph/katana-enterprise/pull/2855

### DIFF
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -921,7 +921,12 @@ katana::PropertyGraph::DeleteNodeIndex(const std::string& column_name) {
       return katana::ResultSuccess();
     }
   }
-  return KATANA_ERROR(katana::ErrorCode::NotFound, "node index not found");
+
+  // TODO(Chak-Pong) make deleteNodeIndex always successful
+  //  before index existence check is available from python side
+  //  return KATANA_ERROR(katana::ErrorCode::NotFound, "node index not found");
+  KATANA_LOG_WARN("the following node index not found: {}", column_name);
+  return katana::ResultSuccess();
 }
 
 // Build an index over edges.


libgraph/src/PropertyGraph.cpp
```diff
@@ -964,6 +968,29 @@ katana::PropertyGraph::DeleteEdgeIndex(const std::string& column_name) {
...
+
+// todo(Chak-Pong) expose it to python side
+katana::Result<bool>
+katana::PropertyGraph::IsNodeIndexPresent(const std::string& column_name) {
+  for (auto it = node_indexes_.begin(); it != node_indexes_.end(); it++) {
+    if ((*it)->column_name() == column_name) {
+      node_indexes_.erase(it);
+      return true;
+    }
+  }
```
arekay: You can use the same idiom in both for consistency.

libgraph/src/PropertyGraph.cpp
```diff
@@ -964,6 +968,29 @@ katana::PropertyGraph::DeleteEdgeIndex(const std::string& column_name) {
   return KATANA_ERROR(katana::ErrorCode::NotFound, "edge index not found");
 }
 
+// todo(Chak-Pong) expose it to python side
```
arthurp: Which python side? You seemed to be talking about `katana.remote` on slack.

libgraph/src/PropertyGraph.cpp
```diff
@@ -921,7 +921,11 @@ katana::PropertyGraph::DeleteNodeIndex(const std::string& column_name) {
       return katana::ResultSuccess();
     }
   }
-  return KATANA_ERROR(katana::ErrorCode::NotFound, "node index not found");
+
+  // TODO(Chak-Pong) make deleteNodeIndex always successful
+  //  before index existence check is available from python side
+  //  return KATANA_ERROR(katana::ErrorCode::NotFound, "node index not found");
+  return katana::ResultSuccess();
```
arekay, line 929: I don't think it is a good idea to always return true as it loses semantics - why return anything at all? If the signature returns a `Result` it should have some information.
Secondly, i think deleting an index is an important enough operation to require acknowledgement that the index was indeed deleted. The alternate is to add a `HasIndex` call after deletion to verify if the index was deleted. This way a single call can be used to check if the index did exist and was indeed deleted.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -870,6 +870,10 @@ class KATANA_EXPORT PropertyGraph {
     return node_iterator(node_id);
   }
 
+  Result<bool> IsNodeIndexPresent(const std::string& column_name);
```
arekay: Isn't this a duplicate for https://github.com/KatanaGraph/katana/blob/36253df6ce8f334b3c4e617a132044eb966091c3/libgraph/include/katana/PropertyGraph.h#L902

arekay (commented):
Some minor comments.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -870,6 +870,10 @@ class KATANA_EXPORT PropertyGraph {
...
+
+  Result<bool> IsEdgeIndexPresent(const std::string& column_name);
```
arthurp: The first method already exists: `HasNodePropertyIndex`. The second does not, but should be called `HasEdgePropertyIndex` and should be implemented like the node version.

Please look more carefully before reimplementing things.

libgraph/src/PropertyGraph.cpp
```diff
@@ -964,6 +968,29 @@ katana::PropertyGraph::DeleteEdgeIndex(const std::string& column_name) {
...
+  }
+  return false;
+}
+
+// todo(Chak-Pong) expose it to python side
+katana::Result<bool>
+katana::PropertyGraph::IsNodeIndexPresent(const std::string& column_name) {
+  for (auto it = node_indexes_.begin(); it != node_indexes_.end(); it++) {
+    if ((*it)->column_name() == column_name) {
+      node_indexes_.erase(it);
```
arthurp: This code will be removed so this is just a comment for future coding you do. This will delete indexes in a method that by it's name should be read-only. This would have been caught if you had made these methods `const` (as they should be). Please be more careful.

arthurp (changes requested):

libgraph/src/PropertyGraph.cpp
chakpongchung: Yes, I agree except that we might not be able to perform the existence check along with the edit(create/delete index) within one single call because the current return type is `void`.

I think We either need to change the return type to be `bool` or expose the existence check to python.  If the former is preferred, I can do it in the follow-up PR. @arekay @arthurp

libgraph/src/PropertyGraph.cpp
chakpongchung: yes, the python side `katana.remote` used here:

https://github.com/KatanaGraph/katana-enterprise/blob/master/python/test/remote_graph_test.py

arekay (approved):
Please check that the create/delete entity/index calls are consistent in the return behavior (either both return void or bools)

libgraph/src/PropertyGraph.cpp
arthurp: If the goal is to switch this back soon, then this should definitely have  warning `KATANA_LOG_WARNING` when it does this.

arthurp (approved):
For consistency, do this for edge indexes too.

denish44 (approved):
LG

PR #799, opened 2022-01-11
allow nlohmann::ordered_json in JsonDump KAT-1157

danielmawhirter:
after #795 (unrelated CI issue)

allow nlohmann::ordered_json to use the same DumpJSON interface as nlohmann::json
required for the (nonstandard json) ordering of keys in query result according to return order

KAT-1157

PR #830, opened 2022-01-24
allow nlohmann::ordered_json in JsonDump KAT-1157 (#799)

danielmawhirter:
allow nlohmann::ordered_json to use the same DumpJSON interface as nlohmann::json
required for the (nonstandard json) ordering of keys in query result according to return order

KAT-1157


roshandathathri (approved):
Thanks!

PR #831, opened 2022-01-25
[KAT-1686] Update libtsuba to accept and return in-memory manifests; commit manifest with TxnContext->Commit()

yige-hu:
* Update libtsuba to accept and return in-memory manifest (open repo tests passed, waiting for enterprise repo tests before merge: PR #831 in katana, PR #3042 in katana-enterprise)

* Added a TxnContext.Commit() function, which is used by (1) single-operations committed as transactions (QueryOperation, AnalyticsOperation, ImportOperation, ImportFromMLOperation), (2) transaction commits and (3) in TxnContext() destructor when auto-commit is on (for tools and tests)

* Now all operation and tests all get delayed RDGManifest commit, this should make a smooth transition towards delaying RDGManifest commit to the Go server

* In the future, graph-worker will no longer commit the manifests for both transactions and single operations as transactions. Instead, `Operation::final()` will send the in-memory manifest to the Go server, which later will write back the manifest to storage.

Made up comment resolving for PR #844 .

Enterprise repo: [PR #3042](https://github.com/KatanaGraph/katana-enterprise/pull/3042)


libgraph/src/BuildGraph.cpp
```diff
@@ -1895,7 +1896,11 @@ katana::WritePropertyGraph(
         "edge prop: ({}) {}", field->type()->ToString(), field->name());
   }
 
-  auto result = prop_graph.Write(dir, "libkatana_graph");
+  // Since this function is only used by tools under
+  // external/katana/tools/graph-convert/, but not any other graph
```
ddn0: in this repo, it is just `tools/graph-convert/`

And if `graph-convert` is the only one who uses this function can we just pull out the write here and put it into `graph-convert`?

libtsuba/include/katana/TxnContext.h
```diff
@@ -61,6 +66,8 @@ class KATANA_EXPORT TxnContext {
 
   void SetTopologyWrite() { topology_write_ = true; }
 
+  void DelayManifestCommit() { commit_manifest_ = false; }
```
ddn0: Do we really need to expose this function to callers?

As I understand it, `commit_manifest_` will only be used during the transition from on-disk to in-memory manifests (and even then, I'm not sure if it really is necessary).

Afterwards, people who want to write their manifests (e.g., tests, tools) can just call "Commit(txn_ctx)" (or whatever we call it) themselves.

ddn0 (commented):
Looking good; some questions...

libgraph/src/BuildGraph.cpp
witchel: You should check with @patrickkenney9801 to see if these tools are still in use.

libgraph/src/BuildGraph.cpp
```diff
@@ -1895,7 +1896,11 @@ katana::WritePropertyGraph(
...
+  // operations that can potentially be involved in transactions,
+  // we commit the RDGManifest here.
+  auto result = prop_graph.Write(dir, "libkatana_graph", txn_ctx);
   if (!result) {
     return result.error().WithContext("writing to fs");
   }
```
witchel: I would write this as `auto result = KATANA_CHECKED(prop_graph.Write(dir, "libkatana_graph", txn_ctx));`
My rule of thumb is to only output a context message if I have data to print in the message.

libgraph/src/PropertyGraph.cpp
```diff
@@ -429,19 +430,21 @@ katana::PropertyGraph::DoWrite(
...
   katana::RDGManifest manifest = KATANA_CHECKED(katana::FindManifest(uri));
   katana::RDGHandle rdg_handle =
       KATANA_CHECKED(katana::Open(std::move(manifest), katana::kReadWrite));
   auto new_file = std::make_unique<katana::RDGFile>(rdg_handle);
 
-  if (auto res = DoWrite(*new_file, command_line, versioning_action); !res) {
+  if (auto res = DoWrite(*new_file, command_line, versioning_action, txn_ctx);
+      !res) {
     return res.error();
   }
```
witchel: This is fine, but it is an opportunity to modernize this code with `KATANA_CHECKED`.

libtsuba/include/katana/TxnContext.h
```diff
@@ -85,9 +92,11 @@ class KATANA_EXPORT TxnContext {
 
   bool GetTopologyWrite() const { return topology_write_; }
 
+  bool GetCommitManifest() const { return commit_manifest_; }
+
 private:
-  std::string ConcatRDGProperty(std::string rdg_dir, std::string prop) {
-    return rdg_dir + kPropSeparator + prop;
+  inline std::string ConcatRDGProperty(std::string rdg_dir, std::string prop) {
+    return Uri::JoinPath(rdg_dir, prop);
```
witchel: Why do you need an inline function that does the same thing as `JoinPath`?

libtsuba/include/katana/TxnContext.h
```diff
@@ -85,9 +92,11 @@ class KATANA_EXPORT TxnContext {
 
   bool GetTopologyWrite() const { return topology_write_; }
 
+  bool GetCommitManifest() const { return commit_manifest_; }
+
```
witchel: I tend to agree with @Amber that `Get `and `Set `are names of last resort.  I think `CommitManifest()` would be a better name.

libtsuba/include/katana/TxnContext.h
```diff
@@ -100,6 +109,8 @@ class KATANA_EXPORT TxnContext {
   bool all_properties_write_{false};
   bool topology_read_{false};
   bool topology_write_{false};
+
+  bool commit_manifest_{true};
```
witchel: What does this Boolean mean?  It seems like `Commit `or `CommitToStorage `would be a function on TxnContext that would commit in-memory state, like a manifest file.  But this could be an `early_commit `flag to write the manifest as soon as we know its name?

libtsuba/include/katana/TxnContext.h
```diff
@@ -85,9 +92,11 @@ class KATANA_EXPORT TxnContext {
...
+  bool GetCommitManifest() const { return commit_manifest_; }
+
 private:
-  std::string ConcatRDGProperty(std::string rdg_dir, std::string prop) {
-    return rdg_dir + kPropSeparator + prop;
+  inline std::string ConcatRDGProperty(std::string rdg_dir, std::string prop) {
+    return Uri::JoinPath(rdg_dir, prop);
   }
 
   static constexpr char kPropSeparator[] = "/propertyseparator/";
```
witchel: Just curious, why is this a character array?

witchel (approved):
This mostly looks like putting TxnContext where it needs to be, which is fine.

When writing code like this, I try to avoid changing behavior, but I do tend to update idioms like `KATANA_CHECKED`.

Nothing in this code seems to actually commit a manifest file that was written in a transaction, right?

libgraph/src/BuildGraph.cpp
patrickkenney9801: I don't think that tooling is still in use, I think @tylershunt exposed graphml importing from another API as well

libtsuba/include/katana/TxnContext.h
yige-hu: I thought it's more versatile if in the future we want to change e.g. the way to concatenate. I've change it to use `JoinPath` directly now.

libtsuba/include/katana/TxnContext.h
yige-hu: I see. I'll change to remove all the `Get`-s. But what did he suggest to deal with `Set`?

libtsuba/include/katana/TxnContext.h
yige-hu: I'm removing this since it's unused. It was a character array since `static` member variables with initialization on declaration only works with `char[]`. I learnt that it only works for `std::string` after c++17. I got a building error with our settings when I used `static constexpr std::string`.

libtsuba/include/katana/TxnContext.h
yige-hu: Do you mean you want `Commit` and `CommitToStorage` to be member functions of `TxnContext`? Right now they are member functions of `DistPropertyGraph`. Graphs are not cached in `TxnContext`. The boolean `commit_manifest_` is used to indicate a function call with a `TxnContext` passed in as parameter whether to commit the manifest to storage. For transaction related operations, such as transactions and single-operations as transactions, they define a `TxnContext` with this boolean set to be false. So that `libtsuba` does not commit the manifest to storage. Otherwise, `commit_manifest_` is by default true, meaning that all tests and tools not related to operations in the server/graph-worker context will commit the manifest to storage in `libtsuba`. 

In this sense, we can remove the setter function for this boolean, and only allow it to be defined in `TxnContext` constructors. Do you want me to remove the setter or just remain it?

libgraph/src/BuildGraph.cpp
witchel: I agree with this suggestion.

yige-hu:
> Nothing in this code seems to actually commit a manifest file that was written in a transaction, right?

@witchel the transaction commit operation right now still calls `DistPropertyGraph->Commit(TxnContext*)` to commit the manifest to storage. The graph is either cached by `QueryClient` (in `QueryOperation`, `ImportOperation` and `ImportFromMLOperation`), or cached by operation itself (`AnalyticsOperation`). But since I'm making changes in the previous commits to use `TxnContext` instead of the original `bool commit_manifest`, there is actually an issue/non-issue -- inside one transaction, the transaction operations (begin, commit and abort) as well as normal operations all share one single `TxnContext`. So right now, all of them actually commits the manifest to storage (since we set `TxnContext.commit_manifest_=true` because the server is not ready). In the future, when then server is ready, we'll set `TxnContext.commit_manifest_ = false`, so that all of them do not commit manifest to storage, and let the server do instead.

libtsuba/include/katana/TxnContext.h
witchel: @yige-hu, this is the same issue I'm getting at when I ask about the `commit_manifest_ `boolean.
I'd like to see a CommitToStorage function that is only called by testing code or utilities.  Maybe only `external/katana/tools/graph-convert/`, with a note that this API is deprecated. Or it sounds like we can just remove graph-convert without consequence. 

But the larger issue here is that every Write and Commit should be delayed until there is a commit on the transaction object (however you write it).  That commit will eventually migrate to the server, but needs to be in the code right now.

witchel:
> > Nothing in this code seems to actually commit a manifest file that was written in a transaction, right?
> 
> @witchel the transaction commit operation right now still calls `DistPropertyGraph->Commit(TxnContext*)` to commit the manifest to storage. The graph is either cached by `QueryClient` (in `QueryOperation`, `ImportOperation` and `ImportFromMLOperation`), or cached by operation itself (`AnalyticsOperation`). But since I'm making changes in the previous commits to use `TxnContext` instead of the original `bool commit_manifest`, there is actually an issue/non-issue -- inside one transaction, the transaction operations (begin, commit and abort) as well as normal operations all share one single `TxnContext`. So right now, all of them actually commits the manifest to storage (since we set `TxnContext.commit_manifest_=true` because the server is not ready). In the future, when then server is ready, we'll set `TxnContext.commit_manifest_ = false`, so that all of them do not commit manifest to storage, and let the server do instead.

This is not what I thought the plan was.

I thought the plan was to provide a legacy API for committing to storage, if necessary.

But the important part of your change is to test the idea that we can defer writing the manifest to commit time without breaking our system.  Eventually "commit time" will happen on the server.  But for now, you should add calls to commit the current transaction.  This way, `commit_manifest_` will always be false and will not be needed.

libtsuba/include/katana/TxnContext.h
yige-hu: @ddn0 @witchel  It's not only `external/katana/tools/graph-convert/`, but tons of other code in all sorts of `test` directories under different libraries (e.g. libgraph, libgluon, libquery) that are outside the scope of transactions. Because of all those things not related to operations or transactions (they all calls the functions which requires the `TxnContext` parameter, and we give them a local one created inside the test cases so that there's no segfault under the call stack) which still need to write back the manifest, I'm adding this `bool commit_manifest_`. If you don't like this, we can add something like `TxnContext->CommitToStorage` for every each of the local `TxnContext`-s for the tests/tools that are not related to transactions. But I think it's uglier than setting this boolean with default value true so that it is more transparent to people who don't care about transactions.

Also, it brings back the previous issue of why I ported `AnalyticsOperation` to use `QueryClient`. `TxnManager` maintains its own `QueryClient`. If operations (except for `AnalyticsOperation`, according to previous implementation done to operations) are inside a transaction, they use the `QueryClient` provided by that transaction and commit at the transaction commit. Otherwise, they use the `QueryClient` created by `graph-worker` and commit at the end of each operation. `QueryClient` caches the graph on which it operates. In this way, `TxnManager` can implement a `CommitToStorage` by committing the manifest of the graph cached by `QueryClient`. Otherwise, I'm going to add another pointer in the `TxnManager` which points to the graph that it operates on (due to the current implementation of katana, we only work on one graph every time) and is solely used by `AnalyticsOperation`. Then, we can commit that graph while committing the transaction, which makes it a wrapper as `TxnContext->CommitToStorage`.

libtsuba/include/katana/TxnContext.h
```diff
@@ -3,53 +3,58 @@
...
+#include "katana/URI.h"
 #include "katana/config.h"
 
 namespace katana {
 
 class KATANA_EXPORT TxnContext {
 public:
+  TxnContext() {}
+
+  TxnContext(bool commit_manifest) : commit_manifest_(commit_manifest) {}
```
ddn0: can you make this constructor `explicit`?

libtsuba/include/katana/TxnContext.h
yige-hu: Also, I don't think we eventually need a `TxnContext->Commit()` API, since ultimately the server is going to write back manifest for operation/transaction related operations. For the tests/tools not related to transactions, I feel it's clearer for the functions to write back manifest as usual, rather than ask the tests/tools to call a commit for each local `TxnContext`. Committing manifest in `graph-worker` using any kind of API is like an intermediate step.

ddn0: >  It's not only external/katana/tools/graph-convert/, but tons of other code in all sorts of test directories under different libraries (e.g. libgraph, libgluon, libquery) that are outside the scope of transactions. 

My main worry is that the semantics and API of `TxnContext(true)` is subtly different than `TxnContext()`. The first writes to storage if any callee eventually calls `DoStore` while the other doesn't write anything unless you call "Commit(txn_ctx)" and that usually is "delayed" to some top level function.

This raises some questions:
- If one calls "Commit(txn_ctx)" with a context created by `TxnContext(true)` should we commit again? 
- If a caller calls `DelayManifestCommit()` should we commit the manifest so far?
- Do people using the `TxnContext(true)` path depend on commits happening at the time `DoStore` is called? And won't that be different than when we have `TxnContext(false)`? Yet in both cases we have a txn_ctx. How are clients supposed to reason about this difference? 

In my mind it sounds like we have two different ideas.

1. The TxnContext that we know and love and
2. An auto-commit policy for parts of the codebase that don't care too much about transactional execution but would like to call into the part of the codebase that does care

A possible implementation would be:

```cpp
template<typename C>
class CommitGuard { // Maybe call this AutoCommit?
  C ctx_;
public:
  CommitGuard() {}

  ~CommitGuard() {
    // Yes this is violent but AutoCommit is only for tests
    KATANA_LOG_ASSERT(ctx_.Commit());
  }

  C& operator() {
    return ctx_;
  }
};
```

and in tests and such one would just write
```
CommitGuard<TxnContext> txn_ctx;
```

instead of
```
TxnContext txn_ctx
```

wherever they first needed a transaction context like thing.

If you don't like the idea of separate classes, you could conflate auto commit with TxnContext itself and get something like the boolean constructor you have now and compared to

```
TxnContext(bool auto_commit)
~TxnContext() {
  if (auto_commit_) { ... }
}
```

what you have now is more transparent because the errors would be returned at `DoStore` rather than when the root TxnContext was disposed, though it still looks more magical than it needs to be. And seen from this perspective, an improvement over the current PR would be something like:

```
class TxnContext {
  friend DoStore
  bool auto_commit_;
}
```

because only DoStore really needs to know about the auto commit policy.

Even then, it is a bug for any test to create a context like

```
TxnContext txn_ctx;
```

They must instead write

```
TxnContext txn_ctx(true);
```

which seems more cryptic than

```
CommitGuard<TxnContext> txn_ctx;
```

witchel: I like the RIAA model of `CommitGuard<TxnContext> txn_ctx;`

I do like this naming better:  `AutoCommit<TxnContext> txn_ctx;`

@yige-hu, I'm not sure I understand your point about `QueryClient `and `TxnManager`, but `QueryClient `doesn't cache the graph anymore.  That is done by `GraphLifetime`.

libgraph/src/BuildGraph.cpp
yige-hu: @ddn0 @witchel 
The function `WritePropertyGraph` is used only by tools under `tools/graph-convert`. But there are lots of other functions or classes in `BuildGraph` used by other parts in the repo, like `libgraph`. There are also lots of structs and functions not used by any other code in the repo and probably need to be removed.

Since the above comment in code is out-of-dated after switching from using the boolean to `TxnContext`, I'll just remove the comment and leave the code in `BuildGraph.cpp`? Or do you still want me to move the function to somewhere like `tools/graph-convert/Transform.h`?

witchel: Let's focus on landing your PR.  We need to keep  `PropertyGraph::Write`,  so I agree you can just change to comment to say this Write is committing directly to storage because this is a non-transactional tool.

I guess my question is about `PropertyGraph::Write` and `PropertyGraph::Commit`.  Do these require a TxnContext?

libtsuba/include/katana/TxnContext.h
yige-hu: @ddn0 @witchel I can use the `AutoCommit<TxnContext> txn_ctx`. Then I need to cache something to commit in `TxnContext`. Do you want me to cache the `RDGManifest` itself and add/expose another function in `libtsuba`, `PropertyGraph` and `DistPropertyGraph` to write back the `RDGManifest`, or cache a pointer to the `DistPropertyGraph` and calls commit? Can I still assume every `graph-worker` instance works only on one graph, or do I need to keep track of a set of graphs in the future?

yige-hu: I think I'm caching the `RDGManifest` in `TxnContext` now. So that all the code remains in `libtsuba`.

yige-hu: Feel like we're going to miss this deadline, since I can finish this change as fast as tonight. Then there needs another round of review and changes.... Also, any form of `RDGManifest` committing from the `graph-worker` is eventually not needed, since transaction/operation-related manifests are going to be written back by the server in the near future, and the tests/tools not related to transactions right now still writes back manifest immediately (by default, `TxnContext()` has its `commit_manifest_` set to false, so they only need to define it as `TxnContext()`).

yige-hu: @ddn0 @witchel Feel like we're going to miss this deadline, since I can finish this change as fast as tonight. Then there needs another round of review and changes.... Also, any form of RDGManifest committing from the graph-worker is eventually not needed, since transaction/operation-related manifests are going to be written back by the server in the near future, and the tests/tools not related to transactions right now still writes back manifest immediately (by default, TxnContext() has its commit_manifest_ set to false, so they only need to define it as TxnContext()).

witchel: It makes sense to me that a TxnContext would have an RDGManifest.  That is important metadata to track for a transaction.

I feel like we are talking in circles a bit.  I don't like the `commit_manifest_` flag because of the reasons Donald outlines.  For example, what happens when you call commit on a TxnContext that has `commit_manifest_ == true`?

Yes, the commit will move to the server, but that is M6, not M5.  For now, having an explicit commit operation lets us have confidence that moving the commit operation to the server will work.

Please schedule a zoom call with me if you want to talk about this more.  I'm available tonight and so on.

libgraph/src/BuildGraph.cpp
yige-hu: These require `TxnContext` mainly for caching the `RDGManifest`. I'm going to write it so that when `Commit` is called, it does not writeback `RDGManifest`, but writes the manifest to `TxnContext` instead. (Before the current changes, it was for checking on `commit_manifest_`)

libtsuba/include/katana/TxnContext.h
yige-hu: I think I know your idea but can't finish changing today. It makes sense if you want the API in M5 before the server writes back. But is the server writing back manifest not a part of M5?

yige-hu: @ddn0 @witchel  I eventually used `bool auto_commit_` in `TxnContext`, and used `TxnContext`'s destructor for the auto commit. I tried the `AutoCommit<TxnContext>` but it doesn't seem to fit in our use cases.
```
template <typename C>
class AutoCommit {
  C ctx_;

public:
  AutoCommit() {}

  ~AutoCommit() {
    // Yes this is violent but AutoCommit is only for tests
    KATANA_LOG_ASSERT(ctx_.Commit());
  }

  operator C&() { return ctx_; }
};
```
With the above code and self-define type converter, it can only convert the following case
```
AutoCommit<TxnContext> a;
TxnContext b = a;
```
But the followed is not possible with C++ self-defined converter, because &a is not a member variable.
```
AutoCommit<TxnContext> a;
TxnContext *b = &a;  // error, not possible
```
  And we need the above case because we call functions in tools and tests as
```
TxnContext txn_ctx;
func(..., &txn_ctx);
```
with AutoCommit, suppose we can convert the pointers,
```
AutoCommit<TxnContext> txn_ctx;
func(..., &txn_ctx);  //error, not possible
```
which needs conversion from `AutoCommit*` to `TxnContext*` if we use AutoCommit with self-defined conversion. Otherwise, the semantics may be ugly and inconvenient for test/tool authors, such as
```
AutoCommit<TxnContext> txn_ctx;
func(..., &((TxnContext)txn_ctx);
```
Or 
```
AutoCommit<TxnContext>  txn_ctx;
func(..., &txn_ctx->GetContext());
```
Therefore, I'd prefer using the `TxnContext` destructor with `bool auto_commit_`.

witchel: I'm not hung up on the AutoCommit class, but couldn't you add a method to it like this?
```
TxnContext* Prt() {return &ctx_;}
```
Which would allow you to write this.
```
AutoCommit<TxnContext> txn_ctx;
func(..., txn_ctx.Ptr());
```
But the main point to me is that `DoStore `does not write anything to storage.  If you have some `TxnContext` objects that are auto-commit and the rest are explicitly committed by the programmer, that is good enough for me for this PR.

libtsuba/src/RDG.cpp
```diff
@@ -81,45 +81,29 @@ WriteProperties(
 }
 
 katana::Result<void>
-CommitRDG(
+ComputeManifest(
```
witchel: Can you call this `SetTnxContextManifest`?  The point to me isn't so much that it is being computed as it is being held in memory as the "result" for this transaction.

libtsuba/src/RDG.cpp
```diff
@@ -81,45 +81,29 @@ WriteProperties(
...
-            manifest_file.string(),
-            reinterpret_cast<const uint8_t*>(curr_s.data()), curr_s.size()),
-        "CommitRDG future failed {}", manifest_file);
-    return katana::ResultSuccess();
-  });
-  if (ret) {
-    handle.impl_->set_rdg_manifest(std::move(new_manifest));
-  }
-  return ret;
+  auto manifest_file = katana::RDGManifest::FileName(
```
witchel, line 106: This request is unrelated to your PR, but can you mark this with a TODO(witchel) that the file names generated by `RDGManifest::FileName` are not correct because they do not include the partitioning policy and the number of partitioned hosts.  I recently discovered this problem trying to debug `tsuba_gc`.  Unfortunately, our manifest naming logic is kind of screwed up right now.

libtsuba/src/TxnContext.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+katana::Result<void>
+katana::TxnContext::Commit() const {
+  if (manifest_file_.empty()) {
+    return katana::ResultSuccess();
+  }
+
+  katana::CommBackend* comm = katana::Comm();
+  comm->Barrier();
+
```
witchel: Why do we need this Barrier?

roshandathathri: Please avoid barriers like the plague in the code ;)

It's better to either use `DistTerminationDetection` or `CombineResultAndReset()`

libtsuba/include/katana/TxnContext.h
yige-hu: I can. I just feel it looks ugly and inconvenient for people who don't care about transactions (,isn't it?). Similar to the above mentioned 
```
AutoCommit<TxnContext>  txn_ctx;
func(..., &txn_ctx->GetContext());
```

libtsuba/include/katana/TxnContext.h
```diff
@@ -3,53 +3,73 @@
...
 
 namespace katana {
 
 class KATANA_EXPORT TxnContext {
 public:
-  void InsertNodePropertyRead(std::string rdg_dir, std::string name) {
-    node_properties_read_.insert(ConcatRDGProperty(rdg_dir, name));
+  TxnContext() {}
+
+  explicit TxnContext(bool auto_commit) : auto_commit_(auto_commit) {}
```
ddn0, line 19: Could you add some documentation about this constructor, e.g.,
```
/// Create a transaction context that commits changes when the context is destroyed.
///
/// This is typically used when calling from transaction unaware code like tests.
```

and the one above it
```
/// Create a transaction context. Changes are only committed when Commit is called.
```

libtsuba/src/TxnContext.cpp
yige-hu: @roshandathathri looks like both of the functions are in enterprise repo and not exposed to the open repo... is there any other substitutes? I right now get a segfault in `katana::Comm()` while calling this `TxnContext.Commit()` in the `~TxnContext()` when the scope of the `TxnContext` is in `main`. If there's no better way, I'll just modify the tests to explicitly call a `TxnContext.Commit()` before the program exits.

yige-hu: The barrier code was copied from the old `CommitRDG` in `libtsuba/src/RDG.cpp`, for making sure the writer of OneHostOnly isn't writing a previous version of the file.

libtsuba/include/katana/TxnContext.h
yige-hu: Done. I set the `auto_commit_{true}` by default.

libtsuba/src/TxnContext.cpp
witchel: Y'know what?  There is a barrier inside `OneHostOnly`.  I bet it wasn't there when this Barrier was added.  @tylershunt might remember off hand.  But I think we can get rid of this `comm->Barrier()`;

The issue still remains that you don't want to call commit after the end of main.

yige-hu: I see... Looking like `OneHostOnly` already calls Comm()->Barrier() by itself. Thanks!

PR #833, opened 2022-01-25
Storage format version 4 KAT-2562

e-mcginnis:
release rdg storage_format_version 4 which contains the following changes:
- removes the redundant header from the node/edge_entity_type_id_arrays

This is the follow up to https://github.com/KatanaGraph/katana/pull/826 and transitions this feature from the unstable storage format to storage_format_version 4

KAT-2562

### DIFF
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -65,26 +65,28 @@ CheckTopology(
 katana::Result<katana::PropertyGraph::EntityTypeIDArray>
 MapEntityTypeIDsArray(
     const katana::FileView& file_view, size_t num_entries,
-    bool rdg_unstable_format) {
-  if (file_view.size() == 0) {
-    return katana::ErrorCode::InvalidArgument;
-  }
-
+    bool is_headerless_entity_type_id_array) {
   // allocate type IDs array
   katana::PropertyGraph::EntityTypeIDArray entity_type_id_array;
   entity_type_id_array.allocateInterleaved(num_entries);
 
   const katana::EntityTypeID* type_IDs_array = nullptr;
 
-  if (KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat) &&
-      rdg_unstable_format) {
+  if (is_headerless_entity_type_id_array) {
     type_IDs_array = file_view.ptr<katana::EntityTypeID>();
   } else {
+    // If we have header, the file_view should not be empty
+    if (file_view.size() == 0) {
+      return katana::ErrorCode::InvalidArgument;
+    }
+
     const auto* data = file_view.ptr<katana::EntityTypeIDArrayHeader>();
     type_IDs_array = reinterpret_cast<const katana::EntityTypeID*>(&data[1]);
   }
 
-  KATANA_LOG_DEBUG_ASSERT(type_IDs_array != nullptr);
+  if (num_entries != 0) {
+    KATANA_LOG_DEBUG_ASSERT(type_IDs_array != nullptr);
+  }
 
   katana::ParallelSTL::copy(
       &type_IDs_array[0], &type_IDs_array[num_entries],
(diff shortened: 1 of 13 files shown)


e-mcginnis:
blocking on https://github.com/KatanaGraph/katana-enterprise/pull/3047 passing as well

e-mcginnis:
includes change from https://github.com/KatanaGraph/katana/pull/834

libtsuba/include/katana/RDGPrefix.h
```diff
@@ -54,6 +54,9 @@ class KATANA_EXPORT RDGPrefix {
       const RDGManifest& manifest, uint32_t partition_id);
 };
 
+/// THIS HEADER WAS DEPRECATED AS OF storage_format_version = 4
+/// DO NOT MODIFY THIS HEADER
+/// This header is used only for backwards compatibility with older storage formats
 /// EntityTypeIDArrayHeader describes the header in the on disk representation
 /// of EntityTypeID arrays, it could be probably be rolled into RDGPrefix but it
 /// has slightly different uses so for now it is separate
```
pyrito, line 62: And so it begins...

libgraph/src/PropertyGraph.cpp
```diff
@@ -100,33 +99,13 @@ WriteEntityTypeIDsArray(
...
-  } else {
-    katana::EntityTypeIDArrayHeader data[1] = {
-        {.size = entity_type_id_array.size()}};
-    arrow::Status aro_sts =
-        ff->Write(&data, sizeof(katana::EntityTypeIDArrayHeader));
-
+  if (entity_type_id_array.size()) {
+    const katana::EntityTypeID* raw = entity_type_id_array.data();
+    auto buf = arrow::Buffer::Wrap(raw, entity_type_id_array.size());
+    arrow::Status aro_sts = ff->Write(buf);
```
pyrito, line 108: nit: is there any reason why buf is `auto` and the other types are defined explicitly?

pyrito (approved):
Very cool to see this all come together!

libgraph/src/PropertyGraph.cpp
e-mcginnis: I don't recall my reasoning originally, but in general when it comes to auto I prefer:
1) don't use auto, as I find it is likely to hide useful information from the programmer. I could figure out what type `entity_type_id_array.data()` using lsp or looking at the method, but reading the declared type is faster. 
2) to use auto only if the type is already clear. eg: `buf` is clearly an `arrow::Buffer`

libtsuba/include/katana/RDGPrefix.h
e-mcginnis: Maybe when we do a full release we can drop support for all older storage_format_versions and start fresh without our old sins to weigh us down. 
That would leave more room for our future sins. :D

libtsuba/src/RDGCore.cpp
```diff
@@ -335,7 +335,7 @@ katana::RDGCore::RemoveEdgeProperty(int i, katana::TxnContext* txn_ctx) {
 
 katana::Result<katana::NUMAArray<katana::EntityTypeID>>
 katana::RDGCore::node_entity_type_id_array(size_t begin, size_t end) const {
-  if (part_header().IsEntityTypeIDsOutsideProperties()) {
+  if (!part_header().IsUint16tEntityTypeIDs()) {
```
scober, line 337: This seems wrong to me. Why are you returning an empty array when the types are 8 bits? You are also trying to load the array when there is no independent type array and therefore no file.

e-mcginnis: since storage_format_version_2 support was dropped a while ago, uint8_t EntityTypeIDs are unsupported, so there is no need to check for them. 
The naming for these `IsBlah` methods is kinda bad imo, and that can only be blamed on me. I'll have to revisit them later. 

to clarify: 
`IsEntityTypeIDsOutsideProperties` basically just says we have a storage_format_version 2 or higher rdg
`IsUint16tEntityTypeIDs` basically just says we have a storage_format_version 3 or higher rdg, so it is a "stronger" check than `IsEntityTypeIDsOutsideProperties`

Because of all of that, only RDGs >= storage_format_version_3 will have EntityTypeIDs outside properties, so this logic checks out.

scober: Can we rename the functions then? Because `IsUint16tEntityTypeIDs()` is confusing to read for me. Also if we don't support 8 bit ids any more then `LoadAndMaybeConvert` should be renamed to `LoadIDArray` and it doesn't need to be templated.

e-mcginnis: yep, we definitely can. I'm not going to in this PR though, its gone on long enough :) 

I did rename and remove the templating from `LoadAndMaybeConvert()` though

scober: Nice, thanks

PR #834, opened 2022-01-25
always write node/edge_entity_type_id_array

e-mcginnis:
We don't actually have any lifetime tracking for the in memory
entity_type_id arrays, which means we don't actually know when the array
on disk is out of date and we should write. For now, just always write the file.
This is correct, but wasteful.

### DIFF
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -415,27 +415,15 @@ katana::PropertyGraph::DoWrite(
 
   KATANA_CHECKED(DoWriteTopologies());
 
-  if (!rdg_.node_entity_type_id_array_file_storage().Valid()) {
-    KATANA_LOG_DEBUG("node_entity_type_id_array file store invalid, writing");
-  }
-
+  //TODO(emcginnis): we don't actually have any lifetime tracking for the in memory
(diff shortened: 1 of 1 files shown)


amberhassaan (approved):
Looks good. Can you comment on what would be a good solution? e.g., can we solve this by providing a  "mutable" interface where you get the mutable type id array and that allows RDG to record the fact that type id arrays have changed?

e-mcginnis:
> Looks good. Can you comment on what would be a good solution? e.g., can we solve this by providing a "mutable" interface where you get the mutable type id array and that allows RDG to record the fact that type id arrays have changed?

That would be one option, another would be to track the state of all persisted objects like we track the loaded/unloaded/dirty/clean state of properties.

witchel:
> > Looks good. Can you comment on what would be a good solution? e.g., can we solve this by providing a "mutable" interface where you get the mutable type id array and that allows RDG to record the fact that type id arrays have changed?
> 
> That would be one option, another would be to track the state of all persisted objects like we track the loaded/unloaded/dirty/clean state of properties.

I love the state machine for properties.  When I try to do something stupid, the state machine is like, "you are stupid."  Then I still feel stupid, but also understood by the world.

PR #835, opened 2022-01-25
Remove PropertyGraphRetractor

rob-isaac:
Removes the PropertyGraphRetractor class, and enables protected access to some private members instead (`rdg_`, `node_entity_type_manager_`, `edge_entity_type_manager_`)


yan-zaretskiy (approved):
LGTM

e-mcginnis:
Did I miss the conversation where we decided PropertyGraphRetractor was a bad idea? I don't feel strongly one way or the other, but I'm curious why we are doing this?

libgraph/include/katana/PropertyGraph.h
```diff
@@ -239,11 +243,11 @@ class KATANA_EXPORT PropertyGraph {
       EntityTypeManager&& edge_type_manager) noexcept
       : rdg_(std::move(rdg)),
         file_(std::move(rdg_file)),
-        node_entity_type_manager_(std::move(node_type_manager)),
-        edge_entity_type_manager_(std::move(edge_type_manager)),
         node_entity_type_ids_(std::move(node_entity_type_ids)),
         edge_entity_type_ids_(std::move(edge_entity_type_ids)),
-        pg_view_cache_(std::move(topo)) {
+        pg_view_cache_(std::move(topo)),
```
e-mcginnis: Is this a style-only change, or accidentally included, or am I missing something?

libgraph/include/katana/PropertyGraph.h
```diff
@@ -114,7 +109,16 @@ class KATANA_EXPORT PropertyGraph {
 
   friend class PGViewCache;
 
-  friend class PropertyGraphRetractor;
+protected:
+  RDG& rdg() { return rdg_; }
+  const RDG& rdg() const { return rdg_; }
+  // TODO(Rob): make these private. Would be good to have protected accessor
+  //            but there is already a public const accessor, so a protected
+  //            non-const accessor throws compile errors
```
ddn0: Why does a non-const acessor throw compile errors?

`void Fn() const` and `void Fn()` are considered different functions: https://godbolt.org/z/vdqEKM3jn

ddn0 (commented):
I don't follow this code much, so maybe this change is better than the alternative, but generally giving access to fields (versus methods) is hard to maintain because it is harder to enforce class invariants and it is not obvious to subclasses how to maintain class hierarchy invariants without a clear interface.

libgraph/include/katana/PropertyGraph.h
roshandathathri: Agree with Donald. Can't we have non-const accessors for these variables instead?

libgraph/include/katana/PropertyGraph.h
```diff
@@ -346,6 +350,8 @@ class KATANA_EXPORT PropertyGraph {
 
   uint32_t partition_id() const { return rdg_.partition_id(); }
 
+  uint32_t partition_policy_id_() { return rdg_.part_metadata().policy_id_; }
```
roshandathathri: Drop suffix `_` in the function name. 

nit: Not in favor of having accessor names that differ from the variable names. Either change the variable name or add a TODO to do so later.

roshandathathri (changes requested):

libgraph/include/katana/PropertyGraph.h
rob-isaac: It doesn't have to do with inheritance. Here is what happens: https://godbolt.org/z/8xrnaoTox

libgraph/include/katana/PropertyGraph.h
rob-isaac: Didn't mean to have the `_` suffix. I think `partition_policy_id` or `PartitionPolicyID` is the right name though, what name do you propose?

libgraph/include/katana/PropertyGraph.h
rob-isaac: those member variables got moved below in the declaration to make them protected, and the initialization needs to match the declaration

libgraph/include/katana/PropertyGraph.h
roshandathathri: Yes, that's fine. I meant the variable name needs to be changed from `policy_id_` to `partition_policy_id_`

libgraph/include/katana/PropertyGraph.h
ddn0: I see. Having accessors with the same name at different protection levels is somewhat annoying.

You'd have to do something like: https://godbolt.org/z/eM91aoa6M to disambiguate these calls. Or you can simply give the mutating accessor a different name. It is protected anyway, subclasses should be ok with a different name for it.

libgraph/include/katana/PropertyGraph.h
rob-isaac: I think that `policy_id_` in the `PartitionMetadata` class pretty clearly refers to the partition policy id, so I am not sure I agree with the more verbose name change

libgraph/include/katana/PropertyGraph.h
rob-isaac: Yeah I don't think forcing people using the public API to const-cast when they want to call that method is a great solution haha. I will just make the protected method `mutable_<var>` instead of `<var>` so there is no ambiguity.

roshandathathri (approved):
LGTM. Thanks!

libgraph/include/katana/PropertyGraph.h
roshandathathri: Ok, this is fine.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -114,7 +114,21 @@ class KATANA_EXPORT PropertyGraph {
...
+protected:
+  RDG& rdg() { return rdg_; }
+  const RDG& rdg() const { return rdg_; }
+  EntityTypeManager& mutable_node_entity_type_manager() {
+    return node_entity_type_manager_;
+  }
+  EntityTypeManager& mutable_edge_entity_type_manager() {
+    return edge_entity_type_manager_;
+  }
+  EntityTypeIDArray& mutable_node_entity_type_ids() {
```
amberhassaan: FYI: I've been against exposing NUMAArrays by reference because there's a potential issue of who owns and frees the (memory for) the array. The alternative to this, which I prefer, is providing buffer like objects so that the contents of the array can be changed by calling code, but calling code can't mess with the memory or lifetime of the array.

amberhassaan (approved):
Looks good. Left a comment on NUMAArray accessors (as usual).

libgraph/include/katana/PropertyGraph.h
rob-isaac: Yeah I wish I didn't have to expose these members, but the manipulation of state by the partitioner requires it and changing partitioner logic is outside the scope of this PR. Will add a TODO though.

witchel:
> Did I miss the conversation where we decided PropertyGraphRetractor was a bad idea? I don't feel strongly one way or the other, but I'm curious why we are doing this?

I would also like to hear the case for this.  The code looks fine.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -98,7 +98,9 @@ class KATANA_EXPORT PropertyGraph {
   using EntityTypeIDArray = katana::NUMAArray<EntityTypeID>;
 
   PropertyGraph(PropertyGraph&& other) = default;
+  PropertyGraph& operator=(PropertyGraph&& other) = default;
 
+public:
```
rob-isaac, line 103: nit: looks like an extra `public` declaration snuck in during the rebase

arthurp: I will fix this next time I touch this code. I have a hunch I'll need to touch it a little in the transformed view wrapping coming soon.

arthurp:
> > Did I miss the conversation where we decided PropertyGraphRetractor was a bad idea? I don't feel strongly one way or the other, but I'm curious why we are doing this?
> 
> I would also like to hear the case for this. The code looks fine.

Retractors provide access which totally break OO encapsulation. So they are bad OO design definitionally. I think this is why it was just assumed. Now that DistPropertyGraph is a subclass it's not needed any more since the subclass can use protected methods. I know a few other places where using the retractor, but honestly, that's exactly why we need to remove it. To force people to use or expand the actual public API instead of breaking encapsulation.

PR #836, opened 2022-01-25
a test case for bfs correctness

wkyu2kg:
This PR adds a testcase (test_bfs_iter) that checks the correctness of BFS algorithm for an input graph tests/data/edge2.csv, compared to networkx.


arthurp (approved):

metagraph-plugin/tests/data/edge2.csv
```diff
@@ -0,0 +1,11 @@
+Source,Destination,Weight
```
gliga: Do you need a file or can you (ideally) this this as part of the code in an array?

metagraph-plugin/tests/test_algorithms.py
```diff
@@ -13,6 +13,13 @@ def test_bfs(networkx_weighted_directed_8_12, kg_from_nx_di_8_12):
     assert bfs2_kg.tolist() == [2, 4, 5, 6, 7]
 
 
+def test_bfs_iter(networkx_weighted_directed_bfs, kg_from_nx_di_bfs):
```
gliga, line 17: if this is failing then you should mark it with xfail.

gliga (approved):
Thanks for including tests.  See a couple of comments.

metagraph-plugin/tests/test_algorithms.py
wkyu2kg: Actually, this is the purpose of this test case. It would fail until the BFS implementation is fixed.

metagraph-plugin/tests/data/edge2.csv
wkyu2kg: Either way is fine. But I used "edge2.csv" for consistency with "edge1.csv" and other graphs in this code.

gliga: If either is fine, can we please build this in memory for now.

metagraph-plugin/tests/test_algorithms.py
gliga: Correct, but my suggestion is to include this test on our master.  The way you can include it is to mark it with xfail and then it will be running but not failing our CI.  Once this is fixed those that fix the issue can go ahead and enable this test (as otherwise they will not be able to have their CI green).

wkyu2kg: Sure. I committed the changes to address both of your comments.

gliga: Thanks.  Can you make sure to update your Jira ticket for this case and mention the unit test you included.

wkyu2kg: They are already linked. I updated the issue with the xfail label.

PR #838, opened 2022-01-26
RDGSlice: implement version indicators

scober:
IsEntityTypeIDsOutsideProperties() and IsUint16tEntityTypeIDs() indicate
the format of the RDG on storage. It is important that RDGSlice expose
that information.


e-mcginnis (approved):
I'll try to keep this requirement in mind as I add more of these
LGTM!

PR #841, opened 2022-01-27
bump arrow to 6.0

tylershunt:
I took over Daniel's PR since we needed to solve issues with out s3 client to make it work.

I added a change to how we handle SIGPIPE. It happens *much* more often with the new sdk version; performance in my limited tests (import ldbc-3k) seems to be unaffected though.

Lookout for the enterprise pr.

Closes #741


katana_requirements.yaml
```diff
@@ -205,11 +205,11 @@ libarrow-python:
     conda: null
     pip: null
     apt: libarrow-python-dev
-  version: &arrow_version [ 4, 5 ]
+  version: &arrow_version [ 6, 7 ]
   labels:
     - apt
   version_overrides: &arrow_version_overrides
-    apt: "=4.0.1-1"
```
arthurp, line 212: Did you check that this is the exact version of the debian arrow packages? Not every version has a fast micro release like 4 did.

arthurp (approved):
LGTM other than the potentially incorrect APT version of the arrow packages.

katana_requirements.yaml
tylershunt: Will check, thanks for the flag.

tylershunt: Ok; `apt list` agrees that this is a valid version (and appears to be the latest for 6):
```
libarrow600/focal 6.0.1-1 amd64
```

ddn0 (approved):

libsupport/test/array-from-scalars.cpp
```diff
@@ -22,9 +22,28 @@ TestNulls() {
...
+  std::vector<std::shared_ptr<arrow::Scalar>> scalars(NUM_ENTRIES);
+  scalars[3] = std::make_shared<arrow::LargeStringScalar>("three");
+  scalars[4] = std::make_shared<arrow::LargeStringScalar>("four");
+  scalars[9] = std::make_shared<arrow::LargeStringScalar>("nine");
+  std::shared_ptr<arrow::Array> array =
+      KATANA_CHECKED(katana::ArrayFromScalars(scalars, arrow::large_utf8()));
+  // NB: length() is signed, size() is not - but they should both have small
+  // enough values that it doesn't matter
+  KATANA_LOG_VASSERT(
+      static_cast<size_t>(array->length()) == scalars.size(),
```
witchel, line 36: It is better to ask to size_t or to the signed type?

witchel:
> I added a change to how we handle SIGPIPE. It happens _much_ more often with the new sdk version; performance in my limited

I found the code about as easy to understand as any signal handling code.  Can you give me the few sentence summary of what is going on here?  Is there an intuition I should have as to why our storage stack is haunted by SIGPIPE?

libsupport/test/array-from-scalars.cpp
tylershunt: I'm not sure @danielmawhirter wrote that commit.

tylershunt:
> > I added a change to how we handle SIGPIPE. It happens _much_ more often with the new sdk version; performance in my limited
> 
> I found the code about as easy to understand as any signal handling code. Can you give me the few sentence summary of what is going on here? Is there an intuition I should have as to why our storage stack is haunted by SIGPIPE?

`SIGPIPE` happens sometimes when there's a network error. Most applications either set the correct socket options so that it doesn't happen, or ignore the signal explicitly (you can find a lot of stack overflow articles where they tell you to do just that) so that they can just handle the errors as bad return codes from operations on sockets. We don't control the sockets in this case; the aws-sdk does, and the approach that they have taken is to throw their hands up since this seems to just happen because of they way they use lib curl:

https://sdk.amazonaws.com/cpp/api/LATEST/struct_aws_1_1_http_options.html#ac56ca429444ca99aeb2dce6aec35f017

witchel:
> > > I added a change to how we handle SIGPIPE. It happens _much_ more often with the new sdk version; performance in my limited
> > 
> > 
> > I found the code about as easy to understand as any signal handling code. Can you give me the few sentence summary of what is going on here? Is there an intuition I should have as to why our storage stack is haunted by SIGPIPE?
> 
> `SIGPIPE` happens sometimes when there's a network error. Most applications either set the correct socket options so that it doesn't happen, or ignore the signal explicitly (you can find a lot of stack overflow articles where they tell you to do just that) so that they can just handle the errors as bad return codes from operations on sockets. We don't control the sockets in this case; the aws-sdk does, and the approach that they have taken is to throw their hands up since this seems to just happen because of they way they use lib curl:
> 
> https://sdk.amazonaws.com/cpp/api/LATEST/struct_aws_1_1_http_options.html#ac56ca429444ca99aeb2dce6aec35f017

Thank you that is very informative.  I don't know if it is worth a brief comment in the code.

> libCurl infects everything with its global state.

There is little I enjoy as much as actively aggressive software documentation.

libsupport/test/array-from-scalars.cpp
danielmawhirter: We know by construction that both are within `[0, max<int64_t>]`, so I have no opinion on the casting convention

libsupport/include/katana/ArrowVisitor.h
```diff
@@ -6,6 +6,7 @@
 #include <arrow/api.h>
 #include <arrow/type.h>
 #include <arrow/vendored/datetime/date.h>
+#include <boost/preprocessor/repetition/repeat.hpp>
```
danielmawhirter: not needed

danielmawhirter (approved):

libsupport/test/array-from-scalars.cpp
witchel: Ok.  I don't ask to be a pain.  I just am a pain.

But I really do wonder about this because it comes up all the time with STL and Arrow.  I used to cast to `size_t`, but now I cast to `int64_t`.  I figure a negative number is more probable than a ridiculously large size.

danielmawhirter: I would agree with that assessment. Casting in either direction does have some theoretical false positives thanks to the sign bit. I guess the most precise form would be something like:
```c++
bool equals(uint64_t u, int64_t i) {
  constexpr uint64_t high_bit = 1ULL << 63;
  if (u & high_bit) {
    // u is too high for an int64_t
    return false;
  } else if (i & high_bit) {
    // i is too low for a uint64_t
    return false;
  }
  // value is in range of both
  return static_cast<int64_t>(u) == i;
}

// code golf submission:
bool equals(uint64_t u, int64_t i) {
  return (!((u | i) & high_bit)) && static_cast<int64_t>(u) == i;
}
```

Or if we wait for c++20:
```c++
bool equals(uint64_t u, int64_t i) {
  return static_cast<std::make_signed_t<uint64_t>>(u) == i;
}
```

In light of those options, I consider some crazy edge case false positives to be an ok compromise. But I think you're right, the crazy edge case where `we have a negative and we shouldn't` is probably more common than `we have a really big value and we shouldn't`

libsupport/src/ArrowVisitor.cpp
```diff
@@ -251,12 +251,17 @@ katana::AppendToBuilder(
...
-
-  return katana::VisitArrow(visitor, builder.get());
+  for (size_t i = 0, n = scalars.size(); i < n; ++i) {
+    if (scalars[i] && scalars[i]->is_valid) {
+      KATANA_CHECKED(builder->AppendScalar(*scalars[i]));
+    } else {
+      KATANA_CHECKED(builder->AppendNull());
+    }
+  }
+  return KATANA_CHECKED(builder->Finish());
```
witchel, line 266: OMG, arrow code that can be understood by a human?  Sacre bleu!

PR #842, opened 2022-01-28
support: Encode URIs

ddn0:
Non-alphanumeric characters are usually percent-escaped in URIs. Escape
these characters when converting a URI into a string.

Uri::path() continues to return raw bytes under the assumption that
callers expect it to behave like a standard filesystem path (with no
special escaping).

Along the way remove Uri::Encode, which base64-encodes a URI. This is a
non-standard thing to do to URIs specifically and is not currently used
by any code.

### DIFF
--- a/libsupport/include/katana/URI.h
+++ b/libsupport/include/katana/URI.h
@@ -24,10 +24,12 @@ public:
   static Result<Uri> MakeFromFile(const std::string& str);
   /// Append a '-' and then a random string to input
   static Result<Uri> MakeRand(const std::string& str);
+
   static std::string JoinPath(const std::string& dir, const std::string& file);
 
-  /// Return the base64 (url variant) encoded version of this uri
-  std::string Encode() const;
+  // Decode returns the raw bytes represented by a URI encoded string.
+  static std::string Decode(const std::string& uri);
+
   /// Hash URI
   struct Hash {
     std::size_t operator()(const Uri& uri) const {
@@ -36,8 +38,15 @@ public:
   };
 
   const std::string& scheme() const { return scheme_; }
+
+  /// path returns the portion of a URI after the scheme. This a concatenation
+  /// of the traditional URI host and path components. Unlike string(), the
+  /// returned value are raw bytes and there is no encoding of special
+  /// characters.
   const std::string& path() const { return path_; }
-  const std::string& string() const { return string_; }
+
+  /// string returns the URI as a URI-encoded string.
+  const std::string& string() const { return encoded_; }
 
   bool empty() const;
 
@@ -60,7 +69,7 @@ private:
 
   std::string scheme_;
   std::string path_;
-  std::string string_;
+  std::string encoded_;
 };
 
 KATANA_EXPORT bool operator==(const Uri& lhs, const Uri& rhs);
--- a/libtsuba/src/LocalStorage.cpp
+++ b/libtsuba/src/LocalStorage.cpp
@@ -26,8 +26,8 @@ namespace fs = boost::filesystem;
 namespace {
 
 katana::Result<void>
-EnsureDirectories(const std::string& uri) {
-  fs::path m_path{uri};
+EnsureDirectories(const std::string& path) {
+  fs::path m_path{path};
   fs::path dir = m_path.parent_path();
   if (!dir.empty()) {
     if (boost::system::error_code err; !fs::create_directories(dir, err)) {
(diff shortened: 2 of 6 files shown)


libsupport/include/katana/URI.h
```diff
@@ -24,10 +24,12 @@ class KATANA_EXPORT Uri {
   static Result<Uri> MakeFromFile(const std::string& str);
   /// Append a '-' and then a random string to input
   static Result<Uri> MakeRand(const std::string& str);
+
   static std::string JoinPath(const std::string& dir, const std::string& file);
 
-  /// Return the base64 (url variant) encoded version of this uri
-  std::string Encode() const;
+  // Decode returns the raw bytes represented by a URI encoded string.
+  static std::string Decode(const std::string& uri);
```
witchel, line 31: While I understand this comment, when I read "raw bytes" I'm thining `std::vector<uint8_t>`.

libtsuba/src/LocalStorage.cpp
```diff
@@ -48,7 +48,10 @@ katana::LocalStorage::CleanUri(std::string* uri) {
   if (uri->find(uri_scheme()) != 0) {
     return;
   }
-  *uri = std::string(uri->begin() + uri_scheme().size(), uri->end());
+
+  std::string path(uri->begin() + uri_scheme().size(), uri->end());
```
witchel: As an idiom to get a file name from a URI, this does not thrill me.

witchel (approved):
The code you wrote is clean.  But I'm a bit worried that our URI class is serving two masters, the network and the file system.  Is it standard to normalize URIs the way realpath normalizes path names?  I think not (e.g., removing redundant '/' characters).  

If all the tests pass, maybe no one is relying on the detailed behavior of .string() for URIs, but I feel like I've been sloppy in assuming that I can call .string() and the result can still be passed to `tsuba::Create`.

libtsuba/src/LocalStorage.cpp
ddn0: I forgot to add this, but `CleanUri` can be replaced with `Uri::path`.

libsupport/include/katana/URI.h
ddn0: I mean what is a string but a buffer of bytes? :p

ddn0:
> But I'm a bit worried that our URI class is serving two masters, the network and the file system. Is it standard to normalize URIs the way realpath normalizes path names? I think not (e.g., removing redundant '/' characters).

I do agree with the sentiment that the URI class is really URIPath. We don't expect query parameters for example.

I think the current semantics of `URI` (assuming it really does mean "URI path thing") is defensible.

If you create a `URI` from a string that does not begin with `scheme://` treat it as a path, `realpath` normalize it, and normalize its scheme to `file://`. Otherwise, treat the string as a URI. All URI operations (join, base name, etc.) do not normalize path components (`.`, `..`, `///`) nor follow symbolic links (thank god).

There is some ambiguity if you really wanted to work on a file named `file://` but we probably have to accept this.

I think that removing `realpath` normalization altogether is also reasonable and consistent. Basically, defer to the underlying scheme (fs plugin) as to the interpretation of the path component. I'm not sure why we have `realpath` in URI.cpp to begin with.

witchel:
> > But I'm a bit worried that our URI class is serving two masters, the network and the file system. Is it standard to normalize URIs the way realpath normalizes path names? I think not (e.g., removing redundant '/' characters).
> 
> I do agree with the sentiment that the URI class is really URIPath. We don't expect query parameters for example.
> 
> I think the current semantics of `URI` is defensible.
> 
> If you create a `URI` from a string that does not begin with `scheme://` treat it as a path, `realpath` normalize it, and normalize its scheme to `file://`. Otherwise, treat the string as a URI. All URI operations (join, base name, etc.) do not normalize path components (`.`, `..`, `///`) nor follow symbolic links (thank god).

Yeah, this seems reasonable to me.  I assume gs and s3 accept percent escaped URIs.

> There is some ambiguity if you really wanted to work on a file named `file://` but we probably have to accept this.

I accept.

> I think that removing `realpath` normalization altogether is also reasonable and consistent. Basically, defer to the underlying scheme (fs plugin) as to the interpretation of the path component. I'm not sure why we have it to begin with.

For example, it is helpful when the implementation has to do things like Join and you need to know if your URI ends in a / or not.

ddn0:
> For example, it is helpful when the implementation has to do things like Join and you need to know if your URI ends in a / or not.

I realized that my statement of

>> Otherwise, treat the string as a URI. All URI operations (join, base name, etc.) do not normalize path components (., .., ///) nor follow symbolic links (thank god).

is a little wrong, we do do conversions like collapsing `//` => `/` and appending `/` when doing join. I think that is a reasonable thing to do for all URIs. It seems right that we don't do anything special for `.` or `..` though as the meaning of that is specific to the scheme.

witchel:
Will this implementation of Uri::string cause problems for `PropertyGraph::rdg_dir`, which looks like this.

`const std::string& rdg_dir() const { return rdg_->rdg_dir().string(); }`

ddn0:
> Will this implementation of Uri::string cause problems for PropertyGraph::rdg_dir, which looks like this.
> const std::string& rdg_dir() const { return rdg_->rdg_dir().string(); }

If `rdg_dir` is really supposed to be a URI, which in all the cases I have seen so far is the case, this should be fine. `rdg_dir` usually is a URI because it needs to have a scheme to identify the access plugin, e.g., `s3://`.

If `rdg_dir` is supposed to be a (local) path, this will be a problem. And also suffers from the ambiguity I mentioned above which is if this a path then `s3://` is supposed to be a file named `s3://`.

witchel:
> > Will this implementation of Uri::string cause problems for PropertyGraph::rdg_dir, which looks like this.
> > const std::string& rdg_dir() const { return rdg_->rdg_dir().string(); }
> 
> If `rdg_dir` is really supposed to be a URI, which in all the cases I have seen so far is the case, this should be fine. `rdg_dir` usually is a URI because it needs to have a scheme to identify the access plugin, e.g., `s3://`.
> 
> If `rdg_dir` is supposed to be a (local) path, this will be a problem. And also suffers from the ambiguity I mentioned above which is if this a path then `s3://` is supposed to be a file named `s3://`.

Yeah, looking at `AnalyticsOperation.cpp` as a random example, we store `rdg_dir()` in a string that gets passed to `FindManifest` where it is made back into a `Uri`.  My concern was just that we haven't been terribly consistent in using Uri.  But your change could help flush out cases where we are assuming a Uri is a local path, which is a good thing.  Over time, we should try to standardize on Uris, but I ran into some complications when I reworked `InformPath`.  There are many interfaces that take strings.

PR #843, opened 2022-01-28
Add drop down version selector to documentation website [KAT-889]

JamesFCoffey:
This adds a drop down menu to the top right corner of the documentation websites.

KAT-889

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -64,13 +64,26 @@ exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", "**.ipynb_checkpoints"]
 
 # -- Options for HTML output -------------------------------------------------
 
+if tags.has("external"):
+    publish_url = "docs.katanagraph.com"
+else:
+    publish_url = "docs.k9h.dev"
+
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 
 html_theme = "pydata_sphinx_theme"
 html_logo = "_static/logo.png"
 html_title = "Katana"
-html_theme_options = {"show_prev_next": False}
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -64,13 +64,26 @@
 
 # -- Options for HTML output -------------------------------------------------
 
+if tags.has("external"):
+    publish_url = "docs.katanagraph.com"
+else:
+    publish_url = "docs.k9h.dev"
```
arthurp, line 70: Should this be in some function to make sure it's not executed before `tags` is populated? I really don't know, top level code is just a bit dangerous because execution order is unpredictable.

arthurp (approved):
This is great!!! I'm so happy to see our docs improving like this.

Please do check on my comment below, but if you are confident it's safe, carry on.

docs/conf.py
JamesFCoffey: @arthurp I think this should be ok.

Is it true that top level code execution is unpredictable in a Python script? I thought it was executed sequentially. I've not run into this before. Is there some circumstance where this happens?

arthurp: Within a module, you are correct. The issue is that module import order is hard to predict. It is technically deterministic, but it's affected by what other modules import your code, so it's really hard to predict. So the issue is whether `tags` could be changed by other code after this code executes invalidating the value of `publish_url`.

One fix would be to use a function `get_publish_url` instead. That way it would always be up to date if tags changes.

arthurp: So I just realized what file this is. Silly me. :-/ This code is fine. Sorry for the confusion.

PR #844, opened 2022-01-28
[KAT-1688] [KAT-3051] Add RDG directory as prefix when tracking property names in TxnContext

yige-hu:
Add RDG directory as prefix when tracking property names in `TxnContext`. This is for graphs with the same property names.

Jira tasks: [KAT-1688] [KAT-3051]
Enterprise repo: [PR #3105](https://github.com/KatanaGraph/katana-enterprise/pull/3105)

[KAT-3051]: https://katanagraph.atlassian.net/browse/KAT-3051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libtsuba/include/katana/TxnContext.h
+++ b/libtsuba/include/katana/TxnContext.h
@@ -9,40 +9,48 @@ namespace katana {
 
 class KATANA_EXPORT TxnContext {
 public:
-  void InsertNodePropertyRead(std::string name) {
-    node_properties_read_.insert(name);
+  void InsertNodePropertyRead(std::string rdg_dir, std::string name) {
+    node_properties_read_.insert(ConcatRDGProperty(rdg_dir, name));
   }
 
   template <typename Container>
-  void InsertNodePropertyRead(const Container& names) {
-    node_properties_read_.insert(names.begin(), names.end());
+  void InsertNodePropertyRead(std::string rdg_dir, const Container& names) {
+    for (auto name : names) {
+      node_properties_read_.insert(ConcatRDGProperty(rdg_dir, name));
+    }
   }
 
-  void InsertNodePropertyWrite(std::string name) {
-    node_properties_write_.insert(name);
-  }
-
-  template <typename Container>
-  void InsertNodePropertyWrite(const Container& names) {
-    node_properties_write_.insert(names.begin(), names.end());
-  }
-
-  void InsertEdgePropertyRead(std::string name) {
-    edge_properties_read_.insert(name);
+  void InsertNodePropertyWrite(std::string rdg_dir, std::string name) {
+    node_properties_write_.insert(ConcatRDGProperty(rdg_dir, name));
   }
 
   template <typename Container>
-  void InsertEdgePropertyRead(const Container& names) {
-    edge_properties_read_.insert(names.begin(), names.end());
+  void InsertNodePropertyWrite(std::string rdg_dir, const Container& names) {
+    for (auto name : names) {
+      node_properties_write_.insert(ConcatRDGProperty(rdg_dir, name));
+    }
   }
 
-  void InsertEdgePropertyWrite(std::string name) {
-    edge_properties_write_.insert(name);
+  void InsertEdgePropertyRead(std::string rdg_dir, std::string name) {
+    edge_properties_read_.insert(ConcatRDGProperty(rdg_dir, name));
   }
 
   template <typename Container>
-  void InsertEdgePropertyWrite(const Container& names) {
-    edge_properties_write_.insert(names.begin(), names.end());
+  void InsertEdgePropertyRead(std::string rdg_dir, const Container& names) {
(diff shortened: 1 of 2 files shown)


libtsuba/include/katana/TxnContext.h
```diff
@@ -78,6 +86,12 @@ class KATANA_EXPORT TxnContext {
   bool GetTopologyWrite() const { return topology_write_; }
 
 private:
+  std::string ConcatRDGProperty(std::string rdg_dir, std::string prop) {
+    return rdg_dir + kPropSeparator + prop;
```
witchel, line 90: I use `katana::Join` or `katana::JoinPath` in `URI.h`.

libtsuba/include/katana/TxnContext.h
```diff
@@ -9,40 +9,48 @@ namespace katana {
...
-    node_properties_read_.insert(name);
+  void InsertNodePropertyRead(std::string rdg_dir, std::string name) {
+    node_properties_read_.insert(ConcatRDGProperty(rdg_dir, name));
   }
 
   template <typename Container>
-  void InsertNodePropertyRead(const Container& names) {
-    node_properties_read_.insert(names.begin(), names.end());
+  void InsertNodePropertyRead(std::string rdg_dir, const Container& names) {
+    for (auto name : names) {
```
witchel, line 18: Can all of these be `const auto&`

libtsuba/src/RDGCore.cpp
```diff
@@ -185,7 +185,8 @@ katana::RDGCore::AddNodeProperties(
   auto written_prop_names = KATANA_CHECKED(AddProperties(
       props, &node_properties_, &part_header_.node_prop_info_list()));
   // store write properties into transaction context
-  txn_ctx->InsertNodePropertyWrite<std::set<std::string>>(written_prop_names);
+  txn_ctx->InsertNodePropertyWrite<std::set<std::string>>(
+      rdg_dir_.string(), written_prop_names);
```
witchel, line 189: @ddn0, this is how a lot of people approach URI::string, which is just give me the string version of this file location.  I don't think they are expecting percent escapes. (This is a reference to Donald's #842.

witchel (approved):
We need this, but be careful about URI::string (which goes double for me).

libtsuba/src/RDGCore.cpp
ddn0: As long as we go through tsuba to read the paths, it should be fine. The string version is a proper URI, the local fs backend knows how to decode URIs.

The problematic case would be if someone stripped the scheme portion of the URI off without decoding the path portion.

witchel: Ah, that is ok with me.

libtsuba/include/katana/TxnContext.h
ddn0: This is the kind of operation that is unsafe to do on bare URIs. If `kPropSeparator` or `prop` contained non-safe URI characters, we would create an invalid URI.

libtsuba/include/katana/TxnContext.h
witchel: I assume you will change these in your next PR?

libtsuba/include/katana/TxnContext.h
witchel: Well, then Yige, I guess my suggestion to use `JoinPath `is a requirement.

ddn0: @yige-hu can you please update this to use `Uri::Join` unless you know that `rdg_dir` did not come from `Uri::string`?

libtsuba/include/katana/TxnContext.h
yige-hu: sorry that I just saw this comment after merging. I'll add the changes in PR#831

libtsuba/include/katana/TxnContext.h
yige-hu: @witchel I feel here '+' operator can work better because I'm concatenating three strings. `katana::JoinPath(rdg_dir, kPropSeparator).Join(prop)` doesn't look very straightforward.

witchel: Wouldn't the call be `katana::JoinPath(rdg_dir, prop)`?

As Donald writes, if `kPropSeparator `or `prop` contained non-safe URI characters, the result would be an invalid URI.  You could argue that you are building a string, not a URI, but honestly, I think you are better off storing URIs.  I use them as cache keys.  Because they have a normalized form, they are better than a string, which will fail to match `gs://bucket/ `and `gs://bucket`, even though those two refer to the same storage location.

PR #845, opened 2022-01-28
RDG: add NUMAArrray type ids interface

scober:
A common pattern is to load the file associated with a type id
array and then copy it into a NUMAArray. Move this pattern into the RDG
itself so it isn't duplicated so many times. Centralizing this in
RDGCore also creates a single location where storage format
complications are handled.

### DIFF
--- a/libtsuba/src/RDGCore.cpp
+++ b/libtsuba/src/RDGCore.cpp
@@ -5,6 +5,7 @@
 #include "katana/ArrowInterchange.h"
 #include "katana/ErrorCode.h"
 #include "katana/ParquetReader.h"
+#include "katana/RDGPrefix.h"
 #include "katana/Result.h"
 
 namespace {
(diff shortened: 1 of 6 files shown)


libtsuba/src/RDGCore.cpp
```diff
@@ -131,6 +132,38 @@ schemify(const std::vector<katana::PropStorageInfo>& prop_info_list) {
...
+LoadAndMaybeConvertTypeIDArray(
+    size_t begin, size_t end, const katana::Uri& types_path,
+    const katana::RDGPartHeader& part_header) {
+  katana::NUMAArray<katana::EntityTypeID> types;
+
+  // NB: we add sizeof(EntityTypeIDArrayHeader) to every range element because
+  // the structure of this file is
+  // [header, value, value, value, ...]
+  // Recent storage formats remove this header
+  size_t header_size = part_header.unstable_storage_format()
```
e-mcginnis, line 147: FYI the unstable format is officially becoming version 4 in https://github.com/KatanaGraph/katana/pull/833
which is just having some CI trouble on the enterprise side of things :/ 

The change here would be just like https://github.com/KatanaGraph/katana/pull/833/files#diff-c5c5eb22f4a9863510a9a228eb80a6239b374834360ea096c3366a66e1240330

scober: Would you like me to wait to merge after #833?

e-mcginnis: Nah, its not a big deal either way. 
Just ping me if this merges before #833 so I can make sure to fix it up

scober: If you don't have any objections, would you approve it?

e-mcginnis (approved):
LGTM!

libtsuba/src/RDGCore.cpp
e-mcginnis: sorry, was mid-review when I got interrupted earlier.

libtsuba/src/RDGCore.cpp
```diff
@@ -131,6 +132,40 @@ schemify(const std::vector<katana::PropStorageInfo>& prop_info_list) {
...
+  katana::FileView fv;
+
+  KATANA_CHECKED_CONTEXT(
+      fv.Bind(types_path.string(), storage_begin, storage_end, true),
+      "loading node type id array, begin: {}, end: {}", begin, end);
+
+  types.allocateInterleaved(end - begin);
+
+  const auto* storage_types = fv.template valid_ptr<StorageType>();
+  katana::do_all(
```
amberhassaan, line 162: We have `ParallelSTL::copy()` which may be slightly more efficient.

libtsuba/src/RDGCore.cpp
```diff
@@ -131,6 +132,40 @@ schemify(const std::vector<katana::PropStorageInfo>& prop_info_list) {
...
+  size_t storage_begin = begin * sizeof(StorageType) + header_size;
+  size_t storage_end = end * sizeof(StorageType) + header_size;
+
+  katana::FileView fv;
+
+  KATANA_CHECKED_CONTEXT(
+      fv.Bind(types_path.string(), storage_begin, storage_end, true),
+      "loading node type id array, begin: {}, end: {}", begin, end);
+
+  types.allocateInterleaved(end - begin);
```
amberhassaan, line 159: Do we expect a new copy being allocated everytime we call the API? Previously, `Bind` would mmap a file in and the user code would decide to copy into a `NUMAArray` or whatever.

amberhassaan (commented):
Sorry for being late to review. Left a couple of comments that I felt were worth thinking about.

libtsuba/src/RDGCore.cpp
scober: That is a good point

libtsuba/src/RDGCore.cpp
scober: Currently many users make this copy anyway. Users that aren't making this copy can still use the original interface. Reducing the number of copies and/or using a NUMA-aware allocator for the original `Bind()` are left as future work.

amberhassaan: OK. This makes sense given the old interface is still there.

PR #846, opened 2022-01-28
Introduce RDGOptionalDatastructures: RDKLSHIndex

e-mcginnis:
RDGOptionalDatastructures are datastructures that may be appended to an RDG
These datastructures are represented by a manifest file, which is json
encoded and stores the information required to load the optional
datastructure.

The only mandatory field in an optional datastructure manifest file is
"paths" which is used to list any additional files that make up the
optional datastructure. The "paths" field is
```
unordered_map<name, file_path>
```
the file_path is relative to the rdg_directory.

The RDGPartHeader tracks these optional data structures in the new field
```
optional_datastructure_manifests_
```
Any amount of optional datastructures may or may not be present in any
RDG. It is expected that if an optional datastructure is not present,
that the user will generate it instead.

The first optional datastructure is RDKLSHIndex.

JIRA: KAT-3696

### DIFF
--- a/libgraph/include/katana/PropertyGraph.h
+++ b/libgraph/include/katana/PropertyGraph.h
@@ -315,6 +315,24 @@ public:
     return *edge_entity_type_manager_;
   }
 
+  Result<std::optional<RDKLSHIndexPrimitive>> LoadRDKLSHIndexPrimitive() {
+    return rdg_->LoadRDKLSHIndexPrimitive();
+  }
+
+  Result<void> WriteRDKLSHIndexPrimitive(RDKLSHIndexPrimitive& index) {
+    return rdg_->WriteRDKLSHIndexPrimitive(index);
+  }
+
+  Result<std::optional<RDKSubstructureIndexPrimitive>>
+  LoadRDKSubstructureIndexPrimitive() {
+    return rdg_->LoadRDKSubstructureIndexPrimitive();
+  }
+
+  Result<void> WriteRDKSubstructureIndexPrimitive(
+      RDKSubstructureIndexPrimitive& index) {
+    return rdg_->WriteRDKSubstructureIndexPrimitive(index);
+  }
+
   const std::string& rdg_dir() const { return rdg_->rdg_dir().string(); }
 
   uint32_t partition_id() const { return rdg_->partition_id(); }
--- a/libtsuba/include/katana/RDGOptionalDatastructure.h
+++ b/libtsuba/include/katana/RDGOptionalDatastructure.h
@@ -1,0 +1,82 @@
+#ifndef KATANA_LIBTSUBA_KATANA_RDGOPTIONALDATASTRUCTURE_H_
+#define KATANA_LIBTSUBA_KATANA_RDGOPTIONALDATASTRUCTURE_H_
+
+#include <cstdint>
+#include <unordered_map>
+#include <vector>
+
+#include "katana/AtomicWrapper.h"
+#include "katana/DynamicBitset.h"
+#include "katana/ErrorCode.h"
+#include "katana/FileView.h"
+#include "katana/JSON.h"
+#include "katana/Logging.h"
+#include "katana/NUMAArray.h"
+#include "katana/Result.h"
+#include "katana/URI.h"
+#include "katana/WriteGroup.h"
+#include "katana/config.h"
+#include "katana/tsuba.h"
+
+namespace katana {
+
+/// Base class for all optional datastructures.
+/// Paths to the RDGOptionalDatastructure files are stored in the RDGPartHeader::optional_datastructure_manifests_
+class KATANA_EXPORT RDGOptionalDatastructure {
+public:
+  std::map<std::string, std::string> paths() const { return paths_; }
+  void set_paths(std::map<std::string, std::string> paths) { paths_ = paths; }
+
+  /// Load() should make the Optional Data structure fully available to the caller,
+  /// Mapping any additional files defined in paths_ into memory as necessary
+  static katana::Result<RDGOptionalDatastructure> Load(
+      const katana::Uri& rdg_dir_path, const std::string& path) = delete;
+
+  /// Write() should immediately persist the optional data structure to disk
+  /// Users of RDGOptionalDatastructures should ensure Write() is called before RDG::Store()
+  katana::Result<std::string> Write(katana::Uri rdg_dir_path) = delete;
+
+  static katana::Result<void> ChangeStorageLocation(
+      const std::string& manifest_relpath, const katana::Uri& old_loc,
+      const katana::Uri& new_loc) {
+    katana::Uri old_manifest_path = old_loc.Join(manifest_relpath);
+    katana::FileView fv;
+    KATANA_CHECKED(fv.Bind(old_manifest_path.string(), true));
+    RDGOptionalDatastructure data;
+    KATANA_CHECKED(katana::JsonParse<RDGOptionalDatastructure>(fv, &data));
+
+    // copy over any extra files the optional datastructure relies on
+    // Assumes that all OptionalDatastructures properly extend the RDGOptionalDatastructure class
+    for (const auto& file : data.paths_) {
+      katana::FileView fvtmp;
+      katana::Uri old_path = old_loc.Join(file.second);
+      KATANA_CHECKED(fvtmp.Bind(old_path.string(), true));
+      katana::Uri new_path = new_loc.Join(file.second);
+      KATANA_CHECKED(katana::FileStore(
+          new_path.string(), fvtmp.ptr<uint8_t>(), fvtmp.size()));
+      KATANA_CHECKED(fvtmp.Unbind());
+    }
+
+    // copy out the manifest itself
+    katana::Uri new_manifest_path = new_loc.Join(manifest_relpath);
+    KATANA_CHECKED(katana::FileStore(
+        new_manifest_path.string(), fv.ptr<uint8_t>(), fv.size()));
+    KATANA_CHECKED(fv.Unbind());
+
+    return katana::ResultSuccess();
+  }
+
+  friend void to_json(nlohmann::json& j, const RDGOptionalDatastructure& data);
+  friend void from_json(
+      const nlohmann::json& j, RDGOptionalDatastructure& data);
+
+protected:
+  // map of exta files this optional datastructure will load
+  // { "file_name" : "rdg-relative_path" }
+  // track these so that when we move the RDG, we also move these extra files
+  std::map<std::string, std::string> paths_;
+};
+
+}  // namespace katana
+
+#endif
--- a/libtsuba/src/RDG.cpp
+++ b/libtsuba/src/RDG.cpp
@@ -5,6 +5,7 @@
 #include <fstream>
 #include <iterator>
 #include <memory>
+#include <optional>
 #include <regex>
 #include <string>
 #include <unordered_set>
@@ -1046,6 +1047,80 @@ katana::RDG::SetEdgeEntityTypeIDArrayFile(
 katana::Result<katana::NUMAArray<katana::EntityTypeID>>
 katana::RDG::edge_entity_type_id_array() const {
   return KATANA_CHECKED(core_->edge_entity_type_id_array());
+}
+
+katana::Result<std::optional<katana::RDKLSHIndexPrimitive>>
+katana::RDG::LoadRDKLSHIndexPrimitive() {
+  if (!KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat)) {
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument,
+        "The UnstableRDGStorageFormat feature flag must be set to use this "
+        "feature");
+  }
+  std::optional<std::string> res =
+      KATANA_CHECKED(core_->part_header().OptionalDatastructureManifest(
+          kOptionalDatastructureRDKLSHIndexPrimitive));
+  if (!res) {
+    return std::nullopt;
+  }
+
+  katana::RDKLSHIndexPrimitive index = KATANA_CHECKED_CONTEXT(
+      katana::RDKLSHIndexPrimitive::Load(rdg_dir(), res.value()),
+      "Failed to load RDKLSHIndexPrimitive located at {}", res.value());
+  return index;
+}
+
+katana::Result<void>
+katana::RDG::WriteRDKLSHIndexPrimitive(katana::RDKLSHIndexPrimitive& index) {
+  if (!KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat)) {
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument,
+        "The UnstableRDGStorageFormat feature flag must be set to use this "
+        "feature");
+  }
+  std::string path = KATANA_CHECKED(index.Write(rdg_dir()));
+  core_->part_header().AppendOptionalDatastructureManifest(
+      kOptionalDatastructureRDKLSHIndexPrimitive, path);
+
+  return katana::ResultSuccess();
+}
+
+katana::Result<std::optional<katana::RDKSubstructureIndexPrimitive>>
+katana::RDG::LoadRDKSubstructureIndexPrimitive() {
+  if (!KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat)) {
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument,
+        "The UnstableRDGStorageFormat feature flag must be set to use this "
+        "feature");
+  }
+  std::optional<std::string> res =
+      KATANA_CHECKED(core_->part_header().OptionalDatastructureManifest(
+          kOptionalDatastructureRDKSubstructureIndexPrimitive));
+  if (!res) {
+    return std::nullopt;
+  }
+
+  katana::RDKSubstructureIndexPrimitive index = KATANA_CHECKED_CONTEXT(
+      katana::RDKSubstructureIndexPrimitive::Load(rdg_dir(), res.value()),
+      "Failed to load RDKSubstructureIndexPrimitive located at {}",
+      res.value());
+  return index;
+}
+
+katana::Result<void>
+katana::RDG::WriteRDKSubstructureIndexPrimitive(
+    katana::RDKSubstructureIndexPrimitive& index) {
+  if (!KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat)) {
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument,
+        "The UnstableRDGStorageFormat feature flag must be set to use this "
+        "feature");
+  }
+  std::string path = KATANA_CHECKED(index.Write(rdg_dir()));
+  core_->part_header().AppendOptionalDatastructureManifest(
+      kOptionalDatastructureRDKSubstructureIndexPrimitive, path);
+
+  return katana::ResultSuccess();
 }
 
 katana::RDG::RDG(std::unique_ptr<RDGCore>&& core) : core_(std::move(core)) {}
--- a/libtsuba/src/RDGPartHeader.cpp
+++ b/libtsuba/src/RDGPartHeader.cpp
@@ -1,16 +1,20 @@
 #include "RDGPartHeader.h"
 
+#include <cstdint>
+#include <unordered_map>
 #include <vector>
 
 #include "Constants.h"
 #include "GlobalState.h"
 #include "PartitionTopologyMetadata.h"
 #include "RDGHandleImpl.h"
+#include "katana/CompilerSpecific.h"
 #include "katana/ErrorCode.h"
 #include "katana/Experimental.h"
 #include "katana/FaultTest.h"
 #include "katana/FileView.h"
 #include "katana/Logging.h"
+#include "katana/RDGOptionalDatastructure.h"
 #include "katana/RDGStorageFormatVersion.h"
 #include "katana/RDGTopology.h"
 #include "katana/Result.h"
@@ -47,6 +51,7 @@ const char* kPartitionTopologyMetadataEntriesKey =
     "kg.v1.partition_topology_metadata_entries";
 const char* kPartitionTopologyMetadataEntriesSizeKey =
     "kg.v1.partition_topology_metadata_entries_size";
+const char* kOptionalDatastructuresKey = "kg.v1.optional_datastructures";
 
 //
 //constexpr std::string_view  mirror_nodes_prop_name = "mirror_nodes";
@@ -274,6 +279,13 @@ katana::RDGPartHeader::ChangeStorageLocation(
   edge_entity_type_id_array_path_ = "";
   topology_metadata_.ChangeStorageLocation();
 
+  // the OptionalDatastructure Files are loaded and stored on demand,
+  // move any that are present, even if they are unloaded, to the new location
+  for (auto& datastructure : optional_datastructure_manifests_) {
+    KATANA_CHECKED(katana::RDGOptionalDatastructure::ChangeStorageLocation(
+        datastructure.second, old_location, new_location));
+  }
+
   return katana::ResultSuccess();
 }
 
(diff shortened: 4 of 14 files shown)


e-mcginnis:
@astolman I used https://github.com/KatanaGraph/katana-enterprise/blob/6095a1b0731d3e5b46726d731377a2fa5efa2b0e/plugins/rdkit/src/RDKLSHIndex.cpp#L16 as reference for the RDKLSHIndex file. Please let me know if I am missing any necessary API. 

These files:
https://github.com/KatanaGraph/katana/pull/846/commits/783201119946d98fdeaeb008119aae42faacec00#diff-f1c2f51f404f1d11de3a7f312315961020411ef71331dd816999405483f095c5
https://github.com/KatanaGraph/katana/pull/846/commits/783201119946d98fdeaeb008119aae42faacec00#diff-caa6c7c1076381c9956fb40e3156fea8f501a50ba7427c7c9da37c63d7607db6
will be the most relevant for using the RDKLSHIndex

This can be tested by executing katana software with the envvar `KATANA_ENABLE_EXPERIMENTAL="UnstableRDGStorageFormat"`
which will enable storing/loading RDKLSHIndex file. 

If there are additional files that RDK needs to have persisted, please let me know.

libtsuba/include/katana/RDKLSHIndex.h
```diff
@@ -0,0 +1,130 @@
+#ifndef KATANA_LIBTSUBA_KATANA_RDKLSHINDEX_H_
```
yshalabi: Why cant design these interfaces such that the RDGOptionalDatastructures that katana-enterprise need can be defined and reside in katana-enterprise?

libtsuba/src/RDG.cpp
```diff
@@ -1026,6 +1031,37 @@ katana::RDG::SetEdgeEntityTypeIDArrayFile(
   return core_->RegisterEdgeEntityTypeIDArrayFile(new_type_id_array.BaseName());
 }
 
+katana::Result<katana::RDKLSHIndex>
+katana::RDG::LoadRDKLSHIndex() {
+  if (!KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat)) {
```
yshalabi, line 1054: Since this is a feature we are presumably wanting now, when do you envision removing this requirement?

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -329,6 +346,43 @@ katana::to_json(json& j, const katana::RDGPartHeader& header) {
...
+        {kNodeEntityTypeIDDictionaryKey,
+         header.node_entity_type_id_dictionary_},
+        {kEdgeEntityTypeIDDictionaryKey,
+         header.edge_entity_type_id_dictionary_},
+        {kNodeEntityTypeIDNameKey, header.node_entity_type_id_name_},
+        {kEdgeEntityTypeIDNameKey, header.edge_entity_type_id_name_},
+        {kPartitionTopologyMetadataKey, header.topology_metadata_},
+        {kOptionalDatastructuresKey, header.optional_datastructure_manifests_}};
+  } else {
+    j = json{
```
yshalabi, line 354: block unnecessary given lines 335-348

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -425,6 +479,13 @@ katana::from_json(const json& j, katana::RDGPartHeader& header) {
     j.at(kTopologyPathKey).get_to(entry.path_);
     header.topology_metadata_.Append(entry);
   }
+
+  // Version 5 added optional data structures
```
yshalabi: version 5 or version 4?

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -609,3 +670,79 @@ katana::to_json(json& j, const katana::PartitionTopologyMetadata& topomd) {
...
 }
+
+void
+katana::from_json(const nlohmann::json& j, katana::DynamicBitset& bitset) {
+  std::vector<uint64_t> vec;
+  size_t size;
+  j.at("bitset_vec").get_to(vec);
+  j.at("bitset_size").get_to(size);
+
+  for (const auto& val : vec) {
```
yshalabi: debug

libtsuba/src/RDG.cpp
```diff
@@ -1026,6 +1031,37 @@ katana::RDG::SetEdgeEntityTypeIDArrayFile(
...
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument,
+        "The UnstableRDGStorageFormat feature flag must be set to use this "
+        "feature");
+  }
+  std::string path =
+      KATANA_CHECKED(core_->part_header().OptionalDatastructureManifest(
+          kOptionalDatastructureRDKLSHIndex));
+  katana::RDKLSHIndex index =
+      KATANA_CHECKED(katana::RDKLSHIndex::Load(rdg_dir(), path));
```
yshalabi: We should return an appropriate error if this fails

libtsuba/src/RDG.cpp
```diff
@@ -1026,6 +1031,37 @@ katana::RDG::SetEdgeEntityTypeIDArrayFile(
...
+katana::Result<katana::RDKLSHIndex>
+katana::RDG::LoadRDKLSHIndex() {
+  if (!KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat)) {
+    return KATANA_ERROR(
+        ErrorCode::InvalidArgument,
+        "The UnstableRDGStorageFormat feature flag must be set to use this "
+        "feature");
+  }
+  std::string path =
+      KATANA_CHECKED(core_->part_header().OptionalDatastructureManifest(
```
yshalabi, line 1061: We should return an appropriate error if this fails

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -609,3 +670,79 @@ katana::to_json(json& j, const katana::PartitionTopologyMetadata& topomd) {
...
+  bitset.clear();
+  bitset.resize(size);
+  for (const auto& val : vec) {
+    bitset.set(val);
+  }
+  for (const auto& val : vec) {
+    KATANA_LOG_VASSERT(bitset.test(val), "val {} unset!!", val);
+  }
+
+  KATANA_LOG_WARN(
```
yshalabi: debug

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -609,3 +670,79 @@ katana::to_json(json& j, const katana::PartitionTopologyMetadata& topomd) {
...
+        "EMCGINNIS from_json vec size = {}, val = {}", vec.size(), val);
+  }
+
+  bitset.clear();
+  bitset.resize(size);
+  for (const auto& val : vec) {
+    bitset.set(val);
+  }
+  for (const auto& val : vec) {
+    KATANA_LOG_VASSERT(bitset.test(val), "val {} unset!!", val);
```
yshalabi: debug

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -609,3 +670,79 @@ katana::to_json(json& j, const katana::PartitionTopologyMetadata& topomd) {
...
+  }
+  j = nlohmann::json{{"bitset_vec", vec}, {"bitset_size", bitset.size()}};
+  KATANA_LOG_WARN(
+      "EMCGINNIS to_json vec size = {}, bitset_size = {}", vec.size(),
+      bitset.size());
+}
+
+void
+katana::from_json(const nlohmann::json& j, katana::RDKLSHIndex& index) {
+  j.at("num_hashes_per_bucket").get_to(index.num_hashes_per_bucket_);
```
yshalabi, line 682: Since we are going this route of storing arbitrary structures in our distributed graph we ought to at least version control the data structures so that we can fail the load if versions change

libtsuba/include/katana/RDKLSHIndex.h
yshalabi: @bowu in the case that this is not done, be aware that these aspects of the index implementations will be placed in katana-open

libtsuba/test/storage-format-version/v5-optional-datastructure-rdklsh-index.cpp
```diff
@@ -0,0 +1,129 @@
+#include <cstdint>
```
yshalabi: Once this lands and katana-enterprise is updated lets make sure to get at least these tests in:
1. LSH index created, saved, and loaded (just like this test)
2. RDG with existing LSH loaded (this will need to be updated with every change to the index implementation)
3. Test emulating case where an RDG contains an outdated LSH index and a load index is attempted from it ( to mimic graceful failing). BTW, right now in the present implementation this case will cause an std::abort.

libtsuba/test/storage-format-version/v5-optional-datastructure-rdklsh-index.cpp
```diff
@@ -0,0 +1,129 @@
...
+    KATANA_LOG_FATAL("katana::InitTsuba: {}", init_good.error());
+  }
+  katana::GaloisRuntime Katana_runtime;
+
+  if (argc <= 1) {
+    KATANA_LOG_FATAL("missing rdg file directory");
+  }
+
+  // Ensure the feature flag is actually set
+  KATANA_LOG_ASSERT(KATANA_EXPERIMENTAL_ENABLED(UnstableRDGStorageFormat));
```
yshalabi: Maybe I am getting my self confused -- but just to make sure --are we requiring that users of this RDG optional structure feature will require someone to enable experimental RDG features or are we saying that users of the LSH Index stored in the RDG will require this flag? This is just until we get to the point where we commit to supporting this long-term -- right?

yshalabi (commented):
Looking good, some questions and comments

libtsuba/include/katana/RDKLSHIndex.h
e-mcginnis: That would be great. I'm not sure our current dependency graph would make that easy to do. I'm sure there is some clever way to achieve it, but this is intended to be a quick implementation to get folks unblocked.

libtsuba/src/RDG.cpp
e-mcginnis: a day or two after this merges to give folks some time to test it before we set this format in stone.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -342,6 +342,12 @@ class KATANA_EXPORT PropertyGraph {
     return edge_entity_type_manager_;
   }
 
+  Result<RDKLSHIndex> LoadRDKLSHIndex() { return rdg_.LoadRDKLSHIndex(); }
```
astolman: So these are the calls I will use as a user to initiate saving and loading?

libtsuba/src/RDGPartHeader.cpp
e-mcginnis: 5 :) 
the PR to transition the current unstable format to 4 has been in flight for a couple days. I expected it to be merged already, but that hasn't worked out. I could combine the changes in 4 with these changes, but I think that would muddy the already dirty water.

astolman (commented):
This looks good to me. I am a little concerned with putting any of the rdk stuff in the open repo though. 

The only big concern I have is that this just deals with the LSH index. There actually is another index, the substructure index, that needs to get saved too. I was thinking you were going to use the file I/O functions I already had, and so I wrote similar save/load methods for the substruct index https://github.com/astolman/katana-enterprise/blob/559f7f56e748d25ce95d59308bfe3f0dfbd1c4f0/plugins/rdkit/src/RDKSubstructIndex.cpp#L136

libtsuba/src/RDGPartHeader.cpp
e-mcginnis: gah, thanks. removed.

libtsuba/src/RDGPartHeader.cpp
e-mcginnis: I'm not sure I understand. This structure is already versioned by:
1) the RDG storage_format_version
2) the json key `kg.v1.rdk_lsh_index`

Any changes to this structure will require bumping the storage_format_version. We can also bump the version in the json key (all of the RDGPartHeader structures have this) but that hasn't been done historically.

e-mcginnis: also, once this lands and is officially in storage format version 5, I'll update the input for https://github.com/KatanaGraph/katana/blob/master/python/test/test_storage_format.py which will then automatically detect changes to the data structure that are not accompanied by either the unstable flag or a format version bump.

libtsuba/test/storage-format-version/v5-optional-datastructure-rdklsh-index.cpp
e-mcginnis: yep, the is only while we are stabilizing what the RDKLSHIndex looks like to avoid needless tinkering with the rdg_storage_format

libgraph/include/katana/PropertyGraph.h
e-mcginnis: yep.

e-mcginnis:
> This looks good to me. I am a little concerned with putting any of the rdk stuff in the open repo though.

I'd love to be wrong, but I don't believe we  have a good way to avoid putting this in the open repo, since our RDG is defined by the open repo code. I'm sure I could figure something out, but this is supposed to be a quick turn around thing :) 
If you feel any terminology is too telling to be in the open repo, please let me know and I will change it. 



> The only big concern I have is that this just deals with the LSH index. There actually is another index, the substructure index, that needs to get saved too. I was thinking you were going to use the file I/O functions I already had, and so I wrote similar save/load methods for the substruct index https://github.com/astolman/katana-enterprise/blob/559f7f56e748d25ce95d59308bfe3f0dfbd1c4f0/plugins/rdkit/src/RDKSubstructIndex.cpp#L136



I'll add support for the substructure index as well. It'll be easy enough now that the plumbing is all in.

libtsuba/src/RDGPartHeader.cpp
yshalabi: Why are we coupling the RDG storage format version with the version of the optional structures?

yshalabi:
> I'm sure I could figure something out, 
Yeah, I think you can. You are already almost there, you have a class that these datastructures are going to inherit from. Am I missing why this cant be solved with some inheritance to get the manifest keys (and whatever else is needed) from the derived class? 

> but this is supposed to be a quick turn around thing

I dont like feeling like we are modifying our RDG format haphazardly. Even if we have that wonderful experimental flag to contain the potential impact :D. At the end of the day the requests which triggered these efforts came out of nowhere. We can take our time to do this right.

Besides, at the end of the day this PR doesnt eliminate the fact that the user will still need to be prepared for the datastructure to fail loading. So, in some ways we are still at status quo.

IMO it is worthwhile to iterate on this implementation to get to the point where we dont have to export these classes to katana open. Also, the coupling of the RDG version with the versions of these data structure is something I would like us to understand more.

e-mcginnis:
>Besides, at the end of the day this PR doesnt eliminate the fact that the user will still need to be prepared for the datastructure to fail loading. 

The consumer of any optional datastructure API should be ready to generate the datastructure if it is not available. Generating could be done automatically, or we could inform the user that they can use `some_function()` to generate them and let them decide if they want to eat the generation cost now.  
This is already the case with optional topologies. They are generated if not available on storage.
The API
```
  Result<RDKLSHIndex> LoadRDKLSHIndex() { return rdg_.LoadRDKLSHIndex(); }
```
can be handled like 
```
auto res = pg.LoadRDKLSHIndex()
if (!res) {
// Generate the RDKLSHIndex instead
}
```
I could see an argument for the API to `throw` instead though. Not really sure what is better in this context. 

> IMO it is worthwhile to iterate on this implementation to get to the point where we dont have to export these classes to katana open

In general, I think we should aim for simplicity if we can. Putting things in open-katana should not be see as a bad thing if it makes implementation easier. That said, if something can't be in open, then it can't be in open. We should discuss with stakeholders on which this is. 

> Also, the coupling of the RDG version with the versions of these data structure is something I would like us to understand more.

From an idealistic standpoint, 
1) Anything in the RDG folder is part of the RDG. 
2) Anything specifying how the RDG is stored on disk is part of the RDGs storage format. 
3) Any changes to the RDG storage format, should necessitate a change to the RDG storage format version. 

Of course, those are just ideals. Our current implementation makes it a little difficult to perfectly enforce them. But, the test https://github.com/KatanaGraph/katana/blob/master/python/test/test_storage_format.py attempts to enforce those ideals.
It takes advantage of the fact that our single-host RDGs are entirely deterministic (aside for the random string attached to file names), and relies on a "maximal" rdg, aka an RDG with all optional parts generated. This maximal rdg is generated by code at the most recent known good storage format version. We then load the RDG with the current code, load all of the optional parts into memory, and store it. 
Since the data didn't change, and we are deterministic, the newly stored RDG should be identical to the original. So we go through all of the files in the 2 RDG directories and ensure their data matches. 
This test runs in CI, and fails if it sees any file in the two RDGs don't match. Bumping the storage format version bypasses this test, as we expect them to not match. 

So the format of these optional data structures is protected by the storage format version of the RDG.

libtsuba/src/RDGPartHeader.cpp
e-mcginnis: The optional structures will, in most cases, get generated from the graphs they are attached to. This means that the optional structures are really only useful on a specific RDG, at a specific RDG storage format version. 
So it makes sense to extend the RDG storage format version to cover optional structures, which extends the protection from unintended changes.
This could be made more complicated later if we like.

e-mcginnis:
@astolman the RDKSubstructureIndex can be loaded/stored in the RDG using the following propertygraph API endpoints:
```
  Result<RDKSubstructureIndex> LoadRDKSubstructureIndex() {
    return rdg_.LoadRDKSubstructureIndex();
  }

  Result<void> WriteRDKSubstructureIndex(RDKSubstructureIndex& index) {
    return rdg_.WriteRDKSubstructureIndex(index);
  }

```

the substructure index is defined at https://github.com/KatanaGraph/katana/pull/846/files#diff-ce0d2256c0d8ef4a88b60c9fb906972f947286b5de90b613ae0841f10336c915

libtsuba/src/RDG.cpp
e-mcginnis: part header api now returns the following error on failure:
```
    return KATANA_ERROR(
        katana::ErrorCode::InvalidArgument,
        "No optional datastructure manifest available matching [{}]",
        optional_datastructure_name);
```

libtsuba/src/RDG.cpp
e-mcginnis: the rdg api now returns the following error on failure:
```
  katana::RDKLSHIndex index = KATANA_CHECKED_CONTEXT(
      katana::RDKLSHIndex::Load(rdg_dir(), path),
      "Failed to load RDKLSHIndex located at {}", path);
```

libtsuba/include/katana/RDGOptionalDatastructure.h
```diff
@@ -0,0 +1,84 @@
...
+
+namespace katana {
+
+/// Base class for all optional datastructures.
+/// Paths to the RDGOptionalDatastructure files are stored in the RDGPartHeader::optional_datastructure_manifests_
+class KATANA_EXPORT RDGOptionalDatastructure {
+public:
+  std::unordered_map<std::string, std::string> paths() const { return paths_; }
+  void set_paths(std::unordered_map<std::string, std::string> paths) {
+    paths_ = paths;
```
witchel: I think it is a valuable principle to avoid non-determinism when it is computationally cheap.  I prefer `std::map` to `std::unordered_map` for small maps that get serialized, because I think the advantages of a deterministic order outweigh any performance benefit.

roshandathathri:
@e-mcginnis My quick two cents.

(1) Can we have a dynamic registry for optional data structures instead of making them static? 

```cpp
  Result<Buffer> LoadOptionalDataStructure(const std::string& name) {
    ...
  }

  Result<void> WriteOptionalDataStructure(const std::string& name, Buffer&& buf) {
    ....
  }
 ```
 
Then the users are responsible for (de)serializing their data structure. Also, we can then move the data-structure details itself to enterprise. 
 
(2) In this case, indexes must exist if they are called with a "name", so if they don't exist, it is indeed an error. If you want it to be truly optional, then you should return `Result<std::optional<...>>`. This will make the API/semantics much more clear. It's bad practice to handle an error as "not available" behavior because the error could be for different reasons.

witchel (commented):
I don't have time to finish this review, I'll just throw in this comment.

e-mcginnis:
@roshandathathri 

>(2) In this case, indexes must exist if they are called with a "name", so if they don't exist, it is indeed an error. If you want it to be truly optional, then you should return Result<std::optional<...>>. This will make the API/semantics much more clear. It's bad practice to handle an error as "not available" behavior because the error could be for different reasons.

Agreed, I will change the API to use std::optional


> Then the users are responsible for (de)serializing their data structure. Also, we can then move the data-structure details itself to enterprise.

I understand the desire from the user side of things to push the RDG layer aside and just handle the (de)serialization, it provides a lot of flexibility. But that flexibility comes with a bunch of drawbacks.

1) No centralization of serialization functions. There is no reason we should maintaining multiple ways to serialize common data strcutures for storage. Libtsuba, and the RDG, already have easy to extend ways to serialize things. We should take advantage of them, and improve on them, when we can. 

2) libtsuba & the RDG are now responsible for maintaining any number of arbitrarily structured, arbitrarily changing files. The RDG is responsible for files inside the RDG directory, and the storage format version protects us from unintended structural changes to those files. Should these optional files be excluded from that protection?
    - If Yes, the optional files should not be protected by the storage format version:
      -  Does every user have to implement their own data structure versioning, and CI tests to catch unintended changes to those specific structures? 
        -  The reliability of these matters because uncaught, unintended changes to these structures can either:  fail spectacularly or silently provide incorrect answers 
    - If No, the optional files should be protected by the storage format version:
      - Now libtsuba is responsible for catching changes to data structures it does not know the definition for. If libtsuba cannot create them, it is difficult to test for their presence let alone changes in them. 


All of this raises the question: What is the RDG for? Is the RDG for persisting a representation of the in memory graph, or is the purpose broader and the RDG is for persisting data necessary for the katana software stack to work efficiently?


All of this should really be discussed further in an RFC, so I'm writing one up. I don't think discussion will be over before we need this though. 

If we decide to go ahead and merge this, I'll be more than happy to rip it out later if we decide to go a different direction.

roshandathathri:
Sorry for the delay in response. 

> 1. No centralization of serialization functions.

Disagree. In terms of design, IMO serialization should be independent of communication/storage interface. That's what our NetworkInterface and gSerialize aim to do. NetworkInterface should/does NOT handle serialization. There's a common set of APIs to do serialization. Can we do something similar for storage as well? @tylershunt 

> 2\. libtsuba & the RDG are now responsible for maintaining any number of arbitrarily structured, arbitrarily changing files

I think this is orthogonal. All I'm saying is that the storage/RDG interface receives a "serialized buffer", a "name/key" for that buffer, and the "RDG" to store it in (maybe whether it updates the version or not). Why would such an interface need to know how that serialized buffer was created?

e-mcginnis:
> That's what our NetworkInterface and gSerialize aim to do. NetworkInterface should/does NOT handle serialization. There's a common set of APIs to do serialization. 

I'm all for having a common set of APIs to do serialization outside of libtsuba, but the fact is right now most serialization for storage happens in libtsuba. Also, comparing our persistent storage serialization to temporary on the wire serialization is flawed. For persistent serialization, we insist on being backwards compatible with all previous serializations. This is not necessarily a requirement for our network serialization. These two cases of serialization are attempting to solve similar problems with different goals, and so likely will look different.

>Why would such an interface need to know how that serialized buffer was created?

Strictly as proposed, the interface doesn't need to. But some layer has to ensure that all previous serializations of the buffer are still properly supported, as backwards compatibility is a concern. 
A more subtle issue is consistency. We want loading an arbitrary data structure, which holds data `A` from storage, which was stored at some previous storage_format_version `N` to look identical in memory to the same data structure, holding data `A`, freshly generated now at storage_format_version `M`. This is a difficult situation to detect breakages in. 

The RDG and libtsuba already have tools to handle these requirements. Why offload the backwards compatibility, and consistency requirements to other parts of the codebase?

astolman (approved):
I can't say much about the changes to libtsuba, etc, but the API looks like it's got what I need, the exposure of rdkit to open looks benign, and I would like this merged ASAP, so I'm giving my approval.

roshandathathri:
> I'm all for having a common set of APIs to do serialization outside of libtsuba, but the fact is right now most serialization for storage happens in libtsuba. Also, comparing our persistent storage serialization to temporary on the wire serialization is flawed. For persistent serialization, we insist on being backwards compatible with all previous serializations. This is not necessarily a requirement for our network serialization. These two cases of serialization are attempting to solve similar problems with different goals, and so likely will look different.

Agree. The serialization functions/API are part of libtsuba. However, the user can invoke it on their data structures and combine them into a single file or multiple files as they need. That logic need not be tied inside libtsuba in my point.

> Strictly as proposed, the interface doesn't need to. But some layer has to ensure that all previous serializations of the buffer are still properly supported, as backwards compatibility is a concern.

I'm not sure I understand. The serialization is still controlled by libtsuba but all the buffers/files and their contents.

PR #847, opened 2022-01-29
[KAT-2749] Add bindings for certain additional entity type manager methods

denish44:
EntityTypeManager have some methods that aren't exposed in the python binding. This PR adds additional methods to the python wrapper which logically makes sense to expose.

Jira: KAT-2749

### DIFF
--- a/libkatana_python_native/src/EntityTypeManager.cpp
+++ b/libkatana_python_native/src/EntityTypeManager.cpp
@@ -34,13 +34,14 @@ struct AtomicEntityType : public EntityType {
   std::string name() { return ToString(); }
 };
 
-katana::SetOfEntityTypeIDs GetSetOfEntityTypeIds(std::vector<EntityType> types) {
-    katana::SetOfEntityTypeIDs type_ids;
-    type_ids.resize(types.size());
-    for (auto t : types) {
-      type_ids.set(t.type_id);
-    }
-    return type_ids;
+katana::SetOfEntityTypeIDs
+GetSetOfEntityTypeIds(std::vector<EntityType> types) {
+  katana::SetOfEntityTypeIDs type_ids;
+  type_ids.resize(types.size());
+  for (auto t : types) {
+    type_ids.set(t.type_id);
+  }
+  return type_ids;
 }
 
 }  // namespace
@@ -108,8 +109,8 @@ katana::python::InitEntityTypeManager(py::module_& m) {
              std::vector<EntityType> types) -> Result<EntityType> {
             auto set_of_type_ids = GetSetOfEntityTypeIds(types);
             return EntityType{
-                self,
-                KATANA_CHECKED(self->GetOrAddNonAtomicEntityType(set_of_type_ids))};
+                self, KATANA_CHECKED(
+                          self->GetOrAddNonAtomicEntityType(set_of_type_ids))};
           })
       .def(
           "type_from_id",
@@ -124,11 +125,12 @@ katana::python::InitEntityTypeManager(py::module_& m) {
           "get_atomic_subtypes",
           [](const katana::EntityTypeManager* self, EntityType type) {
             py::set ret;
-            if(self -> HasEntityType(type.type_id)) {
-              auto type_set = self -> GetAtomicSubtypes(type.type_id);
-              for (EntityTypeID subtype_id = 0; subtype_id < type_set.size(); ++subtype_id) {
+            if (self->HasEntityType(type.type_id)) {
+              auto type_set = self->GetAtomicSubtypes(type.type_id);
+              for (EntityTypeID subtype_id = 0; subtype_id < type_set.size();
+                   ++subtype_id) {
                 if (type_set.test(subtype_id)) {
-                  if(self -> GetAtomicTypeName(subtype_id).has_value()) {
+                  if (self->GetAtomicTypeName(subtype_id).has_value()) {
                     ret.add(AtomicEntityType{{self, subtype_id}});
                   }
                 }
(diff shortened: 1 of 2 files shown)


denish44:
@bowu @arthurp This PR adds methods I think makes sense for exposure. Please let me know if there is need to expose additional things. But think this in general provides enough functionality to do what's needed with entity type manager.

bowu:
I added @sanchveda as a reviewer because he tried to implement NARS in Katana, which needed type info exposed in this PR. @sanchveda Could you please let us know whether the API is sufficient for you to implement NARS?

libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -103,5 +120,42 @@ katana::python::InitEntityTypeManager(py::module_& m) {
...
+                }
+              }
+            }
+            return ret;
+          })
+      .def(
+          "get_num_atomic_types", &katana::EntityTypeManager::GetNumAtomicTypes)
+      .def(
+          "get_num_of_entity_types",
+          &katana::EntityTypeManager::GetNumEntityTypes);
```
arthurp: `GetNumEntityTypes` should be called as `num_types` (all types are entity types). `GetNumAtomicTypes` should be called as `num_atomic_types`.

The `Get` at the beginning is an oddity even in C++.

libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -103,5 +120,42 @@ katana::python::InitEntityTypeManager(py::module_& m) {
...
+          "get_atomic_subtypes",
+          [](const katana::EntityTypeManager* self, EntityType type) {
+            py::set ret;
+            if (self->HasEntityType(type.type_id)) {
+              auto type_set = self->GetAtomicSubtypes(type.type_id);
+              for (EntityTypeID subtype_id = 0; subtype_id < type_set.size();
+                   ++subtype_id) {
+                if (type_set.test(subtype_id)) {
+                  if (self->GetAtomicTypeName(subtype_id).has_value()) {
+                    ret.add(AtomicEntityType{{self, subtype_id}});
```
arthurp: You don't need the check. The type is an atomic type since it was returned from `GetAtomicSubtypes`.

libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -103,5 +120,42 @@ katana::python::InitEntityTypeManager(py::module_& m) {
...
-          });
+          })
+      .def(
+          "get_atomic_subtypes",
+          [](const katana::EntityTypeManager* self, EntityType type) {
+            py::set ret;
+            if (self->HasEntityType(type.type_id)) {
+              auto type_set = self->GetAtomicSubtypes(type.type_id);
+              for (EntityTypeID subtype_id = 0; subtype_id < type_set.size();
+                   ++subtype_id) {
```
arthurp, line 133: This isn't very efficient. It iterates over every bit in the vector, not just the ones. I don't think we should fix it now, but please add a TODO and file an issue against `DynamicBitSet` to add a method that efficiently finds ones (by iterating over 64-bit words and doing bit magic). That would help a lot with this and should be reasonably fast.

arthurp (changes requested):
If there are any methods that will take more than a trivial amount of time (microseconds), add `py::call_guard<py::gil_scoped_release>()` as discussed here: https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil. Make sure the method doesn't access any python objects though.

denish44:
> If there are any methods that will take more than a trivial amount of time (microseconds), add `py::call_guard<py::gil_scoped_release>()` as discussed here: https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil. Make sure the method doesn't access any python objects though.

I don't think we need to worry about this. with the new exposure and old exposure, I think the worst case scenario is O(types) in any method which has a numerical upper bound as we saw so mostly O(1) theoretically.

sanchveda (commented):
@bowu I guess for NARS, the only thing I would require is to get a view from a set of different types and then a way to get the local ids of the individual nodes in the view. That should be sufficient of it's implementation.

denish44:
> @bowu I guess for NARS, the only thing I would require is to get a view from a set of different types and then a way to get the local ids of the individual nodes in the view. That should be sufficient of it's implementation.

node ids? this manager is only for type ids ? I don't think this has anything to do with views unless i misunderstood something. tagging @arthurp too just incase.

bowu:
@denish44 What Sanchayan meant is that he mainly cares about projected view instead of types. I believe Arthur is working on view APIs.

arthurp:
> @denish44 What Sanchayan meant is that he mainly cares about projected view instead of types. I believe Arthur is working on view APIs.

Yep. Though projected views will come slightly after the other views because it's getting a rework in C++ right now. Should land soon.

python/test/test_property_graph.py
```diff
@@ -350,16 +351,18 @@ def test_types(graph):
...
         graph.node_types.get_non_atomic_entity_type([node_atomic_types["Message"], node_atomic_types["Post"]])
         == non_atomic_type
     )
     assert graph.node_types.get_atomic_subtypes(non_atomic_type) == set(
         [node_atomic_types["Message"], node_atomic_types["Post"]]
     )
-    assert graph.node_types.get_supertypes(node_atomic_types["Message"]) == set(
-        [node_atomic_types["Message"], non_atomic_type]
+    assert graph.node_types.get_supertypes(node_atomic_types["Message"]).issuperset(
+        set([node_atomic_types["Message"], non_atomic_type])
```
arthurp: NIT: You shouldn't need the set constructor here. You can pass any collection, I think. Also Set literals are supported in python. They look like this: `{a, b, c}`. They look a LOT like dict literals though (only differing by the colons), so I kinda prefer the constructor form anyway.

arthurp (approved):
LGTM.

PR #848, opened 2022-01-29
tsuba: Fix cache eviction invalidation when adding properties

ddn0:
LoadProperty takes a table named props and a cache and an RDG. Props
itself is just a pointer from an RDG. The cache also maintains a
reference to RDGs. Cache operations may change RDGs which in turn may
change the props table. This can lead to problems if information about
props is computed at some point and a cache operation happens before the
props information is used.

In particular, this can lead to an error adding a column (index >
num_columns) to props when the index i is computed before AddProperties
executes and then is used in the callback for AddProperties.


tylershunt (approved):
Thanks. Yes this makes sense.

witchel (approved):
Thank you.

PR #849, opened 2022-01-29
tsuba: Fix cache eviction invalidation when adding properties

ddn0:
LoadProperty takes a table named props and a cache and an RDG. Props
itself is just a pointer from an RDG. The cache also maintains a
reference to RDGs. Cache operations may change RDGs which in turn may
change the props table. This can lead to problems if information about
props is computed at some point and a cache operation happens before the
props information is used.

In particular, this can lead to an error adding a column (index >
num_columns) to props when the index i is computed before AddProperties
executes and then is used in the callback for AddProperties.

### DIFF
--- a/libtsuba/src/RDG.cpp
+++ b/libtsuba/src/RDG.cpp
@@ -678,10 +678,6 @@ LoadProperty(
     tsuba::NodeEdge node_edge, tsuba::PropertyCache* cache, tsuba::RDG* rdg,
     std::vector<tsuba::PropStorageInfo>* prop_info_list,
     const katana::Uri& dir) {
-  if (i < 0 || i > props->num_columns()) {
-    i = props->num_columns();
-  }
-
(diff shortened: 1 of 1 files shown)


ddn0:
Note: this is on the release branch

witchel (approved):
This is a great catch.  It seems like props comes (half the time) from `RDGCore::node_property()`, which is the loaded properties and by the time the callback gets serviced, the set of loaded properties might have changed.

KAT-2854 is on my plate and eliminates property access via index, but whatever changes I make, I'll go over the design with Tyler to make sure we are confident moving forward.

As discussed with Donald on slack, I will also think about better testing coverage for caching.  I have tests for the case without a cache because that is the code path used on a cache miss, but I will look at configuring a small cache to get coverage of property replacement.

roshandathathri (approved):
Thanks!

PR #851, opened 2022-01-31
Property Graph 2.0 and port graphs to Pybind11

arthurp:
On top of: https://github.com/KatanaGraph/katana/pull/863

With: https://github.com/KatanaGraph/katana-enterprise/pull/3168

~Not a complete PR. But I wanted to give people a chance to look at it before it's complete to speed things along.~ This is now complete and almost ready to merge.


arthurp:
The largest lacking thing at this point is that many tests have not been updated to the PropertyGraph 2.0 API. Also, a few things are still using the old graph object, this needs to be eliminated clearly.

The old graph object is going to stay around though, so it can still be used in enterprise while I port DistPropertyGraph.

libkatana_python_native/include/katana/python/ErrorHandling.h
```diff
@@ -4,30 +4,60 @@
...
+/// Convert @p err to a python exception and raise it.
+/// Always throws pybind11::error_already_set.
+KATANA_EXPORT pybind11::handle RaiseResultException(
+    const katana::ErrorInfo& err);
+
+}  // namespace detail
+
+template <typename T>
+T
+PythonChecked(const katana::Result<T>& src) {
```
denish44: Is this suppose replace KATNA_CHECKED i saw in places?

denish44 (approved):
LG

libkatana_python_native/include/katana/python/ErrorHandling.h
arthurp: It doesn't need to. Though I did think about doing that. I think we should stick to using `KATANA_CHECKED` when possible. But it's not possible in all cases since some functions must have specific return types such as init implemenations.

arthurp:
@pyrito @mindv0rtex This PR is now ready for review.

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -120,27 +137,146 @@ struct DefCopy<T, std::void_t<decltype(T((const T&)std::declval<T>()))>> {
...
+}
+
+namespace detail {
+
+template <typename T>
+using has_size_t = decltype((size_t)std::declval<T>().size());
+
+template <typename ClassT>
+ClassT
+DefLen(ClassT& cls) {
```
yan-zaretskiy: Could you move it up next to other defs?

libkatana_python_native/include/katana/python/CythonIntegration.h
```diff
@@ -132,16 +132,17 @@ namespace katana {
 
 // Add Cython classes here as needed. Remove them when they are moved to
 // pybind11.
-CYTHON_REFERENCE_SUPPORT(katana::PropertyGraph, "katana.local", "Graph");
-CYTHON_REFERENCE_SUPPORT(katana::TxnContext, "katana.local", "TxnContext");
+//CYTHON_REFERENCE_SUPPORT(katana::PropertyGraph, "katana.local", "Graph");
```
yan-zaretskiy: Do you still need to commented out line?

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -206,6 +263,24 @@ DefWithNumba(
...
+constexpr void
+DefWithNumba(
+    pybind11::class_<Cls, ClsExtra...>& cls,
+    std::enable_if_t<
+        std::is_same_v<
+            typename detail::StaticFunctionBinder<func>::StaticFunctionPointer,
+            decltype(func)>,
+        const char>* name,
+    const Extra&... extra) {
+  static_assert(
```
yan-zaretskiy: Here and above, why do you use `static_assert` with the same condition as inside `std::enable_if_t`?

libkatana_python_native/include/katana/python/TemplateSupport.h
```diff
@@ -22,13 +28,41 @@ pybind11::object
...
+template <typename... Types, typename F>
+void
+InstantiateForTypes(pybind11::module m, const std::string& basename, F f) {
+  pybind11::module builtins = pybind11::module::import("builtins");
+  auto make_template_type1 = pybind11::cast<pybind11::function>(
+      pybind11::module::import("katana.native_interfacing.template_type")
+          .attr("make_template_type1"));
+  pybind11::dict types;
+
+  auto t = std::make_tuple(InstantiateForType<Types>(m, basename, f, types)...);
```
yan-zaretskiy: You don't really need to make a tuple here - it's not used anywhere, as far I can see. Could you just fold over the comma operator instead?

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,748 @@
...
+};
+
+LazyDataAccessor::~LazyDataAccessor() = default;
+
+template <typename T>
+class LazyDataAccessorTyped : public LazyDataAccessor {
+public:
+  virtual T at_typed(ssize_t i) const = 0;
+  virtual py::object at(ssize_t i) const final { return py::cast(at_typed(i)); }
+  virtual py::object array(const py::slice& slice) const {
```
yan-zaretskiy: Nitpick: you don't need `virtual` is derived classes, just `override` or `final`.

libkatana_python_native/include/katana/python/CythonIntegration.h
arthurp: I was keeping it as an example: I'll add "For example" to the comment to make that clear.

libkatana_python_native/include/katana/python/TemplateSupport.h
arthurp: I don't think you can do that. Any expanded argument pack needs to be an argument to something else, I think. People often use arrays, I just used tuple here do that there is no need to have a dummy value of a consistent type (eg. `(code, 0)...`).

I'll give it a quick try though.

PS: Pedantry warning. But this is a map not a fold.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: Aaagh. I tried to get this right and failed anyway. I'll fix this. :-) Thanks.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,748 @@
...
+    cls.template def("__getitem__", [](Cls& self, const std::string& v) {
+      return *self.Find(v);
+    });
+    cls.template def("find_all", [](Cls& self, const std::string& v) {
+      return py::make_iterator(self.LowerBound(v), self.UpperBound(v));
+    });
+
+    //    katana::DefConventions(cls);
+
+    return std::move(cls);
```
yan-zaretskiy, line 690: Does this have to be wrapped in `std::move`?

yan-zaretskiy (approved):
Outstanding. I only left very minor nits and a couple of questions.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: It doesn't. I removed the move.

arthurp: I was wrong. It is needed. For some reason I get a "copied despite being returned by name" warning. Doing the move is harmless, so I'll just leave it there.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
...
+
+  py::class_<PropertyGraph, std::shared_ptr<PropertyGraph>> cls(m, "Graph");
+
+  katana::DefCythonSupport(cls);
+  katana::DefConventions(cls);
+  katana::RegisterNumbaClass(cls);
+
+  cls.def(
+      py::init(
+          [](py::object path,
```
pyrito, line 171: Why does this have to be a `py::object`? Can't this just be a `py::str` directly? Or is this so we can catch type errors when casting? Or if we pass in paths like `url.URL` or `pathlib.Path` that have string representations?

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
...
+             std::optional<std::vector<std::string>> node_properties,
+             std::optional<std::vector<std::string>> edge_properties,
+             TxnContext* txn_ctx) -> std::shared_ptr<PropertyGraph> {
+            auto path_str = py::str(path).cast<std::string>();
+            py::gil_scoped_release guard;
```
pyrito, line 176: so we release the GIL for the rest of this function? After this scope is the GIL reacquired?

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
...
+          properties are loaded.
+      )""");
+
+  // Below are the view API methods that we have in mind (not all of them may end up being exposed to Python). We differentiate between 3 classes of identifiers for nodes and edges:
+  // Local ID: This is an ID produced by a specific partitioning of a distributed graph.
+  // Topology handle: It is a handle used to relate to an entity in a topology data structure. It is a handle that various topology-related methods return.
+  // Property index: This is used to access that entity’s properties.
+  // We allow conversions between them in one direction: topology handle -> local ID -> property index.
+
+  katana::DefWithNumba<&PropertyGraph::NumNodes>(cls, "num_nodes");
```
pyrito, line 205: I always forget that we can actually pass references as template parameters...

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
...
+      py::arg("provenance") = py::str("katana.local"),
+      py::call_guard<py::gil_scoped_release>(),
+      R"""(
+      Write the property graph to the specified path or URL (or the original path it was loaded from if path is
+      not provided). Provide lineage information in the form of a command line.
+
+      :param path: The path to which to write or None to use `self.path`.
+      :type path: str or Path
+      :param command_line: Lineage information in the form of a command line.
+      :type command_line: str
```
pyrito, line 647: Will this string be formatted correctly?

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
...
+      KATANA_CHECKED(
+          arrow::py::NumPyDtypeToArrow(pyarray.dtype().ptr(), &dtype));
+      KATANA_CHECKED(arrow::py::NdarrayToArrow(
+          arrow::default_memory_pool(), v.ptr(), nullptr, true, dtype,
+          arrow::compute::CastOptions::Safe(dtype), &array));
+      KATANA_LOG_DEBUG_ASSERT(array);
+    } else {
+      // Convert any other python sequence.
+      arrow::py::PyConversionOptions options;
+      options.from_pandas = true;
```
pyrito, line 55: This is super cool. I wonder if we can do something similar with Dask. It should be possible, but it may just require some trickery.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
...
+#include "katana/python/NumbaSupport.h"
+#include "katana/python/PropertyGraphPython.h"
+#include "katana/python/PythonModuleInitializers.h"
+#include "katana/python/TemplateSupport.h"
+#include "katana/python/TypeBans.h"
+
+namespace py = pybind11;
+
+katana::Result<std::shared_ptr<arrow::Table>>
+katana::python::PythonArgumentsToTable(
```
pyrito, line 29: This is a really cool method and gives me hope that we can more naturally create RDGs from dataframes without having to go through a round-about of calling CSV import.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,742 @@
+#define NPY_NO_DEPRECATED_API NPY_1_11_API_VERSION
```
pyrito, line 1: What's this?

cmake/Modules/BuildCommon.cmake
```diff
@@ -241,9 +241,20 @@ endif ()
 message(STATUS "Building Katana language bindings: ${KATANA_LANG_BINDINGS}")
 
 if (python IN_LIST KATANA_LANG_BINDINGS)
-  include(FindPythonModule)
-  find_package (Python3 COMPONENTS Interpreter Development NumPy)
+  # Make sure any python virtual environment (such as conda) is preferred.
+  set (Python3_FIND_VIRTUALENV FIRST)
+
+  # The NumPy component of FindPython3 does not work correctly in conda-build
+  # (it works find in normal conda environments), so use SP_DIR, which is the
```
amberhassaan: s/find/fine

libkatana_python_native/include/katana/python/CythonIntegration.h
```diff
@@ -151,16 +153,20 @@ DefCythonSupport(pybind11::class_<T, Extra...> cls) {
...
-        auto* ptr = reinterpret_cast<std::shared_ptr<T>*>(addr);
-        return *ptr;
-      });
-  cls.template def_static(
-      "_make_from_address_unique", [](uintptr_t addr) -> std::unique_ptr<T> {
-        auto* ptr = reinterpret_cast<std::unique_ptr<T>*>(addr);
-        return std::move(*ptr);
-      });
+  if constexpr (std::is_same_v<Holder, std::shared_ptr<T>>) {
+    cls.template def_static(
```
pyrito, line 158: What's going on here exactly?

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -241,7 +237,7 @@ GetRangeStep(T& self) -> std::remove_reference_t<decltype(*self.begin())> {
     return 1;
   }
   auto next = self.begin();
-  next++;
+  std::advance(next, 1);
```
pyrito: What's the need for this method as opposed to just incrementing the iterator?

libkatana_python_native/include/katana/python/Conventions.h
```diff
@@ -244,7 +237,7 @@ GetRangeStep(T& self) -> std::remove_reference_t<decltype(*self.begin())> {
...
```
amberhassaan, line 246: This method is unclear to me. Why isn't its return type `std::ptrdiff_t`?

libkatana_python_native/include/katana/python/NumbaSupport.h
```diff
@@ -255,10 +255,6 @@ DefWithNumba(
             decltype(member_func)>,
         const char>* name,
     const Extra&... extra) {
-  static_assert(std::is_same_v<
```
amberhassaan, line 258: why remove these static asserts?

libkatana_python_native/include/katana/python/PropertyGraphPython.h
```diff
@@ -33,4 +34,16 @@ PythonArgumentsToTable(
...
+
+template <typename T>
+struct ::pybind11::detail::type_caster<std::unique_ptr<katana::EntityIndex<T>>>
+    : public katana::python::banned_type_caster<
+          std::unique_ptr<katana::EntityIndex<T>>> {};
+
+template <>
+struct ::pybind11::detail::type_caster<std::unique_ptr<katana::PropertyGraph>>
+    : public katana::python::banned_type_caster<
+          std::unique_ptr<katana::PropertyGraph>> {};
```
pyrito: This is interesting, so basically if someone tries to create a `std::unique_ptr` reference to PropertyGraph they have to cast it in this way and this struct will throw an error when the cast tries to happen? Do you have a clearer example of incorrect usage? It might help me picture how the code here comes into play.

pyrito (approved):
LGTM. Magical stuff

libkatana_python_native/src/PropertyGraph.cpp
arthurp: Yes. I want this to be able to accept `Path`.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: Yes.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: Actually you cannot pass references. What you can pass is pointers to functions (and maybe variables of specific kinds).

libkatana_python_native/src/PropertyGraph.cpp
arthurp: I'm not aware of any formatting requirements. I honestly think we need to get ride of it. I think it has been subsumed by the op log.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: You mean access dask native code from our native code? IT might be possible, I haven't looked. I think dask is almost entirely python though. Not much actual native code other than numpy.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: :-) Yea. You can already write things like `g.add_node_property(pandas_df)`. I think it's tested in the pytests.

libkatana_python_native/src/PropertyGraph.cpp
pyrito: Oh right, yes stupid mistake, you're passing in a function pointer. My bad.

libkatana_python_native/src/PropertyGraph.cpp
pyrito: Not Dask native code, but maybe construct arrow objects by getting the underlying pandas partitions and trying to create arrow tables from there. Something along those lines.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: It tells the numpy native headers what API version I expect. That way it will not issue specific warnings. I cannot exactly remember. https://numpy.org/devdocs/reference/c-api/deprecations.html#deprecation-mechanism-npy-no-deprecated-api

libkatana_python_native/include/katana/python/CythonIntegration.h
arthurp: I'm only adding this method if Holder is `shared_ptr`. This prevents creating objects that will provide lifetime issues since pybind11 is not dynamically flexible about this and has no static or runtime checking.

libkatana_python_native/include/katana/python/Conventions.h
arthurp: It's just the more STL correct way to do it. I think there are some esoteric iterators that support `advance` but not `++`. I think it's for things that should appear as iterators even though they don't natively support the concept. This was just a clean up.

libkatana_python_native/include/katana/python/PropertyGraphPython.h
arthurp: The caster is used by pybind11 when casting objects to/from python objects. This caster causes a *static* assert when this happens.  At least in some cases. There is lengthy documentation in `BanTypes.h`.

This will not affect use of unique_ptr<PropertyGraph> at the C++ level.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: Ah. Yea. We might be able to do that. But it might be easier to get numpy versions of the underlying partitions and work from there. We'll see. Either is possible. pybind11 has rather nice support for numpy.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -0,0 +1,753 @@
...
+      py::overload_cast<GraphTopologyTypes::Node>(
+          &PropertyGraph::OutEdges, py::const_),
+      py::call_guard<py::gil_scoped_release>());
+
+  // OutEdges(NodeHandle, EdgeEntityTypeID) -> OutEdgeHandle iterator - iterator over out-edges of a type for a node
+  cls.def(
+      "out_edge_ids",
+      [](PropertyGraph& self, GraphTopologyTypes::Node n,
+         const EntityType& ty) {
+        return self.BuildView<PropertyGraphViews::EdgeTypeAwareBiDir>()
```
amberhassaan, line 296: so does this mean we can't keep around a `view` object in python?

amberhassaan (commented):
Left some comments.

libkatana_python_native/include/katana/python/NumbaSupport.h
arthurp: They are redundant with the enable if above them.

libkatana_python_native/src/PropertyGraph.cpp
arthurp: Access views are abstracted away in Python yes. You just get the universal view which allows everything. We may have to allow explicit instances of views for numba compiled algorithm implementations, but that should be a special case not the normal usage.

libkatana_python_native/include/katana/python/Conventions.h
arthurp: No. This is a difference between values, not iterators. And the values could be of some other type. I maybe should use a signed variant of the value though.

arthurp: I'll add documentation to the function. The gist is that is returns the difference between values in an iterator which act like `range` does in python. That is returning strides integer values over some range of values.

libkatana_python_native/include/katana/python/PropertyGraphPython.h
pyrito: Gotcha. Thanks for the clarification.

PR #854, opened 2022-02-01
Maintain Shared Pointer to PropertyGraph in PropertyGraphViews [KAT-3405]

rob-isaac:
Currently, views maintain raw pointers to the PropertyGraph rather than shared pointers. This can be unsafe because if the PropertyGraph is destroyed before the View, the view will either segfault or return garbage values.

This PR changes the View structures to maintain shared pointers to the PropertyGraph. It additionally removes the PropertyGraph::BuildView functions in favor of PropertyGraphView::Make functions.

This is WIP primarily because I need to wait until the pybind11 wrappings are in to get the python side working. The c++ code should be about ready.

### DIFF
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -1135,11 +1135,17 @@ private:
 class KATANA_EXPORT ProjectedPropGraphViewWrapper : public GraphTopologyTypes {
 public:
   explicit ProjectedPropGraphViewWrapper(
-      const PropertyGraph* pg,
+      std::shared_ptr<const PropertyGraph> pg,
       std::shared_ptr<const ProjectedTopology> projected_topo) noexcept
-      : prop_graph_(pg), projected_topo_ptr_(std::move(projected_topo)) {
+      : prop_graph_(std::move(pg)),
+        projected_topo_ptr_(std::move(projected_topo)) {
     KATANA_LOG_DEBUG_ASSERT(projected_topo_ptr_);
   }
+
+  static std::unique_ptr<ProjectedPropGraphViewWrapper> Make(
+      std::shared_ptr<PropertyGraph> pg,
+      const std::vector<std::string>& node_types,
+      const std::vector<std::string>& edge_types) noexcept;
 
   auto NumNodes() const noexcept { return topo().NumNodes(); }
 
@@ -1200,7 +1206,9 @@ public:
     return topo().original_to_projected_edge_id(eid);
   }
 
-  const PropertyGraph* property_graph() const noexcept { return prop_graph_; }
+  const PropertyGraph* property_graph() const noexcept {
+    return prop_graph_.get();
+  }
 
   const std::shared_ptr<arrow::Buffer>& node_bitmask() const noexcept {
     return topo().node_bitmask();
@@ -1215,7 +1223,7 @@ protected:
   }
 
 private:
-  const PropertyGraph* prop_graph_;
+  std::shared_ptr<const PropertyGraph> prop_graph_;
   std::shared_ptr<const ProjectedTopology> projected_topo_ptr_;
 };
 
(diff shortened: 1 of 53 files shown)


libgraph/src/analytics/triangle_count/triangle_count.cpp
```diff
@@ -263,15 +263,14 @@ EdgeIteratingAlgo(const SortedGraphView* graph) {
 
 katana::Result<uint64_t>
 katana::analytics::TriangleCount(
-    katana::PropertyGraph* pg, TriangleCountPlan plan) {
+    const SortedGraphView& sorted_view, TriangleCountPlan plan) {
```
arthurp, line 266: Doesn't this break that conventions for analytics operations? Shouldn't this take a property graph and then get the sorted view internally?

libgraph/include/katana/GraphTopology.h
```diff
@@ -1628,181 +1636,48 @@ class BasicPropGraphViewWrapper : public Topo {
...
   explicit BasicPropGraphViewWrapper(
-      PropertyGraph* pg, const Topo& topo) noexcept
-      : Base(topo), prop_graph_(pg) {}
+      std::shared_ptr<PropertyGraph> pg, const Topo& topo) noexcept
+      : Base(topo), prop_graph_(std::move(pg)) {}
+
+  // Note: Can add another Make method which takes a ViewCache
+  //       as a parameter if we ever find that necessary
+  /// Make the PropertyGraph View by calling the appropriate
+  /// BuildOrGet method on the PropertyGraph's ViewCache
```
arthurp, line 1645: Documentation should say what the function does, not how it does it.

libgraph/include/katana/GraphTopology.h
```diff
@@ -1628,181 +1636,48 @@ class BasicPropGraphViewWrapper : public Topo {
...
+  static std::unique_ptr<BasicPropGraphViewWrapper<Topo>> Make(
```
arthurp, line 1646: How does this `unique_ptr` interact with the cache? I thought these views were supposed to be cached?

Even if they are not cached yet, we should probably return a `shared_ptr` here so we don't have to change every user of this code to introduce caching later.

arthurp (commented):
I definitely support using `shared_ptr`. However, I'm kinda confused about the removal of the view builder. Is this just an internal change, where you use a different method for implementing `PropertyGraph::BuildView`? Or is that method gone? If it's gone how is user code supposed to get cached views cleanly?

rob-isaac:
> I definitely support using `shared_ptr`. However, I'm kinda confused about the removal of the view builder. Is this just an internal change, where you use a different method for implementing `PropertyGraph::BuildView`? Or is that method gone? If it's gone how is user code supposed to get cached views cleanly?

PropertyGraph::BuildView is removed to avoid using `std::shared_from_this`. Users instead use `View::Make` and pass in a shared pointer to the PropertGraph. The `Make` methods internally check the PropertyGraph's cache for the *topologies* that it needs (the propertygraph caches topologies, not views) and uses them if available. Otherwise it builds the topology and puts it in the cache.

Basically, the caching behavior is still the same with `View::Make` as it was with `PropertyGraph::BuildView`.

libgraph/src/analytics/triangle_count/triangle_count.cpp
rob-isaac: Yes

libgraph/include/katana/GraphTopology.h
rob-isaac: As I mentioned in the main reply - Views are not cached in the ViewCache b/c view are just a thin wrapper over PropertyGraph + Topologies (and sometimes other DataStructures). The ViewCache caches the expensive objects associated with views (mainly topologies). Thus, views are reconstructed each time, but their members are loaded from cache if possible.

arthurp:
Thanks. This makes sense. BTW, why avoid using `std::shared_from_this` I was actually thinking about introducing it since it would make the pybind11 wrappers significantly less prone to object lifetime mistakes.

libgraph/src/analytics/triangle_count/triangle_count.cpp
arthurp: Are you gonna fix it?

rob-isaac: Yes, this PR is just on the backlog rn

rob-isaac:
> Thanks. This makes sense. BTW, why avoid using `std::shared_from_this` I was actually thinking about introducing it since it would make the pybind11 wrappers significantly less prone to object lifetime mistakes.

I'm not super against it, but I prefer the make semantics for views, which also allows the user to move their DPG into the view if they don't care about it.

rob-isaac:
@mindv0rtex finished this work up in #961

PR #855, opened 2022-02-01
[KAT-2689] Add support for keyword parameter for capture environment

denish44:
This change allows for capturing keyword parameters which can be used in registered call backs for capture environment as needed.

Jira: KAT-2689

### DIFF
--- a/python/katana/bug/environment.py
+++ b/python/katana/bug/environment.py
@@ -206,7 +206,6 @@ def is_interactive():
 
 
 def capture_environment(filename: Optional[Union[str, Path, Any]] = None, **kwargs):
-    # pylint: disable=unused-argument
     """
     Capture the execution and build environment in as much detail as reasonably possible
     and store it to a file. This is used for bug reporting.
@@ -233,7 +232,7 @@ def capture_environment(filename: Optional[Union[str, Path, Any]] = None, **kwar
     try:
         with zipfile.ZipFile(file=file, mode="w", compression=zipfile.ZIP_BZIP2) as zipout:
             for f in _environment_capture_routines:
-                f(zipout)
+                f(zipout, kwargs)
     finally:
         if filename is not None:
             file.close()
--- a/python/katana/bug/environment.py
+++ b/python/katana/bug/environment.py
@@ -19,7 +19,7 @@ def register_capture_routine(f):
 
 
 @register_capture_routine
-def _capture_system(zipout: zipfile.ZipFile):
+def _capture_system(zipout: zipfile.ZipFile, **kwargs):
     import katana  # Imported here to avoid cyclical import
 
     capture_string(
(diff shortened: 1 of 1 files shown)


python/katana/bug/environment.py
```diff
@@ -203,14 +205,16 @@ def is_interactive():
...
+def capture_environment(filename: Optional[Union[str, Path, Any]] = None, **kwargs):
+    # pylint: disable=unused-argument
     """
     Capture the execution and build environment in as much detail as reasonably possible
     and store it to a file. This is used for bug reporting.
 
     :param filename: The file name for the captured environment information.
         (Default: an auto generated file in the system temporary directory)
     :type filename: str or Path or a file-like object or None
+    :key client: Required if using remote enviroment. Expects katana.remote.Client
```
arthurp: I worry about documenting this here because this documentation is related to code that isn't here. But I also see the value in it. Hmm. What do you think?

python/katana/bug/environment.py
```diff
@@ -203,14 +205,16 @@ def is_interactive():
     return hasattr(sys, "ps1")
 
 
-def capture_environment(filename: Optional[Union[str, Path, Any]] = None):
+def capture_environment(filename: Optional[Union[str, Path, Any]] = None, **kwargs):
+    # pylint: disable=unused-argument
```
arthurp: To actually do anything with `kwargs` you need to pass it to the registered routines in `_environment_capture_routines`. I think you will want `f(zipout, **kwargs)`. Otherwise, the enterprise code will not be able to access the `client` arg to do anything useful with it.

arthurp (changes requested):

python/katana/bug/environment.py
denish44: I would say keep it, the user can just ignore this since it specify when required. I think the remote users will recognize since they will do imports with remote and also build a remote client so it should click with them more than other users.

python/katana/bug/environment.py
denish44: oh snap, yes you are correct. forgot the most important part. X.X

python/katana/bug/environment.py
denish44: Maybe we say required when using enterprise remote client. Else maybe we have a separate capture_environment for enterprise with it's own docstring that acts as a pass through to the open source one but the enterprise will have the correct docstring. I don't know how I like hiding it from the user, feel like it should be documented somewhere atleast.

arthurp: I think this is fine.

BTW. You should put full sentences in doc strings. What you wrote here is very stilted because of the lack of articles and stuff.

arthurp (approved):

python/katana/bug/environment.py
denish44: done

PR #858, opened 2022-02-01
NUMAArray: add front() and back()

scober:
I keep reaching for these so I just added them.

### DIFF
--- a/libgalois/include/katana/NUMAArray.h
+++ b/libgalois/include/katana/NUMAArray.h
@@ -130,6 +130,23 @@ public:
   void set(difference_type x, const_reference v) { data_[x] = v; }
   size_type size() const { return size_; }
   bool empty() const { return size() == 0; }
+  // calling front() or back() on an empty array is unsafe
+  reference front() {
+    KATANA_LOG_DEBUG_ASSERT(!empty());
+    return data_[0];
+  }
+  const_reference front() const {
+    KATANA_LOG_DEBUG_ASSERT(!empty());
+    return data_[0];
+  }
+  reference back() {
+    KATANA_LOG_DEBUG_ASSERT(!empty());
+    return data_[size() - 1];
+  }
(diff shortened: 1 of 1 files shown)


e-mcginnis (approved):
LGTM

ps: not a huge fan of hiding potentially unsafe behavior behind nice to use APIs, but I guess that is the CPP way https://www.cplusplus.com/reference/vector/vector/back/ so I'm not going to stop you :)

scober:
> ps: not a huge fan of hiding potentially unsafe behavior behind nice to use APIs, but I guess that is the CPP way https://www.cplusplus.com/reference/vector/vector/back/ so I'm not going to stop you :)

I generally agree (especially because I think `front()` and `back()` are unlikely to be called often or in tight loops). But I was looking at the `std::vector` standard and it said "undefined behavior" so I said "me too". I have also developed some instincts for checking before calling these functions because they have burned me before.

libgalois/include/katana/NUMAArray.h
```diff
@@ -130,6 +130,11 @@ class NUMAArray {
   void set(difference_type x, const_reference v) { data_[x] = v; }
   size_type size() const { return size_; }
   bool empty() const { return size() == 0; }
+  // calling front() or back() on an empty array is unsafe
+  reference front() { return data_[0]; }
+  const_reference front() const { return data_[0]; }
+  reference back() { return data_[size_ - 1]; }
```
amberhassaan: please use `size()` instead of `size_` field. How about adding a debug assert of `!empty()` as that will assist those debugging weird issues.

amberhassaan (approved):
Left some comments.

libgalois/include/katana/NUMAArray.h
e-mcginnis: +1 for a debug assert. Seems better than nothing.

PR #859, opened 2022-02-01
tsuba: Creating RDG interface for python out-of-core import

pyrito:
Previously, out-of-core import produces v1 RDGs which has been difficult to maintain due to its lack of efficient type lookups. This PR makes the appropriate changes needed for out-of-core import to produce v4 RDGs. The major changes needed were the inclusion of the `EntityTypeManager` as well as a wrapper interface to write out RDG part headers. 

I tried to keep the overarching storage interfaces as pure as possible. Please let me know if there's anything I could make clearer in documentation, etc. I'm still working on cleaning this up.


libtsuba/include/katana/RDGPythonInterface.h
```diff
@@ -0,0 +1,35 @@
...
+namespace katana {
+
+struct RDGPropInfo {
+  std::string property_name;
+  std::string property_path;
+};
+
+/// RDGPythonInterface exposes some inner details of RDGs that are needed
+/// by the Python out-of-core import tool
+class RDGPythonInterface {
```
arthurp: Missing `KATANA_EXPORT`.

libtsuba/include/katana/RDGPythonInterface.h
```diff
@@ -0,0 +1,35 @@
...
+public:
+  static katana::Result<void> WriteRDGPartHeader(
+      std::vector<katana::RDGPropInfo> node_properties,
+      std::vector<katana::RDGPropInfo> edge_properties,
+      katana::EntityTypeManager node_entity_type_manager,
+      katana::EntityTypeManager edge_entity_type_manager,
+      const std::string& topology_path, const std::string& rdg_dir);
+};
```
arthurp: No need for a class. You could just have a free loading function.

libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -112,6 +112,15 @@ katana::python::InitEntityTypeManager(py::module_& m) {
...
           })
+      .def(
+          "get_or_add_non_atomic_entity_type_from_strings",
+          [](katana::EntityTypeManager* self,
+             std::set<std::string> type_strs) -> Result<EntityType> {
+            return EntityType{
+                self,
+                KATANA_CHECKED(
+                    self->GetOrAddNonAtomicEntityTypeFromStrings(type_strs))};
+          })
```
arthurp: Don't add this. It was left out intentionally because it can be written trivially in Python like this `m.get_or_add_non_atomic_entity_type({m[n] for n in string_container})`.

libkatana_python_native/src/RDGInterface.cpp
```diff
@@ -0,0 +1,25 @@
...
+// This should NOT be used by anyone else for any kind of purpose since it
+// exposes low level details that users do not need to be concerned with.
+void
+katana::python::InitRDGInterface(py::module& m) {
+  // Define the wrapped interface for PropStorageInfo - needed for RDGPartHeader
+  // Only need the initial constructor since properites will be in memory
+  py::class_<katana::RDGPropInfo> rdg_prop_info_cls(m, "RDGPropInfo");
+  rdg_prop_info_cls.def(py::init([](std::string name, std::string path) {
+    return katana::RDGPropInfo{name, path};
+  }));
```
arthurp: This can just be `rdg_prop_info_cls.def(py::init<std::string, std::string>())`.

libkatana_python_native/src/EntityTypeManager.cpp
pyrito: Not sure if I follow, but I think I do need this because I need to add strings directly to get the entity type.

arthurp: The method `GetOrAddNonAtomicEntityTypeFromStrings` just converts each argument to a type and then gets the intersection of them (the type that has all the given types). So you can just do that yourself with a comprehension. The only reason `GetOrAddNonAtomicEntityTypeFromStrings` exists in C++ is that collections handling is so bad in C++ that writing it out would be really painful.

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -118,37 +118,57 @@ katana::RDGPartHeader::Make(const katana::Uri& partition_path) {
...
+katana::Result<void>
+katana::RDGPartHeader::Write(
+    katana::RDGHandle handle,
+    katana::RDG::RDGVersioningPolicy retain_version) const {
+  std::unique_ptr<FileFrame> ff =
+      KATANA_CHECKED(WriteFileFrame(handle, retain_version));
+  KATANA_CHECKED(ff->Persist());
+  TSUBA_PTP(internal::FaultSensitivity::Normal);
+  return katana::ResultSuccess();
+}
```
arthurp: Why are there two methods that are almost, but not quite the same?

arthurp (approved):
Please address my comments below and get rid of that `from_strings` method. Hit me up if you need guidance on any of it.

libtsuba/src/RDGPartHeader.cpp
pyrito: The first one is the default, current implementation expected by other parts of tsuba. It passes in a WriteGroup which queues up a bunch of async write requests. My method doesn't make use of a WriteGroup b/c I was facing an issue with what it was doing but I actually may be able to get rid of this refactoring. Let me circle back on this with the other changes you requested on the enterprise PR.

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -118,37 +118,57 @@ katana::RDGPartHeader::Make(const katana::Uri& partition_path) {
   return KATANA_CHECKED(MakeJson(partition_path));
 }
 
-katana::Result<void>
-katana::RDGPartHeader::Write(
-    katana::RDGHandle handle, katana::WriteGroup* writes,
+katana::Result<std::unique_ptr<katana::FileFrame>>
+katana::RDGPartHeader::WriteFileFrame(
```
tylershunt: Let's name this `FillFileFrame` instead. The other functions that use the verb `Write` are actually storing things.

libtsuba/src/RDGPartHeader.cpp
tylershunt: In other places the `WriteGroup` is optional. And if it's `nullptr` we just do the write immediately. Let's do that here too.

tylershunt (approved):
Looks good in general; I did have a couple of style nits that I think are worth addressing.

libkatana_python_native/src/RDGInterface.cpp
arthurp: @pyrito This should be fixed.

libtsuba/src/RDGPartHeader.cpp
pyrito: I don't know why I dismissed that idea when I was trying to find clean ways to approach this. That's clearly the simplest solution, but I think I wanted to avoid messing up the existing interface or something (?) Either way, I'll make that change

libtsuba/src/tsuba.cpp
```diff
@@ -307,6 +308,88 @@ katana::CopyRDG(
...
+      edge_entity_type_id_array_path);
+
+  // Set the topology metadata
+  part_header.MakePartitionTopologyMetadataEntry(topology_path);
+
+  katana::RDGTopologyManager topo_manager = KATANA_CHECKED(
+      katana::RDGTopologyManager::Make(part_header.topology_metadata()));
+
+  // get the metadata we need from the topology file
+  KATANA_CHECKED(topo_manager.ExtractMetadata(
```
e-mcginnis: Before this is called, we have written the topology ourselves right? So we shouldn't have to extract the metadata from the topology, since we have the data we need in memory already. 

Since all we care about at this point is the part header, we actually don't need a topology manager at all. 

Instead, we should just fill the PartitionTopologyMetadataEntry ourselves:
```

  // Set the topology metadata
  PartitioTopologyMetadataEntry * entry =  part_header.MakePartitionTopologyMetadataEntry(topology_path);

  // update our metadata entry with what we loaded
  // Assumes the topology is a standard csr topology
  topology_state_ = TopologyKind::kCSR;
  transpose_state_ = TransposeKind::kNo;
  edge_sort_state_ = EdgeSortKind::kAny;
  node_sort_state_ = NodeSortKind::kAny;

 entry->Update(
      num_edges_, num_nodes_,
      /*edge_index_to_property_index_map_present=*/false,
      /*node_index_to_property_index_map_present=*/false,
      /*edge_condensed_type_id_map_size_=*/0,
      /*edge_condensed_type_id_map_present=*/false,
      /*node_condensed_type_id_map_size_=*/0,
      /*node_condensed_type_id_map_present=*/false, 
      topology_state_,
      transpose_state_, 
      edge_sort_state_, 
      node_sort_state_);
```

You could abstract this out into a `FillCSRMetadataEntry` so both this code and `MapMetadataExtract` can use it, if you'd like.

libtsuba/src/tsuba.cpp
```diff
@@ -3,6 +3,7 @@
 #include "GlobalState.h"
 #include "RDGHandleImpl.h"
 #include "RDGPartHeader.h"
+#include "RDGTopologyManager.h"
```
e-mcginnis: drop this include when addressing my other comment

e-mcginnis (changes requested):

libtsuba/src/tsuba.cpp
pyrito: A little confused here - could you clarify. So yes, at this point, we have written the topology out to disk, however I don't think that data is in memory already? I figured that I would use this method with the Topology manager in case we do anything else in the future to add on metadata with.

e-mcginnis: MapMetadataExtract is really just for extracting the metadata from a topology file in a previous storage format RDG, which comes with some assumptions that are different than the ones out-of-core import comes with. 

For example: the topology file MapMetadataExtract expects is always from a storage_format_version_1 RDG.
Moving from a storage format version 1 topology file comes with some challenges, like the number of nodes/edges sometimes being 0 in the metadata, but correct in the topology file. 

Out of core import doesn't have to care about that.

You already have the number of nodes/edges so you don't need to bind the the topology file to read them, and as of now out-of-core import generates a csr topology, that is not transposed, and does not have an edge/node sort state.  so you can just do
```
  auto topology_state = TopologyKind::kCSR;
  auto transpose_state = TransposeKind::kNo;
  auto edge_sort_state = EdgeSortKind::kAny;
  auto node_sort_state = NodeSortKind::kAny;
```
If we change that up later, those will have to be passed in to the WritePartHeader function. 

Since, some of the assumptions are the same right now for both uses, I think it would make sense to abstract that snippet I posted in my previous comment into its own function, calling it in your out-of-core input helper and in MapMetadataExtract. Then if we need to initialize any new topology metadata fields, both paths will get updated properly.

pyrito: Thanks for the clarification. I think that makes sense. I'lll refactor out the topology metadata setting in another method.

libtsuba/src/RDGTopology.cpp
```diff
@@ -273,21 +273,7 @@ katana::RDGTopology::MapMetadataExtract(
       "Extracted num_edges = {} does not match the known num_edges = {}",
       num_edges_, num_edges);
 
-  topology_state_ = TopologyKind::kCSR;
```
e-mcginnis: Even with your change, these lines:
```
topology_state_ = TopologyKind::kCSR;
  transpose_state_ = TransposeKind::kNo;
  edge_sort_state_ = EdgeSortKind::kAny;
  node_sort_state_ = NodeSortKind::kAny;
```
must still be present here as they are setting state in the RDGTopology object itself here.

libtsuba/src/PartitionTopologyMetadata.h
```diff
@@ -93,6 +93,21 @@ class KATANA_EXPORT PartitionTopologyMetadataEntry {
     node_sort_state_ = node_sort_state;
   }
 
+  void FillCSRMetadataEntry(uint64_t num_nodes, uint64_t num_edges) {
```
e-mcginnis, line 96: This looks great, thanks!

e-mcginnis (changes requested):

libtsuba/src/RDGTopology.cpp
pyrito: Oops! You are correct.

e-mcginnis (approved):
Looks great, thanks!

PR #860, opened 2022-02-02
Add global toc [KAT-2560]

JamesFCoffey:
Overriding the `html` template with an extra variable, `startdepth=0`, tricks the Python script behind `pydata-sphinx-theme` to use a the equivalent to `globaltoc.html`.

KAT-2560


origandrew (approved):
LGTM. This is great!

PR #861, opened 2022-02-02
[KAT-3781] Add a warnings config file with a method to disable modin warnings

denish44:
This adds a warnings_config file where we can define methods to disable warnings from certain modules. This will be imported and called in katana.remote on the enterprise repo since that's the only place currently using modin.

Note: this doesn't filter out all of modin warnings but the one's that seemed confusing to the user so far. We will disable additional warnings as people see them and find they aren't relevant.

JIra: KAT-3781

### DIFF
--- a/python/katana/warnings_config.py
+++ b/python/katana/warnings_config.py
@@ -1,0 +1,13 @@
+import warnings
+
+
+def disable_partial_modin_warnings():
+
+    # Ignore any warnings related to defaulting to pandas for any operation.
+    warnings.filterwarnings("ignore", message=r".*[dD]efaulting to pandas implementation.*", module="modin")
+
+    # Ignore warnings related to distribution of object
+    warnings.filterwarnings("ignore", message=r".*[dD]istributing.*object", module="modin")
+
+    # Ignore warnings for implementation requests
+    warnings.filterwarnings("ignore", message=r".*request implementation.*", module="modin")


arthurp (approved):

denish44:
This adds a warnings_config file where we can define methods to disable warnings from certain modules. This will be imported and called in katana.remote on the enterprise repo since that's the only place currently using modin.

python/katana/warnings_config.py
```diff
@@ -0,0 +1,13 @@
+import warnings
+
+
+def disable_partial_modin_warnings():
+
+    # Ignore any warnings related to defaulting to pandas for any operation.
```
gliga, line 6: you had period at the end of this comment but not for other comments below :)

python/katana/warnings_config.py
```diff
@@ -0,0 +1,13 @@
...
+    warnings.filterwarnings("ignore", message=r".*[dD]efaulting to pandas implementation.*", module="modin")
+
+    # Ignore warnings related to distribution of object
```
gliga, line 9: My understanding that filtering "ignores", but what I would like to see documented is the following:
* why does the warning happen in the first place
* why it is safe that we ignore it here

gliga (commented):
Thanks.  Glad to see this.  A couple of comments to consider in the future.

python/katana/warnings_config.py
denish44: That's fair, I will keep this in mind for the next one's we ignore as we hear from people.

python/katana/warnings_config.py
denish44: X.X oops

PR #863, opened 2022-02-02
Misc small changes in preparation for switching PropertyGraph wrapping to Pybind11

arthurp:
With: https://github.com/KatanaGraph/katana-enterprise/pull/3161

I split these out so they can be reviewed while I finalize the big change. Each commit applies to different files and can be reviewed separately.

This PR will not be squashed since each change is independent, so having the separate commits on master may help people understand the history in the future.


denish44 (approved):

libgalois/src/Threads.cpp
```diff
@@ -28,6 +28,10 @@ KATANA_EXPORT unsigned int activeThreads = 1;
 
 unsigned int
 katana::setActiveThreads(unsigned int num) noexcept {
+  // Reset "burn power"/"busy wait" mode since it might be configured for a
+  // different number of threads than we have after this call. That can cause
+  // crashes.
+  katana::GetThreadPool().beKind();
```
arthurp, line 34: @roshandathathri @gurbinder533 @ddn0 Could one of your verify this is OK?

roshandathathri: I'm not sure. `burnPower()` does call `beKind()` internally. Was that not sufficient? I mean was there some bug that you noticed and this fixed?

arthurp: The bug was that calling `burnPower(8)` followed by `setActiveThreads(4)` would create a crash. Calling `burnPower` again would be safe. But that's not what I'm referring to. This was happening in the pytest tests since they switch thread count and mode between tests.

roshandathathri: Ok, that makes sense. This looks good to me. 

@amberhassaan FYI

amberhassaan: Looks OK to me.

amberhassaan (commented):
One comment only.

PR #864, opened 2022-02-02
Replace python checkers tool with a parallel python implemention

arthurp:


scripts/check_python.py
```diff
@@ -0,0 +1,132 @@
...
+        help="Add a pattern to exclude from processing. The pattern can be a path or a glob.",
+    )
+    parser.add_argument("roots", type=str, nargs="+", help="The root paths to search from, or files to check.")
+    args = parser.parse_args()
+
+    if not args.pylint and not args.black and not args.isort:
+        args.pylint = True
+        args.black = True
+        args.isort = True
+
```
srogatch, line 59: Somewhere after these lines, can we print to console the `pylint`, `black` and `isort` versions used? This will help formatting the code locally with exactly the same versions, otherwise different versions may format the code differently and this would cause CI failures.

srogatch (approved):
LGTM, except that we don't know the versions of the tools again.

scripts/check_python.py
```diff
@@ -0,0 +1,132 @@
...
+    exclude_paths = set(args.exclude)
+    # For backwards compat check an old environment variable.
+    exclude_paths.update(os.environ.get("PRUNE_PATHS", "").split(":"))
```
ddn0, line 62: Might be worth a quick check if anyone actually depends on `PRUNE_PATHS` and I'm sure they can learn to pass their arguments on the command line instead.

ddn0 (approved):

scripts/check_python.py
arthurp: :-) OK. I don't like that it prints those versions. I'll add a flag to *disable* printing the versions, but print them by default.

scripts/check_python.py
arthurp: I'm gonna merge as is. If there was no response to my question on code related in a couple days I will create another PR to remove PRUNE_PATHS entirely.

PR #865, opened 2022-02-03
reduce parallelism for conda build

e-mcginnis:
without this change, the default level is 2 which
can cause issues in our CI

example of the kind of failure this fixes: https://github.com/KatanaGraph/katana/runs/5041815770?check_suite_focus=true

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -123,6 +123,7 @@ jobs:
     - name: Build conda Packages
       run: |
         . .github/workflows/activate_miniconda.sh
+        export CPU_COUNT=1
         scripts/build_conda.py --output-folder $HOME/build/
 
     - name: Build conda Packages for Katana Metagraph Plugin


conda_recipe/build.sh
```diff
@@ -14,7 +14,7 @@ rm -rf build
 mkdir build
 cd build
 if [ -z "${CMAKE_BUILD_PARALLEL_LEVEL}" ]; then
-  export CMAKE_BUILD_PARALLEL_LEVEL="$CPU_COUNT"
+  export CMAKE_BUILD_PARALLEL_LEVEL=1
```
ddn0: Looks like `CPU_COUNT` can be overridden from the environment: https://github.com/conda/conda-build/pull/1149 Can you check if that would work rather than changing `build.sh` itself?

cc @arthurp 

https://github.com/conda/conda-build/pull/1149

arthurp: Does cmake honor `CPU_COUNT`? I vaguely remember finding out that it didn't and you had to set `CMAKE_BUILD_PARALLEL_LEVEL`. I think that's why I wrote this code initially.

e-mcginnis: I'll try that out, but I suspect the env var will get unset at some point like `CMAKE_BUILD_PARALLEL_LEVEL` did in my previous attempt at fixing this: https://github.com/KatanaGraph/katana/pull/846/commits/d375267795f5f7ba5efb725416da10093718e1d0

arthurp: As @ddn0, I think you could set CPU_COUNT in the code that calls conda-build instead of here.

arthurp: `conda-build` unsets most variables, but those issues say it passes through `CPU_COUNT`.

e-mcginnis: seems it worked https://github.com/KatanaGraph/katana/runs/5045244371?check_suite_focus=true

ddn0 (approved):

PR #866, opened 2022-02-03
Test construction of transformation topologies

yan-zaretskiy:
A simple test that constructs optional topologies on top of a projected view. Yet it revealed a bug in `CondensedTypeIDMap::MakeFromEdgeTypes`!


e-mcginnis (approved):
this fix, and the tests, look good to me. Awesome that the tests caught a bug.

libgraph/src/GraphTopology.cpp
```diff
@@ -526,7 +526,7 @@ katana::CondensedTypeIDMap::MakeFromEdgeTypes(
   katana::do_all(
       katana::iterate(Edge{0}, topo.NumEdges()),
       [&](const Edge& e) {
-        katana::EntityTypeID type = pg->GetTypeOfEdgeFromPropertyIndex(e);
+        katana::EntityTypeID type = pg->GetTypeOfEdgeFromTopoIndex(e);
```
amberhassaan, line 529: Sorry, why is TopoIndex the right answer and PropertyIndex wrong?

yan-zaretskiy: That's because this loop simply iterates over the edges in a topology from 0 to max value, which is the topology index. Essentially, it means the index into an adjacency vector.

amberhassaan: OK, got it. I hope we can put this ugliness behind us soon.

PR #869, opened 2022-02-05
Memory policies for the memory supervisor

witchel:
Current memory policies get their signal about how much memory is being used from the OS, but use the data it has from the different managers for reclaiming.  Right now we only have a single manager (for properties) which is not used by the rest of the system.

Memory policies encapsulate hairy and heuristic decisions about what constitutes memory pressure.  These will have to be tuned over time.  The current API includes three functions.

- Should we reclaim some standby memory now?
-- Try to avoid memory emergencies.
- Are we under high memory pressure?
--    If so, reject requests for standby memory.
- Are we in imminent danger of OOM?
--    If yes, kill ourselves.  We might want to do more cleanup in this case

I took this opportunity to remove some unused oplog code from graph-convert.


libgalois/include/katana/MemoryPolicy.h
```diff
@@ -0,0 +1,70 @@
...
+  /// Utility function to find out our OOM score from Linux
+  static uint64_t OOMScore();
+  struct MemInfo;
+
+protected:
+  MemoryPolicy();
+  void UpdateMemInfo(MemInfo* mem_info, count_t active, count_t standby) const;
+
+  count_t physical_;
+  double high_used_ratio_threshold_;
```
ddn0: protected fields are not ideal. Protected methods are better because you can dictate the class invariants more cleanly.

witchel: Okey dokey, smokey.

libgalois/include/katana/MemoryPolicy.h
```diff
@@ -0,0 +1,82 @@
...
+class KATANA_EXPORT MemoryPolicyMinimal : public MemoryPolicy {
+public:
+  MemoryPolicyMinimal();
+  count_t ReclaimForMemoryPressure(
+      count_t active, count_t standby) const override;
+  bool MemoryPressureHigh(count_t active, count_t standby) const override;
+  bool KillSelfForLackOfMemory(count_t active, count_t standby) const override;
+};
+
+/// Memory policy that prioritizes performance, i.e., it uses memory aggressively
```
scober, line 72: Could you change the comments on these two policies to differentiate them more?

libgalois/include/katana/PropertyManager.h
```diff
@@ -19,11 +19,15 @@ class KATANA_EXPORT PropertyManager : public Manager {
   const std::string& MemoryCategory() const override { return memory_category; }
   count_t FreeStandbyMemory(count_t goal) override;
 
-  void AddProperty(const katana::Uri& property_path);
+  /// Client wants a property, see if we have it in the cache and if so return it and make the memory active.
+  /// Returns nullptr if manager does not have it in the cache
+  std::shared_ptr<arrow::Table> GetProperty(const katana::Uri& property_path);
```
scober, line 24: Does this mean we have settled on file name as the one true cache key?

libgalois/src/MemoryPolicy.cpp
```diff
@@ -0,0 +1,213 @@
...
+}  // namespace
+
+void
+katana::MemoryPolicy::UpdateMemInfo(
+    MemInfo* mem_info, count_t active, count_t standby) const {
+  mem_info->active = active;
+  mem_info->standby = standby;
+  mem_info->rss_bytes = katana::ProgressTracer::ParseProcSelfRssBytes();
+  mem_info->used_ratio =
+      (physical() == 0) ? 0.0 : (double)mem_info->rss_bytes / physical();
```
scober: I know this is the corner case of all corner cases, but 0.0 seems like a weird value here. Would you consider using `NaN`?

libgalois/src/MemorySupervisor.cpp
```diff
@@ -16,19 +19,28 @@ getTotalSystemMemory() {
...
+               {"standby", standby},
            });
 }
 
 void
-katana::MemorySupervisor::Sanity() {
+KillCheck(katana::MemoryPolicy* policy, count_t active, count_t standby) {
+  if (policy->KillSelfForLackOfMemory(active, standby)) {
+    LogState(oversubscribed_str, active, standby);
+    KATANA_LOG_FATAL("out of memory");
```
scober, line 30: Should we be broadcasting an error here? I thought one of the advantages of killing the process ourselves was that we could do cleanup and not hang when one host got oversubscribed.

libgalois/src/PropertyManager.cpp
```diff
@@ -9,6 +9,15 @@ const std::string katana::PropertyManager::memory_category = "property";
 // Anchor manager vtable
 katana::Manager::~Manager() = default;
 
+namespace {
+
+std::string
+ToGB(uint64_t total) {
+  return fmt::format("{:.3f}", total / 1024.0 / 1024.0 / 1024.0);
```
scober: I think I have seen this function twice now. Could we put this in libsupport?

libgalois/test/os-memory.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+}
+
+uint64_t
+GetTotalSystemMemory() {
+  uint64_t pages = sysconf(_SC_PHYS_PAGES);
+  uint64_t page_size = sysconf(_SC_PAGE_SIZE);
+  return pages * page_size;
+}
```
scober: I think this function is already defined in the memory manager code.

scober (approved):
LGTM

libgalois/src/PropertyManager.cpp
witchel: I bet you always put your dishes in the dishwasher.

libgalois/src/MemorySupervisor.cpp
witchel: Yes, this is messy.  Is there an example of the clean, distributed shutdown you are talking about?

scober: I guess this repo doesn't handle any distributed execution so maybe a clean distributed shutdown should be implemented near the distributed execution logic?

libgalois/src/PropertyManager.cpp
scober: Dishwasher? I wash my dishes with spit and dirt.

libgalois/src/MemoryPolicy.cpp
witchel: No.

But I did ensure a non-zero value.

libgalois/src/MemorySupervisor.cpp
witchel: @roshandathathri said it was ok.  Ya neat freak.

libgalois/include/katana/PropertyManager.h
witchel: I was very excited by this prospect, as the random tag in the file name is the right level of indirection for the cache instead of rdg_dir+version+view.  But there are some subtleties to caching dirty properties that will probably complicate this glorious vision.

PR #873, opened 2022-02-07
Property Graph: require storage location in factory

scober:
The Make overload that accepts a topology and type arrays still needs to
require a storage location. Add an overload that requires a storage
location and deprecated the overload doesn't.

### DIFF
--- a/libgraph/include/katana/PropertyGraph.h
+++ b/libgraph/include/katana/PropertyGraph.h
@@ -269,9 +269,17 @@ public:
   static Result<std::unique_ptr<PropertyGraph>> Make(
       GraphTopology&& topo_to_assign);
 
-  /// Make a property graph from topology and type arrays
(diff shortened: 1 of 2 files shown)


libgraph/include/katana/PropertyGraph.h
```diff
@@ -269,13 +269,21 @@ class KATANA_EXPORT PropertyGraph {
   static Result<std::unique_ptr<PropertyGraph>> Make(
       GraphTopology&& topo_to_assign);
 
-  /// Make a property graph from topology and type arrays
+  /// [[deprecated("You should provide a rdg dir")]]
   static Result<std::unique_ptr<PropertyGraph>> Make(
```
witchel, line 273: Can we send an electric shock to the developers who use this function?

Or send them emails about their car's extended warranty, maybe every 10 minutes?

witchel (approved):
Let the reign of the storage prefix begin!

PR #874, opened 2022-02-08
Add a custom 404 page

bobdc:
James found the page explaining how to configure our documentation's server to point at a custom 404 page, so I made one to replace the "storage: object doesn't exist" message that https://docs.katanagraph.com/ currently shows for unrecognized requests.


JamesFCoffey (approved):
Later we need to get a graphic design for this page.

PR #876, opened 2022-02-08
PropertyGraph: fix interface

scober:
A previous commit marked a PropertyGraph::Make overload as deprecated
but it also mistakenly removed the implementation. Revert that
implementation removal.

### DIFF
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -223,6 +223,20 @@ katana::PropertyGraph::Make(katana::GraphTopology&& topo_to_assign) {
 
 katana::Result<std::unique_ptr<katana::PropertyGraph>>
 katana::PropertyGraph::Make(
+    katana::GraphTopology&& topo_to_assign,
+    NUMAArray<EntityTypeID>&& node_entity_type_ids,
+    NUMAArray<EntityTypeID>&& edge_entity_type_ids,
+    EntityTypeManager&& node_type_manager,
+    EntityTypeManager&& edge_type_manager) {
+  return std::make_unique<katana::PropertyGraph>(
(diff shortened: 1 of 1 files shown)


witchel (approved):
Fine, but isn't the alternative to rip off the bandaid and change the users of this method?

scober:
> Fine, but isn't the alternative to rip off the bandaid and change the users of this method?

I have set myself a strict requirement that I don't break interfaces if I don't have to. So I want to make sure this repo doesn't break any builds and once I have transitioned all users of this method to the new version I will come back and remove the old version.

witchel:
> > Fine, but isn't the alternative to rip off the bandaid and change the users of this method?
> 
> I have set myself a strict requirement that I don't break interfaces if I don't have to. So I want to make sure this repo doesn't break any builds and once I have transitioned all users of this method to the new version I will come back and remove the old version.

maturity: the state, fact, or period of being mature.

libgraph/src/PropertyGraph.cpp
```diff
@@ -221,6 +221,20 @@ katana::PropertyGraph::Make(katana::GraphTopology&& topo_to_assign) {
       EntityTypeManager{}, EntityTypeManager{});
 }
 
+katana::Result<std::unique_ptr<katana::PropertyGraph>>
+katana::PropertyGraph::Make(
```
amberhassaan, line 225: So this got removed somehow?

amberhassaan (approved):

libgraph/src/PropertyGraph.cpp
scober: Yeah. In my head I had added an overload. In my computer I had overwritten the existing overload.

PR #879, opened 2022-02-09
Call MemorySupervisor::GetTotalSystemMemory static function, don't get instance

witchel:
I accidentally created recursion where I didn't want it, so call the static function and don't create a MemorySupervisor instance while creating a MemorySupervisor instance.  The current state of enterprise/open makes testing tough.

### DIFF
--- a/libgalois/src/MemoryPolicy.cpp
+++ b/libgalois/src/MemoryPolicy.cpp
@@ -166,7 +166,7 @@ katana::MemoryPolicyMinimal::MemoryPolicyMinimal()
 
 katana::MemoryPolicy::MemoryPolicy(
     katana::MemoryPolicy::Thresholds thresholds) {
-  physical_ = katana::MemorySupervisor::Get().GetTotalSystemMemory();
+  physical_ = katana::MemorySupervisor::GetTotalSystemMemory();
   // We divide by physical_
   if (physical_ == 0) {
     physical_ = 1;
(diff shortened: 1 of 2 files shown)


ddn0:
Is there a way to test these new classes independently?

witchel:
> Is there a way to test these new classes independently?

I'm not sure what you are getting at.  The problem came up because I can't build enterprise without manually modifying open because currently enterprise and open are incompatible.  I should have done that before committing my PR, but I've done it now.

`os-memory.cpp` is the test for the memory supervisor and the memory policies, though it has some limitations right now and is more geared toward deriving the right policy.  I don't know what it would mean to test e.g., a memory policy on its own.

scober:
I know `os-memory.cpp` can be a little heavy, could we make a shorter-running version that we can run in CI to just sanity-check that the classes can be created and destroyed?

ddn0:
> I'm not sure what you are getting at. The problem came up because I can't build enterprise without manually modifying open because currently enterprise and open are incompatible. I should have done that before committing my PR, but I've done it now.

MemorySupervisor is defined in this repo. If there was a build problem for one of its callers, then I would expect the fix would be to change some public API of MemorySupervisor, yet the change in this PR does not change any public API of MemorySupervisor.

Is the issue at hand a build problem for someone who depends on MemorySupervisor or a bug in MemorySupervisor itself? If the latter, is there a way to test MemorySupervistor without involving the larger set of external callers of MemorySupervisor?

witchel:
I'm deep into building a reasonable graph loading/caching test on open, but I'd like for this PR to be approved so open (without c33ea550d6cc) will work with enterprise and I won't feel stressed about it.

roshandathathri (approved):
Thanks!

PR #880, opened 2022-02-09
Add multi edge support for triangle counting app 

uditagar:
This PR adds the support for multi-edges in our triangle counting app.

### DIFF
--- a/libgraph/src/analytics/triangle_count/triangle_count.cpp
+++ b/libgraph/src/analytics/triangle_count/triangle_count.cpp
@@ -176,7 +176,12 @@ OrderedCountFunc(
         it_n++;
       }
       if (dst_v == graph->OutEdgeDst(*it_n)) {
-        numTriangles_local += 1;
+        edge_iterator multi_it_n = it_n;
+        while (multi_it_n != graph->OutEdges(n).end() &&
+               graph->OutEdgeDst(*multi_it_n) == dst_v) {
+          numTriangles_local += 1;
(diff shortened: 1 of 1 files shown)


roshandathathri (approved):
LGTM. Thanks!

libgraph/src/analytics/triangle_count/triangle_count.cpp
```diff
@@ -176,7 +176,12 @@ OrderedCountFunc(
         it_n++;
       }
       if (dst_v == graph->OutEdgeDst(*it_n)) {
-        numTriangles_local += 1;
+        edge_iterator multi_it_n = it_n;
+        while (multi_it_n != graph->OutEdges(n).end() &&
+               graph->OutEdgeDst(*multi_it_n) == dst_v) {
+          numTriangles_local += 1;
+          multi_it_n++;
+        }
```
aneeshdurg, line 184: why can't it_n just be used directly without creating multi_it_n?

aneeshdurg (approved):

libgraph/src/analytics/triangle_count/triangle_count.cpp
uditagar: because it is required for further checks involving `dst_v`

roshandathathri: Yes, it can't be reused. Another option is to count it once and store a variable. But this looks cleaner.

PR #881, opened 2022-02-09
Convert TransformationView to PropertyGraph

yan-zaretskiy:
This PR moves `TransformationView` logic directly into the `PropertyGraph` class. This is done in support of refactoring the distributed projections.


roshandathathri (approved):
LGTM

Why is this still a draft?

yan-zaretskiy:
> Why is this still a draft?

You know why - I forgot to press the button! 😃

libgraph/include/katana/PropertyGraph.h
```diff
@@ -962,11 +957,13 @@ class KATANA_EXPORT PropertyGraph {
     return *edge_entity_type_ids_;
   }
 
-  // This constructor is meant to be used is situations when you want to share
-  // property and type data (including RDG) with another PropertyGraph instance,
-  // while using a new topology. The topology is assumed to maintain the correct
-  // property index mapping from its nodes/edges to the original property table.
-  PropertyGraph(const PropertyGraph& parent, GraphTopology&& topo) noexcept
+  /// Call this constructor to populate projection data.
+  PropertyGraph(
```
amberhassaan, line 1003: I'm saddened by addition of yet more fields to poor old dumpster of a `PropertyGraph`.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -997,6 +1010,38 @@ class KATANA_EXPORT PropertyGraph {
   Result<void> WriteView(
       const std::string& uri, const std::string& command_line);
 
+  /// Bitmask of nodes included in the transformation view. Should be used to construct arrow tables.
+  std::shared_ptr<arrow::Buffer> NodeBitmask() const noexcept {
+    if (is_transformed) {
+      auto length = original_to_transformed_nodes_.size();
```
amberhassaan: Shouldn't  you back these members by fields so that we don't create a new bitmask on every call (and a new `shared_ptr`)?

libgraph/include/katana/PropertyGraph.h
```diff
@@ -997,6 +1010,38 @@ class KATANA_EXPORT PropertyGraph {
...
+    if (is_transformed) {
+      auto length = original_to_transformed_edges_.size();
+      return std::make_shared<arrow::Buffer>(
+          edge_bitmask_data_.data(), arrow::BitUtil::BytesForBits(length));
+    }
+    return nullptr;
+  }
+
+  /// Return the number of nodes of the original property graph.
+  uint64_t NumOriginalNodes() const {
```
amberhassaan, line 1074: why not use this logic in `NumNodes`? Asking because I'd like to understand why we'd like to keep the original `NumNodes` as is.

amberhassaan (approved):
Looks good. Left some comments.

libgraph/include/katana/PropertyGraph.h
yan-zaretskiy: `NumNodes` is the number of nodes in the topology, which is smaller for projections. We'd need it when we want to iterate over all nodes in the topology. This `NumOriginalNodes` is really only used when adding a new node property to the table, to make sure I allocate the correct number of rows.

libgraph/include/katana/PropertyGraph.h
yan-zaretskiy: It's a cheap operation that is also not on the hot path. Arrow buffer is just a fat pointer. This was my thinking for generating it on the fly. But I guess I can create it once at construction instead. Let me do that.

libgraph/include/katana/PropertyGraph.h
yan-zaretskiy: At least the public interface didn't have to change much...

uditagar:
so now, there is no TransformationView anymore?

yan-zaretskiy:
> so now, there is no TransformationView anymore?

Right, just the `PropertyGraph`

uditagar:
> > so now, there is no TransformationView anymore?
> 
> Right, just the `PropertyGraph`

ok

PR #884, opened 2022-02-11
[KAT-1686] Update libtsuba to accept and return in-memory manifests

yige-hu:
From PR #831 :
* Update libtsuba to accept and return in-memory manifest

* Added a TxnContext.Commit() function, which is used by (1) single-operations committed as transactions (QueryOperation, AnalyticsOperation, ImportOperation, ImportFromMLOperation), (2) transaction commits and (3) in TxnContext() destructor when auto-commit is on (for tools and tests)

* Now all operation and tests all get delayed RDGManifest commit, this should make a smooth transition towards delaying RDGManifest commit to the Go server

* In the future, graph-worker will no longer commit the manifests for both transactions and single operations as transactions. Instead, `Operation::final()` will send the in-memory manifest to the Go server, which later will write back the manifest to storage.

* Made up comment resolving for PR #844 .

New:
* Now support caching in-memory manifest from multiple graphs with `unordered_map` with `rdg_dir()` as the key


yige-hu:
Hey all, you can review only the latest 3 commits in this PR, and the latest commit in the enterprise PR #3042 , which should be the fix.

libtsuba/src/RDG.cpp
```diff
@@ -107,9 +105,15 @@ SetTnxContextManifest(
   auto manifest_file = katana::RDGManifest::FileName(
       handle.impl_->rdg_manifest().dir(),
       handle.impl_->rdg_manifest().viewtype(), new_manifest.version());
-  handle.impl_->set_rdg_manifest(std::move(new_manifest));
 
-  txn_ctx->SetManifestInfo(new_manifest.dir(), manifest_file, new_manifest);
+  KATANA_LOG_DEBUG(
+      "new manifest version {}, dir {}, file {}", new_manifest.version(),
+      handle.impl_->rdg_manifest().dir(), manifest_file);
```
witchel: Can you put information like this in ProgressTrace?  We don't use LOG_DEBUG much anymore.  Something like this.
```
 katana::GetTracer().GetActiveSpan().Log(
      "new manifest",
      {
          {"new_version", new_manifest.version()},
          {"manifest_dir", handle.impl_->rdg_manifest().dir()},
          {"manifest_file", manifest_file},
});
```

libtsuba/include/katana/tsuba.h
```diff
@@ -57,6 +58,9 @@ OpenFlagsValid(uint32_t flags) {
 KATANA_EXPORT katana::Result<RDGManifest> FindManifest(
     const std::string& rdg_name);
 
+KATANA_EXPORT katana::Result<katana::RDGManifest> FindManifest(
```
witchel, line 61: Is this finding the latest manifest? Maybe that should be in the name.

libtsuba/include/katana/TxnContext.h
```diff
@@ -121,8 +121,8 @@ class KATANA_EXPORT TxnContext {
 
   bool TopologyWrite() const { return topology_write_; }
 
-  inline bool ManifestCached(const Uri& manifest_file) const {
-    return manifest_info_.count(manifest_file);
+  inline bool ManifestCached(const Uri& rdg_dir) const {
+    return manifest_info_.count(rdg_dir);
```
witchel: I'm not crazy about the implicit `size_t `to `bool `conversion here.

witchel (approved):
Looks good.

What two manifests end up in the same transaction in our tests?

Also, I'm not crazy about using the rdg_dir as the key.  It should at least be rdg_dir+version.  But maybe this is a larger question.  I'd like to see RDGManifests become the "one true name" for RDGs, even if they have incomplete naming information (i.e., the version number is stale).

libtsuba/include/katana/tsuba.h
yige-hu: It's finding manifest first by checking the cache in the `TxnContext`, if not cached, it calls the original `FindManifest`. The old `FindManifest` finds the manifest from storage if passed with a rdg file name; or the latest version manifest in a directory if a rdg directory is provided. Since the original `FindManifest` does not contain `latest` in its name, I don't think the new `FindManifest` needs it. If we really want a new name, I think `FindManifestCached` is better, but since it also finds the manifest from storage if not cached, I feel `FindManifest` with different signature should be fine.

yige-hu:
> What two manifests end up in the same transaction in our tests?
> 
I lost the error log. But I remember seeing two manifest files, which caused the in-memory manifest's loading and version tracking being wrong.

> Also, I'm not crazy about using the rdg_dir as the key. It should at least be rdg_dir+version. But maybe this is a larger question. I'd like to see RDGManifests become the "one true name" for RDGs, even if they have incomplete naming information (i.e., the version number is stale).

I think using `rdg_dir` is better considering there can be multiple versions for one `rdg_dir`, and when calling `FindManifest`, the processes sometimes doesn't know which version to use -- `FindManifest` loads the manifest from storage if a rdg file name is provided; or loads the latest manifest if a non-rdg file/ rdg directory is provided.

libtsuba/src/RDG.cpp
yige-hu: Strangely I'm getting a `no matching function for call to 'katana::ProgressSpan::Log(const char [13], <brace-enclosed init ializer list>)'` with this. I can't tell any difference by eye compared to other call sites.

yige-hu: More compilation error message:
```
/home/yige/katana-enterprise/external/katana/libtsuba/src/RDG.cpp:117:24: error: no matching function for call to 'katana::ProgressSpan::Log(const char [13], <brace-enclosed initializer list>)'
  117 |                       });
      |                        ^
In file included from /home/yige/katana-enterprise/external/katana/libtsuba/src/RDG.cpp:36:
/home/yige/katana-enterprise/external/katana/libsupport/include/katana/ProgressTracer.h:234:16: note: candidate: 'virtual void katana::ProgressSpan::Log(const string&, const Tags&)'
  234 |   virtual void Log(const std::string& message, const Tags& tags) = 0;
      |                ^~~
/home/yige/katana-enterprise/external/katana/libsupport/include/katana/ProgressTracer.h:234:60: note:   no known conversion for argument 2 from '<brace-enclosed initializer list>' to 'const Tags&' {aka 'const std::vector<std::pair<std::__cxx11::basic_string<char>, katana::Value> >&'}
  234 |   virtual void Log(const std::string& message, const Tags& tags) = 0;
```

yige-hu: Ok, I think it's the katana::Uri failed to convert itself into string in Tags's constructor. It needs explicit string().

PR #887, opened 2022-02-11
Allow different dev env instructions [KAT-3728]

JamesFCoffey:
Removed the link at the top and put an include directive to allow for different dev environment instructions. Also added placeholders for other differences between `katana` and `katana-enterprise`. These extra 3 files would then be different on the `katana-enterprise` repo.

KAT-3728


scober (approved):
Thanks for doing this!

PR #889, opened 2022-02-12
RDG: add (potentially temporary) interface for getting the URI of pro…

tylershunt:
…perties

Because of the way we're staging the work, we need to be able to get
these paths from another part of the code base.

### DIFF
--- a/libtsuba/src/RDG.cpp
+++ b/libtsuba/src/RDG.cpp
@@ -693,6 +693,28 @@ UnloadProperty(
   return KATANA_CHECKED(props->RemoveColumn(i));
 }
 
+katana::Result<katana::Uri>
+GetStorageLocationIfValid(
+    const std::string& name,
+    const std::vector<katana::PropStorageInfo>& prop_info_list) {
+  auto psi_it = std::find_if(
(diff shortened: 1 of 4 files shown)


scober (approved):
LGTM

libtsuba/src/RDG.cpp
```diff
@@ -693,6 +693,28 @@ UnloadProperty(
...
+        katana::ErrorCode::PropertyNotFound, "no property named {}",
+        std::quoted(name));
+  }
+
+  if (!(psi_it->IsAbsent() || psi_it->IsClean())) {
+    return KATANA_ERROR(
+        katana::ErrorCode::AssertionFailed, "the property exists but is dirty");
+  }
+  // TODO(thunt) there's really no reason why we shouldn't always use uri
+  auto path = KATANA_CHECKED(katana::Uri::Make(psi_it->path()));
```
witchel, line 714: My only objection in this code is the use of the term `path`.  The function returns a file name, not a path.  I thought by path, it would be the path name including rdg_dir.  But it isn't.

witchel (approved):
'Tis the harmonic convergence of access to property file names.

PR #890, opened 2022-02-12
Memory API refinement and make sure property caching happens

witchel:
The MemorySupervisor now creates the managers and indexes them by name. The property cache is no longer exposed and is only accessed indirectly via the property manager (which itself is accessed via the memory supervisor). I think this makes our design less "loopy," but I await Donald's judgment.

AddProperties got a Boolean flag because we use it to load properties and other graph metadata (like the local_to_global array). I need the flag to do accurate bookkeeping for property loading.

This makes the ProgressTracer mandatory for more tests (@e-mcginnis). I don't think that is so bad. I like the Progress tracer. Though @patrickkenney9801, am I initializing the text tracer in the best way?

Note to Donald and Scober. I tried to write a test for this functionality in open, but it was just too difficult. I couldn't use RDGLoader and I can't refer to a DistPropertyGraph. I do have a new test in enterprise.


patrickkenney9801:
This usage looks good, the Tracer is only initialized if DistMemSys or SharedMemSys are also initialized

libtsuba/src/RDG.cpp
```diff
@@ -615,6 +615,35 @@ katana::RDG::Store(
...
+}  // namespace
+
+const std::string&
+katana::RDG::NodePropertyFileName(const std::string& name) {
+  return GetLatestFileName(name, core_->part_header().node_prop_info_list());
+}
+const std::string&
+katana::RDG::EdgePropertyFileName(const std::string& name) {
+  return GetLatestFileName(name, core_->part_header().edge_prop_info_list());
+}
```
tylershunt: I'm particularly interested in this interface which I need in some form to merge some LSPG work; I have a version that returns a Uri, but fails if the property isn't clean or absent. If that's acceptable, I can merge #899 an you could use it here; If not I'm happy to use this when it merges.

tylershunt (approved):

libtsuba/src/RDG.cpp
witchel: I can go either way.  I return an empty string if the property is dirty or absent.  So merge away if you want.  Currently, I can't build on HEAD due to arrow issues like this.
`In file included from /home/witchel/katana/external/katana/libtsuba/src/RDG.cpp:35:                                                                                                                                   /home/witchel/katana/external/katana/libtsuba/include/katana/ParquetWriter.h:23:34: error: no member named 'PARQUET_2_6' in 'parquet::ParquetVersion'; did you mean 'PARQUET_2_0'?                                            parquet::ParquetVersion::PARQUET_2_6};                                                                                                                                                                                ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~           `

`
/home/witchel/katana/external/katana/libsupport/include/katana/ArrowVisitor.h:259:17: error: no member named 'INTERVAL_MONTH_DAY_NANO' in 'arrow::Type'                                                                  TYPE_CASE(INTERVAL_MONTH_DAY_NANO)                                                                                                                                                                                 ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~  
`

tylershunt: You need to update arrow.

libtsuba/src/AddProperties.cpp
```diff
@@ -130,44 +126,16 @@ katana::AddProperties(
...
-                  {"approx_size_human",
-                   katana::BytesToStr(
-                       "{:.2f}{}", katana::ApproxTableMemUse(props))},
-              });
-        }
-      }
       KATANA_CHECKED_CONTEXT(
           add_fn(props), "adding {}", std::quoted(prop->name()));
       prop->WasLoaded(props->field(0)->type());
-      if (prop->type()->Equals(arrow::uint8())) {
```
e-mcginnis, line 166: Don't we still need checks to ensure we don't add uint8 properties, or do we enforce this elsewhere now?

libtsuba/src/AddProperties.h
```diff
@@ -17,8 +17,9 @@ KATANA_EXPORT katana::Result<std::shared_ptr<arrow::Table>> LoadPropertySlice(
     const std::string& expected_name, const katana::Uri& file_path,
     int64_t offset, int64_t length);
 
+// is_property is true for properties and false for RDG metadata
```
e-mcginnis, line 20: Could I get some context regarding why we store RDG metadata (aka non-properties) as properties? 
Is this something we are happy with, or are we trying to move away from it?

PR #892, opened 2022-02-12
PropertyGraph: always require storage prefix

scober:
There was one remaining PropertyGraph::Make overload that didn't require
a storage prefix. Add an overload to replace that one that does require
a storage prefix and mark the original one deprecated.

### DIFF
--- a/libgraph/include/katana/PropertyGraph.h
+++ b/libgraph/include/katana/PropertyGraph.h
@@ -266,10 +266,16 @@ public:
       const katana::RDGLoadOptions& opts = katana::RDGLoadOptions());
 
   /// Make a property graph from topology
+  // [deprecated("please provide a storage prefix")]
   static Result<std::unique_ptr<PropertyGraph>> Make(
       GraphTopology&& topo_to_assign);
 
-  /// [[deprecated("You should provide a rdg dir")]]
+  /// Make a property graph from topology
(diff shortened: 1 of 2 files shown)


witchel (approved):
We appreciate your dedication to the cause.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -266,10 +266,16 @@ class KATANA_EXPORT PropertyGraph {
       const katana::RDGLoadOptions& opts = katana::RDGLoadOptions());
 
   /// Make a property graph from topology
+  // [deprecated("please provide a storage prefix")]
```
amberhassaan, line 269: A boring but useful followup task would be to add the `[[deprecated]]` attribute: See https://thek-teamworkspace.slack.com/archives/C02C1F8C9DL/p1644535228882339

amberhassaan (commented):
My bigger question: If we are generating a `PropertyGraph` "In-memory", do we use a default `rdg_dir`?

witchel:
> My bigger question: If we are generating a `PropertyGraph` "In-memory", do we use a default `rdg_dir`?

There is no system default, because it isn't clear what that should be ("./"?, "/tmp"?).  But if you are generating a property graph in memory and your ctest fixture generates a temp dir for your, I've been going with "./".

PR #893, opened 2022-02-12
Build: pin cmake version to <3.22

scober:
The specific issue is that the FindAWS<Something> cmake function that is bundled
with aws-sdk has a while loop that uses syntax that breaks in CMake
3.22.

### DIFF
--- a/conda_recipe/environment.yml
+++ b/conda_recipe/environment.yml
@@ -22,7 +22,8 @@ dependencies:
  - clang-tools>=10
  - clang-tools>=10
  - clangxx>=10
- - cmake>=3.17,<4a0
+   # the aws-sdk cmake is incompatible with version 3.22
+ - cmake>=3.17,<3.22a0
  - conda-build>=3
  - conda-verify>=3
  - coverage>=5.5
--- a/katana_requirements.yaml
+++ b/katana_requirements.yaml
@@ -110,6 +110,9 @@ cmake:
   labels:
     - apt
     - conda/dev
+  version_overrides:
+    conda/dev: ">=3.17,<3.22a0"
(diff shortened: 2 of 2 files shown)


tylershunt (approved):

gliga:
I was able to "fix" local build by changing:

while (NOT TEMP_NAME STREQUAL ${LIB_SEARCH_PREFIX}) 

to

while (NOT TEMP_NAME STREQUAL "${LIB_SEARCH_PREFIX}") 

and I used cmake 3.22.2

Not sure if this is a fix or not but build with conda went fine.

arthurp:
This needs a techdebt issue so we don't end up trapped on an old cmake at some point when we need to use a package that requires a new one.

Also would a cmake policy setting work around the limitation?

scober:
> Also would a cmake policy setting work around the limitation?

I think it probably would.

arthurp:
.... Then why oh why was that not the first thing you tried? It's more specific, requires changes in fewer places, and avoids laying a land mine for ourselves later.

scober:
Because I didn't know about cmake policies until you mentioned them here and I googled it.

arthurp:
:-) Ah. I see. Well then. Sorry I didn't mention it sooner. My review fu is failing me. Yea. Policies are a thing. They are really really weird and kinda scary, but they are better than some other options.

scober:
Yeah. I think they are probably the right solution to this problem. I will work on migrating. Also no need to apologize for suggesting a better solution!

arthurp:
Well I should maybe apologize for being so snarky about it. ;-) Thanks for following up with a better fix after this one.

PR #894, opened 2022-02-12
Enable setting types when creating graphs from numpy or pandas data. [KAT-1509]

arthurp:


libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -35,11 +35,23 @@ katana::python::EntityType::Make(
   }
 }
```
denish44, line 37: Unrelated but shouldn't we also call this file EntityTypeManagerPython.cpp?

libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -136,5 +160,9 @@ katana::python::InitEntityTypeManager(py::module_& m) {
             return ret;
           })
       .def("num_atomic_types", &katana::EntityTypeManager::GetNumAtomicTypes)
-      .def("num_types", &katana::EntityTypeManager::GetNumEntityTypes);
+      .def("num_types", &katana::EntityTypeManager::GetNumEntityTypes)
+      .def_property_readonly(
+          "unknown_type", [](katana::EntityTypeManager* self) {
```
denish44, line 164: When is this possible?

denish44 (approved):
LG

libkatana_python_native/src/EntityTypeManager.cpp
arthurp: We allow nodes and edges to have no defined type.

libkatana_python_native/src/EntityTypeManager.cpp
```diff
@@ -83,7 +95,18 @@ katana::python::InitEntityTypeManager(py::module_& m) {
...
+            return EntityType{
+                self, KATANA_CHECKED(self->AddAtomicEntityType(name))};
+          })
+      .def(
+          "get_or_add_atomic_entity_type",
+          [](katana::EntityTypeManager* self,
+             const std::string& name) -> Result<EntityType> {
+            return EntityType{
+                self, KATANA_CHECKED(self->GetOrAddEntityTypeID(name))};
+          })
```
pyrito: I think I may have wrapped the same method and called it in the enterprise repo. Oops.

python/katana/local/import_data.py
```diff
@@ -21,6 +28,56 @@
 ]
 
 
+class EntityTypeArray(EntityTypeManager):
```
pyrito: Love this class - thanks for doing this!

pyrito (approved):
LGTM overall - only concern is that I have merged in code that already wraps `GetOrAddEntityTypeID`

libkatana_python_native/src/EntityTypeManager.cpp
arthurp: ... How? This class isn't declared there.

python/katana/local/import_data.py
arthurp: Yea. I needed it, but I was pretty sure it would be generally useful.

Oh. I should add it to `__all__`.

libkatana_python_native/src/EntityTypeManager.cpp
pyrito: Oh to clarify, if you rebase you can see that I wrapped the same method in the open repo and just called it in the enterprise repo.

pyrito:
@arthurp my enterprise PR is still turtle crawling through CI. You could rebase, remove the duplicated code, and I could try to pick up these new changes in my enterprise PR. Lazy me doesn't want to do this because I just wanna get the damn thing merged, but I think this would be the right course of action.

libkatana_python_native/src/EntityTypeManager.cpp
arthurp: I was thinking about this. On one hand, I think it's silly to duplicate information from the directories in the filename. On the other hand it might be better to be explicit in the file name because people don't pay attention.
You are right that we should have matching header and source names. The questions is whether they should be `SubsystemPython` or just `Subsystem`. For example, should we have `python/EntityTypeManagerPython.{cpp,h}` or `python/EntityTypeManager.{cpp,h}`.

libkatana_python_native/src/EntityTypeManager.cpp
arthurp: Ah. OK. I'll rebase and handle the difference. It is odd that github thinks this is mergeable though.

pyrito: It's because I named my method differently :)

libkatana_python_native/src/EntityTypeManager.cpp
denish44: I would say explicit so `EntityTypeManagerPython.{cpp, h}`. I think some people can overlook that `python/` easily when searching.

arthurp: Discussed in slack: This will be handled in another PR.

arthurp:
> @arthurp my enterprise PR is still turtle crawling through CI. You could rebase, remove the duplicated code, and I could try to pick up these new changes in my enterprise PR. Lazy me doesn't want to do this because I just wanna get the damn thing merged, but I think this would be the right course of action.

This PR *will* break your PR since I removed the `get_or_add_entity_type_id` method you added since it does not return an ID so it is misnamed. I'm sure I reviewed it when you added it. I just wasn't paying attention I guess. Sorry.

pyrito:
> > @arthurp my enterprise PR is still turtle crawling through CI. You could rebase, remove the duplicated code, and I could try to pick up these new changes in my enterprise PR. Lazy me doesn't want to do this because I just wanna get the damn thing merged, but I think this would be the right course of action.
> 
> This PR _will_ break your PR since I removed the `get_or_add_entity_type_id` method you added since it does not return an ID so it is misnamed. I'm sure I reviewed it when you added it. I just wasn't paying attention I guess. Sorry.

Yes you're right. This will break my PR. I'll fix it though. And no worries. Worked out well enough.

PR #895, opened 2022-02-14
Disable bookkeeping assertion temporarily

witchel:
I hereby resolve to not only run ctest -L quick, but also all integration tests before I submit my PRs.

But in the meantime, this will unblock #3305.


pyrito (approved):
:')

PR #896, opened 2022-02-14
Fallback to installing with conda if mamba fails.

arthurp:
This is needed because there is some weird case where mamba fails to select the correct packages for download. It shouldn't happen often I think.

This PR fixes failures like this:

```
CondaHTTPError: HTTP 404 NOT FOUND for url <https://conda.anaconda.org/conda-forge/noarch/charset-normalizer-2.0.11-pyhd8ed1ab_0.conda>
```

in [`test_requirements`](https://github.com/KatanaGraph/katana/runs/5171337131?check_suite_focus=true#step:7:479).


arthurp:
Reviewers picked at random from people who I know have used conda a bit. One review is plenty.

pyrito (approved):

PR #897, opened 2022-02-14
Add BUILD_CPP_TESTING and BUILD_PYTHON_TESTING to control testing by language

arthurp:
This reduces the time required to link all the tests when you really just need the server and worker binaries to test python code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ add_subdirectory(libtsuba)
 add_subdirectory(libgalois)
 add_subdirectory(libgraph)
 
-if(KATANA_IS_MAIN_PROJECT AND BUILD_TESTING)
+if(KATANA_IS_MAIN_PROJECT AND cpp IN_LIST KATANA_LANG_TESTING)
   add_subdirectory(lonestar)
 else()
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ add_subdirectory(libtsuba)
 add_subdirectory(libgalois)
 add_subdirectory(libgraph)
 
-if(KATANA_IS_MAIN_PROJECT AND BUILD_TESTING)
+if(KATANA_IS_MAIN_PROJECT AND BUILD_CPP_TESTING)
```
ddn0: `BUILD_TESTING` is a semi-official variable: https://cmake.org/cmake/help/latest/module/CTest.html

but `BUILD_CPP_TESTING` and `BUILD_PYTHON_TESTING` are just our things, so I'd prefer that we give them a `KATANA_` prefix, e.g., `KATANA_LANG_TESTING` in correspondence to `KATANA_LANG_BINDINGS`.

arthurp: Good point. I'll update this when I have the time.

tylershunt (approved):

PR #898, opened 2022-02-15
support: Provide accessors to CommBackend fields

ddn0:
Begin the refactor away from direct field access.

### DIFF
--- a/libsupport/include/katana/CommBackend.h
+++ b/libsupport/include/katana/CommBackend.h
@@ -29,16 +29,38 @@ public:
   /// Notify other tasks that there was a failure; e.g., with MPI_Abort
   virtual void NotifyFailure() = 0;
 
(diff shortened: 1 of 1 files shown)


arthurp (approved):
Can you mark the fields deprecated and then suppress the warnings in the accessor methods? Might be a good idea. Though it's not required.

roshandathathri (approved):
Thanks!

ddn0:
> Can you mark the fields deprecated and then suppress the warnings in the accessor methods?

You can do the first but I don't think there is a way to do the second.

arthurp:
> > Can you mark the fields deprecated and then suppress the warnings in the accessor methods?
> 
> You can do the first but I don't think there is a way to do the second.

Aren't there local warning suppression pragmas? But anyway, this can all be later.

PR #901, opened 2022-02-15
fixup arrow, requirements install on ubuntu21.04

e-mcginnis:
on 21.04 the apt version is:
```
$apt-get --version
apt 2.2.4ubuntu0.1 (amd64)
```

which broke the version check in our requirements script.
Add a fallback path to catch this.

### DIFF
--- a/katana_requirements.yaml
+++ b/katana_requirements.yaml
@@ -419,7 +419,7 @@ libnuma:
   labels:
     - apt
 ccache:
-  version: [3, 4]
+  version: [3, 4.3]
   labels:
     - apt/dev
     - conda/apt
--- a/scripts/katana_requirements/__main__.py
+++ b/scripts/katana_requirements/__main__.py
@@ -197,9 +197,10 @@ def setup_install_arguments(parser):
 
 def get_apt_version():
     version_information_str = str(subprocess.check_output(["apt-get", "-v"], stderr=subprocess.DEVNULL), "UTF-8")
-    version_match = re.match(r"apt ([0-9.]+) \(.*\)", version_information_str)
+    version_match = re.match(r"apt ([0-9.]+)(?:ubuntu[0-9.]+)? \(.*\)", version_information_str)
     if not version_match:
         raise RuntimeError(f"apt-get returned unexpected version information:\n{version_information_str}")
+
     return Version(version_match.group(1))
 
 
--- a/scripts/setup_ubuntu.sh
+++ b/scripts/setup_ubuntu.sh
@@ -93,11 +93,13 @@ if [ "$VERSION" == "21" ]; then
     curl "http://archive.ubuntu.com/ubuntu/pool/main/r/re2/libre2-5_20200101+dfsg-1build1_amd64.deb" \
          --output /tmp/libre2-5.deb
     apt install -yq /tmp/libre2-5.deb && rm /tmp/libre2-5.deb
-else
-    curl "https://apache.jfrog.io/artifactory/arrow/$RELEASE_ID/apache-arrow-apt-source-latest-$RELEASE_CODENAME.deb" \
-         --output /tmp/apache-arrow-apt-source-latest.deb
 fi
+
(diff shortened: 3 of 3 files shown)


scripts/katana_requirements/__main__.py
```diff
@@ -199,8 +199,16 @@ def get_apt_version():
...
-        raise RuntimeError(f"apt-get returned unexpected version information:\n{version_information_str}")
-    return Version(version_match.group(1))
+        # Try matching against ubuntu tagged versions instead
+        version_match = re.match(r"apt ([0-9.]+)ubuntu([0-9.]+) \(.*\)", version_information_str)
+        if not version_match:
+            raise RuntimeError(f"apt-get returned unexpected version information:\n{version_information_str}")
+
+        version_string = version_match.group(1).split("u")[0]
+    else:
+        version_string = version_match.group(1)
```
arthurp: I think there regexs can be combined into something like this:

```
pt ([0-9.]+)(?:ubuntu[0-9.]+)? \(.*\)
```
(`(?:` begins a non-capturing group)

Though I'm not sure why you are splitting on `"u"` below. I think that might be a left over from an initial attempt.

arthurp (changes requested):

scripts/katana_requirements/__main__.py
e-mcginnis: that worked, thanks.

katana_requirements.yaml
```diff
@@ -373,7 +373,7 @@ python:
     conan: null
     apt: python3
   # TODO(amp): Unpin this. Currently we only build for python 3.8 so pin to that to reduce confusion.
-  version: [ 3.8, 3.9 ]
+  version: [ 3.8, 3.9.6 ]
```
e-mcginnis: @arthurp I had to bump the allowed versions of python & ccache for 21.04, is that a problem for our standard 20.04?

arthurp: YES!!! Nothing builds on python 3.8 yet. So if you want to do this, we need to upgrade everything to build on newer python.

ccache upgrade is just fine.

scripts/katana_requirements/__main__.py
```diff
@@ -197,9 +197,10 @@ def setup_install_arguments(parser):
 
 def get_apt_version():
     version_information_str = str(subprocess.check_output(["apt-get", "-v"], stderr=subprocess.DEVNULL), "UTF-8")
-    version_match = re.match(r"apt ([0-9.]+) \(.*\)", version_information_str)
+    version_match = re.match(r"apt ([0-9.]+)(?:ubuntu[0-9.]+)? \(.*\)", version_information_str)
     if not version_match:
-        raise RuntimeError(f"apt-get returned unexpected version information:\n{version_information_str}")
+            raise RuntimeError(f"apt-get returned unexpected version information:\n{version_information_str}")
```
arthurp: I think you changed the indent here.

katana_requirements.yaml
arthurp: BTW, I really do want to upgrade to supporting 3.8, 3.9, AND 3.10. But I haven't had time.

e-mcginnis: hm, then full 21.04 support will have to wait for now as keeping python pinned as-is causes apt package conflicts during script run-time. 
All of my builds seem happy enough on 3.9.5, but maybe I'm not touching stuff that would break.
I'll revert the python version change for now.

arthurp: Well you should be able to run everything in conda just fine. Using APT packages is getting deprecated anyway.

I think rolling this back for the moment is safest.

arthurp (approved):

katana_requirements.yaml
e-mcginnis: Yeah, I'm one of the few holdouts for now. I really should switch over to conda, but I have yet to find the time to debug the issues I had with it last time I tried.

arthurp: I understand.

If you can reproduce your problems in a docker container or on a machine I can log into, I would be happy to debug it and fix the tooling. When I can time, which might be a bit. Tooling bugs are bugs. Even if they only happen in some environments.

PR #902, opened 2022-02-16
Simplify module aliasing to fix multiloading issue [KAT-3840]

arthurp:


arthurp:
@JamesFCoffey I have you on here so you know about this since this may mess up some of the fully qualified module names in the documentation. I'm not really sure. I have ideas on how to fix it if it does mess it up.

python/katana/plugin.py
```diff
@@ -71,8 +71,8 @@ def load_module(self, fullname):
                 raise ImportError(
                     f"module {fullname} already exists while trying to insert {aliased_name} at that name."
                 )
```
pyrito, line 73: side note: this confuses me, if we are trying to load the module, we first check whether it's in sys.modules. If it's not, we check that it's in the insert table and we're at the stage where the aliased_name isn't in sys.modules either, so we now import that aliased_name. Now isn't it possible for `fullname` to be in `sys.modules`? Why would that be an error?

arthurp: Hmm. Interesting. Let me look at this. I think I had the idea that a re-entrant call was possible so I wanted to check from it.

pyrito: Yeah I guess your intuition was right because after this else case we also set `sys.modules[fullname]`

arthurp: I think this check is insanely unlikely to actually fail. But I also think it is possible. The case I see if that the importer finds the "fullname" initially, but then during loading of the aliased package it finds a different version of the "fullname" package and loads it, so we now have two packages that want to have the same name and are loaded from different places. I think it's possible. But very very unlikely. But it doesn't hurt to check for it.

JamesFCoffey (commented):
I'll take a look to see if the docs changed after this merges. I plan to get on KAT-1150 soon. Thanks for the head's up.

python/katana/plugin.py
```diff
@@ -71,8 +71,8 @@ def load_module(self, fullname):
...
-        module.__name__ = fullname
-        module.__package__ = fullname
```
pyrito, line 75: What was the original intent of having the module change its name to `fullname`? Why not keep the module name as is?

pyrito (approved):
I think I understand why things could get loaded twice, but still an interesting scenario. LGTM

python/katana/plugin.py
arthurp: Because I was hoping it would make `katana_enterprise.remote` report that it was `katana.remote` to improve error messages and help.

PR #906, opened 2022-02-17
Return nullptr in index_to_type_map_data if no types

aneeshdurg:
This was causing a sanitizer error when committing to an empty graph. It doesn't seem to actually change anything functionally.

### DIFF
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -598,6 +598,9 @@ public:
   void invalidate() noexcept { is_valid_ = false; };
 
   const EntityTypeID* index_to_type_map_data() const noexcept {
+    if (index_to_type_map_.empty()) {
+      return nullptr;
+    }
     return &index_to_type_map_[0];
   }
 


libgraph/include/katana/GraphTopology.h
```diff
@@ -598,6 +598,9 @@ class KATANA_EXPORT CondensedTypeIDMap : public GraphTopologyTypes {
   void invalidate() noexcept { is_valid_ = false; };
 
   const EntityTypeID* index_to_type_map_data() const noexcept {
+    if (index_to_type_map_.empty()) {
+      return nullptr;
```
roshandathathri, line 602: Shouldn't this be an error as we always have at least one entity type ID?

aneeshdurg: I think if you generate a completely empty graph it will have no type ids until the first commit.

aneeshdurg: or at least that was my understanding from looking at the metadata in the RDG

roshandathathri: Will this be called on empty graphs only in our unit tests or is that a user-visible use-case?

e-mcginnis: I think both directions of this mapping should always have the mapping between index 0 and kUnknownEntityType, like EntityTypeManager always has kUnknownEntityType even for empty graphs. 

Looks like the issue is that we only are going over in use edge EntityTypes, https://github.com/KatanaGraph/katana/blob/f83d4ddf9b5fb925fff54a969d0c4e030211c022/libgraph/src/GraphTopology.cpp#L536
so on an empty graph, kUnknownEntityType won't get included.

aneeshdurg: I don't know if we ever want users to create/use empty graphs, but I think it would be useful if we did explicitly support it. I think this solution isn't the best, and ideally this function should return an optional or something, but right now I just want to unblock our use cases of empty graphs.

aneeshdurg: @e-mcginnis are you suggesting that the real fix should be to just always add `kUnknownEntityType` to that set in the code you shared?

e-mcginnis: I'm not sure. I could also see the argument that the CondensedTypeIDMap should be empty on an empty Graph. 
Since the CondensedTypeID map is supposed to only contain of the in-use EntityTypeIDs. 

The more I look at it, the more I think it makes sense for the CondensedTypeIDMap (and so the id_typ_type_map) to be empty for an empty graph, no EntityTypeIDs are in use here, so there is nothing to hold on to.

roshandathathri: I think we need to decide the semantics. Empty graph issue aside, if a graph has nodes without assigning a type to any of the nodes (which is allowed in our system right?), then what is the current and new behavior?

aneeshdurg: no change. There's no functional change in this PR at all, it just makes sanitizer stop complaining.

e-mcginnis: default EntityType for nodes/edges is already kUnknownEntityType
see usages of:
```
katana::PropertyGraph::EntityTypeIDArray
MakeDefaultEntityTypeIDArray(size_t vec_sz) {
  katana::PropertyGraph::EntityTypeIDArray type_ids;
  type_ids.allocateInterleaved(vec_sz);
  katana::ParallelSTL::fill(
      type_ids.begin(), type_ids.end(), katana::kUnknownEntityType);
  return type_ids;
}
```
https://github.com/KatanaGraph/katana/blob/f83d4ddf9b5fb925fff54a969d0c4e030211c022/libgraph/src/PropertyGraph.cpp#L118

So a graph in your example would have a non-empty CondensedTypeIDMap, and so a non-empty index_to_type_map_

so this PR would have no functional change to that example. 
This change only effects empty graphs

roshandathathri: Thanks for your patience and clarification guys!

I'm concerned that if we return `nullptr`, it'll lead to crashes in the uses of this function for empty graphs because the users might NOT check whether it's empty or not. 

@mindv0rtex Can you please add a TODO/doxygen-comments to clarify this in a subsequent PR?

For now, I'll approve to merge this.

roshandathathri (approved):
Thanks

libgraph/include/katana/GraphTopology.h
e-mcginnis: > I'm concerned that if we return nullptr, it'll lead to crashes in the uses of this function for empty graphs because the users might NOT check whether it's empty or not.

agreed, we should not be returning nullptrs here. Getting pointers to potentially empty vectors leaves us in a bad spot .

e-mcginnis (commented):
lgtm

PR #908, opened 2022-02-18
Fixup test-datasets cache logic [KAT-3709]

e-mcginnis:
- retry more than once on failure
- introduce better logging
- cleanup in all fail scenerios

JIRA: KAT-3709

### DIFF
--- a/python/katana/example_data.py
+++ b/python/katana/example_data.py
@@ -2,9 +2,11 @@
 Utilities which download example data for testing and experimentation.
 """
 
+import logging
 import os
 import shutil
 import subprocess
+import time
 import urllib.request
 import zipfile
 from pathlib import Path
@@ -18,6 +20,8 @@ __all__ = ["get_rdg_dataset_at_version", "get_rdg_dataset", "get_csv_dataset", "
 # TODO(emcginnis) it would be really really nice if this got updated automatically
 # when the submodule ref held by open katana is updated
 DATASETS_SHA = "9934dddf3e37ecc14a21cfc814b117eac0266156"
+
+logger = logging.getLogger(__name__)
 
 
 def get_rdg_dataset_at_version(rdg_name, storage_format_version, as_url=False):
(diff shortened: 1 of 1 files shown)


e-mcginnis:
Immediately hit a networking flake in apt, how ironic

python/katana/example_data.py
```diff
@@ -105,17 +106,25 @@ def _get_dataset(rel_path):
...
-        raise ValueError(f"Dataset does not exist at {path}")
+    while not path.exists() or count < max_attempts:
+        try:
+            # lame backoff, costs nothing on first attempt
+            time.sleep(count)
+            datasets_path = _get_test_datasets_directory(invalidate_cache=invalidate_cache)
+            path = datasets_path / rel_path
+            return path
+        except ValueError as e:
+            print(f"Observed {e} while getting test datasets, ignoring and trying again")
```
arthurp: Use `logging`. https://docs.python.org/3/howto/logging.html#logging-basic-tutorial
Please use the `logger = logging.getLogger(__name__)` convention.

python/katana/example_data.py
```diff
@@ -150,6 +159,7 @@ def find_path() -> Path:
         found_path = None
         # If SRC_DIR or KATANA_SOURCE_DIR environment is set, just use it
         env_path = []
+        print(f"os.environ = {os.environ}")
```
arthurp: This shouldn't even be logged like this. Just remove this. If you want to log specific environment variables. That's fine.

python/katana/example_data.py
```diff
@@ -173,13 +183,13 @@ def download_datasets(invalidate_cache=False) -> Path:
...
-                    shutil.rmtree(cache_path)
-                except OSError:
-                    cache_path.unlink()
 
+        # failed to validate, cleanup just in case
+        if cache_path.exists():
+            try:
+                shutil.rmtree(cache_path)
+            except OSError:
+                cache_path.unlink(missing_ok=True)
```
arthurp: Do you want try-finally here?

arthurp (commented):
Please switch to logging before merging this. We don't need print spam in our tests. Looks OK though other than that.

python/katana/example_data.py
pyrito: Try not to use `print`. I think Arthur wanted us to use the `logging` module instead and add a TODO for Arthur there as well.

python/katana/example_data.py
pyrito: Debugging print?

python/katana/example_data.py
```diff
@@ -105,17 +109,25 @@ def _get_dataset(rel_path):
...
-        raise ValueError(f"Dataset does not exist at {path}")
+    while not path.exists() and count < max_attempts:
+        try:
+            # lame backoff, costs nothing on first attempt
+            time.sleep(count)
+            datasets_path = _get_test_datasets_directory(invalidate_cache=invalidate_cache)
+            path = datasets_path / rel_path
+            return path
+        except ValueError as e:
+            logger.info(f"Observed {e} while getting test datasets, ignoring and trying again")
```
arthurp, line 129: NIT: You can include the exception in the log message with some argument which I cannot remember. Not critical though.

arthurp (approved):
LGTM.

PR #911, opened 2022-02-18
Fix bugs affecting projection topologies

yan-zaretskiy:
Fix how projections with no resulting edges are constructed. Earlier implementation did not map to the correct original node ID in case projections resulted in 0 edges.

Additionally, this PR adds the `Enumerate` iterator wrapper, which enables Python-style iteration: `for [i, value] : collection`.

### DIFF
--- a/libgraph/src/PropertyGraph.cpp
+++ b/libgraph/src/PropertyGraph.cpp
@@ -270,23 +270,16 @@ katana::PropertyGraph::Make(
 std::unique_ptr<katana::PropertyGraph>
 katana::PropertyGraph::MakeEmptyEdgeProjectedGraph(
     const PropertyGraph& pg, uint32_t num_new_nodes,
-    const DynamicBitset& bitset) {
+    const DynamicBitset& nodes_bitset,
+    NUMAArray<Node>&& original_to_projected_nodes_mapping,
+    NUMAArray<GraphTopology::PropertyIndex>&&
+        projected_to_original_nodes_mapping) {
   const auto& topology = pg.topology();
 
   NUMAArray<Edge> out_indices;
   out_indices.allocateInterleaved(num_new_nodes);
 
   NUMAArray<Node> out_dests;
-  NUMAArray<Node> original_to_projected_nodes_mapping;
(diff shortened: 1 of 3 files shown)


libsupport/include/katana/Iterators.h
```diff
@@ -323,6 +323,36 @@ MakeDisjointRangesEnd(
       DisjointRangesIterator<I>::RangeKind::kSecond);
 }
 
+template <typename T>
+constexpr auto
+Enumerate(T&& iterable) {
```
arthurp, line 328: Nice. Always good to see C++ code libraries catching up to the 90s.

arthurp (approved):
LGTM. Though I didn't review in all that much detail.

libgraph/src/PropertyGraph.cpp
```diff
@@ -317,8 +310,18 @@ katana::PropertyGraph::MakeEmptyEdgeProjectedGraph(
...
-  return MakeEmptyEdgeProjectedGraph(pg, 0, bitset);
+    const katana::PropertyGraph& pg,
+    const katana::DynamicBitset& nodes_bitset) {
+  const auto& topology = pg.topology();
+  NUMAArray<Node> original_to_projected_nodes_mapping;
+  original_to_projected_nodes_mapping.allocateInterleaved(topology.NumNodes());
+  katana::ParallelSTL::fill(
+      original_to_projected_nodes_mapping.begin(),
+      original_to_projected_nodes_mapping.end(),
+      static_cast<Node>(topology.NumNodes()));
```
uditagar, line 321: but I thought you were going to make this as an identity function

yan-zaretskiy: `original_to_projected_nodes_mapping` has the size of the original nodes, and in this method it maps to nothing. The identity is handled by passing `original_to_projected_nodes_mapping` to `MakeEmptyEdgeProjectedGraph` in case we have no edges but all the nodes.

uditagar: I am still unable to see it. Sorry about that.

uditagar (approved):
looks fine

PR #904, opened 2022-02-17
RDG storage_format_version_6

e-mcginnis:
support for optional datastructures

moves optional datastructures from unstable, to stable

PR #913, opened 2022-02-19
Ensure FileFrame:PaddedWrite padding is initialized

e-mcginnis:
Not doing this causes non-determinism in the padding used, and is generally considered a "bad" thing
https://github.com/KatanaGraph/katana/pull/904#issuecomment-1045338809

### DIFF
--- a/libtsuba/src/FileFrame.cpp
+++ b/libtsuba/src/FileFrame.cpp
@@ -211,8 +211,8 @@ katana::FileFrame::PaddedWrite(
       "adding {} bytes of padding. nbytes = {}, byte_boundry = {}",
       num_padding_bytes, nbytes, byte_boundry);
   if (num_padding_bytes > 0) {
-    uint8_t padding[num_padding_bytes];
-    aro_sts = Write(&padding, num_padding_bytes * sizeof(uint8_t));
+    std::vector<uint8_t> padding(num_padding_bytes, 0);
+    aro_sts = Write(padding.data(), padding.size() * sizeof(uint8_t));
     if (!aro_sts.ok()) {
       return katana::ArrowToKatana(aro_sts.code());
     }


libtsuba/src/FileFrame.cpp
```diff
@@ -212,6 +212,7 @@ katana::FileFrame::PaddedWrite(
       num_padding_bytes, nbytes, byte_boundry);
   if (num_padding_bytes > 0) {
     uint8_t padding[num_padding_bytes];
+    std::fill(padding, padding + num_padding_bytes, 0);
```
witchel: Could you do this with a vector, something like this (I didn't compile it so it will be broken).
```
std::vector<uint8_t>padding(num_padding_bytes, 0);
aro_sts = Write(padding.data(), padding.size());
```

witchel (approved):
Can we make this more c++?

libtsuba/src/FileFrame.cpp
e-mcginnis: hows this?

witchel: I think the `padding.size()` is sufficient.  The `* sizeof(uint8_t)` seems like it is only correct because `sizeof(uin8_t) == 1`.

e-mcginnis: I'm not so sure, I think its the other way around. 
Write wants the number of bytes, not the number of elements:
```
arrow::Status
katana::FileFrame::Write(const void* data, int64_t nbytes) {
```
which would be:
```
vector.size() * sizeof(vector_type)
```

so in this case 

```
aro_sts = Write(padding.data(), padding.size());
```
would only be correct because `sizeof(uin8_t) == 1`

witchel: Ugh, I've been writing my own `size()` methods that return byte counts.  Sorry!

e-mcginnis: heh, no problem, honestly I wish Write just figured this out for us. I'll have to add that sometime.

libtsuba/src/FileFrame.cpp
```diff
@@ -211,8 +211,8 @@ katana::FileFrame::PaddedWrite(
       "adding {} bytes of padding. nbytes = {}, byte_boundry = {}",
       num_padding_bytes, nbytes, byte_boundry);
   if (num_padding_bytes > 0) {
-    uint8_t padding[num_padding_bytes];
-    aro_sts = Write(&padding, num_padding_bytes * sizeof(uint8_t));
```
pyrito, line 215: I didn't know we had some remnants of the dark ages in our code base.

PR #914, opened 2022-02-19
Check format of docstrings [KAT-3169]

JamesFCoffey:
Checks to see that docstrings conform to the Google style. It ignores files in `python/test`.

KAT-3169

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -318,7 +318,7 @@ jobs:
         if [ "${{ steps.filter.outputs.python }}" == "true" ]
         then
            . .github/workflows/activate_miniconda.sh
-           ./scripts/check_python_docstrings.sh ${{ steps.filter.outputs.python_files }}
+           pydocstyle --config=./pyproject.toml ${{ steps.filter.outputs.python_files }}
         else
            echo "No Python files were changed."
         fi
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -291,12 +291,17 @@ jobs:
     - name: Install dependencies
       run: |
         . .github/workflows/activate_miniconda.sh
-        mamba install -y -q -c conda-forge black=19.10b0 pylint=2.8.3
+        mamba install -y -q -c conda-forge black=19.10b0 pylint=2.8.3 pydocstyle=6.1.1
 
     - name: Check formatting
       run: |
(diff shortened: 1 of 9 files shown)


scripts/check_python_docstrings.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+export GIT_ROOT=$(readlink -f $(dirname $0)/..)
+
+BASE="$(cd "$(dirname "$(readlink -f "$0")")"; pwd)"
+"$BASE"/check_python.py --pydocstyle "$@"
```
arthurp: We really don't need this wrapper. I just made the other ones for backward compatibly. No reason to have it for new ones.

.github/workflows/python.yaml
```diff
@@ -303,13 +311,14 @@ jobs:
         list-files: shell
         filters: |
           python:
-            - added|modified: '**.py'  
-    - name: Check for lint
+            - added|modified: '**.py'
+    
+    - name: Check docstrings
       run: |
         if [ "${{ steps.filter.outputs.python }}" == "true" ]
```
arthurp, line 318: This is wrong for doc strings since presumably this check can also handle doc strings in extension modules which come from `.cpp` files.

arthurp:
How are you planning to handle the fact that most of the code doesn't actually honor this style yet?

.github/workflows/python.yaml
JamesFCoffey: @arthurp It looks like it does not check docstring in the extension modules. I am adding the Napoleon Sphinx extension to parse Google style docstrings, but I can't find a check to see if the docstrings are correct / present.

JamesFCoffey: I also added an example

arthurp: You are only checking the python *files* here. If you wanted to check the extensions you would need to when the loaded *modules* instead. I don't know if `pydocstyle` supports that.

arthurp: This can be addressed later of course.

arthurp (approved):

JamesFCoffey:
> How are you planning to handle the fact that most of the code doesn't actually honor this style yet?

A separate JIRA can reformat the doc strings, but this change nudges new and changed code to be fixed gradually.

PR #915, opened 2022-02-19
Ignore distutils packages, as PyYAML is needed for the tool itself

srogatch:
Previously, `setup_ubuntu.sh` script failed if it wants to install in `pip` a package that is already installed in the system (distutils). This could be worked around for some packages by uninstalling them from the system. However, `PyYAML` is required for running the Katana install tool itself, as well as many Ubuntu system packages depend on `python3-yaml`. So uninstalling it in distutils is not an option, and the only option I see is in passing `--ignore-installed` to `pip`.


arthurp (approved):
Good catch.

PR #916, opened 2022-02-20
tracer: add offset time from when a span starts to logs

patrickkenney9801:
Adds more metadata to logs so devs can quickly see how long a span took

### DIFF
--- a/libsupport/src/JSONTracer.cpp
+++ b/libsupport/src/JSONTracer.cpp
@@ -114,15 +114,15 @@ std::string
 BuildJSON(
     const std::string& trace_id, const std::string& span_data,
     const std::string& log_data, const std::string& tag_data,
-    const std::string& host_data) {
+    const std::string& host_data, uint64_t duration) {
   fmt::memory_buffer buf;
   uint32_t host_id = katana::ProgressTracer::Get().GetHostID();
   static auto begin = katana::Now();
   auto msec_since_begin = katana::UsSince(begin) / 1000;
 
   fmt::format_to(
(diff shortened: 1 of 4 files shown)


libsupport/src/JSONTracer.cpp
```diff
@@ -114,15 +114,15 @@ std::string
...
+    const std::string& host_data, uint64_t duration) {
   fmt::memory_buffer buf;
   uint32_t host_id = katana::ProgressTracer::Get().GetHostID();
   static auto begin = katana::Now();
   auto msec_since_begin = katana::UsSince(begin) / 1000;
 
   fmt::format_to(
-      std::back_inserter(buf), R"({{"host":{},"offset_ms":{})", host_id,
-      msec_since_begin);
+      std::back_inserter(buf), R"({{"host":{},"offset_ms":{},"span_ms":{})",
```
ddn0, line 124: I get how this makes it easier to read the raw traces, but why wouldn't we just make the tools that parse traces easier to use?

If people find using Jaeger inconvenient to view these traces, I'd rather just write our own TUI frontend to raw trace data than conflate a trace (data) with how it is viewed or analyzed.

patrickkenney9801: @l-hoang is the primary asker for this feature

patrickkenney9801: And the offset_ms fields are also just cruft to make the logs easier to use and aren't used to assemble traces

patrickkenney9801:
Closed waiting further review on adding features to tracing

PR #917, opened 2022-02-21
expand the list of non-coding URI characters

wkyu2kg:
This PR expands the list of non-coding URI characters based RFC-3986 (https://www.rfc-editor.org/rfc/rfc3986).

### DIFF
--- a/libsupport/src/URI.cpp
+++ b/libsupport/src/URI.cpp
@@ -117,6 +117,9 @@ DoJoinPath(std::string_view dir, std::string_view file) {
   return fmt::format("{}{}", dir, file);
 }
 
+// As defined by https://www.rfc-editor.org/rfc/rfc3986,
+// we encode characters outside this list
+// "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/?#[]@!$&'()*+,;="
 bool
 ShouldURLEncode(int c) {
   if ('a' <= c && c <= 'z') {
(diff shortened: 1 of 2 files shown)


libsupport/src/URI.cpp
```diff
@@ -140,12 +143,57 @@ ShouldURLEncode(int c) {
   if (c == '~') {
     return false;
   }
-
-  // We encode whole paths, so in addition to the standard unencoded characters
-  // above, we should not encode '/' either.
+  if (c == ':') {
```
ddn0: nit: can you convert this to a switch statement? At least for the parts that aren't compactly represented as a range?

ddn0 (approved):

libsupport/src/URI.cpp
wkyu2kg: They form several discontinuous segments of ranges. Using switch-case statements to list them out explicitly has better readability. 

```
!
#$
&'()*+,-./
:;
=
?@
[
]
_
~
```

ddn0: I mean simply replace

```
if ('A' <= c && c <= 'Z') {
  return false;
}
....
if (c == '!') {
  return false;
}

if (c == '$') {
  return false;
}
```

with
```
if ('A' <= c && c <= 'Z') {
  return false;
}

switch (c) {
  case '!': 
  case '$':
  ....
    return false;
}
```

> Using switch-case statements to list them out explicitly has better readability.

I'm not sure what you mean here as the current PR doesn't use switch-case statements at all.

wkyu2kg: I have done what you intended. My comment showed up before I push my commit (which is still under testing).

PR #918, opened 2022-02-21
Projected views in katana.local [KAT-2411]

arthurp:
* Adds iterators to DynamicBitset because I'm sick of not having them.
* Cleans up the PropertyGraph::MakeProjectedGraph method to accept type IDs and allow null selection.
* Add `Graph.project` to python to create a projected graph view.


denish44 (approved):
LG

arthurp:
@mindv0rtex I'd like your review of the C++ side of the code. Whenever you have time.

python/test/test_property_graph.py
```diff
@@ -325,3 +325,23 @@ def test_types(graph):
...
+    assert projected_graph.num_nodes() == 0
+    assert projected_graph.num_edges() == 0
+    projected_graph = graph.project([graph.node_types.atomic_types["Person"]])
+    assert projected_graph.num_nodes() == 45
+    assert projected_graph.num_edges() == 58
+    projected_graph = graph.project([graph.node_types.atomic_types["Person"]], edge_types=[])
+    assert projected_graph.num_nodes() == 45
+    assert projected_graph.num_edges() == 0
+    projected_graph = graph.project(
+        [graph.node_types.atomic_types["Message"]], [graph.edge_types.atomic_types["REPLY_OF"]]
```
bowu, line 348: This interface is really nice and useful. Can we have a test using at least two node/edge types?

libgraph/include/katana/PropertyGraph.h
```diff
@@ -279,12 +279,22 @@ class KATANA_EXPORT PropertyGraph {
...
-  static std::unique_ptr<PropertyGraph> MakeProjectedGraph(
+  [[deprecated]] static std::unique_ptr<PropertyGraph> MakeProjectedGraph(
       const PropertyGraph& pg, const std::vector<std::string>& node_types,
       const std::vector<std::string>& edge_types);
 
+  /// Make a projected graph from a property graph. Shares state with
+  /// the original graph.
+  static Result<std::unique_ptr<PropertyGraph>> MakeProjectedGraph(
+      const PropertyGraph& pg, const std::vector<std::string>* node_types,
+      const std::vector<std::string>* edge_types);
```
yan-zaretskiy: Could we pass `std::optional<std::vector>` instead, since we're improving the API?

arthurp: I would LOVE that. I assumed I would get a stink about that because of the copy. I'll change it. :-)

python/test/test_property_graph.py
arthurp: You got it. :-) You are correct that that would be a good test to have.

PR #919, opened 2022-02-22
tsuba: move EntityTypeManager out of support

tylershunt:
This is a prerequisite for moving away from slow bitset but I haven't
done that yet because the code has drifted such that DynamicBitset is
not a drop-in replacement any longer. I gave up after my effort met some
resistance in serialization.

The real change that I wanted here was to expose these functions that
interface between EntityTypeManager and storage.


e-mcginnis (approved):
This looks amazing to me, so happy to see this done.

re: DynamicBitset(Slow) serialization

I assume you are referring to 
```
  using tt_is_copyable = int;
```
and the pain of removing it?

tylershunt:
> hange that I wanted here was to expose these functions that
> interface between EntityTypeManager and storage.

Mostly that I couldn't get this to build because of enterprise's `gSerialize`. I suspect it's not that it's such a big problem since all of these pieces are serializable on their own, but it was taking me longer to figure out than my rat-hole timer would allow :D.

PR #920, opened 2022-02-22
Better memory management policies

witchel:
We provide four memory management policies (null, performance, meek, and
minimal). And a framework for evaluating them while reading a sequence of
"property" files consisting of large arrow tables.

The memory supervisor no longer tracks active memory, we rely on the OS
for that.  The problem with tracking active memory is that a property
might become active, then change size, then get returned to the
supervisor, making bookkeeping difficult.

Use MemoryAvailable from the OS, correct PutProperty, fix louvain_clustering.

### DIFF
--- a/libgalois/include/katana/MemoryPolicy.h
+++ b/libgalois/include/katana/MemoryPolicy.h
@@ -17,34 +17,34 @@ public:
 
   /// Given the current memory counts and whatever OS sources the policy consults,
   /// how much standby memory should we reclaim right now?
-  virtual count_t ReclaimForMemoryPressure(
-      [[maybe_unused]] count_t active,
-      [[maybe_unused]] count_t standby) const = 0;
+  virtual count_t ReclaimForMemoryPressure(count_t standby) const = 0;
 
   /// Given the current memory counts and whatever OS sources the policy consults,
   /// should we refuse discretionary allocations?
-  virtual bool MemoryPressureHigh(
-      [[maybe_unused]] count_t active,
-      [[maybe_unused]] count_t standby) const = 0;
+  virtual bool IsMemoryPressureHigh(count_t standby) const = 0;
 
   /// Given the current memory counts and whatever OS sources the policy consults,
   /// should we clean up and exit?
-  virtual bool KillSelfForLackOfMemory(
-      [[maybe_unused]] count_t active,
-      [[maybe_unused]] count_t standby) const = 0;
+  virtual bool KillSelfForLackOfMemory(count_t standby) const = 0;
+
+  void LogMemoryStats(const std::string& message, count_t standby);
 
   /// Utility function to find out our OOM score from Linux
   static uint64_t OOMScore();
+  /// Utility function to find out available memory in the machine
+  static uint64_t AvailableMemoryBytes();
 
   struct MemInfo;
   struct Thresholds {
     double high_used_ratio_threshold;
     double kill_used_ratio_threshold;
+    count_t kill_self_oom_threshold;
+    count_t high_pressure_oom_threshold;
   };
 
 protected:
   MemoryPolicy(Thresholds thresholds);
-  void UpdateMemInfo(MemInfo* mem_info, count_t active, count_t standby) const;
+  void UpdateMemInfo(MemInfo* mem_info, count_t standby) const;
 
   count_t physical() const { return physical_; }
   double high_used_ratio_threshold() const {
@@ -53,31 +53,53 @@ protected:
   double kill_used_ratio_threshold() const {
     return thresholds_.kill_used_ratio_threshold;
   }
+  count_t kill_self_oom_threshold() const {
+    return thresholds_.kill_self_oom_threshold;
+  }
+  count_t high_pressure_oom_threshold() const {
+    return thresholds_.high_pressure_oom_threshold;
+  }
 
 private:
   count_t physical_;
   Thresholds thresholds_;
 };
 
-/// Memory policy that just tries to avoid the OOM killer and does little to evict
-/// unused data from memory
+/// Memory policy that just tries to avoid the OOM killer.  Unfortunately, it is aggressive about
+/// dumping memory when the OOM score is high, which can be an over reaction.
 class KATANA_EXPORT MemoryPolicyMinimal : public MemoryPolicy {
 public:
   MemoryPolicyMinimal();
-  count_t ReclaimForMemoryPressure(
-      count_t active, count_t standby) const override;
-  bool MemoryPressureHigh(count_t active, count_t standby) const override;
-  bool KillSelfForLackOfMemory(count_t active, count_t standby) const override;
+  count_t ReclaimForMemoryPressure(count_t standby) const override;
+  bool IsMemoryPressureHigh(count_t standby) const override;
+  bool KillSelfForLackOfMemory(count_t standby) const override;
 };
 
 /// Memory policy that prioritizes performance, i.e., it uses memory aggressively
 class KATANA_EXPORT MemoryPolicyPerformance : public MemoryPolicy {
 public:
   MemoryPolicyPerformance();
-  count_t ReclaimForMemoryPressure(
-      count_t active, count_t standby) const override;
-  bool MemoryPressureHigh(count_t active, count_t standby) const override;
-  bool KillSelfForLackOfMemory(count_t active, count_t standby) const override;
+  count_t ReclaimForMemoryPressure(count_t standby) const override;
+  bool IsMemoryPressureHigh(count_t standby) const override;
+  bool KillSelfForLackOfMemory(count_t standby) const override;
+};
+
+/// Minimize use of memory, but take free memory when it is available.
+class KATANA_EXPORT MemoryPolicyMeek : public MemoryPolicy {
+public:
+  MemoryPolicyMeek();
+  count_t ReclaimForMemoryPressure(count_t standby) const override;
+  bool IsMemoryPressureHigh(count_t standby) const override;
+  bool KillSelfForLackOfMemory(count_t standby) const override;
+};
+
+/// Do nothing to ever shed memory.  This will OOM if we occupy too much memory.
+class KATANA_EXPORT MemoryPolicyNull : public MemoryPolicy {
+public:
+  MemoryPolicyNull();
+  count_t ReclaimForMemoryPressure(count_t standby) const override;
+  bool IsMemoryPressureHigh(count_t standby) const override;
+  bool KillSelfForLackOfMemory(count_t standby) const override;
 };
 
 }  // namespace katana
--- a/libgalois/include/katana/MemorySupervisor.h
+++ b/libgalois/include/katana/MemorySupervisor.h
@@ -6,6 +6,7 @@
 #include "katana/Cache.h"
 #include "katana/Manager.h"
 #include "katana/MemoryPolicy.h"
+#include "katana/ProgressTracer.h"
 #include "katana/Result.h"
 #include "katana/config.h"
 
@@ -22,6 +23,12 @@ namespace katana {
 /// The MS does not manage per-allocation tokens, it only manages sizes
 /// Clients are trusted to call the proper functions or the MS will make
 /// bad decisions.
+///
+/// The MS does not track active memory.  The problem is that data structures (e.g.,
+/// properties) change in size while they are active.  If client code loads a 100 byte
+/// property and adds a value, it can store back a 102 byte property.  The MS sees 100
+/// bytes made active and 102 bytes made inactive, which makes the count of active
+/// bytes negative.
 
 class PropertyManager;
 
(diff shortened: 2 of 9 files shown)


libgalois/include/katana/MemoryPolicy.h
```diff
@@ -17,34 +17,34 @@ class KATANA_EXPORT MemoryPolicy {
...
-  virtual bool KillSelfForLackOfMemory(
-      [[maybe_unused]] count_t active,
-      [[maybe_unused]] count_t standby) const = 0;
+  virtual bool KillSelfForLackOfMemory(count_t standby) const = 0;
+
+  void LogMemoryStats(const std::string& message, count_t standby);
 
   /// Utility function to find out our OOM score from Linux
   static uint64_t OOMScore();
+  /// Utility function to find out availble memory in the machine
```
pyrito: nit: typo for available

libsupport/include/katana/Cache.h
```diff
@@ -279,12 +281,12 @@ class KATANA_EXPORT Cache {
 
   ReplacementPolicy policy_;
   // for kLRUSize number of entries kLRUBytes it is byte total
-  size_t capacity_{0};
-  size_t total_bytes_{0};
+  int64_t capacity_{0};
+  int64_t total_bytes_{0};
```
pyrito, line 285: Why signed integers instead of unsigned?

libgalois/include/katana/MemoryPolicy.h
```diff
@@ -53,31 +53,53 @@ class KATANA_EXPORT MemoryPolicy {
...
+  bool KillSelfForLackOfMemory(count_t standby) const override;
+};
+
+/// Do nothing to ever shed memory.  This will OOM if we occupy too much memory.
+class KATANA_EXPORT MemoryPolicyNull : public MemoryPolicy {
+public:
+  MemoryPolicyNull();
+  count_t ReclaimForMemoryPressure(count_t standby) const override;
+  bool IsMemoryPressureHigh(count_t standby) const override;
+  bool KillSelfForLackOfMemory(count_t standby) const override;
```
pyrito, line 102: The interface you've written in this file is pretty clean. Is it necessary to explicitly redefine the virtual functions and override them or can you just define the function in MemoryPolicy.cpp directly?

libgalois/include/katana/MemorySupervisor.h
```diff
@@ -22,6 +23,12 @@ namespace katana {
 /// The MS does not manage per-allocation tokens, it only manages sizes
 /// Clients are trusted to call the proper functions or the MS will make
 /// bad decisions.
+///
+/// The MS does not track active memory.  The problem is that data structures (e.g.,
+/// properties) change in size while they are active.  If client code loads a 100 byte
+/// property and adds a value, it can store back a 102 byte property.  The MS sees 100
+/// bytes made active and 102 bytes made inactive, which makes the count of active
+/// bytes negative.
```
pyrito, line 31: Ah this is a bummer. I assume any solution to this would be more intrusive and time intensive the implement.

libgalois/include/katana/MemorySupervisor.h
```diff
@@ -38,32 +45,32 @@ class KATANA_EXPORT MemorySupervisor {
...
 
-  /// Notify MS that manager freed \p bytes of active memory.
-  void ReturnActive(const std::string& name, count_t bytes);
   /// Notify MS that manager freed \p bytes of standby memory.
   void ReturnStandby(const std::string& name, count_t bytes);
 
   /// Manager \p name wants to transition \p bytes from active to standby
-  /// Returns the number of standby bytes allowed by MS, possibly 0.
-  count_t ActiveToStandby(const std::string& name, count_t bytes);
+  void ActiveToStandby(const std::string& name, count_t bytes);
```
pyrito, line 56: What is "standby" memory? I assume active is for data structures that are in scope and being used.

libgalois/include/katana/MemorySupervisor.h
```diff
@@ -78,19 +85,20 @@ class KATANA_EXPORT MemorySupervisor {
...
+      return 0;
+    }
+    return physical_ - rss;
+  }
 
   std::unique_ptr<MemoryPolicy> policy_;
-  /// Sum of all active memory across all managers
-  count_t active_{};
-  void ActiveMinus(ManagerInfo& info, count_t bytes);
-  void ActivePlus(ManagerInfo& info, count_t bytes);
```
pyrito, line 93: We still have the method `ActiveToStandby` but we're no longer keeping track of active memory in the supervisor?

libgalois/include/katana/MemorySupervisor.h
```diff
@@ -78,19 +85,20 @@ class KATANA_EXPORT MemorySupervisor {
...
-    count_t active{};
     count_t standby{};
   };
   std::unordered_map<std::string, ManagerInfo> managers_;
 
-  count_t Used() { return active_ + standby_; }
-  count_t Available() { return physical_ - Used(); }
+  count_t Available() {
+    auto rss = static_cast<katana::count_t>(
+        katana::ProgressTracer::ParseProcSelfRssBytes());
```
pyrito, line 94: Does `rss` include memory that is swapped out or is it just what's paged in at the moment?

libgalois/src/MemoryPolicy.cpp
```diff
@@ -24,25 +25,31 @@ namespace {
...
+  LogIt(message, &mem_info);
+}
+
+void
+katana::MemoryPolicy::UpdateMemInfo(MemInfo* mem_info, count_t standby) const {
   mem_info->standby = standby;
   mem_info->rss_bytes = katana::ProgressTracer::ParseProcSelfRssBytes();
+  mem_info->available_bytes = AvailableMemoryBytes();
   mem_info->used_ratio = (double)mem_info->rss_bytes / physical();
   mem_info->oom_score = OOMScore();
```
pyrito, line 54: TIL about OOMScore!

libgalois/src/MemoryPolicy.cpp
```diff
@@ -67,61 +74,58 @@ katana::MemoryPolicyPerformance::MemoryPressureHigh(
...
 
 katana::MemoryPolicyPerformance::MemoryPolicyPerformance()
-    : MemoryPolicy(
-          {.high_used_ratio_threshold = 0.85,
-           .kill_used_ratio_threshold = 0.93}) {}
+    : MemoryPolicy({
+          .high_used_ratio_threshold = 0.85,
+          .kill_used_ratio_threshold = 0.95,
+          .kill_self_oom_threshold = 1280,
+          .high_pressure_oom_threshold = 1100,
```
pyrito, line 119: hold on, is it even possible to have a OOM score that is larger than 1000? If the score is usage * 10, wouldn't that be the max? Or can usage exceed 100% when we consider swap as well?

libgalois/src/MemoryPolicy.cpp
```diff
@@ -67,61 +74,58 @@ katana::MemoryPolicyPerformance::MemoryPressureHigh(
...
-    count_t active, count_t standby) const {
+    count_t standby) const {
   MemInfo mem_info;
-  UpdateMemInfo(&mem_info, active, standby);
+  UpdateMemInfo(&mem_info, standby);
 
-  if (mem_info.oom_score > 1200 ||
-      mem_info.used_ratio > kill_used_ratio_threshold()) {
+  if ((mem_info.oom_score > kill_self_oom_threshold() ||
+       mem_info.used_ratio > kill_used_ratio_threshold()) &&
```
pyrito, line 106: Is there ever a case where the used ratio is greater than the threshold but the OOM score doesn't exceed the threshold? It seems like they should both go hand in hand?

libgalois/include/katana/MemorySupervisor.h
```diff
@@ -38,32 +45,32 @@ class KATANA_EXPORT MemorySupervisor {
...
-  /// Application cannot continue if it does not get memory
-  void BorrowActive(const std::string& name, count_t bytes);
   /// Request permission to allocate \p bytes for standby memory
   /// Returns the number of bytes granted, possibly 0
   count_t BorrowStandby(const std::string& name, count_t goal);
 
-  /// Notify MS that manager freed \p bytes of active memory.
-  void ReturnActive(const std::string& name, count_t bytes);
   /// Notify MS that manager freed \p bytes of standby memory.
   void ReturnStandby(const std::string& name, count_t bytes);
```
pyrito: For some reason I expected this function to return some value. Perhaps the name is a little misleading.

libgalois/src/PropertyManager.cpp
```diff
@@ -52,53 +52,48 @@ katana::PropertyManager::GetProperty(const katana::Uri& property_path) {
...
-  if (goal >= static_cast<count_t>(cache_->size())) {
-    total = static_cast<count_t>(cache_->size());
-    MemorySupervisor::Get().ReturnStandby(Name(), std::min(goal, total));
-    cache_->clear();
-  } else {
-    total = static_cast<count_t>(cache_->Reclaim(goal));
-    MemorySupervisor::Get().ReturnStandby(Name(), std::min(goal, total));
-  }
+  auto reclaim = static_cast<count_t>(cache_->Reclaim(goal));
+  MemorySupervisor::Get().ReturnStandby(Name(), reclaim);
```
pyrito: Can the amount of standby memory change by the time we try to reclaim X bytes that are passed in? Or is that the point of `Reclaim` in the property cache?

libgalois/src/MemorySupervisor.cpp
```diff
@@ -17,18 +16,19 @@ const std::string oversubscribed_str = "memory manager oversubscribed";
...
                {"standby", standby},
+               {"reclaimed", bytes_reclaimed},
            });
 }
 
 void
-KillCheck(katana::MemoryPolicy* policy, count_t active, count_t standby) {
-  if (policy->KillSelfForLackOfMemory(active, standby)) {
-    LogState(oversubscribed_str, active, standby);
+KillCheck(
```
pyrito, line 28: This name is a little confusing but I think it's better than renaming the function to `ShouldIKillMyself`

pyrito (commented):
I really enjoyed reviewing this PR. One of the few where most use of abstractions were intuitive and clear. Mostly questions/comments.

libsupport/include/katana/Cache.h
witchel: Good question.  It is so I can detect underflow by testing for negative values.

libgalois/include/katana/MemoryPolicy.h
witchel: I need to tell the compiler that I am overriding them because they are pure virtual in the base class.  But if they weren't pure virtual and I omitted their type signature in the derived class, I'd get the base implementation, which I don't want.

libgalois/include/katana/MemorySupervisor.h
witchel: I think so.  I want to drop this feature for properties at least.

libgalois/include/katana/MemorySupervisor.h
witchel: Standby memory is a cache for storage.  It can always be purged without harming functional correctness.

libgalois/include/katana/MemorySupervisor.h
witchel: Yep.  The application and the memory supervisor have the concept of active memory and there are some state transitions we need, but the supervisor isn't tracking the amount of active memory.

libgalois/include/katana/MemoryPolicy.h
pyrito: Right, I was just wondering if you could override and define the functions in .cpp without having to redefine the interface here, if that makes sense. But perhaps in order for the vtable lookup to work, it needs it defined in the headers as well? Just seems repetitive.

libgalois/include/katana/MemorySupervisor.h
witchel: It is only resident.  I have a TODO in the code to worry about swap, but I think that is a second-order effect.

libsupport/include/katana/Cache.h
pyrito: Gotcha. Probably a smart way to catch bugs instead of scratching your head at overflow wrap arounds.

libgalois/src/MemoryPolicy.cpp
witchel: It is strictly inferior to `/sys/fs/cgroup/memory/memory.pressure_level` which gives a direct interface to the kernel's judgment of a process' level of memory pressure, but that interface seems deprecated.

libgalois/src/MemoryPolicy.cpp
witchel: It can be greater than 1000.  I don't know what the scale is.

libgalois/include/katana/MemorySupervisor.h
pyrito: Would properties be standby if they were just being cached but not actively used?

libgalois/src/MemoryPolicy.cpp
witchel: Not necessarily.  The OOM score goes up a lot when a process starts allocating a lot of memory, so there is a velocity associated with the score.  But I don't have a detailed understanding of the components that go into the OOM score.  I should just read the kernel code.

libgalois/src/MemoryPolicy.cpp
pyrito: Interesting. Some sources I read online say that the max and others say that it can be greater than 1000. I wonder if it's based on the kernel version.

libgalois/include/katana/MemorySupervisor.h
witchel: That is a good point.  I was going for `Borrow/Return`.  I changed it to `Get/Put`.

libgalois/src/PropertyManager.cpp
witchel: The code is single-threaded, so I don't anticipate the amount of standby memory changing.

libgalois/src/MemorySupervisor.cpp
witchel: Yeah, I hear you.  It is an internal function, along with `SanityCheck`.

libgalois/include/katana/MemoryPolicy.h
witchel: I'm not a c++ expert, but I think it is necessary.

libgalois/include/katana/MemorySupervisor.h
witchel: Yes.

libgalois/src/MemoryPolicy.cpp
witchel: I am worried about being robust across different kernel versions.  This code has changed a bunch in 10+ years.  There is also `/proc/self/oom_adj` (deprecated) and `/proc/self/oom_score_adj`.  I don't deal with these right now.

libgalois/src/MemorySupervisor.cpp
pyrito: Fair enough

pyrito (approved):
LGTM. I hope at some point we can have some memory stress tests

PR #922, opened 2022-02-23
Fix a UB bug in BuildArray; avoid commiting projected topologies

yan-zaretskiy:
This PR fixes two issues: 
1) It prevents the projected topologies from being committed to the RDG, which would otherwise conflict with the original topologies.
2) It fixes a bug in BuildArray, whereby the `builder.AppendValues()` call could have been passed an empty vector, which causes an UB.


uditagar (approved):
looks fine, man.

PR #923, opened 2022-02-23
Correct optional datastructure file RDG move paths

e-mcginnis:
just another reason getting the list of rdg files needs a test! But thats going to have to wait for now as it will be non-trivial to implement...

added some useful loglines as well

### DIFF
--- a/libtsuba/src/RDGManifest.cpp
+++ b/libtsuba/src/RDGManifest.cpp
@@ -9,6 +9,7 @@
 #include "katana/ErrorCode.h"
 #include "katana/FileView.h"
 #include "katana/JSON.h"
+#include "katana/Logging.h"
 #include "katana/ParquetReader.h"
 #include "katana/RDGOptionalDatastructure.h"
 #include "katana/Result.h"
@@ -286,7 +287,7 @@ katana::RDGManifest::FileNames() {
       }
 
       for (auto it : header.optional_datastructure_manifests()) {
-        fnames.emplace(it.first);
+        fnames.emplace(it.second);
         KATANA_CHECKED(AddOptionalDatastructureSubfiles(
             fnames, katana::Uri::JoinPath(dir().string(), it.second)));
       }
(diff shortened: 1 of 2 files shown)


yshalabi:
> just another reason getting the list of rdg files needs a test! But thats going to have to wait for now as it will be non-trivial to implement...
> 
> added some useful loglines as well

I tried to add a test for this in katana-enterprise #2093 but when the underlying interface was refactored before I could merge I decided to close it and just revisit later when the interface stabilizes.... of course I forgot. IDK if there is anything useful there but you can check it out to see if it is worth resuscitating.

e-mcginnis:
I'm thinking it will look a bit like the storage format version test, basically check the actual on disk version of the maximal rdg against the FileNames output

libtsuba/src/RDGManifest.cpp
```diff
@@ -286,7 +287,7 @@ katana::RDGManifest::FileNames() {
       }
 
       for (auto it : header.optional_datastructure_manifests()) {
-        fnames.emplace(it.first);
+        fnames.emplace(it.second);
```
yshalabi, line 290: These kinds of bugs are harder to make if we have wrappers around the iterator type which return the desired value. e.g: FileNameFrom...

libtsuba/src/RDGPartHeader.cpp
```diff
@@ -345,10 +345,14 @@ katana::to_json(json& j, const katana::RDGPartHeader& header) {
         "RDGPartHeader.unstable_storage_format_ is true");
   }
 
-  KATANA_LOG_DEBUG(
+  KATANA_LOG_WARN(
```
yshalabi, line 348: helpful stuff I think

yshalabi (approved):
LGTM

libtsuba/src/RDGManifest.cpp
pyrito: Dang

PR #924, opened 2022-02-23
Pin markupsafe since 2.1 is not compatible with jinja2.

arthurp:
jinja2 isn't limiting the versions enough. And markupsafe
removed a module in a minor release.

Mirrors: https://github.com/KatanaGraph/katana-enterprise/pull/3398

### DIFF
--- a/conda_recipe/environment.yml
+++ b/conda_recipe/environment.yml
@@ -38,6 +38,7 @@ dependencies:
  - isort>=5.9.3,<5.10a0
  - jemalloc>=5.2.1
  - jinja2>=2.11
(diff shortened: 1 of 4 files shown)


jdent23 (approved):
Why is the constraint here `markupsafe<2.1` versus on enterprise it's `markupsafe==2.0`?

arthurp:
> Why is the constraint here `markupsafe<2.1` versus on enterprise it's `markupsafe==2.0`?

I just messed up. I actually think `markupsafe<2.1` is better. I'm gonna use that in enterprise. Good catch.

PR #925, opened 2022-02-24
Remove ambiguous overload

rob-isaac:
GetOutEdges(<node>) was implemented in the base class as well as the derived class (with the same implementation). Since the derived class declared `using Base::GetOutEdges` it was getting two indistinguisable declarations for the same function, leading to an compiler error for ambiguity when trying to use that function.

### DIFF
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -1163,7 +1163,9 @@ public:
     return Base::out().DoesEdgeTypeExist(edge_type);
   }
 
+  using Base::InDegree;
   using Base::InEdges;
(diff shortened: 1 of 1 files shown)


yan-zaretskiy:
Good catch. Could you add `using Base::InDegree` and `using Base::OutDegree` to this PR?

yan-zaretskiy (approved):

amberhassaan (approved):
Agree with Yan's suggestion about degree methods.

rob-isaac:
> Good catch. Could you add `using Base::InDegree` and `using Base::OutDegree` to this PR?

Done

PR #930, opened 2022-02-25
support: Rename Uri to URI

ddn0:


roshandathathri (approved):
Didn't look at the code. Seems like a find-and-replace. Approve the change!

tylershunt (approved):

ddn0:
> Seems like a find-and-replace.

It was manual because I didn't want to break existing code which used the name `FooUri`. I only changed the type.

pyrito (approved):

PR #931, opened 2022-02-25
Fix copying a projected PG detached from original properties [KAT-4107]

yan-zaretskiy:
Clustering operations require copying a property graph to the new purely in-memory instance that only contains the topology. To do it correctly, we have to discard any property index indirection when copying the topology.

I also added bounds checking to `TypedPropertyGraph` and `TypedPropertyGraphView` methods, which would have caught this issue.

### DIFF
--- a/libgraph/include/katana/analytics/ClusteringImplementationBase.h
+++ b/libgraph/include/katana/analytics/ClusteringImplementationBase.h
@@ -489,16 +489,11 @@ struct ClusteringImplementationBase {
  */
   static katana::Result<std::unique_ptr<katana::PropertyGraph>>
   DuplicateGraphWithSameTopo(const katana::PropertyGraph& pfg_from) {
-    const katana::GraphTopology& topology_from = pfg_from.topology();
-
-    katana::GraphTopology topo_copy = GraphTopology::Copy(topology_from);
-
-    auto pfg_to_res = katana::PropertyGraph::Make(std::move(topo_copy));
-    if (!pfg_to_res) {
-      return pfg_to_res.error();
-    }
-    return std::unique_ptr<katana::PropertyGraph>(
-        std::move(pfg_to_res.value()));
+    // The resulting PropertyGraph is detached from the property table of the input graph.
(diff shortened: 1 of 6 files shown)


libgraph/include/katana/analytics/ClusteringImplementationBase.h
```diff
@@ -489,16 +489,11 @@ struct ClusteringImplementationBase {
...
-    if (!pfg_to_res) {
-      return pfg_to_res.error();
-    }
-    return std::unique_ptr<katana::PropertyGraph>(
-        std::move(pfg_to_res.value()));
+    // The resulting PropertyGraph is detached from the property table of the input graph.
+    // Thus any property index indirection of the original topology has to be dropped.
+    katana::GraphTopology topo_copy =
+        GraphTopology::CopyWithoutPropertyIndexes(pfg_from.topology());
+    return katana::PropertyGraph::Make(std::move(topo_copy));
```
uditagar, line 496: what will happen to the underlying edge weights in the copied graph?

uditagar (changes requested):
I am going to put a blocker on this until I fully understand the change in `ClusteringImplementationBase.h `file.

libgraph/include/katana/analytics/ClusteringImplementationBase.h
yan-zaretskiy: This function creates a brand-new property graph that contains a copy of the topology and nothing else. There are no edge weights. I didn't change any of that logic. What this PR does is that it makes sure that when a new property is created on this new property graph, we can index into the property column correctly, by dropping the original projected-to-original mappings.

uditagar: then either this function is not being used anymore or something else is going on. I will check this then. thanks

uditagar: ok, it seems that edge property is calculated in a separate function, so it seems fine. thanks

uditagar (approved):
looks fine

PR #932, opened 2022-02-28
experimental: use string lookup rather than a static variable [KAT-4123]

tylershunt:
experimental: switch to global variables and a string registry

    At first, we avoided this method because I wanted to prioritize failing
    to compile when the developer mistyped a feature flag. That caused some
    problems with multiple declarations that we decided were more important.

    This version will cause a debug build to fail when a developer uses a
    flag that was not registered.

### DIFF
--- a/libsupport/include/katana/Experimental.h
+++ b/libsupport/include/katana/Experimental.h
@@ -1,37 +1,48 @@
 #ifndef KATANA_LIBSUPPORT_KATANA_EXPERIMENTAL_H_
 #define KATANA_LIBSUPPORT_KATANA_EXPERIMENTAL_H_
 
+#include <iomanip>
 #include <memory>
 #include <mutex>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
+#include "katana/Logging.h"
 #include "katana/config.h"
 
-namespace katana {
-namespace internal {
+namespace katana::internal {
 
 /// ExperimentalFeature tracks the state of feature flags set in the environment;
 /// It is not intended to be used directly; please see the macro KATANA_EXPERIMENTAL_FEATURE
 /// below
 class KATANA_EXPORT ExperimentalFeature {
 public:
-  static const ExperimentalFeature* Register(
-      const std::string& feature_name, const std::string& filename,
-      int line_number);
+  ExperimentalFeature(const char* name, const char* filename, int line_number)
+      : name_(name), filename_(filename), line_number_(line_number) {
+    std::lock_guard<std::mutex> lock(registered_features_mutex_);
+    registered_features_.emplace(name_, this);
+  }
 
-  /// report the feature flags that were checked on codepaths that were executed and
-  /// the flag was set to true
+  static bool IsEnabled(const char* feature_name) {
+    auto it = registered_features_.find(feature_name);
+    if (it == registered_features_.end()) {
+      KATANA_LOG_DEBUG_VASSERT(
+          false, "reference to unregistered feature flag {}", feature_name);
+      return false;
+    }
+    it->second->CheckEnv();
+    return it->second->is_enabled_;
+  }
+
+  /// report the feature flags that were enabled
   static std::vector<std::string> ReportEnabled();
 
-  /// report the feature flags that were used but stayed false
+  /// report the feature flags that were disabled
   static std::vector<std::string> ReportDisabled();
 
   /// report the feature flags that were provided but did not match any registered flag
   static std::vector<std::string> ReportUnrecognized();
-
-  bool IsEnabled() const { return is_enabled_; }
 
   ExperimentalFeature(const ExperimentalFeature& no_copy) = delete;
   ExperimentalFeature& operator=(const ExperimentalFeature& no_copy) = delete;
@@ -44,13 +55,6 @@ public:
   int line_number() const { return line_number_; }
 
 private:
-  ExperimentalFeature(std::string name, std::string filename, int line_number)
-      : name_(std::move(name)),
-        filename_(std::move(filename)),
-        line_number_(line_number) {
-    CheckEnv();
-  }
-
   void CheckEnv();
 
   std::string name_;
@@ -59,12 +63,11 @@ private:
   bool is_enabled_{};
 
   static std::mutex registered_features_mutex_;
-  static std::unordered_map<std::string, std::unique_ptr<ExperimentalFeature>>
+  static std::unordered_map<std::string, ExperimentalFeature*>
       registered_features_;
 };
 
-}  // namespace internal
-}  // namespace katana
+}  // namespace katana::internal
 
 /// KATANA_EXPERIMENTAL_FEATURE creates a flag that can be set from the environment.
 /// The macro takes a feature_name which should be an unquoted, unique string that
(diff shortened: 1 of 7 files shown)


ddn0:
I see warnings like this locally:

```
[1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
[1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
[1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/DeleteOperator.cpp:20
[1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
[1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
[1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/RemoveOperator.cpp:18
[1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
[1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
[1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/UpdateOperator.cpp:24
```

will this change fix these?

tylershunt:
> I see warnings like this locally:
> 
> ```
> [1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
> [1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
> [1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/DeleteOperator.cpp:20
> [1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
> [1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
> [1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/RemoveOperator.cpp:18
> [1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
> [1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
> [1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/UpdateOperator.cpp:24
> ```
> 
> will this change fix these?

Hmm no not this case, that's an example of the behavior I thought would be good to warn folks about; I was seeing warnings where it was complaining about two paths to the same header file. Perhaps I should just disable the warning all together? It doesn't seem to be buying us much and is just more of an annoyance.

tylershunt:
> > I see warnings like this locally:
> > ```
> > [1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
> > [1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
> > [1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/DeleteOperator.cpp:20
> > [1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
> > [1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
> > [1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/RemoveOperator.cpp:18
> > [1,0]<stderr>:WARNING: /home/ddn/w/katana/external/katana/libsupport/src/Experimental.cpp:74: UseLogStructuredForUpdates declared in multiple places:
> > [1,0]<stderr>:  here: /home/ddn/w/katana/libquery/src/CreateOperator.cpp:32
> > [1,0]<stderr>:  and here: /home/ddn/w/katana/libquery/src/UpdateOperator.cpp:24
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will this change fix these?
> 
> Hmm no not this case, that's an example of the behavior I thought would be good to warn folks about; I was seeing warnings where it was complaining about two paths to the same header file. Perhaps I should just disable the warning all together? It doesn't seem to be buying us much and is just more of an annoyance.

This PR will get rid of those warnings. https://github.com/KatanaGraph/katana-enterprise/pull/3435

ddn0:
Wouldn't these issues be mostly addressed by making sure `KATANA_EXPERIMENTAL_FEATURE` is used only in places where it makes sense to define a variable (i.e., in cpp files and not in headers)?

You might have to change `KATANA_EXPERIMENTAL_ENABLED` to declare the variable (but not define it) so that users don't have to worry about ordering issues. But you'd still get a linker error if you tried to enable a feature that wasn't defined.

tylershunt:
> Wouldn't these issues be mostly addressed by making sure `KATANA_EXPERIMENTAL_FEATURE` is used only in places where it makes sense to define a variable (i.e., in cpp files and not in headers)?
> 
> You might have to change `KATANA_EXPERIMENTAL_ENABLED` to declare the variable (but not define it) so that users don't have to worry about ordering issues. But you'd still get a linker error if you tried to enable a feature that wasn't defined.

Maybe, my original thought process was that the tradeoff was worth it to only have to use one macro. In your proposal we need at least two; one to declare then a definition somewhere.

ddn0:
> Maybe, my original thought process was that the tradeoff was worth it to only have to use one macro. In your proposal we need at least two; one to declare then a definition somewhere.

Isn't the pair `KATANA_EXPERIMENTAL_FEATURE` and `KATANA_EXPERIMENTAL_ENABLED` two macros?

My proposal keeps the same number of macros and number of uses as well.

Somewhere (even in a cpp file without any other reference to it), write
```
KATANA_EXPERIMENTAL_FEATURE(...);
```

Elsewhere, you can write `KATANA_EXPERIMENTAL_ENABLED` wherever you want, and you don't need to even reference the original `KATANA_EXPERIMENTAL_FEATURE` definition directly except by name because `KATANA_EXPERIMENTAL_ENABLED` is basically:

```
extern katana::internal::...#foo
return katana::internal::...#foo.IsEnabled()
```

If you call `KATANA_EXPERIMENTAL_ENABLED` without having a `KATANA_EXPERIMENTAL_FEATURE` you will get a linker error. If you define `KATANA_EXPERIMENTAL_FEATURE` multiple times, the current registration code will emit a warning at runtime.

tylershunt:
> 



> extern katana::internal::...#foo
> return katana::internal::...#foo.IsEnabled()

Makes sense! that hadn't occurred to me.

tylershunt:
> > 
> 
> > extern katana::internal::...#foo
> > return katana::internal::...#foo.IsEnabled()
> 
> Makes sense! that hadn't occurred to me.

I was worried about the linker not catching instances that span shared libraries, but of course I could just export the symbol as well.

tylershunt:
I think this now implements the compromise we discussed @ddn0

ddn0 (approved):

PR #934, opened 2022-02-28
libgalois: fix DynamicBitset count/not bug

scober:
DynamicBitset::bitwise_not() applies bitwise not to every element in
DynamicBitset's internal vector. DynamicBitset::count() calls popcount
on every element in DynamicBitset's internal vector and takes a sum.

The last element in the vector has some unused bits in it if the size of
the bitset is not a multiple of 64. When the vector is constructed,
these bits are set to 0. So calling popcount on that element gives only
the set and used bits. But once bitwise_not has been called, those
unused bits are set to 1. And calling popcount on the last element gives
the number of set and used bits plus the number of unused bits.

One solution is to change count to not count the unused bits. One
solution is to change bitwise_not to not invert the unused bits. Change
the behavior of not to maintain the invariant that the unused bits are
always 0 because that gives nice semantics when resizing the bitset.

### DIFF
--- a/libgalois/src/DynamicBitset.cpp
+++ b/libgalois/src/DynamicBitset.cpp
@@ -41,6 +41,8 @@ katana::DynamicBitset::bitwise_not() {
   katana::do_all(
       katana::iterate(size_t{0}, bitvec_.size()),
       [&](size_t i) { bitvec_[i] = ~bitvec_[i]; }, katana::no_stats());
+
+  RestoreTrailingBitsInvariant();
 }
 
 void
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -52,6 +52,24 @@ private:  // variables
   katana::PODVector<TItem> bitvec_;
   size_t num_bits_{0};
 
+  /// DynamicBitset must maintain the invariant that the unused bits in the last
+  /// element of bitvec_ are 0. This invariant is required for count() to return
+  /// the correct value and also simplifies resizing bitsets to larger sizes.
+  /// Most mutating methods maintain this invariant (for example, bitwise_and()
+  /// will bitwise and the last entries of its operands together and if the
+  /// inputs both have 0s in the unused bits the output will as well) but
+  /// bitwise_not() must explicitly restore this invariant.
+  void RestoreTrailingBitsInvariant() {
+    if (size() > 0) {
+      KATANA_LOG_DEBUG_ASSERT(!bitvec_.empty());
+
+      size_t last_entry_offset =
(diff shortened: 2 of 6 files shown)


libgalois/test/dynamic-bitset.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+  for (size_t i = 0; i < test.size(); i += 3) {
+    test.set(i);
+  }
+
+  return test;
+}
+
+katana::Result<void>
+TestNotAndCount() {
```
ddn0: whooo!!!

roshandathathri (approved):
LGTM. Thanks!

Can you also fix DynamicBitsetSlow? @e-mcginnis FYI

libgalois/src/DynamicBitset.cpp
```diff
@@ -41,6 +41,12 @@ katana::DynamicBitset::bitwise_not() {
   katana::do_all(
       katana::iterate(size_t{0}, bitvec_.size()),
       [&](size_t i) { bitvec_[i] = ~bitvec_[i]; }, katana::no_stats());
+
+  // invariant: the unused bits in the last element of bitvec_ are always 0
+  size_t last_entry_offset =
```
ddn0: Would it make sense to pull this out in a function like "mask non zeros" (or whatever) and then it is a little more obvious why only "not" needs to do this?

scober: Why would pulling this out into a function make it more clear why only "not" needs to do it?

ddn0: why don't we need to mask the result of xor?

scober: We only support `xor` on bitsets of the same size. And `xor` will maintain the invariant. If you pass in a bitset where all unused bits are 0, you will get out a bitset where all unused bits are 0.

ddn0: > Why would pulling this out into a function make it more clear why only "not" needs to do it?

Because you can write this:

```cpp
/// MaskTrailingNonZeros sets any non-zeros outside of the nominal size of the bitset to zero. Count
/// depends on such extra bits to be zero to return the correct value.
//
// As a micro-optimization, we can skip calling this when the trailing bits are known to be zero by construction.
void MaskTrailingNonZeros() {
 ...
}

void bitwise_not(...) {
  ...
  MaskTrailingNonZeros();
}
```

The more you can describe the abstraction rather than the implementation, the more likely future readers and writers will able to maintain the datatype invariants rather than seeing code as a bunch of unrelated statements.

Though, I'd just call `MaskTrailingNonZeros` all the time because it is basically free to call and ensures that we maintain the class invariant now and in the future.

PR #937, opened 2022-03-03
libgalois: fix DynamicBitset bug

scober:
RestoreTrailingBitsInvariant was constructing a mask of all zeroes in
the case that the size of the bitset was a mutliple of 64. This resulted
in masking off the entire last element in the vector.

Fix this bug and add some more unit tests that give broader coverage of
DynamicBitset and act as a regression test for this bug in particular.

### DIFF
--- a/libgalois/include/katana/DynamicBitset.h
+++ b/libgalois/include/katana/DynamicBitset.h
@@ -65,8 +65,10 @@ private:  // variables
 
       size_t last_entry_offset =
           size() % (CHAR_BIT * sizeof(decltype(bitvec_)::value_type));
-      uint64_t last_entry_mask = (1UL << last_entry_offset) - 1;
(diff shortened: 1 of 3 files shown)


yan-zaretskiy:
Thanks! Did you check the previously failing enterprise tests?

scober:
> Thanks! Did you check the previously failing enterprise tests?

I checked the one that failed deterministically.

yan-zaretskiy (approved):
My inner voice tells me there's probably a way to do this branchlessly, but I don't see it and I am not sure whether this code path is that performance sensitive.

scober:
> My inner voice tells me there's probably a way to do this branchlessly, but I don't see it and I am not sure whether this code path is that performance sensitive.

The big issue I see is that `n % n == 0`. If `n % n == n` then the original code would have worked in this case. And most of the branchless ideas I had ran up against this problem.

PR #938, opened 2022-03-03
ensure path is set before use on error path when locating datasets

e-mcginnis:


pyrito (approved):
Beautiful. I'm surprised the linter didn't catch this. It's yelled at me for similar stuff before but I think that was for making use of an index in the for-loop that wasn't previously defined.

e-mcginnis:
> Beautiful. I'm surprised the linter didn't catch this. It's yelled at me for similar stuff before but I think that was for making use of an index in the for-loop that wasn't previously defined.

I am just as surprised. I guess I trusted the linter too much :/

PR #941, opened 2022-03-04
DynamicBitset: optimizes set-bit getting function, adds test

l-hoang:
Getting a list of set bits in a dynamic bitset can be useful; the
implementation of it before this commit did a loop over all bits which
was not ideal.

This commit revises the function to optimize the loop over the bitset by
dealing with a few special cases: no set bits, a single set bit, and all
set bits. The idea is that the bitset is very likely going to be sparse
or very dense, and these special cases can be optimized.

Other changes:
- Adds serial version of the offset getting function for times when you
  need to call it in a parallel section
- Refactors counting code
- Adds a test for count/offset
- Renames the slow DynamicBitset test to free up the name for the actual
  dynamic bitset test


libgalois/test/dynamic-bitset.cpp
```diff
@@ -135,6 +135,106 @@ const Invariant NotValues =
 
 const std::vector<Invariant> invariants = {NotAndCount, NotValues};
 
+katana::Result<void>
+CountAndOffsetsTest() {
```
scober, line 139: FWIW, I think this test could be reframed as an invariant that is agnostic to its input (the way the the other tests in the file are structured).

libgalois/src/DynamicBitset.cpp
```diff
@@ -96,90 +141,140 @@ katana::DynamicBitset::count() const {
...
+    if (int_to_examine == 0) {
+      // nothing set, skip
+      continue;
+    }
+
+    uint64_t offset = bit_index * katana::DynamicBitset::kNumBitsInUint64;
+
+    // TODO(l-hoang) optimize 63 set bits case? (find only 0, add the rest)
+    // optimize special corner cases case
+    if (int_to_examine == 1) {
```
scober, line 256: I'm confused by this - if this is integer equality don't you know that `int_to_examine == 0x0000000000000001` inside this block? Same comment for parallel version.

libgalois/src/DynamicBitset.cpp
```diff
@@ -96,90 +141,140 @@ katana::DynamicBitset::count() const {
...
+    } else if (int_to_examine == std::numeric_limits<uint64_t>::max()) {
+      // all set: add all, no testing required
+      for (size_t i = 0; i < katana::DynamicBitset::kNumBitsInUint64; i++) {
+        set_elements->emplace_back(offset + i);
+      }
+    } else {
+      // loop through all; hope is that this is rare
+      // TODO(l-hoang) is there a better way to do this, i.e. is test required?
+      for (size_t i = 0; i < katana::DynamicBitset::kNumBitsInUint64; i++) {
+        if (bitset.test(offset + i)) {
```
scober, line 269: While we are splitting optimization hairs, I don't think it makes sense to call `bitset.test()`. You already have the relevant element of `bitvec_`, why not just compare your index against that instead of having the bitset redo the math you already did? You can look at `DynamicBitsetIterator` for an example of this. I have the same comment for the parallel version.

libgalois/src/DynamicBitset.cpp
```diff
@@ -96,90 +141,140 @@ katana::DynamicBitset::count() const {
...
+      set_elements->emplace_back(offset + CountTrailingZeroes(int_to_examine));
+    } else if (int_to_examine == std::numeric_limits<uint64_t>::max()) {
+      // all set: add all, no testing required
+      for (size_t i = 0; i < katana::DynamicBitset::kNumBitsInUint64; i++) {
+        set_elements->emplace_back(offset + i);
+      }
+    } else {
+      // loop through all; hope is that this is rare
+      // TODO(l-hoang) is there a better way to do this, i.e. is test required?
+      for (size_t i = 0; i < katana::DynamicBitset::kNumBitsInUint64; i++) {
```
scober, line 268: You could try calling popcount on this element and breaking this loop early once you find all of the set bits. Same comment for parallel version.

libgalois/src/DynamicBitset.cpp
```diff
@@ -96,90 +141,140 @@ katana::DynamicBitset::count() const {
...
   // calculate prefix sum of bits per thread
-  for (uint32_t i = 1; i < activeThreads; ++i) {
-    tPrefixBitCounts[i] += tPrefixBitCounts[i - 1];
+  for (uint32_t i = 1; i < active_threads; ++i) {
+    thread_prefix_bit_counts[i] += thread_prefix_bit_counts[i - 1];
   }
-
   // total num of set bits
-  Integer bitsetCount = tPrefixBitCounts[activeThreads - 1];
+  Integer bitset_count = thread_prefix_bit_counts[active_threads - 1];
```
scober, line 185: nit: use `thread_prefix_bit_counts.back()`.

scober (commented):
I left a few comments regarding the optimizations, but I have a few bigger-picture questions:
1. Did you measure this to see what the performance speedup was on some realistic bitsets?
2. `DynamicBitset` has an iterator over set bits now and I have caught myself using `GetOffsets` when using the iterator would be faster/more efficient. Would making the iterator random-access (and therefore more efficient in `do_all`s) be a better use of our optimization efforts?

libgalois/src/DynamicBitset.cpp
l-hoang: yeah, this is a mistake on my part: int to examine used to be a count of the number of set bits but I changed it
incorrectly here

l-hoang:
1) I haven't measured perf here, but I do know that this resulted in some rather good speedup back when
it was changed for Galois for my older GNN work which needed this (don't have the numbers right now,
but IIRC it was significant, esp. when the bitset is nontrivially sized and sparse since otherwise you're just
testing through 0s).

2) Makes sense; then you wouldn't have to materialize it.

If that's the path we want to take then I can close this and come back to it another time (or someone else
can do it; the thing that I needed this for has been dropped for the moment as it didn't give
much perf improvement, so I'm moving on to something else).

l-hoang:
Closing this as it probably makes more sense to optimize iteration over the bitset instead.

PR #942, opened 2022-03-04
Bytes occupied by EntityTypeManager

witchel:
Before #3475.

I suspect my code is wrong since the size I'm getting is tiny.  Oh, I guess it doesn't capture the `EntityTypeID `arrays.

### DIFF
--- a/libtsuba/include/katana/EntityTypeManager.h
+++ b/libtsuba/include/katana/EntityTypeManager.h
@@ -597,6 +597,19 @@ public:
   /// std::string PrintTypes() IS A TESTING ONLY FUNCTION, DO NOT EXPOSE THIS TO THE USER
   std::string PrintEntityTypes() const;
 
+  size_t ApproxByteCount() const {
+    size_t count = 0;
+    count += atomic_entity_type_id_to_type_name_.size() *
+             sizeof(EntityTypeIDToAtomicTypeNameMap::value_type);
+    count += atomic_type_name_to_entity_type_id_.size() *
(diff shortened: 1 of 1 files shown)


libtsuba/include/katana/EntityTypeManager.h
```diff
@@ -597,6 +597,19 @@ class KATANA_EXPORT EntityTypeManager {
   /// std::string PrintTypes() IS A TESTING ONLY FUNCTION, DO NOT EXPOSE THIS TO THE USER
   std::string PrintEntityTypes() const;
 
+  size_t ApproxByteCount() const {
```
amberhassaan, line 600: Is measuring this (memory usage of EntityTypeManager) important? If yes, then we should include the size of dynamic allocated things like strings and dynamic bitsets in the maps.

amberhassaan (approved):

libtsuba/include/katana/EntityTypeManager.h
witchel: Because I was measuring the space consumed by properties, I figured I would measure types also.  But the encapsulation is poor here, so I'm thinking it is not worthwhile.

amberhassaan: I agree. This would only matter if we really had a lot of Entity types with really long names. On the other hand, it does make sense to account for Entity Type ID Arrays in `PropertyGraph`. I assume you're doing that with `graph-info`?

witchel: I do not account for EntityTypeID arrays in graph-info.  You can add that feature!

PR #946, opened 2022-03-07
Add property view sorted by edge property and refactor edge sorted topology

nicelhc13:
[Goal]
This PR adds new policy of edge sorting by edge property then edge destination id.
This PR also contains refactoring to support various edge property sorting policies, which was limited.

[Usecase]
Dist-minimum-spanning-tree needs this edge sorting for optimization to reduce edge iterations.

[Problem]
The current `EdgeShuffleTopology` has multiple policies for edge sorting. However, users only could create views of subsets of them since creating a new `PGViewBuilder` for other edge sorting policies was hard.
(https://github.com/KatanaGraph/katana/blob/71e3fd6b14749e0a328f1d1118daa2e035e7f23d/libgraph/include/katana/GraphTopology.h#L1392)
`PGViewEdgesSortedByDestID` is a representative view for `EdgeShuffleTopology`. It means that we cannot create new views like `PGViewEdgesSortedByProperty` or `PGViewEdgesSortedByDegree` as other designs adopted, since they would be the equivalent types and a compiler would complain about redefinition errors. It only assumes the policy of edges sorting by destination id even though `EdgeShuffleTopology` has multiple policies (https://github.com/KatanaGraph/katana/blob/71e3fd6b14749e0a328f1d1118daa2e035e7f23d/libgraph/include/katana/GraphTopology.h#L334)

[This PR's approach]
To distinguish and support multiple edge sorting policy, this PR adds new parameters for policy (and property name if it needs; sorting by edge property needs the target property name) at the existing BuildView().
(https://github.com/KatanaGraph/katana/blob/71e3fd6b14749e0a328f1d1118daa2e035e7f23d/libgraph/include/katana/GraphTopology.h#L1398)

This PR adds execution paths to support this.

[Design detail information]
This PR has some TODOs, which I would like to discuss about.
I was trying to reuse the existing APIs as much as possible.
I tried to avoid duplicated/complex parameters from user side, like property name and property type.
Edge property sorting could be called from both `TypedPropertyGraph` and `PropertyGraph`.
In case of the `PropertyGraph`, it uses the original type that was specified during `ConstructEdgeProperties()`.

### DIFF
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -285,6 +285,13 @@ public:
            (kind == edge_sort_state());
   }
 
+  bool has_edges_sorted_by_property(
+      const RDGTopology::EdgeSortKind& kind,
+      const std::string& property_name) const noexcept {
+    return (kind == edge_sort_state()) &&
+           (property_name == property_name_for_sorting_);
+  }
+
   using Base::GetLocalEdgeIDFromOutEdge;
 
   static std::shared_ptr<EdgeShuffleTopology> MakeTransposeCopy(
@@ -311,6 +318,26 @@ public:
     return ret;
   }
 
+  static std::shared_ptr<EdgeShuffleTopology> Make(
+      PropertyGraph* pg, const RDGTopology::TransposeKind& tpose_todo,
+      const RDGTopology::EdgeSortKind& edge_sort_todo,
+      const std::string property_name) noexcept {
+    std::shared_ptr<EdgeShuffleTopology> ret;
+
+    if (tpose_todo == RDGTopology::TransposeKind::kYes) {
+      ret = MakeTransposeCopy(pg);
+      KATANA_LOG_DEBUG_ASSERT(
+          ret->has_transpose_state(RDGTopology::TransposeKind::kYes));
+    } else {
+      ret = MakeOriginalCopy(pg);
+      KATANA_LOG_DEBUG_ASSERT(
+          ret->has_transpose_state(RDGTopology::TransposeKind::kNo));
+    }
+
+    ret->sortEdges(pg, edge_sort_todo, property_name);
+    return ret;
+  }
+
   static std::shared_ptr<EdgeShuffleTopology> Make(RDGTopology* rdg_topo);
 
   katana::Result<RDGTopology> ToRDGTopology() const;
@@ -330,6 +357,14 @@ protected:
 
   void SortEdgesByDestType(
       const PropertyGraph* pg, const PropIndexVec& node_prop_indices) noexcept;
+
+  void SortEdgesByPropertyThenDest(
+      const PropertyGraph* pg, const std::string& prop_name) noexcept;
+
+  template <typename ArrowArrayTy>
+  void RunSortingEdgesByPropertyThenDest(
+      const PropertyGraph* pg,
+      std::shared_ptr<ArrowArrayTy> edge_prop_view) noexcept;
 
   void sortEdges(
       const PropertyGraph* pg,
@@ -352,6 +387,19 @@ protected:
     }
   }
 
+  void sortEdges(
+      const PropertyGraph* pg, const RDGTopology::EdgeSortKind& edge_sort_todo,
+      const std::string& property_name) noexcept {
+    switch (edge_sort_todo) {
+    case RDGTopology::EdgeSortKind::kSortedByProperty:
+      SortEdgesByPropertyThenDest(pg, property_name);
+      return;
+    default:
+      KATANA_LOG_FATAL("switch-case fell through");
+      return;
+    }
+  }
+
   EdgeShuffleTopology(
       const RDGTopology::TransposeKind& tpose_todo,
       const RDGTopology::EdgeSortKind& edge_sort_todo,
@@ -365,9 +413,25 @@ protected:
         edge_sort_state_(edge_sort_todo),
         is_valid_(true) {}
 
+  EdgeShuffleTopology(
+      const RDGTopology::TransposeKind& tpose_todo,
+      const RDGTopology::EdgeSortKind& edge_sort_todo,
+      AdjIndexVec&& adj_indices, EdgeDestVec&& dests,
+      PropIndexVec&& edge_prop_indices, PropIndexVec&& node_prop_indices,
+      const std::string& property_name) noexcept
+      : Base(
+            std::move(adj_indices), std::move(dests),
+            std::move(edge_prop_indices), std::move(node_prop_indices)),
+        tpose_state_(tpose_todo),
+        edge_sort_state_(edge_sort_todo),
+        property_name_for_sorting_(property_name),
+        is_valid_(true) {}
+
 private:
   RDGTopology::TransposeKind tpose_state_{RDGTopology::TransposeKind::kNo};
   RDGTopology::EdgeSortKind edge_sort_state_{RDGTopology::EdgeSortKind::kAny};
+
+  std::string property_name_for_sorting_;
 
   bool is_valid_ = true;
 };
@@ -1168,6 +1232,34 @@ public:
   }
 };
 
+template <typename Topo>
+class SortedByEdgePropertyTopologyWrapper : public BasicTopologyWrapper<Topo> {
+  using Base = BasicTopologyWrapper<Topo>;
+
+public:
+  using typename Base::Node;
+
+  explicit SortedByEdgePropertyTopologyWrapper(
+      std::shared_ptr<const Topo> t,
+      [[maybe_unused]] const std::string& property_name) noexcept
+      : Base(std::move(t)) {
+    KATANA_LOG_DEBUG_ASSERT(Base::topo().has_edges_sorted_by_property(
+        RDGTopology::EdgeSortKind::kSortedByProperty, property_name));
+  }
+
+  auto FindEdge(const Node& src, const Node& dst) const noexcept {
+    return Base::topo().FindEdge(src, dst);
+  }
+
+  auto HasEdge(const Node& src, const Node& dst) const noexcept {
+    return Base::topo().HasEdge(src, dst);
+  }
+
+  auto FindAllEdges(const Node& src, const Node& dst) const noexcept {
+    return Base::topo().FindAllEdges(src, dst);
+  }
+};
+
 class KATANA_EXPORT EdgeTypeAwareBiDirTopology
     : public BasicBiDirTopoWrapper<
           EdgeTypeAwareTopology, EdgeTypeAwareTopology> {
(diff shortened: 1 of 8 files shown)


amberhassaan:
@nicelhc13 : Before we implement this change, I want to understand your use case clearly. Can you describe why you need this (in terms of a use case like Minimal Spanning Tree)?

libgraph/include/katana/TypedPropertyGraph.h
```diff
@@ -275,6 +275,11 @@ class TypedPropertyGraphView : public PGView {
       const std::vector<std::string>& edge_properties);
   static Result<TypedPropertyGraphView<PGView, NodeProps, EdgeProps>> Make(
       const PGView& pg_view);
+  static Result<TypedPropertyGraphView<PGView, NodeProps, EdgeProps>> Make(
+      PropertyGraph* pg, const std::vector<std::string>& node_properties,
+      const std::vector<std::string>& edge_properties,
+      const RDGTopology::EdgeSortKind edge_sort_kind,
```
amberhassaan: I'm against this change. It muddies the interface for all analytics apps using `TypedPropertyGraphView` and adds burden on the user to provide the `sort kind` which was so far hidden. Let's find a way that does not change existing interface for analytics apps.

nicelhc13:
> @nicelhc13 : Before we implement this change, I want to understand your use case clearly. Can you describe why you need this (in terms of a use case like Minimal Spanning Tree)?

I am implementing Boruvka algorithm. This finds the lightest edge for each MST. And to do this, as the first step, each member node of the MST finds the lightest edge weight from its immediate neighbors. In this case, if the topology is sorted by edge weight (one of the edge properties) and we memoize the last edge pointing to different MST, we can jump to the last visited edge since the previous edges are already on the same MST and the current edge is the ligthest one. 
You can also check shared memory spanning tree implementation:https://github.com/KatanaGraph/katana/blob/302731b1a7ebb6cceb934d537a338d72733fea3e/lonestar/analytics/cpu/spanningtree/Boruvka.cpp#L116

But besides this, I believe this should be updated. We have multiple policies, but most of them are hidden.

libgraph/include/katana/TypedPropertyGraph.h
nicelhc13: I can think of two options.
1) Use default value; IMO, it is equivalent to the old version
2) Use `enable_if` or `enable_if_t` to make different execution paths and hide from users?

Do you have any better suggestion?
This was also my question and my conclusion was we need a user parameter since it is algorithmic dependent information.

libgraph/include/katana/GraphTopology.h
```diff
@@ -276,10 +281,27 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
...
+  /// and the target property name was `property_name`, then return true.
+  /// Otherwise return false.
   bool has_edges_sorted_by(
-      const RDGTopology::EdgeSortKind& kind) const noexcept {
-    return (kind == RDGTopology::EdgeSortKind::kAny) ||
-           (kind == edge_sort_state());
+      const RDGTopology::EdgeSortKind& kind,
+      const std::optional<std::string> property_name =
+          std::nullopt) const noexcept {
+    bool is_sorted{false};
```
amberhassaan: You've made something that was simple, more verbose and complex. See https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return . 
A better alternative would have been to define a new overload of `has_edge_sorted_by` that takes two parameters.

libgraph/include/katana/TypedPropertyGraph.h
amberhassaan: You can use `enable_if` and define a new overload for your case that takes a property name.

libgraph/include/katana/TypedPropertyGraph.h
```diff
@@ -349,6 +354,32 @@ TypedPropertyGraphView<PGView, NodeProps, EdgeProps>::Make(
...
+template <typename PGView, typename NodeProps, typename EdgeProps>
+Result<TypedPropertyGraphView<PGView, NodeProps, EdgeProps>>
+TypedPropertyGraphView<PGView, NodeProps, EdgeProps>::Make(
+    PropertyGraph* pg, const std::vector<std::string>& node_properties,
+    const std::vector<std::string>& edge_properties,
+    const RDGTopology::EdgeSortKind edge_sort_kind,
+    const std::optional<std::string> prop_name) {
+  auto pg_view = pg->BuildView<PGView>(edge_sort_kind, prop_name);
+  KATANA_LOG_DEBUG_ASSERT(pg);
+  auto node_view_result =
```
amberhassaan: Use `KATANA_CHECKED` here and on line 372.

libgraph/src/GraphTopology.cpp
```diff
@@ -422,6 +422,118 @@ katana::EdgeShuffleTopology::SortEdgesByTypeThenDest(
...
+        "PropertyGraph does not have property: {}", property_name.value());
+  }
+  auto edge_property = edge_prop_ret.value();
+  // TODO(lhc): For now, this finds a property twice through two APIs,
+  //            by GetEdgeProperty() and GetEdgePropertyTyped().
+  //            Later, implement and merge those two functions, and hide
+  //            getting types from developers.
+  //            This is because first, its focus is not performances,
+  //            and second, want to use the existing APIs as much as possible.
+  switch (edge_property->type()->id()) {
```
amberhassaan, line 472: See `libsupport/include/ArrowVisitor.h`. This seems like a case where an ArrowArrayVisitor can help. You'll have to look up some example to learn how to use it.

amberhassaan (changes requested):
High-level change that I'm going to insist on: 

Don't use (or abuse) `std::optional` for property name. Rather define a new overload of the method with `property_name` so that existing interface don't need to change. Similarly, don't change (at least yet) existing sorted views that are named specifically for sort types, such as EdgesSortedByDestID etc.

nicelhc13:
@amberhassaan 
I don't get it about this point:

> Similarly, don't change (at least yet) existing sorted views that are named specifically for sort types, such as EdgesSortedByDestID etc.

It is related to this point of the problems that I listed:

> PGViewEdgesSortedByDestID is a representative view for EdgeShuffleTopology. It means that we cannot create new views like PGViewEdgesSortedByProperty or PGViewEdgesSortedByDegree as other designs adopted, since they would be the equivalent types and a compiler would complain about redefinition error

To define `EdgesSortedByDestID` separately, this specific policy basically should be a different class or should be templatized.
Could you please let me know a way that in your mind?

The main motivation of this PR is due to the above limitation.
I don't want to change interfaces as much as possible, but still don't understand why we are choosing limited interface.
(Again, EdgeShuffleTopo is only available for DestID, and not for others, according to the current implementation)

amberhassaan:
Just fyi: Hochan and I had a quick call. Hochan will define a new class similar to `SortedTopologyWrapper` for his case and that should allow us to separate code paths enough that we don't touch existing interfaces.

nicelhc13:
@amberhassaan 
Could you reevaluate this PR?
I tried to follow your suggestions as many as possible.
Mainly, I tried not to touch other execution paths, but just add new path.

There are two things that I failed to reflect (I need your suggestion, and if you think it is not critical, I will leave TODOs and revisit it later; since deadline is coming)
* I failed to hide Make() getting property name through enable_if on TypedPropertyGraph. (But I could hide Make() on PropertyGraph). This function is static and cannot think about 'clear' way..
* I failed to reflect ArrowVistior style. I think I understood what you wanted, but the problem was CType was not declared for some types. For example, struct TypeTraits of Arrow for arrow::UINT32Type does not provide C++ uint32_t. So I could not avoid hard coding.

Other than them, I think I reflected your comments. I hope this PR is not too much far from your expectation.
Thank you.

libgraph/include/katana/GraphTopology.h
```diff
@@ -285,6 +285,13 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
            (kind == edge_sort_state());
   }
 
+  bool has_edges_sorted_by_property(
```
amberhassaan, line 288: I don't think you need the first parameter as it should always be "EdgesSortedByProperty" (or whatever the name of enum value you added in RDGTopology.h)

libgraph/include/katana/GraphTopology.h
```diff
@@ -311,6 +318,26 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
     return ret;
   }
 
+  static std::shared_ptr<EdgeShuffleTopology> Make(
+      PropertyGraph* pg, const RDGTopology::TransposeKind& tpose_todo,
+      const RDGTopology::EdgeSortKind& edge_sort_todo,
```
amberhassaan, line 323: Same here. enum isn't needed for now. You can rename the method to `MakeEdgesSortedByPropety`

libgraph/include/katana/GraphTopology.h
```diff
@@ -352,6 +387,19 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
     }
   }
 
+  void sortEdges(
```
amberhassaan, line 390: Same here. Drop the enum, rename to sorteEdgesByProperty

libgraph/include/katana/GraphTopology.h
```diff
@@ -365,10 +413,26 @@ class KATANA_EXPORT EdgeShuffleTopology : public GraphTopology {
...
+        tpose_state_(tpose_todo),
+        edge_sort_state_(edge_sort_todo),
+        property_name_for_sorting_(property_name),
+        is_valid_(true) {}
+
 private:
   RDGTopology::TransposeKind tpose_state_{RDGTopology::TransposeKind::kNo};
   RDGTopology::EdgeSortKind edge_sort_state_{RDGTopology::EdgeSortKind::kAny};
 
+  std::string property_name_for_sorting_;
```
amberhassaan, line 434: I don't think you need this field. Drop please.

libgraph/include/katana/GraphTopology.h
```diff
@@ -1608,10 +1730,20 @@ class KATANA_EXPORT PGViewCache {
...
+  std::shared_ptr<EdgeShuffleTopology> BuildOrGetEdgeShuffTopoImpl(
+      PropertyGraph* pg, const RDGTopology::TransposeKind& tpose_kind,
+      const RDGTopology::EdgeSortKind& sort_kind, bool pop,
+      const std::string& property_name) noexcept;
+
   std::shared_ptr<EdgeShuffleTopology> BuildOrGetEdgeShuffTopo(
       PropertyGraph* pg, const RDGTopology::TransposeKind& tpose_kind,
       const RDGTopology::EdgeSortKind& sort_kind) noexcept;
 
+  std::shared_ptr<EdgeShuffleTopology> BuildOrGetEdgeShuffTopo(
```
amberhassaan, line 1742: Drop the enum parameter and rename to convey that it's for building edge sorted by property.

libtsuba/include/katana/RDGTopology.h
```diff
@@ -259,6 +260,11 @@ class KATANA_EXPORT RDGTopology {
       TopologyKind topology_state, TransposeKind transpose_state,
       EdgeSortKind edge_sort_state, NodeSortKind node_sort_state);
 
+  static katana::RDGTopology MakeShadow(
+      TopologyKind topology_state, TransposeKind transpose_state,
+      EdgeSortKind edge_sort_state, NodeSortKind node_sort_state,
```
amberhassaan, line 265: drop `edge_sort_state` here too.

amberhassaan (approved):
@nicelhc13 : Sorry for the delay in getting back to you. 
This looks good overall. Please address the final set of comments which basically say that when you're producing an "EdgeSortedByProperty" type topology, the user doesn't need to pass the enum `edge_sort_kind`. It's already implied.

amberhassaan:
@nicelhc13 : I actually forgot to mention, please see: 
https://github.com/KatanaGraph/katana-enterprise/pull/3666

libgraph/include/katana/GraphTopology.h
nicelhc13: Question; this field can be any property name. users can choose different properties during application. do you really want to assume that the application always uses one property name to sort a topology?

amberhassaan: OK, my bad. I now see why you need this field, which is to answer `has_edges_sorted_by_property(property_name)`.

PR #947, opened 2022-03-08
Bump version to 0.2.0qa on master

tylershunt:
Previous version 0.2.0.
(Manual edit by release manager)


gliga (approved):
I actually have no idea why we do this particular change, but I will assume this is for some consistency.

PR #951, opened 2022-03-08
Store parent graph for transformation views.

yan-zaretskiy:
This commit stores a pointer to the parent graph when constructing a transformation view. We also defer `PropertyGraph`'s `Commit()` and `Write()` implementation to the parent is it exists.

### DIFF
--- a/libgraph/include/katana/PropertyGraph.h
+++ b/libgraph/include/katana/PropertyGraph.h
@@ -279,20 +279,19 @@ public:
       const katana::RDGLoadOptions& opts, katana::TxnContext* txn_ctx);
 
   [[deprecated]] static std::unique_ptr<PropertyGraph> MakeProjectedGraph(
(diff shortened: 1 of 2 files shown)


libgraph/include/katana/PropertyGraph.h
```diff
@@ -1125,7 +1127,7 @@ class KATANA_EXPORT PropertyGraph {
   PGViewCache pg_view_cache_;
 
   // Transformation related data.
-  bool is_transformed{false};
+  PropertyGraph* parent_{nullptr};
```
roshandathathri, line 1130: This seems unsafe. Better to switch to `shared_ptr`.

roshandathathri (approved):
LGTM. Thanks!

e-mcginnis (approved):
Looks great!

PR #954, opened 2022-03-10
Storage prefixes always

scober:
I propose (via PR) complicating both `RDG` and `PropertyGraph`. They can both now optionally be constructed as "ephemeral". An ephemeral graph is backed by a storage location that will be deleted when the in-memory graph goes out of scope. It cannot be committed to that location but it can be written to another location. An existing graph can not be made ephemeral but an ephemeral graph can be made non-ephemeral by writing it to a new location. This functionality is used in three primary places:
1. By users directly via the `katana.local.Graph` python class
2. By analytics functions like the clusterers that create `PropertyGraph`s on the fly as intermediate state
3. By tests

### DIFF
--- a/libgraph/include/katana/PropertyGraph.h
+++ b/libgraph/include/katana/PropertyGraph.h
@@ -250,17 +250,31 @@ public:
       const katana::RDGLoadOptions& opts = katana::RDGLoadOptions());
 
   /// Make a property graph from topology
-  // [deprecated("please provide a storage prefix")]
-  static Result<std::unique_ptr<PropertyGraph>> Make(
-      GraphTopology&& topo_to_assign);
-
-  /// Make a property graph from topology
   static Result<std::unique_ptr<PropertyGraph>> Make(
       const URI& rdg_dir, GraphTopology&& topo_to_assign);
 
+  /// Make a property graph from topology and associate it with an ephemeral
+  /// storage prefix. This is approximates an in-memory graph.
+  static Result<std::unique_ptr<PropertyGraph>> MakeEphemeral(
+      GraphTopology&& topo_to_assign);
+
+  /// Make a property graph from topology
+  [[deprecated("please provide a storage prefix")]] static Result<
+      std::unique_ptr<PropertyGraph>>
+  Make(GraphTopology&& topo_to_assign);
+
+  /// Make a property graph from topology and type arrays and associate it with
+  /// an ephemeral storage prefix. This approximates an in-memory graph.
+  static Result<std::unique_ptr<PropertyGraph>> MakeEphemeral(
+      GraphTopology&& topo_to_assign, EntityTypeIDArray&& node_entity_type_ids,
+      EntityTypeIDArray&& edge_entity_type_ids,
+      EntityTypeManager&& node_type_manager,
+      EntityTypeManager&& edge_type_manager);
+
   /// Make a property graph from topology and type arrays
-  // [deprecated("please provide a storage prefix")]
-  static Result<std::unique_ptr<PropertyGraph>> Make(
+  [[deprecated("please provide a storage prefix")]] static Result<
+      std::unique_ptr<PropertyGraph>>
+  Make(
       GraphTopology&& topo_to_assign, EntityTypeIDArray&& node_entity_type_ids,
       EntityTypeIDArray&& edge_entity_type_ids,
       EntityTypeManager&& node_type_manager,
--- a/libkatana_python_native/src/ImportData.cpp
+++ b/libkatana_python_native/src/ImportData.cpp
@@ -39,7 +39,7 @@ katana::python::InitImportData(py::module& m) {
       [](py::array_t<PropertyGraph::Edge> edge_indices,
          py::array_t<PropertyGraph::Node> edge_destinations)
           -> Result<std::shared_ptr<PropertyGraph>> {
-        return KATANA_CHECKED(katana::PropertyGraph::Make(
+        return KATANA_CHECKED(katana::PropertyGraph::MakeEphemeral(
             TopologyFromCSR(edge_indices, edge_destinations)));
       },
       R"""(
@@ -75,7 +75,7 @@ katana::python::InitImportData(py::module& m) {
             edge_types_owned.begin());
         EntityTypeManager node_type_manager_owned = node_type_manager;
         EntityTypeManager edge_type_manager_owned = edge_type_manager;
-        return KATANA_CHECKED(katana::PropertyGraph::Make(
+        return KATANA_CHECKED(katana::PropertyGraph::MakeEphemeral(
             TopologyFromCSR(edge_indices, edge_destinations),
             std::move(node_types_owned), std::move(edge_types_owned),
             std::move(node_type_manager_owned),
--- a/libtsuba/include/katana/EphemeralStoragePrefix.h
+++ b/libtsuba/include/katana/EphemeralStoragePrefix.h
@@ -1,0 +1,76 @@
+#ifndef KATANA_LIBTSUBA_KATANA_EPHEMERALSTORAGEPREFIX_H_
+#define KATANA_LIBTSUBA_KATANA_EPHEMERALSTORAGEPREFIX_H_
+
+#include <memory>
+
+#include "katana/URI.h"
+#include "katana/file.h"
+
+namespace katana {
+
+/// EphemeralStoragePrefix uses RAII to manage ephemeral storage locations. It
+/// creates a URI under the katana temporary storage prefix and when the object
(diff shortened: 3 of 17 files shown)


libkatana_python_native/src/ImportData.cpp
```diff
@@ -39,7 +39,7 @@ katana::python::InitImportData(py::module& m) {
       [](py::array_t<PropertyGraph::Edge> edge_indices,
          py::array_t<PropertyGraph::Node> edge_destinations)
           -> Result<std::shared_ptr<PropertyGraph>> {
-        return KATANA_CHECKED(katana::PropertyGraph::Make(
+        return KATANA_CHECKED(katana::PropertyGraph::MakeEphemeral(
             TopologyFromCSR(edge_indices, edge_destinations)));
```
arthurp, line 43: A reasonable usage of this python function is to call it and then call `write` on the result to store it to a specific location. Will that work correctly? It relies on writing the RDG to a different location compared to where it started (in the tmp dir).

libtsuba/include/katana/EphemeralStoragePrefix.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    std::unordered_set deletable_files(files.begin(), files.end());
+    auto delete_res = FileDelete(prefix_.path(), deletable_files);
+    if (!delete_res) {
+      KATANA_LOG_WARN(
+          "unable to delete files, not cleaning up ephemeral storage: {}",
+          delete_res.error());
+    }
+  }
```
arthurp, line 49: I just realized something you may not be aware of. On Linux/Unix, you can create, open, and immediately delete a file. The file disappears to anyone listing the directly, but it actually still exists as long as it is still open. You can use that for temp files, that you don't need to close and reopen. It avoids the need to explicitly clean up the files.

scober: The contract here is that `PropertyGraph` has a storage prefix that it is free to do whatever it wants with. So I don't think we will be able to use those sorts of temporary files without invasive changes to the storage layer.

libkatana_python_native/src/ImportData.cpp
scober: I thought a lot about that use case when I was writing this so it should work correctly. But I need to check to see if we are testing it currently.

arthurp: I'm pretty sure I test it from Python.

scober: Could you point me in the right direction to short-circuit my search a little?

libtsuba/include/katana/EphemeralStoragePrefix.h
arthurp: OK. We might need to setup a signal handler to do this work then, since destructors don't run if the program crashes. But for now we should just document that our handling of temporary files is not complete and we will leak temp files in a number of cases. One case, BTW, is python interpreter exit with a live graph reference in the global scope. Python finalizers are not guaranteed to be called in that case, so C++ destructors may not be either. This could very easily happen with someone restarting their Jupyter "kernel" (python interpreter). The result could be creating a new set of temp files every time they restart the kernel.

libkatana_python_native/src/ImportData.cpp
arthurp: https://github.com/KatanaGraph/katana/blob/95ad4cfcacf9b0d5b89b9dbff801c83c1339e179/python/test/test_convert_graph.py#L227

Through the power of searching for `write` then looking at the tests for importing.

tylershunt:
I have two thoughts:
* this could be a lot simpler if we just designated an ephemeral storage prefix and then used some kind of recursive remove to just nuke that when we die (Becomes a check in the commit path to "is this in the ephereral place")
* I would rather the above be what happens when you call `Make` without a URI prefix.

scober:
>     * this could be a lot simpler if we just designated an ephemeral storage prefix and then used some kind of recursive remove to just nuke that when we die (Becomes a check in the commit path to "is this in the ephereral place")

That is true. One downside is that in that model we don't nuke the ephemeral place unless we die so we could potentially build up a lot of cruft there.

>     * I would rather the above be what happens when you call `Make` without a URI prefix.

`MakeEphemeral` has the same arguments as the current `Make`s that don't accept a `URI`. My goal is to remove those `Make`s and replace their usage with `MakeEphemeral` because that makes it clear to callers that they are doing something different.

tylershunt:
> One downside is that in that model we don't nuke the ephemeral place unless we die so we could potentially build up a lot of cruft there.

If that's a concern we could choose random sub directories of that prefix and keep the parts of one property graph, and auto remove that sub-directory when the property graph goes away (leaving the global cleanup on library unload so that we don't have to know about all the live ephemeral objects when we die).

But the check in libtsuba would be just as simple.

tylershunt:
> My goal is to remove those `Make`s and replace their usage with `MakeEphemeral` because that makes it clear to callers that they are doing something different.

This part confused me. Callers of `Make` without any mention of a storage location expected there graphs to be stored somehow?

scober:
> > One downside is that in that model we don't nuke the ephemeral place unless we die so we could potentially build up a lot of cruft there.
> 
> If that's a concern we could choose random sub directories of that prefix and keep the parts of one property graph, and auto remove that sub-directory when the property graph goes away (leaving the global cleanup on library unload so that we don't have to know about all the live ephemeral objects when we die).
> 
> But the check in libtsuba would be just as simple.

This is actually the current state of this PR (sort of). I don't know if this is an actual concern but since I had already written some code to manage per-graph locations I implemented exactly the hybrid you described. I didn't actually write the signal handler to do clean up but I can tack that onto this PR before I merge.

scober:
> > My goal is to remove those `Make`s and replace their usage with `MakeEphemeral` because that makes it clear to callers that they are doing something different.
> 
> This part confused me. Callers of `Make` without any mention of a storage location expected there graphs to be stored somehow?

I don't think it is impossible that we would get this complaint. But I don't feel all that strongly and it is definitely a cleaner interface to have all the `Make` functions have the same name.

amberhassaan:
I'm still confused as to why we need Ephemeral graphs? Why isn't an in-memory PropertyGraph not enough? @scober @arthurp

arthurp:
> I'm still confused as to why we need Ephemeral graphs? Why isn't an in-memory PropertyGraph not enough? @scober @arthurp

Because we need to be able to return ephemeral graphs to the remote API as handles and we cannot guarantee that the workers will still be running when the next request related to the graph comes in. So the ephemeral graphs need to be persistented for the client session or something like that.

scober:
> I'm still confused as to why we need Ephemeral graphs? Why isn't an in-memory PropertyGraph not enough? @scober @arthurp

There is no way to keep a `PropertyGraph` object from trying to write to storage. That is true in principle now but it is becoming more and more true in practice as well (see `MemorySupervisor` for an example of this).

So the proper (but maybe a little snotty) answer to your question is that there is no such thing as an in-memory `PropertyGraph`. We can hide the existence of ephemeral graphs from users more than I am in the current implementation by not using the word "Ephemeral" in the `Make` function name, but the current design is that every `PropertyGraph` has an associated writable storage location.

amberhassaan:
So is the idea with `MemorySupervisor` that we can unload a graph to storage to reduce our in-memory footprint?

tylershunt (approved):
My preference would be to drop `Ephemeral` from the name of the factory function and just promote the `Make` variant without a path to be what (I think) everyone expected it was anyway.

libgraph/include/katana/PropertyGraph.h
```diff
@@ -250,17 +250,31 @@ class KATANA_EXPORT PropertyGraph {
       const katana::RDGLoadOptions& opts = katana::RDGLoadOptions());
 
   /// Make a property graph from topology
-  // [deprecated("please provide a storage prefix")]
   static Result<std::unique_ptr<PropertyGraph>> Make(
+      const URI& rdg_dir, GraphTopology&& topo_to_assign);
+
+  /// Make a property graph from topology and associate it with an ephemeral
+  /// storage prefix. This is approximates an in-memory graph.
+  static Result<std::unique_ptr<PropertyGraph>> MakeEphemeral(
```
amberhassaan, line 258: I support Tyler's suggestion of keeping the name `Make`

amberhassaan (approved):
All good from my side.

scober:
> So is the idea with `MemorySupervisor` that we can unload a graph to storage to reduce our in-memory footprint?

Yes. In particular `MemorySupervisor` expects to be able to write certain large objects (I think just property arrays for now) to storage. So it may never persist the whole graph but it will want to write files.

PR #956, opened 2022-03-10
Support graph views from Numba [KAT-4236]

arthurp:
The support here is not great, but it should serve in the short term. The main problem is that [numba functions cannot be overloaded](https://katanagraph.atlassian.net/browse/KAT-4362) using the framework I have. The result is that methods have ugly names in numba. These same names can also be used in Python for consistency as well.


arthurp:
@JamesFCoffey I put you on here because I started writing some documentation for these methods and I introduced some reST replacements. If you could make sure that's OK with how documentation is going and that you don't have a different approach you want to take that would be great.

JamesFCoffey (commented):
The documentation end looks good to me.

arthurp:
@amazloumi This PR is ready for review. If you can do a review that would help speed things along.

pyproject.toml
```diff
@@ -22,6 +22,7 @@ line_length = 120
 [tool.pydocstyle]
 convention = "google"
 match = "(?!python/test).*"
+add-ignore = "D105"
```
arthurp, line 25: @JamesFCoffey This disables the requirement that "magic methods" like `__repr__` need to be documented. I think documenting them is silly since they have well defined semantics at the language level. How do you feel about this?

amazloumi (approved):
changes look fine to me, specifically separating `out_edge_ids_for_node` from `out_edge_ids` totally makes sense and I like it better.

arthurp:
> specifically separating `out_edge_ids_for_node` from `out_edge_ids` totally makes sense and I like it better.

This is actually something I was going to roll back once we can support overloading in numba. I generally like overloading in cases like this. And I'm a believer in consistency, so either use overloading as appropriate, or don't use it at all. Not some mix. I'll start a discussion on #property-graph-2 to make sure we are all on the same page.

docs/user-guides/algorithms_in_python.ipynb
```diff
@@ -122,7 +122,7 @@
     "    :return:\n",
     "    \"\"\"\n",
     "    # Iterate over the out edges of our node\n",
-    "    for edge_id in graph.out_edge_ids(node_id):\n",
+    "    for edge_id in graph.out_edge_ids_for_node(node_id):\n",
```
denish44, line 125: i think the original name was better `out_edge_ids`

denish44 (approved):

docs/user-guides/algorithms_in_python.ipynb
arthurp: I very much agree. I cannot overload methods in numba right now because it would take extra work and another week at least. This will change back when I can: KAT-4362.

PR #958, opened 2022-03-10
build: move minimum cmake version to 3.22.3

scober:
CMake 3.22.1 had an issue with the aws sdk. 3.22.3 is known to work with
that package so bump to 3.22.3.

### DIFF
--- a/katana_requirements.yaml
+++ b/katana_requirements.yaml
@@ -106,13 +106,13 @@ breathe:
   labels:
     - conda/dev
 cmake:
-  version: [ 3.17, 4 ]
+  version: [ 3.22.3, 4 ]
   labels:
     - apt
     - conda/dev
   version_overrides:
-    conda/dev: ">=3.17,<3.22a0"
-    conda: ">=3.17,<3.22a0"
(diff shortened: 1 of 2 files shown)


katana_requirements.yaml
```diff
@@ -111,8 +111,10 @@ cmake:
     - apt
     - conda/dev
   version_overrides:
-    conda/dev: ">=3.17,<3.22a0"
-    conda: ">=3.17,<3.22a0"
+    # This unfortunate mess is due to the fact that apt doesn't have a CMake 3.22.3 package available yet
+    # TODO (scober)(KAT-3901): bump the minimum CMake version to 3.22.3
+    conda: ">=3.17,<4a0,!=3.22.1,!=3.22.2"
+    conda/dev: ">=3.17,<4a0,!=3.22.1,!=3.22.2"
```
arthurp: You could override the apt version instead of messing up conda. We are dropping support for APT anyway.

scober: Good idea

arthurp (approved):

PR #961, opened 2022-03-11
Use shared pointers to PropertyGraph in PropertyGraph views [KAT-3405]

yan-zaretskiy:
This is the open portion of KAT-3405. I still need to update the `DistPropertyGraph` accordingly.

### DIFF
--- a/libgraph/test/storage-format-version-optional-topologies.h
+++ b/libgraph/test/storage-format-version-optional-topologies.h
@@ -35,19 +35,19 @@ TestOptionalTopologyStorageEdgeShuffleTopology(const katana::URI& inputFile) {
   KATANA_LOG_WARN("***** Testing EdgeShuffleTopology *****");
 
   katana::TxnContext txn_ctx;
-  katana::PropertyGraph pg = LoadGraph(inputFile);
+  std::shared_ptr<katana::PropertyGraph> pg = LoadGraph(inputFile);
 
   // Build a EdgeSortedByDestID view, which uses GraphTopology EdgeShuffleTopology in the background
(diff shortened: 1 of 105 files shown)


tools/generate-maximal-storage-format-rdg/generate-maximal-storage-format-rdg.cpp
```diff
@@ -88,23 +87,23 @@ CleanRelocateGraphStore(katana::PropertyGraph* g, std::string& output_path) {
 
 katana::Result<void>
 MaximizeGraph(std::string& input_rdg, std::string& output_path) {
-  katana::PropertyGraph g_tmp = CleanRelocateGraphLoad(input_rdg);
+  auto g_tmp = CleanRelocateGraphLoad(input_rdg);
```
arthurp: I think we prefer explicit types for cases like this where the type and lifetime is not obvious.

arthurp (approved):

libgraph/test/storage-format-version-optional-topologies.h
```diff
@@ -35,19 +35,19 @@ TestOptionalTopologyStorageEdgeShuffleTopology(std::string inputFile) {
   KATANA_LOG_WARN("***** Testing EdgeShuffleTopology *****");
 
   katana::TxnContext txn_ctx;
-  katana::PropertyGraph pg = LoadGraph(inputFile);
+  auto pg = std::make_shared<katana::PropertyGraph>(LoadGraph(inputFile));
```
rob-isaac: Odd that `LoadGraph` returns an instance rather than a smart pointer

rob-isaac (approved):
LGTM

PR #963, opened 2022-03-13
build numeric arrays in-place

danielmawhirter:
ArrowRandomAccessBuilder stores both data and vaildity bytes in std::vector for thread safety

For numeric types store the data in arrow-managed memory instead. This does not affect thread safety, and has lower peak memory usage.

### DIFF
--- a/libgraph/include/katana/ArrowRandomAccessBuilder.h
+++ b/libgraph/include/katana/ArrowRandomAccessBuilder.h
@@ -17,6 +17,8 @@ public:
 
   katana::Result<void> Resize(size_t length) {
     KATANA_CHECKED(data_.Resize(length));
+    data_.bytes_builder()->UnsafeAdvance(
+        data_.bytes_builder()->capacity() - data_.bytes_builder()->length());
     valid_.resize(length, false);
     return katana::ResultSuccess();
   }
@@ -155,7 +157,7 @@ template <typename ArrowType>
 class ArrowRandomAccessBuilder
     : public internal::BuilderTraits<ArrowType>::Type {
 public:
-  ArrowRandomAccessBuilder(size_t length) {
+  explicit ArrowRandomAccessBuilder(size_t length) {
     KATANA_LOG_ASSERT(this->Resize(length));
   }
 };
--- a/libgraph/include/katana/ArrowRandomAccessBuilder.h
+++ b/libgraph/include/katana/ArrowRandomAccessBuilder.h
@@ -1,17 +1,40 @@
-#include <arrow/api.h>
+#include <arrow/array.h>
+#include <arrow/builder.h>
 
-#include "katana/ErrorCode.h"
 #include "katana/Logging.h"
-#include "katana/NUMAArray.h"
-#include "katana/Properties.h"
 #include "katana/Result.h"
 
 namespace katana {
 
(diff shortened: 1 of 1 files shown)


danielmawhirter:
Reviewers note: NullableBuilder became VectorBackedBuilder, the rest is a rewrite

libgraph/include/katana/ArrowRandomAccessBuilder.h
```diff
@@ -8,69 +8,77 @@
...
-namespace {
+namespace internal {
 
-/// NoNullBuilder uses std::vector for storage
-/// Finalize() makes a copy of the data
-/// Does not support null values
-template <typename ValueType, typename StorageType, typename ArrowType>
-class NoNullBuilder {
+template <typename ArrowType>
+class NumberBuilder {
```
scober: nit: I would prefer `FixedSizeBuilder` or similar because this will be useful for fixed-size-non-numeric types.

libgraph/include/katana/ArrowRandomAccessBuilder.h
```diff
@@ -126,72 +116,50 @@ class NullableBuilder {
...
 
-  katana::Result<std::shared_ptr<arrow::Array>> Finalize() {
-    std::shared_ptr<arrow::Array> array;
-    auto res = Finalize(&array);
-    if (!res) {
-      return res.error();
-    }
-    return array;
+/// ArrowRandomAccessBuilder encapsulates the concept of building
+/// an arrow::Array from <index, value> pairs arriving in unknown order
```
scober: It might be worth making a stronger comment about how concurrent access to distinct entries is allowed but concurrent access to a single entry is not.

scober (approved):
Woohoo!

libgraph/include/katana/ArrowRandomAccessBuilder.h
danielmawhirter: I added a big block up top for this purpose

libgraph/include/katana/ArrowRandomAccessBuilder.h
```diff
@@ -1,76 +1,111 @@
...
-  reference operator[](size_t index) {
+  // NOTE this operator has side-effects. It can safely be used in two ways:
+  // 1) builder[index] = value; where it creates a non-null entry
+  // 2) value = builder[index]; ONLY IF option 1 has already used that index
+  value_type& operator[](size_t index) {
     KATANA_LOG_DEBUG_VASSERT(
         index < size(), "index: {}, size: {}", index, size());
-    return static_cast<ValueType*>(data_.data())[index];
+    valid_[index] = true;
+    return reinterpret_cast<value_type*>(data_.mutable_data())[index];
```
amansavani, line 56: ~Is the operator used for setting or getting data, or both?~
Is a check for validity needed before getting the data?

danielmawhirter: This is mostly used for setting data, but it can be used to read data from already-set positions. Once we return a reference to the caller we can't really control which way it's used without building a bunch of wrapper logic that would add overhead

amansavani (approved):

tylershunt (approved):

PR #964, opened 2022-03-14
Removing warning about deleting projected graph [KAT-4435]

bobdc:
Remove warning about deleting graph projections as discussed at https://thek-teamworkspace.slack.com/archives/C02BGT1MRHQ/p1647030897007439?thread_ts=1647027574.034349&cid=C02BGT1MRHQ .

(Part 2 of 2, where part 1 was the same thing for the `katana-enterprise` repo.)


arthurp:
... I feel something is lacking from the PR name. For instance, content. ;-)

docs/reference/python/katana.local.graph.rst
```diff
@@ -14,8 +14,7 @@ When working with graph projections, keep in mind that:
 
 - The resulting `Graph` projection object shares property data with the original graph. If a user runs an analytics routine such as PageRank on the projection, this will add a new node property to both the projection and the original graph.
 
-- The projection will only be functional while the original graph remains in memory. Deleting the original graph while the projected one is still in use may cause erratic behavior with the projection. Make sure to delete any projections using Python’s `del` command before doing the same with the original graph.
-
+- The projection will only be functional while the original graph remains in memory. Deleting the original graph while the projected one is still in use may cause erratic behavior with the projection.
```
arthurp: Deleting the original graph while the view is still accessible shouldn't be possible from Python. That was the point of our discussion. I will make sure this is really true in the wrapping code. I created an issue for this: KAT-4441.

arthurp (changes requested):

arthurp (approved):

PR #970, opened 2022-03-16
Initial PG Topology and View interface with Opaque IDs

amberhassaan:
[no test]

@mindv0rtex : This is the initial interface. Doesn't build yet. I'll ping you once `libgraph` starts to build and then we can divide up the build error & bug chasing.

### DIFF
--- a/libgraph/include/katana/GraphTopology.h
+++ b/libgraph/include/katana/GraphTopology.h
@@ -13,6 +13,7 @@
 #include "katana/Iterators.h"
 #include "katana/Logging.h"
 #include "katana/NUMAArray.h"
+#include "katana/OpaqueID.h"
 #include "katana/RDGTopology.h"
 #include "katana/Result.h"
 #include "katana/config.h"
@@ -29,6 +30,7 @@ class KATANA_EXPORT PropertyGraph;
 
 /// Types used by all topologies
 struct KATANA_EXPORT GraphTopologyTypes {
+  /*
   using Node = uint32_t;
   using Edge = uint64_t;
   using PropertyIndex = uint64_t;
@@ -43,6 +45,59 @@ struct KATANA_EXPORT GraphTopologyTypes {
   using EdgeDestVec = NUMAArray<Node>;
   using PropIndexVec = NUMAArray<PropertyIndex>;
   using EntityTypeIDVec = NUMAArray<EntityTypeID>;
+  */
+
+  struct Node: public OpaqueIDLinear<Node, uint32_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct Edge: public OpaqueIDLinear<Edge, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct OutEdgeHandle: public OpaqueIDLinear<OutEdgeHandle, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct InEdgeHandle: public OpaqueIDLinear<InEdgeHandle, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct InOutEdgeHandle: public OpaqueIDLinear<InOutEdgeHandle, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  // TODO(Amber/emcginnis): Can this be uint32_t?
+  struct NodePropertyIndex: public OpaqueIDLinear<NodePropertyIndex, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct EdgePropertyIndex: public OpaqueIDLinear<EdgePropertyIndex, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  // TODO(Amber): Opaque IDs should define a separate iterator type instead of
+  // mixing the concepts in one class. 
+  using node_iterator = OpaqueIDCountingIterator<Node>;
+  using nodes_range = StandardRange<node_iterator>;
+
+  using edge_iterator = OpaqueIDCountingIterator<OutEdgeHandle>;
+  using edges_range = StandardRange<edge_iterator>;
+  // TODO(Amber): Remove because we have Nodes() method to iterate over all nodes
+  using iterator = node_iterator;
+
+  //TODO(emcginnis): Each of these *Vec types should really be *Array since they are not resizable
+  // NOTE(Amber): using underlying type so that (1) tusba & gluon partitioner interface doesn't
+  // need to change (2) We can repurpose the adj indices for InOutEdgeHandle in
+  // UndirectedTopology, and InEdgeHandle in Bidirectional topology 
+  using AdjIndexVec = NUMAArray<OutEdgeHandle::underlying_type>;
+  using EdgeDestVec = NUMAArray<Node::underlying_type>;
+  using NodePropIndexVec = NUMAArray<NodePropertyIndex::underlying_type>;
+  using EdgePropIndexVec = NUMAArray<EdgePropertyIndex::underlying_type>;
+  // TODO(Amber): Convert EntityTypeID into a Opaque type
+  using EntityTypeIDVec = NUMAArray<EntityTypeID>;
+  // TODO(Amber/emcginnis): Add static assert checks that tsuba interface topology
+  // pointer types match the array data types above
 };
 
 class KATANA_EXPORT EdgeShuffleTopology;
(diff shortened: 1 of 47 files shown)


libgraph/src/analytics/bfs/bfs.cpp
```diff
@@ -159,7 +159,7 @@ AsynchronousAlgo(
   katana::GAccumulator<size_t> bad_work;
   katana::GAccumulator<size_t> WL_empty_work;
 
-  (*node_data)[source] = 0;
+  (*node_data)[source.value()] = 0;
```
gurbinder533, line 162: This is problematic, right? `source.value` are not contiguous and `size of node_data == num_nodes.` Similar things exist for many other apps as well.

amberhassaan: Yes, it will be a problem when we switch to LSPG. Not a problem today, obviously.

libsupport/include/katana/OpaqueID.h
```diff
@@ -278,6 +302,83 @@ struct OpaqueIDLinear : public OpaqueIDOrderedWithValue<_IDType, _Value> {
...
+  OpaqueIDCountingIterator& operator=(const OpaqueIDCountingIterator& other) noexcept = default;
+  OpaqueIDCountingIterator& operator=(OpaqueIDCountingIterator&& other) noexcept = default;
+
+private:
+
+  bool equal(const OpaqueIDCountingIterator& that) const noexcept {
+    return value_ == that.value_;
+  }
+
+  typename Base::reference
```
yan-zaretskiy, line 363: Do you need to add `const` here?

libgraph/include/katana/GraphTopology.h
```diff
@@ -43,6 +45,59 @@ struct KATANA_EXPORT GraphTopologyTypes {
...
+  using node_iterator = OpaqueIDCountingIterator<Node>;
+  using nodes_range = StandardRange<node_iterator>;
+
+  using edge_iterator = OpaqueIDCountingIterator<OutEdgeHandle>;
+  using edges_range = StandardRange<edge_iterator>;
+  // TODO(Amber): Remove because we have Nodes() method to iterate over all nodes
+  using iterator = node_iterator;
+
+  //TODO(emcginnis): Each of these *Vec types should really be *Array since they are not resizable
+  // NOTE(Amber): using underlying type so that (1) tusba & gluon partitioner interface doesn't
```
yan-zaretskiy, line 90: Ah, so that's how you escaped the extra pain :-)

libgraph/include/katana/GraphTopology.h
```diff
@@ -43,6 +45,59 @@ struct KATANA_EXPORT GraphTopologyTypes {
...
+  // need to change (2) We can repurpose the adj indices for InOutEdgeHandle in
+  // UndirectedTopology, and InEdgeHandle in Bidirectional topology 
+  using AdjIndexVec = NUMAArray<OutEdgeHandle::underlying_type>;
```
yan-zaretskiy, line 93: A random thought as I'm reading the code - I'd alias `OutEdgeHandle::underlying_type` as smth like `OutEdgeHandle_t` and so on, too hard on the eyes, IMHO.

libgraph/include/katana/GraphTopology.h
```diff
@@ -43,6 +45,59 @@ struct KATANA_EXPORT GraphTopologyTypes {
...
+  struct Node: public OpaqueIDLinear<Node, uint32_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct Edge: public OpaqueIDLinear<Edge, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct OutEdgeHandle: public OpaqueIDLinear<OutEdgeHandle, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
```
yan-zaretskiy, line 59: What do you think about making the constructors of all these handle types private and `GraphTopology` to be their friend? No code outside of topology should be able to construct a handle.

libgraph/include/katana/GraphTopology.h
```diff
@@ -43,6 +45,59 @@ struct KATANA_EXPORT GraphTopologyTypes {
...
+  };
+
+  struct InEdgeHandle: public OpaqueIDLinear<InEdgeHandle, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
+  struct InOutEdgeHandle: public OpaqueIDLinear<InOutEdgeHandle, uint64_t> {
+    using OpaqueIDLinear::OpaqueIDLinear;
+  };
+
```
yan-zaretskiy, line 69: What about the `NodeHandle`? It's for situations where we sort nodes.

libsupport/include/katana/OpaqueID.h
amberhassaan: There's a boost library issue here that prevents me from returning const ref. It's the way the iterator_facade code is written.

libgraph/include/katana/GraphTopology.h
amberhassaan: Agree, I've had to do this in analytics code. However `_t` name is not specific enough. Let me suggest `OutEdgeHandleInt` and `NodeInt` and so on.

libgraph/include/katana/GraphTopology.h
amberhassaan: Good suggestion but let's leave it for near future work.

libgraph/include/katana/GraphTopology.h
amberhassaan: Good point. We should add it.

libgraph/include/katana/GraphTopology.h
amberhassaan: There is an added benefit too. `std::vector` has to call a constructor for the entire vector when you create one with non-zero size. This construction is serial and can become a hidden time sink (and performance issue). So, we should be careful constructing large vectors of opaque types. `NUMAArray` on the other hand has a `constructAt` for this very reason (and can be invoked in a parallel `do_all`)

libgraph/include/katana/GraphTopology.h
```diff
@@ -994,51 +1084,70 @@ class KATANA_EXPORT UndirectedTopologyWrapper : public GraphTopologyTypes {
...
     }
-    return out().OutEdgeDst(eid);
+    return out().OutEdgeDst(OutEdgeHandle{eh.value()});
+  }
+
+  // TODO(amber): It is possible to call UndirectedEdgeNeighbor with an
+  // OutEdgeHandle returned by Edges() below. Which leads me to suggest that we
+  // should drop UndirectedEdgeNeighbor vs OutEdgeDst and stick to a generic name
+  // like GetEdgeDest
+  Node UndirectedEdgeNeighbor(const OutEdgeHandle& eh) const noexcept {
```
yan-zaretskiy, line 1152: It's possible, but why do we need to add it? You can return `InOutEdgeHandle` from `Edges()`

libgraph/include/katana/GraphTopology.h
```diff
@@ -1303,21 +1425,23 @@ class KATANA_EXPORT EdgeTypeAwareBiDirTopology
 
 template <typename Graph>
 using has_undirected_t =
-    decltype(std::declval<Graph>().UndirectedEdges(typename Graph::Node()));
+    decltype(std::declval<Graph>().UndirectedEdges(std::declval<typename Graph::Node>()));
 
-template <typename Graph>
-auto
-Edges(const Graph& graph, typename Graph::Node node) {
-  if constexpr (katana::is_detected_v<has_undirected_t, Graph>) {
```
yan-zaretskiy: Do you prefer SFINAE to `constexpr if`? It's the same result with more code and ugly templates, unless I am missing something.

yan-zaretskiy: I see you didn't remove it below, so there must be something I'm missing.

libgraph/include/katana/GraphTopology.h
amberhassaan: Yes, this comment is outdated. I ended up returning the same type of `edges_range` (i.e. InOutEdgeHandle type) and you're right, this invariant should be established by each topology/view class. Can't have some methods returning a different edge range.

libgraph/include/katana/GraphTopology.h
amberhassaan: Return types are different, hence two different overloads.

amberhassaan: It could be that the compiler can handle this case. I'm yet to find out.

yan-zaretskiy: `auto` functions with `constexpr if` support different branches returning different types, e.g. https://godbolt.org/z/4MPcTrssK

amberhassaan: Disturbing :). But I don't mind if you restore to `constexpr if`.

amberhassaan:
Closed and moved all changes to enterprise PR https://github.com/KatanaGraph/katana-enterprise/pull/3762

PR #972, opened 2022-03-17
docs: Add replacements globally

ddn0:
Without this change, "make docs" fails with:

    docstring of katana.local_native.PyCapsule.in_degree:1:Undefined substitution referenced: "lazy_compute".


JamesFCoffey (approved):
LGTM. Since we are looking at the Google Style guides for Python and C++, I suppose we should also follow suit with filenames ("Use all lowercase, and separate words with hyphens.")

ddn0:
> LGTM. Since we are looking at the Google Style guides for Python and C++, I suppose we should also follow suit with filenames ("Use all lowercase, and separate words with hyphens.")

The Python style is
> must not contain dashes

https://google.github.io/styleguide/pyguide.html#3163-file-naming

While the C++ style is 
> Filenames should be all lowercase and can include underscores (_) or dashes (-). Follow the convention that your project uses. If there is no consistent local pattern to follow, prefer `_`.

https://google.github.io/styleguide/cppguide.html#File_Names

I don't really care as long as we are consistent. If we change to `_`, it should be done separately though.

(Interesting note: the anchors in the Python style guide use `-` and the C++ use `_`)

docs/conf.py
```diff
@@ -136,6 +136,15 @@
...
 copyright = "Katana Graph, Inc. 2022"
 
+rst_epilog = """
+.. |supports_compiled_operator| replace::
+    This method may be used in compiled operators with some restrictions.
+
+.. |lazy_compute| replace::
+    The information required to perform this operation efficiently is computed lazily on the first call to this
+    method. This information is shared between related methods when possible.
+"""
```
arthurp, line 146: I do not love this. I would much prefer to have directory level sets of replacements or something. Since there replacements are not global, they only apply to a few specific classes.

arthurp (approved):
This is a fine hack to fix it. But I would rather have a better way. @JamesFCoffey do you know a better way to scope the replacements as we need them?

docs/conf.py
ddn0: I assumed the issue was that the doc strings are being used outside the scope of the original rst file:

`katana.local_native.PyCapsule.in_degree` isn't [katana.local.graph.rst](https://github.com/KatanaGraph/katana/pull/972/files#diff-74e0246d9861f9edeea27115f8a5a18470ffc8d61a50016fd600db92be689af1). I was assuming this gets pulled in because of subclassing or something.

If I am right, replacements are not directory scoped. An alternative would be to organize the replacement names themselves in some hierarchy.

arthurp: Yea. I bet you are right. And it probably is subclassing (distributed.Graph is a subclass of local.Graph). I'm not sure the best way to do this.

ddn0: I'm going to go ahead with this current hack; if y'all figure out something better, we can do it down the line.

PR #973, opened 2022-03-17
Add script to check build environment for issues [KAT-4926]

arthurp:


scripts/check_build_environment.py
```diff
@@ -0,0 +1,145 @@
...
+            ["conda", "compare", yml], stderr=subprocess.STDOUT, stdout=subprocess.PIPE, encoding="UTF-8"
+        ) as p:
+            output = p.stdout.read()
+            p.wait()
+            if p.returncode != 0:
+                show_problem(f"Conda environment needs update with {yml}\n{output}", CONDA_ENV_UPDATE_INSTRUCTIONS)
+    return has_problem
+
+
+def main():
```
ddn0: Honestly, I'd rather add a subcommand in `dev` like `dev host build`

1. We already have to refer to the enterprise repo so this change is not purely open repo related.
2. `dev` currently does containerized builds but it seems pretty natural to have `dev build` and `dev host build` as they are trying to do the same thing
3. We have a lot of scripts and it is getting to the point where people don't know which ones exist or how/when/why to use them. I think most of them can be folded into a few high-level commands "build", "lint", etc.

arthurp: Fair. I just think this should be in open. The more we put out tooling in enterprise the less we actually have an open source offering. We are already not looking great on that front.

Is moving dev to open still in the works? It seems pretty important to have our main tooling open for any kind of serious open source offering.

ddn0: > Is moving dev to open still in the works?

No one is actively working on this. Though the blocker here is what exactly the division between the two repos should be.

If you want to at least unify the current scripts we have into a single "dev for non containerized builds" and `dev` can just call `dev_for_non_containerized_builds build/lint/check` as part of its `dev host builb/lint/check` and containerized variants that seems a better path than more scripts.

arthurp: OK. What do you think we should do with this specific tool? I don't want to spend too much time on it and I do think this is important as it would have saved at least two people some pain in the last week.

ddn0: If you want people to regularly run this check, I think we should add a command like `dev host build`, which ensures a configured build directory, activated conda environment if using conda, warns that submodules are out of date, run checks, etc. and does a build.

That command can call whatever scripts are appropriate, perhaps the one in this PR.

As a second-order concern, it would be nice for the external script(s) that `dev host build` calls to map one-to-one with `dev host build` , so that the latter command is simply `exec <external-script> $*` with maybe some more arguments like `--conda-env-file` or whatever.

Bonus points for doing the same with linting.

arthurp: So you are saying we can add this script in a form that is cleaned up to match `dev`. Then we can updated `dev` to call it as a subcommand. Right?

Then I propose this:

* I'll update this script to use `click` and match the style of `dev` as best I can.
* I'll add parameters that `dev` might need to set (like explicit build directory or source directory).
* Then we can integrate this with `dev` as a subcommand either by calling or, down the line, just by moving this code into `dev`.

How does that sound?

I think this would work well as it gets this out there reasonably quickly, but still supports the direction you want to take the custom dev scripts.

ddn0: Sounds good to me.

arthurp:
@ddn0 I think I updated things to match out plan. Can you confirm?

arthurp:
@bowu @chakpongchung You can download the script from https://raw.githubusercontent.com/KatanaGraph/katana/e83c0a063432e55a8026395d86d9c2e0c31aafe7/scripts/check_build_environment.py and put it in `external/katana/scripts` and try it. You may get issues where you always see a problem from either the open or enterprise environment file. This is because before this PR they actually disagree slightly and cause different packages to be installed.

ddn0 (approved):

amazloumi (approved):
let's remove mismatches

PR #976, opened 2022-03-18
[KAT-4444] Add useful string representation for non-atomic types

denish44:
The non-atomic types will now convert to atomic types and create a useful string representation by concatenating the atomic type names with '&'.  This is done recursively as non-atomic types can be built from another non-atomic type. The end result is flat list of unique atomic type names.

jira: KAT-4444

### DIFF
--- a/libtsuba/include/katana/EntityTypeManager.h
+++ b/libtsuba/include/katana/EntityTypeManager.h
@@ -527,6 +527,30 @@ public:
     return std::nullopt;
   }
 
+  /// \returns the names of the atomic supertypes for the given non-atomic type
+  /// \p entity_type_id is an non-atomic type
+  std::set<std::string> GetNonAtomicTypeName(
+      EntityTypeID entity_type_id) const {
+    std::set<std::string> result;
+    auto atomic_types = GetAtomicSubtypes(entity_type_id);
+    for (EntityTypeID super_type_id = 0; super_type_id < atomic_types.size();
+         ++super_type_id) {
+      if (atomic_types.test(super_type_id)) {
+        auto val = GetAtomicTypeName(super_type_id);
+        if (val) {
+          result.insert(val.value());
+        } else {
+          // Seems it's possible to have non-atomic from another non-atomic.
+          auto atomic_type_names = GetNonAtomicTypeName(super_type_id);
+          // Set will avoid any duplicate names.
+          result.insert(atomic_type_names.begin(), atomic_type_names.end());
+        }
+      }
+    }
+
+    return result;
+  }
+
   /// \returns a vector containing all atomic type IDs
   std::vector<EntityTypeID> GetAtomicEntityTypeIDs() const {
     std::vector<EntityTypeID> type_vec;
--- a/libtsuba/include/katana/EntityTypeManager.h
+++ b/libtsuba/include/katana/EntityTypeManager.h
@@ -527,7 +527,7 @@ public:
     return std::nullopt;
   }
 
-  /// \returns the names of the atomic supertypes for the given non-atomic type
+  /// \returns the names of the atomic subtypes for the given non-atomic type
   /// \p entity_type_id is an non-atomic type
   std::set<std::string> GetNonAtomicTypeNames(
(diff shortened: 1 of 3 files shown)


libtsuba/include/katana/EntityTypeManager.h
```diff
@@ -527,6 +527,30 @@ class KATANA_EXPORT EntityTypeManager {
...
+    auto atomic_types = GetAtomicSubtypes(entity_type_id);
+    for (EntityTypeID super_type_id = 0; super_type_id < atomic_types.size();
+         ++super_type_id) {
+      if (atomic_types.test(super_type_id)) {
+        auto val = GetAtomicTypeName(super_type_id);
+        if (val) {
+          result.insert(val.value());
+        } else {
+          // Seems it's possible to have non-atomic from another non-atomic.
+          auto atomic_type_names = GetNonAtomicTypeNames(super_type_id);
```
denish44: @roshandathathri this is why i was asking on property-graph-2 channel. If I understood correctly I have to recursively check. I did a test to confirm this and it seems it's possible. (see [test_property_graph.py](https://github.com/KatanaGraph/katana/pull/976/files#diff-ca79ca88ad2a58b4d04d59925b6352ab90930da91d5bd171f073d1662f12db01) changes)

Just want to make sure I didn't misinterpret the meaning of your reply.

arthurp: If this recursive traversal is needed then that should be handled in `GetAtomicSubtypes`. Otherwise it's not actually correct. This function should definitely not have to do anything other than getting the names.

arthurp: If this recursive traversal is needed then that should be handled in `GetAtomicSubtypes`. Otherwise it's not actually correct. This function should definitely not have to do anything other than getting the names.

libtsuba/include/katana/EntityTypeManager.h
```diff
@@ -527,6 +527,30 @@ class KATANA_EXPORT EntityTypeManager {
...
 
+  /// \returns the names of the atomic supertypes for the given non-atomic type
+  /// \p entity_type_id is an non-atomic type
+  std::set<std::string> GetNonAtomicTypeNames(
+      EntityTypeID entity_type_id) const {
+    std::set<std::string> result;
+    auto atomic_types = GetAtomicSubtypes(entity_type_id);
+    for (EntityTypeID super_type_id = 0; super_type_id < atomic_types.size();
+         ++super_type_id) {
+      if (atomic_types.test(super_type_id)) {
```
arthurp: I think I added iteration to `DynamicBitset` and friends. So you don't need to do this, you can use a for-each loop.

arthurp: I think I added iteration to `DynamicBitset` and friends. So you don't need to do this, you can use a for-each loop.

libtsuba/include/katana/EntityTypeManager.h
denish44: I taught about that too but the original method expects a atomic type. a non atomic type would return a null in the code. Now i can change that but I am not sure how other code trace could be affected by this change. that was my only fear

libtsuba/include/katana/EntityTypeManager.h
denish44: Oh interesting. Let me take a look, that would be a nice change

libtsuba/include/katana/EntityTypeManager.h
denish44: @roshandathathri I been thinking about what you said and I want to make sure I am not confused by my understanding. Can you confirm if my test case is even a sensible thing?

libtsuba/include/katana/EntityTypeManager.h
denish44: changed.

libtsuba/include/katana/EntityTypeManager.h
denish44: fixed based on meeting discussions

arthurp (approved):

PR #978, opened 2022-03-21
Use katana::URI to name graph location

witchel:
You ever pull a thread and end up naked?  I wanted to eliminate 12 lines in `RDGLoader `but ended up with this more comprehensive change to use `katana::URI` to name graph locations.  Is it a good idea?  I guess I hope so, but I've really lost perspective, so please let me know.

@arthurp, I pattern matched something for python, but I'd appreciate it if you could give that a careful look.

There will be an enterprise component.

### DIFF
--- a/libkatana_python_native/src/PropertyGraph.cpp
+++ b/libkatana_python_native/src/PropertyGraph.cpp
@@ -363,6 +363,7 @@ DefPropertyGraph(py::module& m) {
              std::optional<std::vector<std::string>> edge_properties,
              TxnContext* txn_ctx) -> std::shared_ptr<PropertyGraph> {
             auto path_str = py::str(path).cast<std::string>();
+            auto path_uri = PythonChecked(URI::Make(path_str));
             py::gil_scoped_release guard;
             katana::RDGLoadOptions options = katana::RDGLoadOptions::Defaults();
             options.node_properties = node_properties;
@@ -370,7 +371,7 @@ DefPropertyGraph(py::module& m) {
             TxnContextArgumentHandler txn_context_handler(txn_ctx);
             KATANA_LOG_DEBUG("{}", reinterpret_cast<uintptr_t>(path.ptr()));
             return PythonChecked(PropertyGraph::Make(
-                path_str, txn_context_handler.get(), options));
+                path_uri, txn_context_handler.get(), options));
           }),
       py::arg("path"), py::kw_only(), py::arg("node_properties") = std::nullopt,
       py::arg("edge_properties") = std::nullopt,
@@ -1034,7 +1035,8 @@ DefPropertyGraph(py::module& m) {
          const std::string& provenance, TxnContext* txn_ctx) {
         TxnContextArgumentHandler txn_handler(txn_ctx);
         if (path) {
-          return self.Write(*path, provenance, txn_handler.get());
+          auto path_uri = PythonChecked(URI::Make(*path));
+          return self.Write(path_uri, provenance, txn_handler.get());
         }
         return self.Commit(provenance, txn_handler.get());
       },
@@ -1051,7 +1053,9 @@ DefPropertyGraph(py::module& m) {
       :param command_line: Lineage information in the form of a command line.
       :type command_line: str
       )""");
-  cls.def_property_readonly("path", &PropertyGraph::rdg_dir);
+  cls.def_property_readonly("path", [](PropertyGraph& self) -> std::string {
+    return self.rdg_dir().string();
+  });
 }
 
 template <typename node_or_edge>
--- a/libsupport/include/katana/URI.h
+++ b/libsupport/include/katana/URI.h
@@ -63,6 +63,9 @@ public:
 
   URI operator+(char rhs) const;
   URI operator+(const std::string& rhs) const;
+  bool operator<(URI const& o) const {
+    return scheme_ < o.scheme_ || path_ < o.path_;
+  }
 
 private:
   URI(std::string scheme, std::string path);
--- a/libtsuba/test/storage-format-version/v3-optional-topologies.cpp
+++ b/libtsuba/test/storage-format-version/v3-optional-topologies.cpp
@@ -49,17 +49,17 @@ CSRPresent(katana::RDG* rdg) {
 }
 
 void
-CleanupRDGDirs(std::vector<std::string> dirs) {
+CleanupRDGDirs(std::vector<katana::URI> dirs) {
   for (auto rdg_dir : dirs) {
     KATANA_LOG_DEBUG("removing rdg dir: {}", rdg_dir);
-    fs::remove_all(rdg_dir);
+    fs::remove_all(rdg_dir.path());
   }
 }
 
 /// Load a graph that was stored without optional topology support
 /// Ensure it survives a store/load cycle
 katana::Result<void>
-TestGraphBackwardsCompatabilityRoundTrip(const std::string& rdg_name) {
+TestGraphBackwardsCompatabilityRoundTrip(const katana::URI& rdg_name) {
   KATANA_LOG_DEBUG("***** Testing Backwards Compatability *****");
 
   KATANA_LOG_ASSERT(!rdg_name.empty());
@@ -71,7 +71,7 @@ TestGraphBackwardsCompatabilityRoundTrip(const std::string& rdg_name) {
   KATANA_CHECKED(CSRPresent(&rdg));
 
   // write out converted rdg
-  std::string rdg_dir1 = KATANA_CHECKED(WriteRDG(std::move(rdg)));
+  auto rdg_dir1 = KATANA_CHECKED(WriteRDG(std::move(rdg)));
   KATANA_LOG_ASSERT(!rdg_dir1.empty());
 
   // load converted rdg
@@ -87,7 +87,7 @@ TestGraphBackwardsCompatabilityRoundTrip(const std::string& rdg_name) {
 /// Ensure added optional topology didn't change
 /// Since we added the optional topology to a graph that originally did not have
 katana::Result<void>
-TestGraphComplexOptionalTopologyRoundTrip(const std::string& rdg_name) {
+TestGraphComplexOptionalTopologyRoundTrip(const katana::URI& rdg_name) {
   KATANA_LOG_DEBUG(
       "***** Testing Complex Optional Topology Support Roundtrip *****");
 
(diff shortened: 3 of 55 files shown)


libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -363,14 +363,19 @@ DefPropertyGraph(py::module& m) {
              std::optional<std::vector<std::string>> edge_properties,
              TxnContext* txn_ctx) -> std::shared_ptr<PropertyGraph> {
             auto path_str = py::str(path).cast<std::string>();
+            auto res = URI::Make(path_str);
+            if (!res) {
+              KATANA_LOG_FATAL("uri problem {}: {}", path_str, res.error());
```
arthurp: Never use KATANA_LOG_FATAL. You can use `PythonChecked` here. It throws exceptions in the failure case which works well with python.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -1030,11 +1035,11 @@ DefPropertyGraph(py::module& m) {
 
   cls.def(
       "write",
-      [](PropertyGraph& self, const std::string* path,
+      [](PropertyGraph& self, const katana::URI& path,
```
arthurp: This is not gonna work since `katana::URI` isn't wrapped for python. Ping me tomorrow and I'll have a better way to handle this.

libkatana_python_native/src/PropertyGraph.cpp
```diff
@@ -1030,11 +1035,11 @@ DefPropertyGraph(py::module& m) {
...
          const std::string& provenance, TxnContext* txn_ctx) {
         TxnContextArgumentHandler txn_handler(txn_ctx);
-        if (path) {
-          return self.Write(*path, provenance, txn_handler.get());
+        if (!path.empty()) {
+          return self.Write(path, provenance, txn_handler.get());
```
arthurp: The idea of "empty" objects doesn't really make any sense in python. So we'll take either a real value or `None`. Pybind11 is nice enough to convert `None` to `null` if you take a pointer instead of a reference, so that's the way to go.

libsupport/include/katana/URI.h
```diff
@@ -63,6 +63,9 @@ class KATANA_EXPORT URI {
 
   URI operator+(char rhs) const;
   URI operator+(const std::string& rhs) const;
+  bool operator<(URI const& o) const {
+    return path_ < o.path_ || (path_ == o.path_ && scheme_ < o.scheme_);
+  }
```
arthurp: I would expect scheme to come first so this is closer to lexical order.

arthurp (changes requested):

libkatana_python_native/src/PropertyGraph.cpp
arthurp: The way I think about empty object is that they are "null" in some sense. That means that any value that allows this is implicitly `std::optional<T>` even when written `T`. In some cases it's better to think of the "empty" value as `0`, but this has the problem that `0` is in fact a valid value, and often the empty value is not actually valid.

libkatana_python_native/src/PropertyGraph.cpp
witchel: I wrote a `URI.pxd` file.  It probably has problems.  Is that what you mean by wrapping?  We can follow up on slack, if that is easier.

libtsuba/test/storage-format-version/v3-optional-topologies.cpp
```diff
@@ -325,6 +325,15 @@ TestGraphBasicRoundTrip(const std::string& rdg_name) {
   return katana::ResultSuccess();
 }
 
+katana::Result<void>
+Run(const std::string& rdg_str) {
```
e-mcginnis, line 329: Big fan of wrapping all of the test cases in a `Run` func where we can use the usual `KATANA_CHECKED`

libkatana_python_native/src/PropertyGraph.cpp
arthurp: pxd and pybind11 are totally separate. Currently some things appear in both places. However I think in this case we don't actually need any of the Cython side changes since all that code is legacy and I think can actually be removed (I'll make a PR to do that).

e-mcginnis (approved):
I have no objections to the cpp code. I can see how it would make sense to use our URI class to describe RDG paths instead of strings, and centralizing the error checking to the URI::Make funcs might be argument enough for this change.

libkatana_python_native/src/PropertyGraph.cpp
witchel: Hmm.  This used to be a pointer to an RDG location and the check was whether the pointer was null.

I changed it to a reference to a URI.  I figure if the URI is empty, then we commit rather than write to a new location.  I'm not sure what your suggestion is.  Only pass a pointer to a URI to maintain the pointer check?  `empty()` is defined on URIs and makes sense to me as something to check to see if we are specifying a new location for the RDG.

arthurp: In C++ this is fine. It doesn't make sense in Python because everything is always a "pointer" (actually a reference, but that term is overloaded in C++). And this is Python interfacing code, so we need to handle things the Pythonic way.

libsupport/include/katana/URI.h
witchel: Yeah, I actually thought about this a little.  I figured file://a < http://b is more intuitive because the scheme is almost out of band information and not lexicographically meaningful.

But your suggestion is reasonable.  I was also worried if file:// schemes are sometimes represented as empty.  That doesn't look to be a problem.

arthurp:
I realized I was making you wait and I just shouldn't. For now go ahead and just take a string in the Python API. Like this `const std::string*`. Check it for null to catch the no value provided case. The default can be null if that arg has a default. I think this is how the code was before. you can convert that to URI before passing it to C++.

[Using URIs on the Python side will have to come later](https://katanagraph.atlassian.net/browse/KAT-5006). Sorry to block you if I did.

witchel:
> I realized I was making you wait and I just shouldn't. For now go ahead and just take a string in the Python API. Like this `const std::string*`. Check it for null to catch the no value provided case. The default can be null if that arg has a default. I think this is how the code was before. you can convert that to URI before passing it to C++.
> 
> Using URIs on the Python side will have to come later. Sorry to block you if I did.

No problem. Can do!

libsupport/include/katana/URI.h
arthurp: `file://` isn't a valid URL, right? File URLs must always be absolute and are written with an empty host part, so they always start with `file:///` (*three* slashes not two).

arthurp: Random guy on the internet agrees with me: https://stackoverflow.com/questions/7857416/file-uri-scheme-and-relative-files

python/katana/cpp/libsupport/URI.pxd
```diff
@@ -0,0 +1,19 @@
...
+        Result[URI] Make(const string& str)
+        string path() const
+        string string() const
+        bool empty() const
+
+cdef inline URI handle_result_URI(Result[URI] res) nogil except *:
+    if not res.has_value():
+        with gil:
+            raise_error_code(res.error())
+    return res.value()
```
witchel: @arthurp, do I not need this file?  I am dealing with URIs in python code, does that mean I need a pybind wrapper?

libsupport/include/katana/URI.h
witchel: I meant to write this.  Sorry to confuse.

Yeah, I actually thought about this a little. I figured file:///a < http:///b is more intuitive because the scheme is almost out of band information and not lexicographically meaningful.

But your suggestion is reasonable. I was also worried if file:/// schemes are sometimes represented as empty. That doesn't look to be a problem.

python/katana/cpp/libsupport/URI.pxd
arthurp: You do not need this file. And you shouldn't need a pybind wrapper for URI at all. Instead just convert to URI in the wrappers for functions that now take URI.

libsupport/include/katana/URI.h
arthurp: The danger of sentinels that are part of the type. :-) With correct use of `std::optional` that shouldn't be an issue anyway.

tylershunt (approved):
LGTM!

yige-hu (approved):
LGTM

witchel:
Hey @arthurp,

Sorry, I'm getting a python error I don't understand.  Where is this python code getting its definition of a Graph?  Not Graph.pxd? Where is path defined?  I looked around, but I can't figure it out.  Thanks!

```
=================================== FAILURES ===================================
___________________________ test_load_graphml_write ____________________________
TypeError: Unregistered type : katana::URI

The above exception was the direct cause of the following exception:

    @pytest.mark.required_env("KATANA_TEST_DATASETS")
    def test_load_graphml_write():
        input_file = Path(get_misc_dataset("graph-convert/movies.graphml"))
        pg = from_graphml(input_file)
        with TemporaryDirectory() as tmpdir:
            pg.write(tmpdir)
            del pg
            graph = Graph(tmpdir)
>           assert graph.path == f"file://{tmpdir}"
E           TypeError: Unable to convert function return value to a Python type! The signature was
E           	(arg0: katana.local_native.Graph) -> katana::URI
```

arthurp:
You need to convert the URI to a string before returning it to Python since we don't have a URI type there. Converting do `std::string` in the pybind11 wrappers should fix this.

witchel:
> You need to convert the URI to a string before returning it to Python since we don't have a URI type there. Converting do `std::string` in the pybind11 wrappers should fix this.

I want to be cool, I want to be hip, I want to convert URIs.  But I don't even know what file you are talking about.  Help!

arthurp:
> > You need to convert the URI to a string before returning it to Python since we don't have a URI type there. Converting do `std::string` in the pybind11 wrappers should fix this.
> 
> I want to be cool, I want to be hip, I want to convert URIs. But I don't even know what file you are talking about. Help!

I think `rdg_dir` now returns a URI. So the code that wraps `rdg_dir` now needs to convert it. The code is here: https://github.com/KatanaGraph/katana/blob/bb07219dca2c84144bd6724c02bea64636487735/libkatana_python_native/src/PropertyGraph.cpp#L1056

You will  need to replace the method pointer with a lambda that calls the method and converts and returns the results. This file is where all the PropertyGraph pybind11 wrappers are so you should see some examples, and you have already changed a couple of things in the file.

witchel:
@arthurp, I think you have a hold on this PR.  Are there outstanding changes you want?

arthurp (approved):

PR #984, opened 2022-03-23
Enforce all rdgs in test-datasets have cmake variables

aneeshdurg:


scripts/check_test_datasets.py
```diff
@@ -0,0 +1,49 @@
+#! /usr/bin/env python3
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
```
e-mcginnis, line 8: nit: I'm a big fan of the convention of writing a block comment at the top of these things stating what its use is, for the uninitiated it can be unclear what this is checking.

scripts/check_test_datasets.py
```diff
@@ -0,0 +1,49 @@
...
+
+@dataclass
+class RDGDir:
+    path: str
+
+    @property
+    def dir_name(self) -> str:
+        return self.path.split("/")[-1]
```
e-mcginnis: nit: this is probably fine, but pathlib.name() is pretty nice for this

e-mcginnis (approved):
lgtm, thanks for this!

PR #987, opened 2022-03-24
Verify TestDatasetsRDGs.cmake in CI

aneeshdurg:
All PRs will fail until this is checked in

This PR adds a new workflow to check that TestDatasetsRDGs.cmake is up-to-date.


scripts/check_test_datasets.py
```diff
@@ -46,6 +47,7 @@ def cmake_definition(self) -> str:
         "d",
     ]
     rdgs_list = subprocess.check_output(get_rdgs_cmd).decode().split()
+    rdgs_list.sort()
```
e-mcginnis, line 50: we love sorting

e-mcginnis (approved):
lgtm!

PR #991, opened 2022-03-25
Added outputs and tests to ksssp

dqawami:


python/test/test_cpp_algos.py
```diff
@@ -153,6 +154,29 @@ def test_sssp(graph: Graph):
...
+        assert paths[i][-1] == report_node
+
+        for _, j_path in enumerate(paths, start=i):
+            if len(i_path) == len(j_path):
+                unique_path = False
+                for (elem_i, elem_j) in zip(i_path, j_path):
+                    if elem_i != elem_j:
+                        unique_path = True
+                        break
+                assert unique_path
```
arthurp: This is complex enough to document what you are doing so people don't have to figure out you are checking that the paths are different.

Also I think you can write this whole thing as follows:
```suggestion
    for i, i_path in enumerate(paths):
        assert paths[i][-1] == report_node
        assert all(list(i_path) == list(j_path) for j_path in paths[i:])
```
`list(x)` pretty much just iterates `x` and puts it's elements in a list. While the paths could in theory be O(|E|), it's unlikely that will be the case for testing.

python/katana/local/analytics/_ksssp.pyx
```diff
@@ -192,18 +213,61 @@ def ksssp(pg, str edge_weight_property_name, size_t start_node,
...
+            raise_error_code(res.error())
+    return res.value()
+
+
+cdef class KssspStatistics(Statistics):
+    """
+    Compute the :ref:`statistics` of a kSSSP computation on a graph
+    """
+    cdef _KssspStatistics underlying
+
```
arthurp, line 287: This class needs properties to expose the actual statistics. At the moment Python cannot actually access any of the statistical values.

dqawami: For now KssspStatistics just stores the paths as arrays and a weight for each path, any suggestions for other stats to calculate?

python/test/test_cpp_algos.py
dqawami: I like the suggestion, I think the main difference would be for the last assert to do `list(i_path) != list(j_path)` given the last check is meant to see if all paths are unique

python/katana/local/analytics/_ksssp.pyx
arthurp: No. But those stats are not exposed to python. :-) That's the issue you need to actually wrap the accessors using properties.

python/test/test_cpp_algos.py
arthurp: Oh. Yea.That was a dumb on my part. :-)

python/katana/local/analytics/_ksssp.pyx
arthurp: There are still no statistics exposed to python on this class. However there are definitely statistics here: https://github.com/KatanaGraph/katana/blob/4370ed81177ea2793daab468d379f8db511c6334/libgraph/include/katana/analytics/k_shortest_paths/ksssp.h#L104-L105

I guess we can just omit the statistics from the Python API for the moment, but if that's the case there is no reason to include `KssspStatistics` in the Python interface at all.

arthurp: My point is that just printing the statistics doesn't seem useful. You can do that already, but being able to programmatically access them seems pretty important.

dqawami: Apologies, didn't realize what was meant by variables being exposed, is it good now?

python/katana/local/analytics/_ksssp.pyx
```diff
@@ -192,18 +222,116 @@ def ksssp(pg, str edge_weight_property_name, size_t start_node,
...
+        """
+        Path from a source to destination node
+
+        :rtype: List[int]
+        """
+        res = []
+        for i in range(self.underlying.path.size()):
+            res.append(self.underlying.path.at(i))
+
+        return res
```
arthurp, line 270: NIT: I think you can just return `self.underlying.path` and Cython will just convert it. But it doesn't matter this will all get discarded once we move to pybind11.

arthurp (approved):

python/test/test_cpp_algos.py
```diff
@@ -153,6 +155,27 @@ def test_sssp(graph: Graph):
...
+def test_ksssp(graph: Graph):
+    weight_name = "workFrom"
+    start_node = 0
+    report_node = 10
+    num_paths = 5
+
+    table = ksssp(graph, weight_name, start_node, report_node, num_paths)
+    stats = KssspStatistics(graph, weight_name, table, report_node)
+    paths = stats.paths
+    assert len(paths) <= num_paths
```
uditagar, line 167: since this is deterministic, can you test how many paths are present from 0 to 10 in this graph and have a specific test? thanks

python/test/test_cpp_algos.py
```diff
@@ -153,6 +155,27 @@ def test_sssp(graph: Graph):
...
+
+    for i, i_path in enumerate(paths):
+        i_nodes = i_path.path
+        assert i_nodes[0] == start_node
+        assert i_nodes[-1] == report_node
+        # Assert that all paths are unique
+        for j_path in paths[(i + 1):]:
+            assert i_path.weight <= j_path.weight
+            assert list(i_nodes) != list(j_path.path)
```
uditagar: you can have same set of nodes in a multi-edge graph

python/katana/local/analytics/_ksssp.pyx
dqawami: Tried returning self.underlying.path at [this commit](https://github.com/KatanaGraph/katana/blob/56743a16f4703196ca1478da32ef1e6a59a7ca32/python/katana/local/analytics/_ksssp.pyx) but it failed to compile due to not automatically converting, so had to do this

python/test/test_cpp_algos.py
dqawami: Just to confirm, do you mean calculating the number of paths in the graph and validating it? Given from what I understand there is a chance that there will be less paths than num_paths

PR #996, opened 2022-03-29
Fix TOC to only put border on the lowest active element

JamesFCoffey:
Improvement to the TOC from High Seas.

### DIFF
--- a/docs/_static/style.css
+++ b/docs/_static/style.css
@@ -185,12 +185,18 @@ input#search-input::placeholder{
 }
 
 .bd-sidebar .nav li>a {
-	padding: 10px 10px;
+	padding: 10px 15px;
+	font-weight: 400;
 	margin-top: 5px;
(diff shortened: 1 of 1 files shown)


bobdc:
Is there an HTML demo of this change that we can see somewhere?

JamesFCoffey:
> Is there an HTML demo of this change that we can see somewhere?

@bobdc Here you go.

**Currently:**
toc_current

**This PR:**
toc_new

JamesFCoffey:
Keep in mind for the first image, it is of docs.k9h.dev and thus has more tabs than this `katana` repo.

bobdc (approved):
Looks good. Good idea.

JamesFCoffey:
Due to conversion of `katana` to a subtree. This has been reopened here: https://github.com/KatanaGraph/katana-enterprise/pull/3761

PR #997, opened 2022-04-12
tsuba: Require URI in interfaces instead of strings

tylershunt:


tylershunt:
Grabbed #998

PR #998, opened 2022-04-13
Add full node and edge schemas to Python graph [KAT-5513]

arthurp:
Cherry-pick of https://github.com/KatanaGraph/katana-enterprise/pull/3903

### DIFF
--- a/python/test/test_property_graph.py
+++ b/python/test/test_property_graph.py
@@ -16,6 +16,23 @@ def test_load(graph):
     assert len(graph.loaded_edge_schema()) == 3
 
 
+def test_schemas(graph):
+    assert len(graph.loaded_node_schema()) == 17
+    assert len(graph.loaded_edge_schema()) == 3
+    assert len(graph.node_schema()) == 17
+    assert len(graph.edge_schema()) == 3
+    node_schema = graph.loaded_node_schema()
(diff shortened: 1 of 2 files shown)


python/test/test_property_graph.py
```diff
@@ -16,6 +16,23 @@ def test_load(graph):
     assert len(graph.loaded_edge_schema()) == 3
 
 
+def test_schemas(graph):
+    assert len(graph.loaded_node_schema()) == 17
```
gliga, line 20: why do `loaded` start with 17; should it be 0 initially (or maybe something is happening when before the graph comes here as the argument, which is the drawback of reviewing code without context :).

gliga (approved):
Nice!

python/test/test_property_graph.py
arthurp: The test graph has all properties loaded by default. I think mostly for historical reasons, but it doesn't seem like a big issue to me.

gliga: That is totally fine.  If somebody updates that eventually this test will fail so they will have to fix, so all is good.

roshandathathri (approved):
LGTM

tylershunt:
Merged in #997

PR #1002, opened 2022-06-07
libtsuba: maintain viewtype/viewargs invariant [KAT-6319]

scober:
When view type is set to the full view specifier in CopyRDG, the view
args needs to be emptied to maintain the invariant.


scober:
cc @KatanaGraph/storage 
cc @bhargavivadali

